PR #502, opened 2014-12-20
Csg node

zeffii:
CSG node, current python implementation located in `utils/csg_*`


zeffii:
@shulmang I've pushed a small fix that should allow vertices made from tuples as well as lists..

could probably move the `elif len(args) == 1 and isinstance(args[0], (list, tuple)):` to the first check, so the other two aren't evaluated first. may improve performance a little bit..

zeffii:
image

shulmang:
Thanks, @zeffii!

However, I'm still seeing the problem even with this commit.  On Blender 
2.73a on OSX the isinstance(args[0], Vector) seems to be returning false 
when a Vector is instantiated with another Vector instance resulting in 
the new Vector having x, y, & z = 0.0 and thus causing the divide by 0.

Here is a PR which simplifies the initialization of Vector to eliminate 
the need for the problematic isinstance(args[0], Vector): 
https://github.com/nortikin/sverchok/pull/594

zeffii:
The documentation for the features of this Node was never completed. it is very simple. The only interesting feature that people might be struggling with is that it also outputs the intersection surface as an unattached mesh. Most people don't need this surface, but there is a scenario where the surface is the only intersecting part of the intersection.

If you don't want this disjoint geometry, stick the output of the node into a Remove Doubles node. The node is not broken.

Issue #517, opened 2015-01-09
New nodes?


portnov:
Hi all.
After playig with Sverchok a bit, it seems to me it lack some number of higher-level nodes. I wonder why? Is it like in contradiction with Sverchok's philosophy or just because no one managed to implement them yet? In last case, i beleive i can help :)
Most of these can be implemented via combination of existing nodes. But such implementation will cause maintainability problems (it's harder to work with a 100-nodes tree than with a 20-nodes), and, potentially, performance problems. Moreover, i beleive such nodes could become sort of higher-level building blocks, and used several times even in one tree.
So here a propose a list of such nodes. Let's discuss, maybe some of this functionality will be better to have in some of existing nodes, or split one item into two to have more orthogonal set of nodes.
Order is random, i put numbers only for futher reference.

1. "N-Gon". Generate a regular polygon with N sides. This is, in my mind, a generalization of Circle node. Added features: 1) optional randomization of vertices location, with separate control over randomization amplitude in R (radius) and Phi (angle) directions; 2) option to connect n'th vertex not with (n+1), but with (n+k) — this will create star-shaped polygons. The last option, in most cases, will be usable only in combination with existing "Intersect Edges" node. I have an implementation of this — i'll do a pull request if you wish.
2. "Zip vertices". Given two lists of vertices, produce a list of edges, which connects n'th vertex from first list with n'th vertex from second (obviously, it will need to re-enumerate vertices first). Now it can be implemented with "Range Int" + "List Length" + "List shift" + "List match" + "List Zip". Probably, it will be enough to add new output slot to List Match node, which will output pairs of indicies of matched items (not items themselve, but their indicies).
3. "Shuckhov". Given two lists of vertices, produce a list of edges, which connects n'th vertex from first list with (n+k)'th and with (n-k)'th vertex from second list (do `wrap' if needed). For example, for two circles of 10 vertices each and with k=3 or 4 this will produce some part of one-sheet hyperboloid (smth like Shukhov's tower http://dic.academic.ru/pictures/wiki/files/83/Shukhov_tower_by_Sergei_Arsenyev.jpg) 
![shukhov](https://cloud.githubusercontent.com/assets/284644/5678747/35144e8c-9817-11e4-85be-35dfdaa5d623.png)
. Useful mostly in comibnation with "Intersect Edges". I have an implementation of this as a ScriptNode.
[...]

PR #518, opened 2015-01-09
Implement NGon node.

portnov:
refs #517.
Please review :)

### DIFF
--- a/docs/nodes/beta/beta.rst
+++ b/docs/nodes/beta/beta.rst
@@ -6,3 +6,5 @@ Beta
    :maxdepth: 2
 
    image_decompose
+   ngon
+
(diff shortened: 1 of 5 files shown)


zeffii:
As pull requests we've decided to allow new nodes into the beta menu first, or alpha if you want. But to leave beta we ask that you supply documentation.

portnov:
Seems all fixed.

zeffii:
ok looks good, btw  image syntax for rst is 
```
.. image:: images/ball1.gif
```

zeffii:
I propose to have seed as an integer instead, similar to Random Vector.  Strictly speaking it doesn't matter, but I don't see seed as a float almost anywhere, (except Random number in sverchok..but it's too late to change that without replacing the node)

portnov:
ok, will do.

portnov:
See last commit.

PR #525, opened 2015-01-10
Implementation of Randomize node

portnov:
Shortly it was described in #517 as No.7.
Please review.

### DIFF
--- a/docs/nodes/beta/beta.rst
+++ b/docs/nodes/beta/beta.rst
@@ -9,5 +9,6 @@ Beta
    fibonacci
    exponential
    ngon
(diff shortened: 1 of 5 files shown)


zeffii:
My only comment about this is that there is potential to increase the descriptive power of the name. "Randomize" is perhaps too indescript. Mentioning _what_ it randomizes might help people use the search feature. "Randomize input Vertices" or "Vector randomizer".  Additionally `SvRandomizeNode` , may be too weak.

portnov:
Done.

portnov:
Done

PR #530, opened 2015-01-13
Implementation of `Objects along edge' node

portnov:
See portnov#6 for some description.
Please review and test :).

### DIFF
--- a/docs/nodes/beta/beta.rst
+++ b/docs/nodes/beta/beta.rst
@@ -10,5 +10,6 @@ Beta
    exponential
    ngon
    randomize
+   objects_along_edge
    iterate
 
--- a/docs/nodes/beta/objects_along_edge.rst
+++ b/docs/nodes/beta/objects_along_edge.rst
@@ -1,0 +1,93 @@
+Duplicate along Edge
+====================
+
+*destination after Beta: Modifier Change*
+
+Functionality
+-------------
+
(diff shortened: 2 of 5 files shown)


zeffii:
```python
def get_axis_idx(self):
    return 'XYZ'.index(self.orient_axis_)
```

maybe it's just me but I prefer to see stuff indented as far to the left as possible.
```python
    orient_axis_ = EnumProperty(name = "Orientation axis",
                        description = "Which axis of donor objects to align with recipient edge",
                        default = "X",
                        items = axes, update=orient_axis_change)

# vs


    orient_axis_ = EnumProperty(
        name="Orientation axis",
        description = "Which axis of donor objects to align with recipient edge",
        default="X",
        items=axes, 
        update=orient_axis_change)

```
else it's just artificially indenting so far

zeffii:
i'm a fan of all the other stuff tho, but that indentation seems a bit jarring

portnov:
I fixed formatting.
And also I added a parameter "padding" to left ends of recipient edge free - this can be useful to put something at vertices, or just to avoid intersections between generated objects.

zeffii:
coool! let's make some neat renders to show this off, or find out to break it :)

portnov:
I did some minor fix in UI.
See last screens in portnov#6 :)

PR #540, opened 2015-01-19
Fix rotation angle detection bug in `objects along edge' node

portnov:
refs portnov#14
Sorry for pull-request spamming :)

I also considered usage of mathutils.Vector.rotation_difference instead of "hand-made" rotation angle detection, but for my testcases my variant did better. You can try uncomment line 207 and comment line 206 and try with your examples, if you have some.


zeffii:
We are fine with not using `mathutils`, usually component-wise hand rolled solutions tend to be slightly more efficient

PR #543, opened 2015-01-21
Vector input and output nodes for cylindrical/spherical coordinates

portnov:
These nodes are analogous to pair of Vector In/Vector Out, but use spherical or cylindrical coordinates. 
refs portnov#11
Please review.

### DIFF
--- a/docs/nodes/beta/beta.rst
+++ b/docs/nodes/beta/beta.rst
@@ -8,6 +8,8 @@ Beta
    image_decompose
    fibonacci
    exponential
+   vector_polar_in
+   vector_polar_out
    ngon
    randomize
    objects_along_edge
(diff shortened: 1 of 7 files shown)


zeffii:
the slightly verbose nature of the check in `process`
```
outputs = self.outputs
if not (outputs['rho'].is_linked or outputs['phi'].is_linked or outputs[self.coordinates].is_linked):
    return
```
I'm thinking we should perhaps offer a lightweight query function for that to allow something like

```
from sv_utils import query_outputs
if not (query_outputs(self, any, ['rho', 'phi', 'self.coordinates']):
   return
```
or something...

ly29:
I think this is short enough, no?
If any can be linked
```python
if not any(socket.is_linked for socket in outputs):
    return
```
If a named subset needs to be checked.
```python
names = ["socket0", "socket1", ...]
if not any(outputs[name].is_linked for name in names):
    return
```

At least until make intended and perhaps required sockets inspectable.

zeffii:
@ly29 yeah, those approaches are much neater, where was my head? :)

PR #553, opened 2015-01-24
Bevel node implementation

portnov:
This uses bmesh.ops.bevel() operator.
See also last comments in portnov#16.
Please review.


zeffii:
Looks OK, looking forward to finding a use for it.

PR #555, opened 2015-01-29
Fix "outdated internal index table" error in Offset node

portnov:
for Blender 2.73:
Node Offset Node had exception BMElemSeq[index]: outdated internal index table, run ensure_lookup_table() first

### DIFF
--- a/nodes/modifier_make/offset.py
+++ b/nodes/modifier_make/offset.py
@@ -164,6 +164,9 @@ class SvOffsetNode(bpy.types.Node, SverchCustomTreeNode):
                 if round(degrees(ang_)) == 180 or round(degrees(ang_)) == 0.0:
                     bme.verts.new(q)
                     bme.verts.index_update()
+                    if hasattr(bme.verts, "ensure_lookup_table"):
+                        bme.verts.ensure_lookup_table()
+
                     list_2.append(bme.verts[-1])
                     dict_0[j].append(bme.verts[-1])
                 else:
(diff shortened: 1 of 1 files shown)


zeffii:
I thought I already committed the fix, guess not..

zeffii:
perhaps I postponed because it felt like there were too many such function calls.. but whatever works, I didn't give it much deep thought as soon as it started working again.

zeffii:
It might be more sane to tests ones if such function call exists. and refuse to process any geometry if it doesn't exist, then we don't need to use the `hasattr()` in subsequent calls. Not using updated blender version isn't something we want to encourage (i think?) @ly29, @nortikin

portnov:
In general, i think, we need to have a list of supported Blender versions stated somewhere. Maybe in README. I suppose not everyone got updated to 2.73 yet, so maybe we need to support 2.72 or even 2.71 (?). For example, Debian testing has only 2.72, stable has even 2.63. I think we obviously not going to support two-years-old blender versions; on the other hand, i think we need to support at least one previous version.

zeffii:
Those are fair points, currently the readme does state 2.71, but I suspect this is due to laziness on our part more than a real effort to retain support. But I see your point, and that's another reason why we have the check elsewhere.

nortikin:
guys, last i have not time and it seems not soon. agree, we need support latest version also.
What is refuse? just not working node? or we need to fix its behaviour? or we need global function that lookup in one line to execute in code?

zeffii:
The commit is essentially the same as this https://gist.github.com/zeffii/56ed885a3b4fd920f7f7 which is my first attempt at fixing the offset node.

If the offsetnode has to process heavy meshes with many radial points for the fillet, it migh be a good idea to rewrite parts of it to avoid the `ensure_lookup_table` being called for every new vertex added.

```python
                    for o in range(n_ + 1):
                        q5 = self.a_rot((rot_ang * o / n_), rp_, axis_, q4)
                        bme.verts.new(q5)
                        bme.verts.index_update()
                        if hasattr(bme.verts, "ensure_lookup_table"):
                            bme.verts.ensure_lookup_table()

                        dict_0[j].append(bme.verts[-1])
                        list_3.append(bme.verts[-1])
```
Also aliasing the `bme.verts` as `bme_verts` is worth doing.

zeffii:
https://github.com/nortikin/sverchok/pull/558 I propose something more radical like this:

PR #557, opened 2015-01-31
Implementation of `Calculate normals' node

portnov:
Given a mesh, calculate normals for its faces and vertices.
Partly functionality of this node intersects with "Centers polygons" node, but: 1) "Centers polygons" cannot calculate normals of vertices, 2) "Centers polygons" cannot calculate normals for meshes without faces, i.e. for curves. On the other hand, "calculate normals" has nothing to do with matrices, as "centers polygons" does.
See also portnov#20.
Please review.

### DIFF
--- a/docs/nodes/beta/beta.rst
+++ b/docs/nodes/beta/beta.rst
@@ -9,6 +9,7 @@ Beta
    bevel
    fibonacci
    exponential
+   calc_normals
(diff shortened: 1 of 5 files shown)


zeffii:
excellent use of the bmesh module

nortikin:
Needed long time ago

PR #558, opened 2015-02-01
i propose something more radical

zeffii:


zeffii:
Probably best not to merge this yet, as I haven't checked if anything is broken by it. But this avoids quite a lot of unnecessary index updates, which probably stem from when Bmesh was new and less well understood

PR #562, opened 2015-02-07
Implementation of `Mesh filter` node

portnov:
This node filters vertices, edges and faces by several criterias, available via bmesh api: boundary vs interior, convex vs concave.
refs portnov#21
Please review and test.

### DIFF
--- a/docs/nodes/beta/beta.rst
+++ b/docs/nodes/beta/beta.rst
@@ -10,6 +10,7 @@ Beta
    fibonacci
    exponential
    calc_normals
+   mesh_filter
    vector_polar_in
    vector_polar_out
    ngon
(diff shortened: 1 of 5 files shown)


portnov:
Please note that this node has intersecting functionality with bmesh_verts.py module from @Kosvor branch. So their functionality probably needs to be merged in one node, to avoid duplication.

Kosvor:
Let them be separated. To me, they are quite different from each other. Also, because their code will be easier to maintain this way.

zeffii:
classnames and id names may be too generic in this case, to avoid potential clashing consider adding the `sv` prefix.

portnov:
Done.

zeffii:
Some really nice code in this. Another clean example for newcomers to study.

Issue #563, opened 2015-02-08
Apply matrices to vertices, edges and faces


portnov:
This corresponds to nortikin#517, No.6. Options:

* to create "Mk2" version of Matrix Apply node, which would not only apply matrices to vertices, but also repeat edges and faces from input to output, so in case of several matrices whole meshes would be duplicated, not only vertices.
* To create "Mk2" of Mesh Join node, which would automatically repeat input edges and faces.

These options are compatible though, i.e. we can have both.

I started implementation of Option1 for now, as it seems simpler for use.

![1](https://cloud.githubusercontent.com/assets/284644/6096652/ac13659e-afbf-11e4-83c9-e13b75c0e346.png)

The code is in apply_and_join branch in my repo.

refs portnov#9

nortikin:
but we have allready this node
apply matrix

zeffii:
Matrix Apply should have optional duplication, not implicit. Because often Matrix Apply will be a final node, and it is more efficient to let the Viewer obtain the edges on the fly, than prefetch.

zeffii:
But i suspect there will be no extra processing per socket, unless something is connected right?

portnov:
@zeffii if edges and/or faces input is not connected, then this input works like Matrix Apply.

zeffii:
yes, cool!

portnov:
![2](https://cloud.githubusercontent.com/assets/284644/6096654/b300fbfa-afbf-11e4-901b-1361a44238c2.png)

zeffii:
Somewhere in the docs for this node, we should mention (due to this change), that unless there is further processing going on which explicitly require the duplicated topology, that letting the _viewer draw_ or _bmesh viewer_ nodes auto repeat the index lists of the topology is slightly more efficient.

portnov:
Ok.

nortikin:
like this one. Why i did not that at beginning? maybe i'm too lazy

portnov:
Closing since this is implemented and merged.

PR #566, opened 2015-02-08
Implementation of `Apply matrix to mesh` node

portnov:
refs #563

### DIFF
--- a/docs/nodes/beta/beta.rst
+++ b/docs/nodes/beta/beta.rst
@@ -16,6 +16,7 @@ Beta
    ngon
    bricks
    randomize
+   matrix_apply_and_join
    objects_along_edge
    extrude_separate
    iterate
--- a/docs/nodes/beta/matrix_apply_and_join.rst
+++ b/docs/nodes/beta/matrix_apply_and_join.rst
@@ -1,0 +1,37 @@
+Apply Matrix to Mesh
+====================
+
+Functionality
+-------------
+
+Applies a Transform Matrix to a list or nested lists of vertices, edges and faces. If several matrices are provided on the input, then this node will produce several meshes.
+
+
+Inputs
+------
+
+This node has the following inputs:
+
+- **Vertices**. Represents vertices or intermediate vectors used for further vector math.
+- **Edges**
+- **Faces**
+- **Matrices**. One or more, never empty.
+
+Outputs
+-------
+
+This node has the following outputs:
+
+- **Vertices**.  Nested list of vectors / vertices, matching the number nested incoming *matrices*.
+- **Edges**. Input edges list, repeated the number of incoming matrices. Empty if corresponding input is empty.
+- **Faces**. Input faces list, repeated the number of incoming matrices. Empty if corresponding input is empty.
+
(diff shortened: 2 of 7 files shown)


zeffii:
but
```python
        result_edges = edges * n
        result_faces = faces * n

        self.outputs['Vertices'].sv_set(result_vertices)
        if self.outputs['Edges'].is_linked:
            self.outputs['Edges'].sv_set(result_edges)
        if self.outputs['Faces'].is_linked:
            self.outputs['Faces'].sv_set(result_faces)
```
Performs `edges*n` and `faces*n` even if no sockets are connected. Though I expect this is not much processing really..

Also I wonder about the Vector_Degenerate if it would not be simpler (or even faster) to
```python
    def apply(self, vecs, mats):
        out = []
        lengthve = len(vecs)-1
        for i, m in enumerate(mats):
            out_ = []
            k = i
            if k > lengthve:
                k = lengthve
            out_ = [(m*v)[:] for v in vecs[k]]  # vector direct to tuple
            out.append(out_)
        return out
```
and ditch the vector_degenerate step?

zeffii:
I don't quite understand, the _Join_ part is not implemented merely the _Repeat_ part?

portnov:
Yes, atm it in fact outputs N whole meshes, to be joined by Mesh Join. Do you think it should also do the work of Mesh Join?

zeffii:
Optionally, yes. Else it might be a misnomer to include _Join_ in the name of the node. My preference is that it could do Join but implemented as a module in `utils/sv_mesh_utils.py` so both a separate Join node and this Apply node could reuse the same code.

The node is handy without the feature though, but then I would call it something like _ApplyAndRepeat_.

portnov:
Ok, will try to implement that

zeffii:
The current code also has a limiting assumption that if there are several nested vert lists, that they all have the same edges/faces, which they might not.

portnov:
From my point of view, usage of Vector_degenerate once at last moment produces much cleaner code than [[:]] mumbo-jumbo :) And I don't think this will do any noticable difference in performance.

I pushed some refactorings; please review.

zeffii:
https://github.com/portnov/sverchok/blob/master/data_structure.py#L392-L393
under the hood the mumbo jumbo is the same, almost.

portnov:
Yep, but it's under the hood, that's whole point :)

zeffii:
I was speculating, which it still is unless someone speedtests.

portnov:
@zeffii sorry, i found a bug introduced while refactoring. I pushed a fix.

portnov:
Btw, currently when Join option is used, the node outputs unwrapped output. Just list of vertices and so on. Should it do wrapping of outputs for Join mode?

zeffii:
maybe @nortikin or @ly29 have opinion about that, I don't at the moment

zeffii:
pick one, if it turns out to be better to have the other, we switch.

portnov:
Ok, then let's wait for other opinions.

nortikin:
from my practic i see wrapping has to be option and not by default. by default it need to be unwrapped as usual.

PR #567, opened 2015-02-08
Fixed edge ordering in circle generator

pathim:
In the circle generator the last edge was defined the opposite way.
This creates inconsistend normals when extruding.
The commit reverses the last edge so all edges point towards the "higher" vertex index (with zero being higher than the last vertex).


portnov:
+1 :)

PR #583, opened 2015-02-22
Implementation of `Extrude edges' node

portnov:
See http://www.blender.org/api/blender_python_api_2_72_1/bmesh.ops.html#bmesh.ops.extrude_edge_only
refs portnov#23
Please review.

### DIFF
--- a/docs/nodes/modifier_change/extrude_edges.rst
+++ b/docs/nodes/modifier_change/extrude_edges.rst
@@ -1,0 +1,45 @@
+Extrude Edges
+=============
+
+*This node testing is in progress, so it can be found under Beta menu*
+
+Functionality
+-------------
+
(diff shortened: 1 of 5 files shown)


zeffii:
looks fine to me. am wondering if we should have a helper function in `utils.sv_bmesh_utils` for these
```python
faces = [f for f in new_geom if isinstance(f, bmesh.types.BMFace)]
```
like:
```python
def filter_by_type(iterable, geom_type=bmesh.types.BMFace):
    return [elem for elem in iterable if isinstance(elem, geom_type)]
```
allowing us to just write:
```python
faces = filter_by_type(iterable)
edges = filter_by_type(iterable, bmesh.types.BMEdge)
# etc..
```

Issue #585, opened 2015-02-23
`Angles at edges' node


portnov:
Given a mesh, calculate angle at each edge.
See http://www.blender.org/api/blender_python_api_2_72_1/bmesh.types.html#bmesh.types.BMEdge.calc_face_angle
This partially corresponds to https://github.com/nortikin/sverchok/blob/obj_sockets/nodes/analyzer/bmesh_verts.py

portnov:
![1](https://cloud.githubusercontent.com/assets/284644/6328298/c00b10be-bb87-11e4-9ae8-75deb3aed81e.png)

portnov:
done.

PR #588, opened 2015-02-23
Implementation of `Angles at the Edges' node

portnov:
refs #585 
Please review.

### DIFF
--- a/docs/nodes/analyzer/analyzer.rst
+++ b/docs/nodes/analyzer/analyzer.rst
@@ -14,4 +14,5 @@ Analyzer
    raycast
    volume
    weights
+   edge_angles
    calc_normals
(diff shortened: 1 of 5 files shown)


portnov:
Most computationally complex part of this appears to be... Reordering of angles.
The problem is: BMesh appears to store edges of mesh in order of it's own preference. So I receive angles in, strictly speaking, random order. So I have to reorder these angles to be in the same order as input edges are. I do it in untangle_edges() function, which is nearly O(N^2).

Kosvor:
random order does not make sense. Campbell Barton should pay attention to it.

portnov:
I think it is not really random - it is probably order simplest / optimal to use by internal bmesh data structures. But from external point of view, it's random.

zeffii:
The unordered state of edges/faces after bmesh.ops is one reason we only use a subset of the bmesh module, it's not a flaw of the module but the result of optimal algorithms where the exact idx of edges and faces is less relevant than the outcome.. even bmesh.ops used for Lathe results in unordered faces/edges, it's usually fine.

PR #590, opened 2015-03-03
Refactoring proposal: factor some input/output boilerplate code out from nodes to core code

portnov:
I propose to simplify code of many nodes by removing duplicated boilerplate for standard inputs/outputs sockets processing routine.

Summary of changes:
* Decorator for checking if mandatory inputs/outputs are connected — checking_links.
* Classes for node input/output sockets metainformation.
* Such metainformation enables us to write default implementation of sv_init().
* Shortcut decorators for most "standard" (i beleive) routine of input/output data processing and matching of input lists — match_inputs and std_inputs_processing.
* Shortcut decorator for nodes which work by converting input lists to BMesh and then BMesh to lists again.

I also changed one pretty simple node to use changed API, for proof-of-concept and testing purposes. As you can see this change allows us to remove about 30 lines of boilerplate code from each node which uses "standard" inputs processing. Metainformation about input and output sockets can be also used later by upgrade system.
I tried to maintain backwards compatibility. All "old" nodes, which do not use new features, should work as they were.

Please review and test. Any suggestions are welcome.

### DIFF
--- a/data_structure.py
+++ b/data_structure.py
@@ -265,6 +265,10 @@ def sv_zip(*iterables):
         yield result
 
 def checking_links(process):
+    '''Decorator for process method of node.
+    This decorator does stanard checks for mandatory input and output links.
+    '''
+
     def real_process(node):
         # check_mandatory_links() node method should return True
         # if all mandatory inputs and outputs are linked.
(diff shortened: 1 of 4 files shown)


zeffii:
This I quite like! if it indeed maintains backward compatibility even better. I especially like the input/output descriptors with defaults / is_mandatory. Unfortunately I don't have time to test this at the moment.

nortikin:
decorators is unexplored by me as should be. Will check with interest.

PR #594, opened 2015-03-12
adding CSG to geom class because these names seem a bit overloaded in blender, simplifying CSGVector initialization

shulmang:


zeffii:
good catch! and thanks for the commit! Enjoy

PR #630, opened 2015-05-09
Init work

zeffii:


zeffii:
I think this silences a few errors on F8 and adds the ascii logo with colour to utils.

perhaps i'll ditch the execjs error message.. or the node entirely.. it didn't go anywhere

image

nortikin:
Ыыыы!!!
Cool, like that kind of stuff

PR #633, opened 2015-05-10
Pragma2

zeffii:
for your consideration and testing. this is not extensively tested but the code is very simple and `me.transform(matrix)` is a very fine way to pump matrix transformation onto vertices.


zeffii:
@nortikin , @ly29   I plan to push this unless there are objections.

ly29:
Looks good but havn't tested it.

nortikin:
i'm ok with it.

PR #645, opened 2015-05-20
Obj sockets new

Kosvor:


nortikin:
```python
def second_as_first_cycle(F,S):
    if len(F)>len(S):
        return list(map(list, zip(*zip(*[F, itertools.cycle(S)]))))
    else:
        return [F,S]
```
please, add description. zipzip ))

PR #650, opened 2015-05-23
Generative art nodes

elfnor:
Added 'SvGenerativeArtNode' to text_editor_plugins.py to allow CTRL-ENTER update

Replaced quit blender on bad xml with ValueError exceptions

Set max matrices default to 1000.


zeffii:
i'm OK with this!

PR #689, opened 2015-06-13
Generative art nodes

elfnor:
need to sort names of objects to get consistent mapping from shapes in xml to mask output and objects in Sverchok


nortikin:
ok, check if it fits https://github.com/nortikin/sverchok/blob/master/nodes/basic_data/objects.py#L53-L57
Or i missing something?

nortikin:
long time i want output objectID from  old GetObjects node. it can take vertex groups now if you check coresponding flag, than we can sort allready in GetObjects node

PR #705, opened 2015-09-06
Add eval node as a beta node

taxpon:
# Eval Node
## Appearance
image

## Function
This node enable users to user 'eval' function in python. This node is so powerful when you handle data structure :)

PR #718, opened 2015-09-13
Add UdpClientNode

taxpon:
I have created UDP Client node to communicate with other programs as I mentioned #705. Please review it.

# Appearance
image

# Function
- Send UDP message given form inlet to server which has specified ip address and port number.
- If the node receive any message from server, received message will directly send to outlet (receive).

# Example
image
- This is a short usage example of this node. UdpClient node receives current frame number from inlet and send it to the remote server. In this example, I built server as a simple echo server so server returns frame number to the node.
- I am currently planing to use this node to communicate with another program which is analyzing music sound date in real-time to show sound wave in sverchok like [this](https://twitter.com/taxpon/status/638680331684438016)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ docs/_build
 docs/.make.bat.swp
 
 nodes/generator/meta_balls.py
+
+# IDE
+.idea
--- a/menu.py
+++ b/menu.py
@@ -231,6 +231,10 @@ def make_node_cats():
         ["ConverterNode",       "SocketConvert"],
     ]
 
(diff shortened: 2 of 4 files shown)


zeffii:
very cool. i made a small [custom node tree for SuperCollider](https://github.com/zeffii/SoundPetal/blob/master/utils/osc_panel.py), with an OSC client / server

zeffii:
i'm OK with adding this!

taxpon:
:+1:

PR #743, opened 2015-11-23
update centres matrix code

elfnor:
The Centres Polygon node can produce the wrong matrix in some cases. These tend to be for faces whose normal is aligned with an axis.

See the image in the comment on my blog http://elfnor.com/simple-sverchok-01-centers-polygons.html#comment-2371038204

I've rewritten the polygons_centers.py to simplify it and fix this problem

elfnor


nortikin:
you did double job, elfnor. some days ago i did the same bugfix.
Sorry for that

nortikin:
but i donno what solution the best..

nortikin:
it seems you solution more clear, it is me, who made double job

PR #750, opened 2015-11-30
Update list_mask_in.rst

sfire:
Add some description and example for the list mask in node.


zeffii:
Thank you @sfire ! 

there are minor formatting and grammar issues but, some doc is better than no doc :) For most people reading the docs their first language isn't English . :)

nortikin:
more contribution more fun

ly29:
Even more fun since it is a different usage than originally intended.

PR #769, opened 2016-02-28
Generative Art - named inputs and outputs

elfnor:
Update to Generative Art node.

Visible changes:
*  variable node inputs are now named according to field name (```{myvar}```) in xml.
*  separate named matrix list output for each type of shape in xml
*  constants can be defined in xml ```<constants myconst="1.618" />```
*  updated docs

Other:
Changed the way the system state matrix accumulates when a rule has multiple call statements.  This makes translated examples from Structure Synth work as expected. 

Note we can change the node name to "LSystem" if preferred. I used "Generative Art" as there was already some Sverchok scripted node code called lsystem.


zeffii:
Happy to merge. should there be any problems let us know. You kept the patch self-contained so no red flags.

I have several (2d and now 3d) Lsystem nodescripts, which in an ideal world I would beef up as a proper node too. I have no problem with you renaming the node, but perhaps to help classification later, use a more descriptive ..something like "LSystem StructureSynth compatible", "LSystem SS"..(I dunno, maybe that's a crap idea) 

that way leaving the namespace available for other implementations.. --- but saying that, it's no big deal if you really want to use "LSystem" as the top level node name instead of "Generate Art".

elfnor:
I'm happy to stay with Generative Art, and leave lsystem for another node.

enzyme69:
Have not seen Elfnor for a while but anyways Structure Synth is now on Browser as Browser Synth:


https://kronpano.github.io/BrowserSynth/?fbclid=IwAR3cqgWjO5EeRk1c0gL3k7TOgWbCxbONJmGg1tZqkEd0rvqks9lcLppD6UQ

PR #795, opened 2016-08-14
Add check for `self.plane` in `CSGNode.flip`

r24y:
The `clone` method checks for the existence of `self.plane` before calling `self.plane.clone()`. This commit adds the same check to the `invert` method.


ly29:
This look sane.

Issue #805, opened 2016-08-16
Show current git branch in node editor header


ly29:
This would be so extremely useful for development...

For me the following works well, in the blender console.

```python
import subprocess
import os
import sverchok
res = subprocess.run(["git", "rev-parse", "--abbrev-ref", "HEAD"],
                            stdout=subprocess.PIPE, 
                            cwd=os.path.dirname(sverchok.__file__))
branch = str(res.stdout.decode("utf-8"))
print(branch)
```

If it works for you I will implement it.

ly29:
![skarmavbild 2016-08-16 kl 11 41 06](https://cloud.githubusercontent.com/assets/6241382/17694912/6aa541bc-63a6-11e6-8f23-f1faa33cf719.png)
[branch_side_panel](https://github.com/nortikin/sverchok/tree/branch_side_panel)

ly29:
Perhaps the node editor after template would be a better place.

ly29:
![skarmavbild 2016-08-16 kl 12 02 40](https://cloud.githubusercontent.com/assets/6241382/17695556/67563338-63a9-11e6-8baf-128543bf79b1.png)

ly29:
Actually it would be easy to switch branches with this also but for now I would be happy with something like this.

zeffii:
lovely. do it :)

ly29:
@zeffii 
Does the command above work for you?

zeffii:
no
```
fatal: Invalid symlink 'C:/Users/zeffi/Desktop/scripts/addons_contrib/sverchok': Function not implemented
```

i'm totally fine with it being added in a more prominent place.. as you first indicated.

zeffii:
i have a folder called GITHUB, which i cloned `sverchok` into
from there I symlinked sverchok to a folder called `...../scripts/addons_contrib`

ly29:
I totally fine accepting some limitations for when this works.
I have symlink from deep in my ```~/Documents``` directory where i have my sverchok to ```~/Library/Application Support/blender/2.77/scripts/addons``` or something.

[fatal: Invalid symlink](https://github.com/msysgit/msysgit/wiki/Why-am-I-getting-fatal-Invalid-symlink-....-Function-not-supported.-error-when-initcloneing) 
Some easy way to resolve the symlink?

zeffii:
this doesnt work `path = os.readlink( path )`

zeffii:
[...]

PR #808, opened 2016-08-17
Branch side panel

ly29:
#805

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -87,7 +87,10 @@ ui_modules = [
     "color_def", "sv_IO_panel", "sv_templates_menu", "sv_panels", "nodeview_space_menu",
     # bgl modules
     "viewer_draw", "viewer_draw_mk2", "nodeview_bgl_viewer_draw",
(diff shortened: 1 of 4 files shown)


zeffii:
each time i start blender or f8, this appears in the messages / console
```python
** Sverchok loaded with 183 nodes **
added keyboard items to Text Editor.
fatal: Invalid symlink 'C:/Users/zeffi/Desktop/scripts/addons_contrib/sverchok': Function not implemented
** version:  (0, 5, 6, 3)  **
```

this as discussed in a thread somewhere is windows/python not being able to use the first route to get the git branch..  am slightly puzzled that it prints any message in the first place since there are no prints in the except of either try/except..
https://github.com/nortikin/sverchok/blob/master/ui/development.py

ly29:
Probably standard error or what it is called on Windows. 
Change order of the tests perhaps

zeffii:
or make it os specific, on windows do my thing, and posix do yours

ly29:
Also good

PR #820, opened 2016-08-22
Json templates fixes

zeffii:


ly29:
I thinks it is easier to just add them to back to the project and consider them binary files. But if you want to sort out the conflicts go ahead. 
Since they are more of generated data than source files I think it makes sense even...

zeffii:
meged directly into master by copying files over.

PR #821, opened 2016-08-22
adds copy button

zeffii:


ly29:
if it works do it

zeffii:
yep.. tho maybe the operator should be somewhere else..

ly29:
Maybe, but since it is so simple we could keep it there.

PR #845, opened 2016-09-01
Bmv groups

zeffii:


zeffii:
this either never worked, or stopped working.

Not all collections can take a second argument in the `.get(   )` , they should arguable all behave like that, but also the `data.text` blocks don't. This seems to suppress the issue, at least made it harder to reproduce the problem..

PR #855, opened 2016-09-03
Gistfixes

zeffii:
don't accept yet!

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -88,7 +88,7 @@ utils_modules = [
     # UI text editor ui
     "text_editor_submenu", "text_editor_plugins",
     # UI operators and tools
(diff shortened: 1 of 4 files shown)


zeffii:
yeah, a message to indicate it was copied to clipboard makes complete sense.

zeffii:
will wrap the upload in a try/except too

zeffii:
image

zeffii:
could possible make it a tabbing system so they aren't stacked but display Export or Import...

zeffii:
io_enum

zeffii:
io_enum

zeffii:
i'm thinking of letting the user make a Frame with a certain label (description), then the gist uploaded will be passed the content of that description.. it has to be a single line..

PR #876, opened 2016-09-14
Monad fixes

ly29:


ly29:
I want to push this but will have zero time to deal with problem coming days, so will hold off.

PR #877, opened 2016-09-17
Node cats preprocessor

zeffii:


zeffii:
@ly29  @nortikin   what do you think?  simplifies the node cat creation i think..

comments could be added as `> this is a comment`

ly29:
It certainly is nicer than having to write the lists etc
I would almost like a completely automatic method using ```.__subclasses__()``` and the module name and perhaps  have beta as tag on the node. But there is also something nice to say for being able to organize the nodes automatically...

zeffii:
you mention exactly what I was thinking... we could add extra `sv_category = "Scene"`  , but I don't see how we could also then force an order in the menu (not without something pretty convoluted).

i'd love to go further and drop the need to use the `,` ,  but then we enter fixed position line

     bl_idname = line[:30].strip()
     item_shortname =line[30: ...]
     ugly....

we can certainly auto populate `/nodes/__init__.py`,  order being irrelevant I think...

ly29:
What I meant to say, the automation isn't worth the trouble. The ```nodes/__init__``` would be nicer to get rid of.

zeffii:
Seen as this isn't accepted yet, i 'll go ahead and drop the `comma` divided but rather use a character positional system. The more I think of this the more relaxing it looks. The commas as the most likely thing to trip us up anyway. We just inforce rules 

  - (no tabs..etc) 
  - indent a node using 4 spaces
  - use an editor with a ruler set to 80 char line width, set the icon rightaligned to the last character
  - follow the existing convention you'll be fine.
  - if you must use comments use `> this is a comment`

ly29:
I like it, getting rid of ```nodes/__init__``` is a separate issue, Some nice comments in a header explaining said rules

zeffii:
do any nodes import from other nodes? the viewers might... ifso then functionality must be pushed to a util/____somename__.py file.

zeffii:
sublime even has a nice package to display markdown neatly, also this looks reasonable

https://github.com/nortikin/sverchok/blob/node_cats_preprocessor/index.md

ly29:
Perhaps use ```.bl_label``` as default and only require ```.bl_idname```

zeffii:
yeah, it does make sense..  i think many bl_labels do not correspond to the shortname, though it would simplify things in some ways.

ly29:
The ```bl_label``` **should** correspond in general I think

zeffii:
it should.. :)  but i'm fairly certain i've never restricted myself to do that..

zeffii:
i mean.. it should.. ideally.  and we can defo go that route.

zeffii:
it simplifies the `index.md` , and icon could then go on `column 35`

zeffii:
i must head out. bbl.

ly29:
It is worth doing a quick survey of ```.bl_label``` and the menu label.

zeffii:
which would be so much easier if the bl_idname and classname where identical for all nodes. which they aren't.

ly29:
You can get the class from bpy.types but yeah also bad form.

zeffii:
I think it's about time we rectify this. but i don't know if changing bl_label of all deviating nodes will be issue free.

going forward it would simplify the whole index/menu setup

ly29:
.bl_label should be fine, .bl_idname is the only magical property.

zeffii:
cool, wasn't positive.

probably won't be today tho..

zeffii:
not as bad as I thought.. but many shortname as very different to the initial bl_label

zeffii:
here the gist https://gist.github.com/zeffii/2eaea534ff714855b683613c286680a4

nortikin:
sorry. today i write docs and move nodes from beta to other menu. Just a moment, please.
I will say when it is ready.

zeffii:
no problems @nortikin

zeffii:
assuming we don' t automate the node_cats generator yet...
```
lines = '''\
    SvBoxRoundedNode              Rounded Box
    HilbertNode                   Hilbert
    Hilbert3dNode                 Hilbert3d
    HilbertImageNode              Hilbert image
    ImageNode                     Image                              FILE_IMAGE
    SvProfileNode                 ProfileParametric
    SvGenerativeArtNode           Generative Art
    SvScriptNode                  Scripted Node                   SCRIPTPLUGINS
    SvScriptNodeMK3               Script 3 Node                   SCRIPTPLUGINS
'''

temp_list = []
for l in lines.split('\n'):
    if not l.strip():
        continue
    
    f = l.strip().split(' ', 1)
    bl_idname = f[0]
    second_part = f[1].strip()
    verbose_mode = (len(second_part) == 45)

    if verbose_mode:
        icon_length = len(second_part[::-1].split(' ')[0])
        shortname = second_part[:-icon_length].strip()
        icon = second_part[-icon_length:]
        print('bl_idname =', bl_idname, ',shortname =', shortname, ',icon =', icon)
    else:
        shortname = second_part
        print('bl_idname =', bl_idname, ',shortname =', shortname)

```

nortikin:
did it!!
Now menu changed

zeffii:
i will resolve conflicts manually @nortikin  keep working away :)

nortikin:
folder reordering for nodes needed. Just to do it. maybe have time tomorrow to reorder. folders as in menu now. + sorting available maybe. But icons??? what to do with icons?

zeffii:
i like icons, and so do other people.  especially beginners.

nortikin:
so... separate icon's list?

zeffii:
it _is_ separate.  icons are optional for scripters to add, and optional to display.  this way I keep it in a simple list inside the index/menu

We have had this discussion before, that's why there's a tickbox in userpreferences _show icons in ctrl+space menu_

zeffii:
I've moved the beta/alpha machines to their new destinations in index.md too

zeffii:
but if we could store the icon information inside the node, i would be amicable to that too. in-fact it would be preferable.

zeffii:
we can totally do this automatically, without adding icons to the index.md / menu.py

```python
>>> sverchok.node_tree.bpy.types.SvLimitedDissolve.bl_icon
AttributeError: type object 'SvLimitedDissolve' has no attribute 'bl_icon'

>>> sverchok.node_tree.bpy.types.SvScriptNodeMK3.bl_icon
'OUTLINER_OB_EMPTY'
```
svlimited dissolve doesn't declare an icon, scriptnode3 does.

It's not a lot of work to update those nodes that have a useful icon. then fix space menu do use that..

zeffii:
I would like to merge this first..

zeffii:
or maybe tomorrow i'll edit the nodes bl_icon and merge that instead...  it's late. brain stopped.

nortikin:
why not just make icons for categories? not nodes. so nodes will inherit icons from "parant"... (sadly object-oriented language).

zeffii:
only 30 nodes have an icon in space_bar menu, i'll experiment tomorrow with dropping the icon from index.md and instead adding bl_icon to those nodes and loading them at runtime.

overal it is simpler, letting scripters suggest their preferred node_icon from their node's code is a good feature i think.

zeffii:
> why not just make icons for categories? not nodes. so nodes will inherit icons from "parant 
because then you have 20 nodes with the same icon, then there's no point to adding an individual icon.. defeating the purpose of icons.

zeffii:
note to self:

if bl_icon == 'OUTLINER_OB_EMPTY' , skip

zeffii:
this is superseded by https://github.com/nortikin/sverchok/pull/878

PR #878, opened 2016-09-17
Node cats deeper

zeffii:


zeffii:
index.md looks like this now: 

https://github.com/nortikin/sverchok/blob/node_cats_deeper/index.md

pretty clean. no icons. icons are pushed to the responsibility of the node.

zeffii:
next and final step for this 'project' would be to also take the shortname from the bl_label, as mentioned in the previous thread/PR.  This does require an en-mass replace of `menu shortnames -> bl_label`, likely tomorrow.

zeffii:
turns out by not passing the shortname, nodeitem will use the bl_label, i didn't know this. until reading the implementation of nodeitems_utils.py.

            # items=[NodeItem(props[0], props[1]) for props in nodes]))
            items=[NodeItem(props[0]) for props in nodes]))  # uses bl_label instead.

ly29:
The ```.bl_label``` is a sensible default.

zeffii:
it looks like we didn't include the groups submenu in space_menu , will add later.

zeffii:
i expect to be done within the hour..

ly29:
Could also be worth testing if the ```NodeItemCustom``` would be good enough for the space menu
https://github.com/nortikin/sverchok/blob/master/menu.py#L344
https://github.com/nortikin/sverchok/blob/master/menu.py#L374-L383

ly29:
Also striking is how many nodes don't use the ```Sv``` prefix even though they are MK2 or even 3 and definitely older than adapting that standard...

ly29:
All these fixes and adjustments makes the code more agile and easier to work with long term, thank you for taking the time.

nortikin:
offtop
shift+G works in node tree ))

zeffii:
i had to leave for a bit,  back now... still a few to do..    this is like spring cleaning :)

zeffii:
there are still a few inconsistent bl_labels  , but i'm eager to merge and put this to rest for now.

zeffii:
I am ready to merge unless there are objections.

PR #881, opened 2016-09-18
Automat

zeffii:
not full implementation yet.

### DIFF
--- a/nodes/__init__.py
+++ b/nodes/__init__.py
@@ -2,6 +2,7 @@ import os
 from os.path import dirname
 from os.path import basename
 from collections import defaultdict
+import pprint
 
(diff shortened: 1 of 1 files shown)


zeffii:
this seems to work.. i haven't found anything broken as a result..

zeffii:
any objections to this?

ly29:
I have some lingering doubts but I think we should do this. 

Next boilerplate we could remove is all the register class in each file and use register module instead,

zeffii:
at any rate, we won't be hand enumerating  the content of the nodes directories anymore. even if the code to populate `nodes_dict` eventually is nothing like this PR.

ly29:
Exactly.

PR #884, opened 2016-09-19
List modifier

zeffii:
tentative implementation, with room to add a few modes in between. I'll write the docs before pushing/merging.

I have mixed feelings about doing dynamic add/remove of sockets. I would like to propose not implementing that here. 

- If the mode requires 2 inputs, data1 and data2 are used
- if mode requires only one, then i propose (as lukast does in geometry nodes) to use data1, if linked, else data2 if linked. Something bugs me about auto-disconnects


ly29:
I am against dynamic sockets also

zeffii:
prelim docs..

### List Modifier

This node offers an assortment of list modification functions. The node has both Unary and Binary modes.

- In the Unary mode the input socket is the linked one  
- if data1 isn't linked then data2 is used as the input if it is linked  
- if both are linked data1 is used.

### Behaviour

Modes | inputs | Behaviour Description
-------- | -------- | --------------
Set   |   unary  |  turns the valid input into a set
Ordered Set by input |  unary | only unique numbers but ordered by the original input sequence
Unique Consecutives | unary | no consecutive repeats
Sequential Set | unary | unique input values, order by their value 
Sequential Set Rev | unary | unique input values, order by their value, reversed
Normalize | unary | scales down the values in the list so it ranges -1.0 to 1.0
Accumulating Sum | unary | see `intertools.accumulate`
Mask Subset (A in B) | binary | generates a mask to indicate for each value in A whether it appears in B
Intersection | binary | returns the set of items that appear in both a AND b
Union | binary | returns the set of items A joined with B 
Difference | binary | returns the set of items that appears in A but not B
Symmetric Diff | binary | returns the set of elements of both A and B that don't appear in Both

nortikin:
lovely

zeffii:
binary modes don't work yet btw , i keep getting called away.

zeffii:
or this one, i don't know..  
image

zeffii:
working example. @enzyme69  you might like this one.
image

zeffii:
@ly29  the `get_f` is a bit embarrasing.. but apparently i'm too braindead to think of a more sensible approach..  ideas?

ly29:
@zeffii don't have much time today but will try think about it

zeffii:
- [ ] need to add bl_label setting too for minimized.

zeffii:
OK! will look

zeffii:
almost ready, as far as i'm concerned:
https://github.com/nortikin/sverchok/blob/list_modifier/docs/nodes/list_main/list_modifier.rst

zeffii:
i'm ready to push @nortikin , @ly29  unless objections.

ly29:
No objections about interface and usage, that part is ready.
The ```get_f``` should be a method on the class since it has a ```self``` argument. 
Also we could cut the functions down by half if we did something like below.

```python
def get_f(self, operation, unary):
    makes_lists = self.listify
    if makes_lists:
        post_process = list
    else:
        post_process = lambda x : x # identify function
    if unary:
        if self.func_ in SET_OPS:
            def f(d):
                if isinstance(d[0], (int, float)):
                    return post_process(operation(d))
                else:
                    return [f(x) for x in d]
    else: # binary
#...
```

zeffii:
line count isn't an important metric in consideration for me until an obviously leaner alternative emerges. Both your suggested alternatives are fine, and yeah these should be methods of the class. 

The good thing is that the extra overhead by using an identity function or a makes_list check inside the   function is that these loops aren't really so tight to be of any real consequence.

I'll probably go with something similar to your suggestions, if not in code then definitely in spirit.

zeffii:
I too prefer the first suggestion..

ly29:
Line count isn't that important in itself, being clear is however. 
Avoiding relating code blocks with small small difference is important however, but even that has to be balanced against other constraints. 

Also the inner loop is the actual operation as stated, furthermore in any reasonable node tree this node won't be the slowest.

zeffii:
i'd really like to cache the function, instead of getting it on each update, only acquiring a new function if the mode is updated.

ly29:
I guess that is possible but I wouldn't bother with it, if the node class instance was a real class so that it was unique for every node I would do it but this lookup is so trivial I don't think it is worth it.

zeffii:
maybe
```python
        do_post = (self.func_ in SET_OPS) and makes_lists
        post_process = list if do_post else lambda x : x # identity function

        if unary:
            def f(d):
                if isinstance(d[0], (int, float)):
                    return post_process(operation(d))
                else:
                    return [f(x) for x in d]
        else: 
            def f(a, b):
                if isinstance(a[0], (int, float)):
                    return post_process(operation(a, b))
                else:
                    return [f(*_) for _ in zip(a, b)]

        return f
```

zeffii:
not sure if it makes sense to add a `match_long_repeat`

ly29:
I guess it should look something like ```recurse_fxy``` but with a different test. https://github.com/nortikin/sverchok/blob/master/utils/sv_itertools.py#L62-L77

zeffii:
I think i'll push and if you want to make changes, be my guest!

ly29:
I will make some tests with it tomorrow

PR #899, opened 2016-10-06
update to new py API (2.77 and above)

Kosvor2:


nortikin:
please, use https://github.com/nortikin/sverchok/blob/master/docs/contribute.rst 
SOME RULES 
14 - instructions to change node.

nortikin:
Made it just now, better lateer than never.

zeffii:
oh oops. I should update the contribute.rst too, 

- we now use only the `index.md`, 
- put your node's bl_idname where you want it
- for convenience and consistency make your bl_idname the same (identical) as the `class` name of the Node.

PR #908, opened 2016-10-09
Skin specials update2

zeffii:
not ready to pull yet


zeffii:
this is untested, but I think most of the logic is in place. no time today to play with it.

PR #927, opened 2016-10-21
Zipsaver

zeffii:


zeffii:
image

handy for sharing .blends as .zips on github's issue tracker. This copies the filepath of the zip to the clipboard too.

zeffii:
we can offer alternative archive formats if needed.

PR #931, opened 2016-10-23
add - no need to first cast coordinate as Vector for Kdtree

zeffii:
not sure if tuple is manditory, but the docs say so.. will test with List too, if that works the whole `(vx, vy, vz)` can be reduced to `xyz` .


zeffii:
should be noticeably faster will large vert count

PR #950, opened 2016-11-06
Snlite drawfunc

zeffii:
attempts prelim draw function 'append'.

to solve (is easy, but brain is slow this week) :  

- [ ] correct socket updates if v is turned to s or socketname is altered..

### DIFF
--- a/index.md
+++ b/index.md
@@ -33,6 +33,7 @@
     SvImageComponentsNode
     SvScriptNode
     SvScriptNodeMK3
+    SvScriptNodeLite
 
 ## Analyzers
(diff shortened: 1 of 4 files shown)


zeffii:
i'd probably use this node as a go to _script node_ now, but I think the things stopping me from merging to alpha/beta are
- no sliders ( i can live with )
- no way to declare classes that are persistent .. i was thinking of another preprocessor line like

```python
...
...
classes [filename] [variables]
```
so... maybe
```python
...
...
classes myclass.py < instance1 = MyClass(var1, var2), ... >
```

Issue #942, opened 2016-10-31
ScriptNode1 - Lite


zeffii:
### todo  (in no particular order)

- [ ] operator function requires a docstring as first line, else the line count (for hoisting) isn't correct. i'd like to fix that but it's proven a bit tricky. (or i've had insufficient time to tackle it)
- [ ] min/max on float/int sockets
- [ ] add bgl refresh if imported:
  
    ```
    def free(self):
        v3dBGL.callback_disable(node_id(self))

    # reset n_id on copy
    def copy(self, node):
        self.n_id = ''
    ```

# SNLite Documentation

![image](https://cloud.githubusercontent.com/assets/619340/20427206/1649e3f8-ad83-11e6-9663-3650236fe863.png)

# Script Node Lite

The script node implementation is intended for quick, non serious, experimentation. It can do many of the things ScriptNode MK1 can do, but it currently lacks support for a few things -- these are listed in the _issues_ section. Unlike SN MK3 this implementation has no illusions of becoming a first class citizen of the node tree, there's a bit of overhead per update.

## Justification

I still like SN1 a lot, but I don't like adding all the unwrapping and indexing code just to get the node to preprocess/precompile -- that was never part of the vision. A coders perogative is to make their life as easy as possible. Sometimes that means spending a bit of time making tools. SNlite isn't perfect but i'll write scripts with this until SN MK3 has had time to mature.

## Features

 - easy template importing directly to node, or to text block (the latter is useful for making changes to a template before loading... as is needed with the `custom_draw_complex` example )
    ![image](https://cloud.githubusercontent.com/assets/619340/25421401/ff29eed2-2a5c-11e7-9594-f85295178f57.png)
 - doesn't force an output socket, but an input socket is expected (else how does it update?!)
 - does allow you to set some defaults : numbers, and lists of stuff
 - does allow you to set the level of nestedness; means you don't need to unwrap/index stuff via code
 - does allow you to declare an override to the node's draw_button function, well... think of it more like an `append` to the draw code. There's an example of how you might make a light weight Curve mapping node by using the ui component of a RGB curve from a material node tree. it's a little convoluted, but it's usefulness musn't be dismissed.
 - has the option to _auto inject_ the list of variable references as _`parameters`_ much like javascript does for 'arguments` inside a function. In essence implemented like this 

    ```python
    parameters = eval("[" + ", ".join([i.name for i in self.inputs]) + "]")
    ```
[...]

PR #971, opened 2016-11-17
Snlite plus

zeffii:
- less verbose script statement
- ..has the possibility to append content to the `self.draw_buttons` in UI.
- execution may be slower than the non-lite Scripted Nodes.


zeffii:
i've created markdown formatted documentation on the [snlite thread](https://github.com/nortikin/sverchok/issues/942) to help people figure out how it works. Can't guarantee i'll answer questions if they are already dealt with in that document.

zeffii:
before pushing I want to fix gist saving/loading

PR #973, opened 2016-11-19
adds switch to load or import template

zeffii:


zeffii:
this adds the ability to either just load a template to the TextBlocks,  or import it straight away into the active node. Handy if you want to pick/mix from template content without needing to load an empty snlite each time to load the template..

PR #980, opened 2016-11-21
Random number gen

zeffii:
image


enzyme69:
​That's insane! Thanks~

PR #1015, opened 2016-12-04
add update to mode switch, also freeze during transit

zeffii:


zeffii:
@ly29  what do you think? inside `def change_inputs(self, context)` 

this

```
self.id_data.freeze
stuff
self.id_data.unfreeze 
updateNode(self, context)
```
seems to work nicely, and prevent's `process` from trying to `process` too early when the mode switch adds a new socket.

ly29:
Seems sane

PR #1046, opened 2016-12-14
Objid set mk2

Kosvor2:


Kosvor2:
Now it also creating additional lists of values, if their amount is less than amount of nested object lists. So no need to use "list join" node or such, if all value lists must be the same.

Kosvor2:
Controlling particles of several objects is much easier now:

particle_sys

Issue #1047, opened 2016-12-15
objects_mk2 freeze


zeffii:
let me know if you experience this freeze, I will revert the last pullrequest. If i had known this was happening I would not have pushed in the first place.

reproduce

1. add any four mesh objects to the scene
2. add objects_mk2 to node tree
3. select two of the mesh objects , and press Get Selection
4. try to delete one of the two currently unselected objects.  

Freezes here, but _why_ is beyond me.

zeffii:
I reverted for simplicity..

zeffii:
OK.. i rolled back and exactly the same thing happens before my edits... i think it's a windows thing.  Can anyone test this sequence of events to see if it happens.. on linux/mac  @enzyme69  @portnov  @ccamara  @taxpon  @ly29

enzyme69:
Got Torus not found error on Mac. But no freezing.

I might test this out on Windows 7, maybe later.

zeffii:
let me be more accurate in my description of the scenario that freezes ..

1.  start a fresh .blend with a new empty NodeTree
2.  add 4 Objects ( Cube, Plane, Torus, Monkey , move them around a little )
3.  add objects_mk2 to nodetree
4.  select Cube and Plane in 3dview, press Get Selection on objects_mk2
5.  hook up objects_mk2 to Viewer Draw Mk2 (inlcuding matrices)
6.  observe that the object data is being displayed by VDMK2
7.  try deleting Torus or Monkey   (yep, not the objects that we collected for objects_mk2....)

this freezes blender here..

zeffii:
![unngif](https://cloud.githubusercontent.com/assets/619340/21257587/3efc7d7c-c379-11e6-817f-44e64dd3e06d.gif)

zeffii:
but it doesn't freeze if I first disconnect objects_mk2 from Viewer Draw, and then delete the torus.

zeffii:
helpful irc triage by lijenstina 

    <lijenstina> zeffii, i've tried a 2.77.3 build and it works
    <lijenstina> from 2016-08-26

zeffii:
possibly https://github.com/lukastoenne/blender/compare/cc8132b0c83e60d46e24ecec93744f0dd2b5db90...841f3e49725685b64f54c8dff6f29dab7bc223ee#diff-406550b469dd898c4c5989bb23b0fd2dR1844

zeffii:
[...]

PR #1048, opened 2016-12-15
Revert "Objmk2 test"

zeffii:
Reverts nortikin/sverchok#1044

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -45,7 +45,7 @@ bl_info = {
         "AgustinJB, Zeffii, Kosvor, "
         "Portnov, Elfnor"
     ),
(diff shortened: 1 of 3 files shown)


nortikin:
Why?

zeffii:
@nortikin  there's a bug that happens only in windows builds, so i was trying to figure out what part it was happening in. https://github.com/nortikin/sverchok/issues/1047#issuecomment-267623020

it's clear that the code I wrote to bypass one bug, isn't good enough to prevent another related crash/freeze..   I could re-apply the objmk2_test, with less defensive code :) and ask windows users to only user Blender builds before the bug was introduced.

Issue #1060, opened 2016-12-27
matrix have to understand vertex as location


nortikin:
Where to write this understanding?
whan you plug vertex to any matrix socket, it have to be flatten in levels in one object and making matrices for every vertex. In .sv_get() function.
Can we?

nortikin:
https://gist.github.com/nortikin/10ead76843211aaf49d5d78a933e3592
trying to do

zeffii:
completely agree! makes sense to do the conversion in the socket

zeffii:
i was thinking the following..
```python
locations = []

def get_all(a_list):
    for item in a_list:
        if isinstance(item, (tuple, list)) and len(item)==3 and isinstance(item[0], (float, int)):
            location.append(matrix_from_location(item))
        else:
            get_all(item)
````

zeffii:
this would generate a flat list of _matrices_.   i don't think we generally use nested matrix lists ,  at least i've not used any lately..

zeffii:
something like this .. but it doesn't work yet :)

https://github.com/nortikin/sverchok/tree/verts_as_matrices

zeffii:
Perhaps this needs to be set in SvSetSocket instead..

zeffii:
viewer_draw_mk2 kind of expects the socket connected to the matrix socket, to me of type matrix...
https://github.com/nortikin/sverchok/tree/verts_as_matrices

needs a small fix too

zeffii:
![image](https://cloud.githubusercontent.com/assets/619340/21579967/68e9eb7a-cfd0-11e6-9309-15830746f5aa.png)

zeffii:
I think we took out most of these checks ( check if input socket was of correct type ) about 2 years ago.. any that remain should be easy to track down..

zeffii:
![image](https://cloud.githubusercontent.com/assets/619340/21580887/1ec76486-d017-11e6-9980-1a998db1addf.png)

@nortikin  want to try it ? see if it breaks older layouts?

zeffii:
https://github.com/nortikin/sverchok/tree/verts_as_matrices

Kosvor2:
Seems ok. I thought you were going to get rid of all sockets checks. That would be cool.

![rid_off_checks](https://cloud.githubusercontent.com/assets/22656834/21582478/3ce26d00-d06a-11e6-98b6-823a97086ae9.png)

zeffii:
[...]

PR #1076, opened 2017-01-01
Verts as matrices

zeffii:


zeffii:
The only problem i see with that is that some nodes still use `SvGetSocket` directly.. maybe this will force a more complete push-through of the `sv_get`  @ly29

ly29:
@zeffii 
I find one any references to ```SvGetSocket``` or ```SvSetSocket``` outside base system.
However many use the older interface of ```SvGetSocketAnyType``` and ```SvSetSocketAnyType```

```python
def SvGetSocketAnyType(self, socket, default=None, deepcopy=True):
    out = SvGetSocket(socket, deepcopy)
    if socket.is_linked:
        return SvGetSocket(socket, deepcopy)
    elif default:
        return default
    else:
        raise LookupError

def SvSetSocketAnyType(self, socket_name, out):
    SvSetSocket(self.outputs[socket_name], out)
```
However here we could just redirect to ```sv_get``` anyway....

```python
def SvGetSocketAnyType(node, socket, default=None, deepcopy=True):
    return socket.sv_get(default=default, deepcopy=deepcopy)

def SvSetSocketAnyType(node, socket_name, data):
    node.outputs[socket_name].sv_set(data)
```

ly29:
120 hits for ``SvGetSocketAnyType``` which is trivial but boring.

zeffii:
> I find one any references to SvGetSocket or SvSetSocket outside base system.
However many use the older interface of SvGetSocketAnyType and SvSetSocketAnyType

i probably confused the two function names then, i think the point still stands because there's an extra level of indirection. 

Nonetheless, a trivial matter if we really want to settle this code.

ly29:
Yeah it would be extra level redirects but the advantage is that we would have one true interface, and behind one implementation instead of two interfaces. 
But best would be to kill the extra interface directly of course.

enzyme69:
What is exactly Verts as Matrices?

zeffii:
@enzyme69  using a vertex list / lists as if it were a Matrix list. A Vector can be transformed into a Matrix, that carries only Location information.

part of this thread ( with visual examples ) https://github.com/nortikin/sverchok/issues/1060

ly29:
Interpreting vertices as location for matrix, skipping the matrix node when you only use location, which seems to be the most common use case

zeffii:
image

right now it says  Matrix 1 , but i've changed it to indicate the number of stored matrices.

zeffii:
maybe a `data_structure_helper` module is appropriate, i don't feel OK just adding these helper functions into the `data_structure` file.

zeffii:
maybe a `data_structure_helper` module is appropriate, i don't feel OK just adding these helper functions into the `data_structure` file.

But .. there's quite an eclectic mix of functionality in the data_structure file.. so maybe no harm?

ly29:
Make a ```socket_helper``` file in core and move all socket/data storage functionality into there perhaps except the old interface.
```data_structure``` should be dismanteled bit by bit I think but this is a different task.

zeffii:
right on.

ly29:
it is a bit obfuscated but quite simple system.

zeffii:
am leaning towards getting the verts2matrix to master before doing too much to `data_structure`..

ly29:
That sounds sane

zeffii:
forcing `deepcopy=False` for this operation btw. else we'll get unintended geometry :)

zeffii:
i haven't tried running the last 4 commits yet..but logic looks sound. hooking them up later

zeffii:
image

ly29:
Having looked things over I some reorganisation is sane

The test for the other socket should be moved into the class definition for sockets with ```cross_test_socket``` into the base class and using them into the relevant class. 

Also, even if no solution is found now - as definitely not a show stopper, some thought of how this could support proper OO code would be good. Refactor of ```.sv_get``` into a base case and use ```super()```. This is perhaps better left for later actually.

I for pushing this into master and getting it tested.

zeffii:
I have more issues with this, but doing docs for easing now.

of course the socket cross over stuff should be part of the socket interface, i see this implementation more as experimental prior to the design phase.

zeffii:
i'd be happy to push but to make the user aware that crossover connections will remain in a state of flux for a while. behaviour may change (un-nesting automatically etc)

Durman:
I do not see sense to wright two figures in socket labels. In the particular case number of matrixes can be either two or five but not both simultaneously.

zeffii:
it's not 51. zoom in.  the number to the right is the combined count of all internal vectors

image

zeffii:
that's 3 years ago, i don't remember exactly why that convention was settled on, but i find it useful for the following reason:
it's useful to see the number of 'container' lists represented in the socket without having to navigate to the origin of a socket.

i don't consider this broken, but maybe under-documented is a fair assessment.

PR #1081, opened 2017-01-02
Krtree edges mk2

zeffii:
skip is a bit of a crazy parameters anyway, but at least the skip=0 case should work just fine.


zeffii:
hmmm.. i changed my mind on this. am going to upload an alternate solution

PR #1093, opened 2017-01-06
Data structure cleanup

ly29:
Not quite ready for merging, please test if you have time.


zeffii:
have only had time to read about half of this. 

- i never even saw Cnode, whatever that is.
- looks like elman neuro node uses the handle caching? `from sverchok.data_structure import handle_read, handle_write`   what on earth for? :)

ly29:
I had missed that.
It is just a interface for a dict, it uses it for caching. I guess. 
It could be replaced by a dict in the file without any problem, I have never used the node so I cannot say of what it is.

zeffii:
before i forget..  i've been staring at this in `def SvSetSocket(socket, out)`

```python
     global socket_data_cache
     if not socket.is_output:
         print("Warning, {} setting input socket: {}".format(socket.node.name, socket.name))
     if not socket.is_linked:
         print("Warning: {} setting unconncted socket: {}".format(socket.node.name, socket.name))
```
wanting to change it to

```python
     global socket_data_cache
     global DEBUG_MODE
     if DEBUG_MODE:
         if not socket.is_output:
             print("Warning, {} setting input socket: {}".format(socket.node.name, socket.name))
         if not socket.is_linked:
             print("Warning: {} setting unconnected socket: {}".format(socket.node.name, socket.name))
```

ly29:
```python

    def process(self):
        handle_name = self.name + self.id_data.name
        handle = handle_read(handle_name)
        ###
        handle_write(handle_name, props)
```

This is just a complicated interface for dict. Especially the fact that it stores a dict ```{"prop": data}``` instead of just the data confuses me a bit 

```python
temp_handle = {}

def handle_read(handle):
    if not (handle in temp_handle):
        return (False, [])

    return (True, temp_handle[handle]['prop'])



def handle_write(handle, prop):
    handle_delete(handle)
    temp_handle[handle] = {"prop" : prop}
```

ly29:
Yeah that makes sense

zeffii:
from   `list_main/join.py`  can drop   ` SvGetSocketAnyType, SvSetSocketAnyType ` from import you ditched them alreay from `process`.

zeffii:
but I understand that a more dedicated pass through all current nodes is desired to tackle `.sv_get` usage.

zeffii:
@ly29  there's no mechanism currently implemented for sverchok nodetree that triggers an update  if scene objects are deleted, is there?. i looked at the handlers, altho a partial scene_update handler implementation exists it's not added as a handler.  (just wondering )

ly29:
No it there isn't, no reason it couldn't be done. 

Yeah reserving that for a major pass instead

zeffii:
the `update_scene_post`  handler triggers waaay too often.

ly29:
It sure does

zeffii:
checking out this branch now, see if I can break anything accidentally. Didn't encounter obvious issues by reading locs alone.

zeffii:
```python
            if Vertices.is_linked:
                Vertices.sv_set(vers_out)
```
am thinking about a helper to keep all that to one line..
```python
        Vertices.sv_set_if_linked(vers_out)
```

The only advantage to setting an unlinked socket is if we are communication with a node from some kind of 'wireless' arrangedment (different node system, a script..whatever, just without direct connection). But those kind of connections would fail anyway due to my hardcoding the  `is_linked` stuff..

zeffii:
maybe my issue here is that it seems counter intuitive to set socket data if it's not being read downstream.

Is the alternative a `node.property` which can be switched on externally in exactly those cases where wireless comms are desired?

```python
    # inside process  
        some_socket.sv_set(some_data)

    # inside sv_set
        node = node_from_socket(self)
        if self.is_linked or node.force_sv_set:
            SvSetSocket(self, data)
```

ly29:
From a practical standpoint it doesn't really matter if is linked or not unless you can avoid some large computation. So unless you do that it doesn't matter, you do save some memory but for most cases I think it is fine.

We could have lower level check and just drop data silently.

```python

def SvSetSocket(socket, out):
    global socket_data_cache
     if DEBUG_MODE:
    if not socket.is_output:
        print("Warning, {} setting input socket: {}".format(socket.node.name, socket.name))

    if not socket.is_linked:

        print("Warning: {} setting unconncted socket: {}".format(socket.node.name, socket.name))

    if not socket.is_linked:
        return

    s_id = socket_id(socket)
    s_ng = socket.id_data.name
    if s_ng not in socket_data_cache:
        socket_data_cache[s_ng] = {}
    socket_data_cache[s_ng][s_id] = out
```

Wireless communications are like gotos and global variables to me.

zeffii:
Ah that is a valid point, if the computation has already been done all that really happens during sv_set is putting a reference of that data in the cache. I think it's time to drop `is_linked` from my routines - will be cleaner anyway.

ly29:
So I want to do some of the proposed changes  from above as well as a few fixes. 
Hopefully tomorrow or Monday

zeffii:
I have not tested this branch yet because i'm doing the ob3 stuff, but want to push ob3 to get it out of the way and test this

zeffii:
testing now

zeffii:
```
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\data_structure.py", line 93
    """return matched list, using the last value to fill lists as needed
    longest list matching [[1,2,3,4,5], [10,11]] -> [[1,2,3,4,5], [10,11,11,11,11]]
    """"
```

zeffii:
@ly29  you don't use a linter for python?

zeffii:
image
```
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\node_tree.py", line 265, in draw
    layout.label(t + '. ' + SvGetSocketInfo(self))
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\core\socket_data.py", line 52, in SvGetSocketInfo
    s_id = socket_id(socket.other)
NameError: name 'socket_id' is not defined
```

ly29:
It had  some trouble with annotations and attitude so I disabled it. 
But I will enable it again. There are some other problems as well but strange that didn't trigger for me.

zeffii:
there's a few more, but i hesitate to push commits to your branch because it's easier for you to fix them..

zeffii:
```diff
diff --git a/core/socket_data.py b/core/socket_data.py
index 616d23b..39a44ee 100644
--- a/core/socket_data.py
+++ b/core/socket_data.py
@@ -16,6 +16,9 @@
 #
 # ##### END GPL LICENSE BLOCK #####

+from sverchok.data_structure import socket_id
+
+
 #####################################
 # socket data cache                 #
 #####################################
diff --git a/core/update_system.py b/core/update_system.py
index 1d971ae..c74e625 100644
--- a/core/update_system.py
+++ b/core/update_system.py
@@ -23,7 +23,7 @@ import bpy
 from mathutils import Vector

 from sverchok import data_structure
-from sverchok.core.socket_data import SvNoDataError
:...skipping...
diff --git a/core/socket_data.py b/core/socket_data.py
index 616d23b..39a44ee 100644
--- a/core/socket_data.py
+++ b/core/socket_data.py
@@ -16,6 +16,9 @@
 #
 # ##### END GPL LICENSE BLOCK #####

+from sverchok.data_structure import socket_id
+
+
 #####################################
 # socket data cache                 #
 #####################################
diff --git a/core/update_system.py b/core/update_system.py
index 1d971ae..c74e625 100644
--- a/core/update_system.py
+++ b/core/update_system.py
@@ -23,7 +23,7 @@ import bpy
 from mathutils import Vector

 from sverchok import data_structure
-from sverchok.core.socket_data import SvNoDataError
+from sverchok.core.socket_data import SvNoDataError, reset_socket_cache
 import sverchok

 import traceback
@@ -367,7 +367,7 @@ def build_update_list(ng=None):
         out = [make_update_list(ng, s, deps) for s in node_sets]
         update_cache[ng.name] = out
         partial_update_cache[ng.name] = {}
-        data_structure.reset_socket_cache(ng)
+        reset_socket_cache(ng)


 def process_to_node(node):
diff --git a/data_structure.py b/data_structure.py
index 4637419..cef85b3 100755
--- a/data_structure.py
+++ b/data_structure.py
@@ -90,7 +90,7 @@ def repeat_last(lst):
 def match_long_repeat(lsts):
     """return matched list, using the last value to fill lists as needed
     longest list matching [[1,2,3,4,5], [10,11]] -> [[1,2,3,4,5], [10,11,11,11,11]]
-    """"
+    """
     max_l = 0
     tmp = []
:
```

zeffii:
wait. that diff doesn't include everything

zeffii:
[ugit.zip](https://github.com/nortikin/sverchok/files/691466/ugit.zip)

zeffii:
image

zeffii:
rudimentary testing seems OK if those changes are applied. I 'd approve immediately..

ly29:
Also needed, 
```python
import sverchok.data_structure

# change
DEBUG_MODE
# to
data_structure.DEBUG_MODE
```

zeffii:
maybe DEBUG_MODE should be in update system, and perhaps should be done with logging module (going from bool to n-levels of debug)

zeffii:
however, debug mode is something I almost never use..

ly29:
allow me to dream an say that the top level module would be a good place, also making it more useful would be good. Using the logging module and defining what it does...
```
sverchok.DEBUG_MODE
```

zeffii:
can be broken up further if we really want..

ly29:
Now I think it is in order.

zeffii:
I think calling the function `replace_socket` is a little redundant in the name department, when writing `self.inputs[0].replace_socket(.....)`. Not a big deal, just `inputs[0]` would already indicated that the function is  going to have relevance to a socket.

```python
    def replace(self, new_type, new_name=None, new_pos=None):
        replace_socket(self, new_type, new_name, new_pos=new_pos or self.index)
```

zeffii:
i updated my previous comment.

ly29:
There is move function in the inputs/outputs. To many options exposed in that case perhaps

ly29:
Since I found some small bugs I will give it a little bit of extra review

PR #1103, opened 2017-01-09
Alt replace socket

zeffii:


zeffii:
i'll sleep on this. and push tomorrow If i don't wake up in a cold sweat.

ly29:
As you noted before changing the name of the socket method to replace would probably be sane.

zeffii:
OK, good, relatively painless modification...

zeffii:
I will leave it as `.replace_socket` now.

PR #1104, opened 2017-01-10
Handler update

ly29:
So this seems to be working quite nicely without any problems.

I am sure there are some however, since this changes how edit changes are handled.

@zeffii 
Would be nice if you could test if this resolves the crash you experienced.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -45,7 +45,7 @@ bl_info = {
         "AgustinJB, Zeffii, Kosvor, "
         "Portnov, Elfnor"
     ),
-    "version": (0, 5, 8, 4),
+    "version": (0, 5, 9, 0),
     "blender": (2, 7, 8),
     "location": "Nodes > CustomNodesTree > Add user nodes",
     "description": "Parametric node-based geometry programming",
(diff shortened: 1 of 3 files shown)


ly29:
I thinks is pretty much good to go, but would like to hear a second opinion.

zeffii:
I aim to test after breakfast.

zeffii:
think you forgot to push a few commits
```python
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\node_tree.py", line 387
    if not self.is_frozen()
                          ^
SyntaxError: invalid syntax
BCPrompt: operators registered
```

ly29:
Indeed I did, getting sloppy.

ly29:
Turned on linter now btw.

zeffii:
A fresh connection to bmeshviewer misses an update , reproduce

- make snlite (petal_sine), 
- connect verts to bmeshviewer,   (obj is created)
- now connect edges,  and the bmviewer doesn't generated edges until it is triggered to update from petal sine.

I'm not sure if this was always the behaviour, but the same is not the case for vdmk2

ly29:
Switching to pre seems to fix that for me.

ly29:
Many small things will likely pop up, sadly, but avoiding crashes and then try squash other changes of behaviour seems like the sane way forward for me.

zeffii:
not worried about the many smaller things that might arise. The ability to use a new Blender build is worth it :)  

Thanks!

PR #1106, opened 2017-01-10
vmath mk2

zeffii:
- This removes noise options from vector math node (these should be their own node anyway. in noisemk2)
- adds vector ui element and scalar ui element when sockets are unconnected (yay@!!)
- drastically cuts the loc to about 170 from 320. (means easier to think about and more welcoming for new contributors)


zeffii:
\o/  i think this solves all my problems with this node.

PR #1119, opened 2017-01-12
maintenance p3 removing old svGet/Set

zeffii:


zeffii:
that's enough for today.

PR #1133, opened 2017-01-16
Skin view mk1 b

zeffii:


zeffii:
am inclined to push this as the same python file name, because mk1 was only released 2 days ago.. literally only a handful of .blend files will contain it.

PR #1134, opened 2017-01-16
Skin view mk1 b (ui updates)

zeffii:


zeffii:
any  updates to this version will be in draw_buttons only, and be purely cosmetic. Any other kind of change shall be a MK2.

PR #1146, opened 2017-01-17
Line mk2

zeffii:


zeffii:
more likely to be useful with a mode switch, 

 - one that expects a list of lists containing various step distances  `[[1.2, 2.2],[3.2, 1.2]]`
 - one that just expects a single list of step widths  `[[2.0, 2.3, 4.3]]

I don't think i'd ever need to make my setup so complicated as to need both modes mixed,

nortikin:
than makeline put to geom.py file ???

nortikin:
than series of lines - from point to point, from point to direction, ortho

enzyme69:
So... Line MK2 will not happen? I'll use List Split in the mean time. [[ []
[] [] ]]

zeffii:
i thought about it more, and while it makes sense to have an axis picker for all generators (where it makes sense) , a simpler solution for now is a single rewire node.

image

it solved a more general problem.

zeffii:
image

enzyme69:
That's cool, I like it!

zeffii:
also making it so you can send 0 to X,Y,or Z  (0 or -1, 1 )

PR #1151, opened 2017-01-18
Stethoscope mk2

zeffii:
not for pushing .


nortikin:
would it be view by element depending on level? or only objects?

nortikin:
and also would it be graphical branches view?

zeffii:
@nortikin  i want it to be able to zoom into all kinds of nested-ness 

so 

- show number of elements in top level  ( .sv_get() )
- allow each element to be further selected  ( .sv_get()[n] )
    - of these show number of elements in this level  ....etc.
    - show all or selection , or high level  
    -   maybe     (not decided yet)
                - ( " list contains n items, each is len 3 " )
                - ( " list contains n items, x are len 3, y are len 4 " )

like nodes ListItem, ListSlice and Sthethoscope combined

zeffii:
i'd like to promote this node already in its current state, in replacement of current stethoscope. I hope to do this lateron today, and remove stethoscope 1 to old_nodes.

I will keep updating `scopeMK2`, it's already more useful.

Issue #1152, opened 2017-01-18
More places to revisit recalc_face_normals


zeffii:
update by @ly29

The following nodes, but this isn't as easy just changing, I think for example volume then recalc face normals might make more sense (or not):

pass 1  [committed](https://github.com/nortikin/sverchok/tree/recalc_normals)

- [x] polygons adaptive :octocat: 🉑 
- [x] solidfy  :octocat: 🉑 

pass 2 https://github.com/nortikin/sverchok/tree/normals_pass2

- [x] holes_fill  :octocat: 🉑 
- [x] offset :octocat: 🉑 
- [x] volume  :octocat: 🉑 
- [x] recalc_normals :octocat: 🉑 
- [x] vertex/normal :octocat:  🉑 
- [x] wire frame :octocat: 🉑

ly29:
Yeah some of them more urgent than others.
Recalc was probably the wrong option...

zeffii:
offset probably is next..

ly29:
Polygons adaptive also

zeffii:
ahhhyes! polygons adaptive, that's the one we've had issues with defo.

ly29:
So moved the code to ```bmesh_from_pydata``` but didn't work out in adaptive polygons by default.

zeffii:
I think i'd prefer this instead
```python
if faces:
        add_face = bm.faces.new
        if face_normals:
            for face in faces:
                add_face(tuple(bm.verts[i] for i in face)).normal_update()
        else:
            for face in faces:
                add_face(tuple(bm.verts[i] for i in face))
```
more code to read, but less instructions to compute

ly29:
Yeah that works out as well.

zeffii:
but don't change that unless add_face is magnitudes slower than testing `if face_normals`  : )

ly29:
Nothing more will happen with this today from me anyway. Feel free to add things.

zeffii:
yeah, not today for me either.

ly29:
Actually some of these might need vertex normals which I guess also needs to be updated in a similar way after the faces...

ly29:
```
if faces:
        add_face = bm.faces.new
        if face_normals:
            for face in faces:
                add_face(tuple(bm.verts[i] for i in face)).normal_update()
        else:
            for face in faces:
                add_face(tuple(bm.verts[i] for i in face))
        if face_normals and vertex_normal:
            for i in range(len(bm.verts)):
[...]

PR #1160, opened 2017-01-20
Recalc normals

ly29:
No really ready for merge perhaps but makes polygon adaptive better in certain scenarios and introduces the new ```bmesh_from_pydata``` interface #1152 

Please test.

### DIFF
--- a/nodes/modifier_make/polygons_adaptative.py
+++ b/nodes/modifier_make/polygons_adaptative.py
@@ -83,7 +83,7 @@ class AdaptivePolsNode(bpy.types.Node, SverchCustomTreeNode):
             versD = Vector_generate(versD_)
             ##### it is needed for normals of vertices
             polsR, polsD, versD = match_long_repeat([polsR, polsD, versD])
-            bm = bmesh_from_pydata(versR, [], polsR, face_normals=True)
+            bm = bmesh_from_pydata(versR, [], polsR, normals=True)
 
             #bmesh.ops.recalc_face_normals(bm, faces=bm.faces[:])
             bm.verts.ensure_lookup_table()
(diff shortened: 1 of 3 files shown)


zeffii:
https://github.com/nortikin/sverchok/blob/recalc_normals/nodes/modifier_make/solidify.py#L39  can be removed :)

I was thinking **(but totally not planning to do it myself, or expecting you to)**  but there's a point to be made to also offer to mode of normal calculation for bmesh_from_pydata, in the same call.

```
   def bmesh_from_pydata(verts, edges, faces, normals='', mode=''):
       # where   mode options are 'fast' / 'correct'
```

zeffii:
according to x https://twitter.com/raymondh/status/822296239156920321
```python
bmesh_from_pydata(verts=[], edges=[], faces=[], normals=""):
```
we should be writing

```python
    bmesh_from_pydata(verts=None, edges=None, faces=None, normals=""):
```

ly29:
The same for normal or it is okay since string isn't mutable?

zeffii:
yeah, that's ok i think.. but will dig around.

zeffii:
i don't know about `normals=''` , the language some of these guys use to discuss what _is_ or what _is not_ allowed is a little presumptive regarding the scope of my vocabulary.

zeffii:
but this kind of suggests we ought to avoid it too.  https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/

ly29:
Ok, that is something that we do in many places actually.  Oops

zeffii:
mutable types not ok, immutable types fine :)

zeffii:
not allowing a string as default param would be a crippling python feature. : )

ly29:
Not my favourite python feature... This can lead to so many bugs...

```python
>>> def f(d={}):
...     print(d)
...     d['a']=True
...     
>>> f()
{}

>>> f()
{'a': True}
```

ly29:
Also I guess verts shouldn't have a default argument since it is needed...

PR #1178, opened 2017-01-25
Normals pass2

zeffii:
- removed option fast/accurate for normal recal in solidfy, it is now correct + fast by default :)
- any node that was formerly doing recalc_normals now does a `bm.normal_update()` instead.
- removed the extra logic from `sv_bmesh_utils/  bmesh_from_pydata`
- removed the `[:]` from `sv_bmesh_utils/ pydata_from_bmesh` , not sure how that got there..?


zeffii:
https://github.com/nortikin/sverchok/issues/1152

almost ready to push this, have tested these nodes with basic stuff.

zeffii:
The only node i'm not confident about is the (surprisingly ) "Recalc Normals"  node,  please tell me if this produces the same output as before.

Should this be called ' Flip Normals ' node?

PR #1182, opened 2017-01-27
add fractal node

zeffii:


zeffii:
instead of calling a function to call another function once, i might be inclinded to write a wrapper function that takes a list of verts instead - in the end this would result in less overhead due to less total number of function calls.

kalwalt:
Do you receive an output from 'fractal'? i receive only zeros

kalwalt:
With voronoi 'noise_type' works maybe should be set some values correctly,

zeffii:
image

kalwalt:
We have Fractals inside sverchok! it is only a question of right settings...
schermata del 2017-01-27 12-23-55

kalwalt:
But does make it sense to name it `Vector Fractal`?

zeffii:
image

the input is a Vector.. but other than that I have no opinion about what to name it..

zeffii:
maybe `Fractal from Vector` , and place it in the Numbers subheading (If you write docs for it... else it stays in beta forever )

zeffii:
note that the socket / slider enable setting only kicks if the node's connected...(else `process` isn't called, and the greying out doesn't happen)   so we have to wrap the updateNode function

kalwalt:
Looks very good now!

> maybe Fractal from Vector , and place it in the Numbers subheading (If you write docs for it... else it stays in beta forever )

yes, maybe it's more adequate. But there are also from mathutils.noise `noise.turbulence noise.cell `to be implemented, maybe all this noise should all stay together?
I can write the docs, i will be honored to do.

zeffii:
My preference is to group things as much as possible, but not grouped to the extreme where the entire noise module needs to be worked into a node.

- with Vector Noise MK2 we have this seed socket because the underlying functions respond to seed (and we can take the return as a Vector or Scalar) . In effect the node stays relatively simple because it has limited scope.
- with Vector Fractal you have collected those functions that have similar signatures and output only Floats, the internals of the node stay simple. 

The problem I see with adding the other noise functions to either of those is simply that the internal working of the node become more complicated than is comfortable. (We already have crazy complicated nodes... this should not be one of them)

You are welcome to code whatever you want and do a pull request, if it looks good (code) we'll judge it on that.

kalwalt:
> The problem I see with adding the other noise functions to either of those is simply that the internal working of the node become more complicated than is comfortable. (We already have crazy complicated nodes... this should not be one of them)

Sorry, i express my thinking in a wrong way. what i would say is that there are also turbulence and other function that could be implemented in other separate nodes. I was refering only to group them all the noise nodes in the layout tree in such a way is easy to find them.  I agree with you that is better to let things simpler as possible.

> You are welcome to code whatever you want and do a pull request, if it looks good (code) we'll judge it on that.

thank's, it's a good practice for me!

zeffii:
@kalwalt  if you take a random sampling of our 190+ nodes you'll be able to judge if we are pragmatic or dogmatic with our own code style, I think we favour practical over pure. This means you'll see some weird code and some nice code, and sometimes a mix.

the question of where to best put the noise nodes is another matter.. one i've not put much brain power into.

PR #1186, opened 2017-01-28
make vector lerp node

zeffii:


nortikin:
than evaluate line has to rip in old_nodes?

zeffii:
I don't fully understand why VectorEvaluate is coded that way, but VectorLerp is not hugely different. I could give Lerp a mode to make it behave like Evaluate, like a tickbox or something -- Then we could remove VectorEvaluate node.

zeffii:
@nortikin  i think line_evaluate can now go to old_nodes :) . This node now has functionality of Both :) Will write docs asap.

zeffii:
@ly29 if you have time, check the code, essentially it should behave the same. not sure about speed..

ly29:
Will check tomorrow.

zeffii:
### Vector Lerp

A linear interpolation between two vectors, or streams of vectors.
If we have two vectors A and B, and a factor 0.5, then the output of the node will be exactly half way on the imaginary finite-line between A and B. Values beyond 1.0 or lower than 0.0 will be extrapolated to beyond the line A-B.

image 1   0.5
image 2   -0.5
image 3  1.5
image 4  [-1, 0, 1, 2]
image 5 evaluate mode vs lerp mode

zeffii:
I may have to change the mode selection to an Enum, like  `| Lerp | Evaluate |`

Issue #1185, opened 2017-01-28
reroute socket doesn't handle certain socket inputs /  outputs 


zeffii:
```
location: <unknown location>:-1
Traceback (most recent call last):
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\node_tree.py", line 279, in draw
    layout.label(t + '. ' + SvGetSocketInfo(self))
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\core\socket_data.py", line 54, in SvGetSocketInfo
    s_id = socket.other.socket_id
AttributeError: 'NoneType' object has no attribute 'socket_id'
```
try  
```
Sphere.verts -> Viz.verts
Sphere.edges -> Reroute -> Viz.edges
```

ly29:
So ```get_other_socket``` fails somehow.

https://github.com/nortikin/sverchok/blob/master/data_structure.py#L523-L543

Looking into view_draw I see the following that won't work, we should use ```socket.other``` here.

```python
        # this should catch accidental connections which otherwise will cause
        # an unrecoverable crash. It might even be an idea to have step in between
        # new connections and processing, it could auto rewire s->s v->v m->m.
        def check_origin(to_socket, socket_type):
            origin_socket_bl_idname = inputs[to_socket].links[0].from_socket.bl_idname

            if isinstance(socket_type, str):
                return origin_socket_bl_idname == sock_dict.get(socket_type)
            else:
                return reverse_sock_dict.get(origin_socket_bl_idname) in socket_type
```
https://github.com/nortikin/sverchok/blob/master/nodes/viz/viewer_mk2.py#L369-L378

ly29:
This shouldn't cause the problem anyway...

ly29:
https://github.com/nortikin/sverchok/tree/view_draw_fix

ly29:
Okay that should fix the issues I think, please test

zeffii:
cool. I hadn't looked at the code yet. will test ASAP.

zeffii:
tested fine. please commit

PR #1187, opened 2017-01-28
View draw fix

ly29:
Should handle a case during reconnection where an linked socket don't have an socket at the other end of the link.
#1185


zeffii:
yes, seems fine. commit !

Issue #1095, opened 2017-01-07
Vector Noise and Seed


enzyme69:
Random Num Gen and Random Vector has handy _seed_ input, sometimes I think Vector Noise needs that Seed value?

It makes me wonder if Vector Noise actually understand vectorization yet or not... 

From my understanding currently Vector Noise is like a function node, it takes Vector position at different place and will generate random value anyway, and moving the Vector point inputs will push the point through vector to generate noise as scalar or vector outputs.

enzyme69:
Also looking into this old issue:
https://github.com/nortikin/sverchok/issues/949

I'll attempt to do similar thing, but in a different set of mind.

zeffii:
have a look at the Vector Noise's code.

enzyme69:
Thanks for reminder, I am gonna look now.

zeffii:
There's a lot of potential in the noise module, I think the Vector Noise node barely scratches the surface. Excellent opportunity for someone who wants to learn bpy / writing nodes to get stuck in.

enzyme69:
https://github.com/nortikin/sverchok/blob/d37532f5777292a12872e4e8115848d6371b85fc/nodes/vertex/noise.py

zeffii:
anyway, there does appeart to be a seed property for the noise module

https://www.blender.org/api/blender_python_api_current/mathutils.noise.html

### mathutils.noise.seed_set(seed)

- Sets the random seed used for random_unit_vector, random_vector and random.
- Parameters:	seed (Int) – Seed used for the random generator. When seed is zero, the current time will be used instead.

not sure if that `random_vector` means `mathutils.noise.noise_vector`, if so the bpy docs are outdated

zeffii:
a simple test snlite ..
```python
"""
in   seed        s  d=0 n=2
in   inputvecs   v  d=[[]] n=1
out  noisevecs   v
"""
out = [[]]

import mathutils
from mathutils import noise

PERLIN = noise.types.STDPERLIN
noise.seed_set(seed)
out[0] = [noise.noise_vector(v, noise_basis=PERLIN) for v in inputvecs]
```

zeffii:
a simple snlite noise vector test confirms this.
```python
"""
in   seed        s  d=0 n=2
in   inputvecs   v  d=[] n=1
out  noisevecs   v
"""
noisevecs = [[]]

import mathutils
from mathutils import noise
from mathutils import Vector as V

PERLIN = noise.types.STDPERLIN
[...]

Issue #105, opened 2014-04-26
@vectorize decorator (one pass fits all)


zeffii:
As this beast grows, we learn that it has needs. The dilemma we face I think @nortikin and @ly29 agree, is it could be less complicated to write a function for a Node that can take any nested/unnested input and produce nested/unnested output.

``` text
input(s) -> ( some_function ) -> output(s)

input(s) -> (@vectorized some_function) -> outputs(s)
[input(s), inputs(s)] -> (@vectorized some_function) -> [output(s), outputs(s)]
```

My knowledge and experience with decorators is limited, but this seems like exactly what they are for.

PR #1188, opened 2017-01-28
Turbulence Node

kalwalt:
I started creating a Turbulence Node, it's not finished (again some troubles with verts and lists :disappointed: ) and also i didn't pulled the latest from the origin/upstream and so aren't in sync, Tomorrow i will fix this.
Thanks!

Walter

noise docs: [mathutils.noise.html](https://docs.blender.org/api/blender_python_api_current/mathutils.noise.html?highlight=turbulence#mathutils.noise.turbulence)


zeffii:
stick a Range Float ( count 2) into a Circle Node's Radius socket, and hook that into your Turbulance node, then stick a print statement directly under 
```
    verts = inputs['Vertices'].sv_get()
    print(verts)
    print(len(verts))
    for i in verts:
        print(i)
```

zeffii:
it's best you figure it out :) this is the hardest part of node code. coming to terms with the levels of incoming lists (nestedness)

kalwalt:
I think that is working, but i receive only zero for noise_basis=0 or =1 or =2, i receive non zero values if i choose noise_basis=3. I think it is a matter of right settings. And yes i agree withyou  this is the hardet part of coding. I use also experimenting with the Python IDLE to figure out how lists works.

to prove the turbulence issue i also coded this little script:

```
from mathutils import noise

def sv_main( verts=[], octaves=3, hard=1, noise_basis=1, amp=0.5, freq=2.0 ):

    data = []

    in_sockets = [
        ['v','Vector', verts],
        ['s', 'octaves', octaves],
        ['s', 'Hard ', hard],
        ['s', 'Noise Basis', noise_basis],
        ['s', 'Amplitude', amp],
        ['s', 'Frequency', freq]
        
    ]

    out_sockets = [
       ['s', 'Float Data', [data]]
    ]

    if verts and verts[0]:    
        for v in verts[0]:
             out = noise.turbulence(v, octaves, hard, noise_basis, amp, freq )
             data.append(out)
             print(data)
                     
    return in_sockets, out_sockets
```

zeffii:
have a look at the turbulence source in Blender,

zeffii:
looks like the other noise_basis work fine here ( i ran the node script )

zeffii:
image

kalwalt:
yes the right settings make the magic:

turbulencenode

i need to fix this line now:
`out.append([turbulence_function(v, octaves, hard, _noise_type, amp, freq) for v in vertices])`

verts are ok but other args are packed in lists.

zeffii:
please struggle with this for a while, and dig through the other nodes to see what patterns we've used.

remember 
- if you connect a number node to a socket    
      - self.inputs['my_socket'].sv_get() will return `[[your_number]]`
- if you connect a range node (for example making 0,1,2,3,4,5,6)
      - self.inputs['my_socket'].sv_get() will return `[[0,1,2,3,4,5,6]]

do a lot of prints and some sequences to get an understanding of what the data is like.

kalwalt:
yes that was the problem!

kalwalt:
In the last commit i cleaned the code but i have deleted something or there is some indentation error that i can't find, can check someone? anyway should be ok.
Thank's

zeffii:
why is this function wrapped?
```
def turbulence( verts, octaves, hard, _noise_type, amp, freq ):
    return noise.turbulence( verts, octaves, hard, _noise_type, amp, freq )
```

kalwalt:
better this? 

`turbulence_f = {'SCALAR': noise.turbulence, 'VECTOR': noise.turbulence_vector}`

and deleting 

`def turbulence( verts, octaves, hard, _noise_type, amp, freq ):
    return noise.turbulence( verts, octaves, hard, _noise_type, amp, freq )`

zeffii:
if the func signature match, no need to wrap.

zeffii:
also remove unused imports

zeffii:
notice you are defining `vertices` twice here, once outside the loop, once as the `element` of the for loop
```python
        vertices = []  # <--- has no effect
        for vertices in verts[:]:
            out.append([turbulence_function(v, m_octaves, m_hard, _noise_type, m_amp, m_freq) for v in vertices])
                                
```

zeffii:
also notice that `verts` will be usually a list wrapping a list of vertices. `[v_list]`.

`verts[:]` just makes a copy of whatever `verts` is. I think for testing you might want to still do something like :  

```
if verts and verts[0]:
    for vertices in verts[0]:
        out.append( ......  )
```

zeffii:
we do list wrapping to allow us to pass multiple collections of verts to a node. That's what the little numbers beside the socket indicated, when connected.
image

kalwalt:
I think i have to do a lot of experiments with the lists and other simpler nodes and code until really understand and increase my familiarity with this aspect of programming...

index.md
```diff
@@ -151,8 +151,9 @@
     SvInterpolationNodeMK2
     SvInterpolationNodeMK3
     SvVertSortNode
-    SvNoiseNodeMK2
+    SvNoiseNodeMK
```
zeffii: MK2

nodes/vector/turbulence.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+        if verts and verts[0]:
+            for vertices in verts:
+                out.append([turbulence_function(v, m_octaves, m_hard, _noise_type, m_amp, m_freq) for v in vertices])
+
+        if 'Noise V' in outputs:
+            outputs['Noise V'].sv_set(Vector_degenerate(out))
+        else:
+            outputs['Noise S'].sv_set(out)
```
zeffii:
```python
if 'Noise V' in outputs:
    outputs['Noise V'].sv_set(Vector_degenerate(out))
else:
    outputs['Noise S'].sv_set(out)
```
consider the alternative
```python
if 'Noise V' in outputs:
    out = Vector_degenerate(out)
outputs[0].sv_set(out)
```

index.md
kalwalt: Sorry,i will fix it!

nodes/vector/turbulence.py
kalwalt: Tested! it does the same.

nodes/vector/turbulence.py
```diff
@@ -0,0 +1,135 @@
...
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software Foundation,
+#  Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# ##### END GPL LICENSE BLOCK #####
+
+import inspect
+import operator
```
zeffii: these aren't needed

kalwalt:
I would add an example image to illustate the node (a screenshot) but how can i attach it in a .rst file? shoul d I upload it before in github and then copy thr link in the doc? Excuse my ignorance but i never did it before.

kalwalt:
vector_fractal_example
image for fractal.rst example. i will post here the images and the i will create links in the files.

zeffii:
https://github.com/nortikin/sverchok/issues/394

kalwalt:
Cellualr node almost ready!
scrren_cellularnode

kalwalt:
I think cellular name is confusing a bit... If we want a cellular shape we shoyld use the Voronoi F1, and so the node name should be changed simply in Cell node.

kalwalt:
@zeffii @ly29 I'm trying ot had a random seed to the turbulence Node (and i will also do for fractal and cel Node) as discussed in this issue https://github.com/nortikin/sverchok/issues/1190, i looked at the ant_landscape addon also this is what i done so far:

````python

        if verts and verts[0]:
            for i,vertices in enumerate(verts[0]):
                print('verts: {0}'.format(verts))
                print(vertices)
                print(vertices[0])
                coords = []
                print('index: {0}'.format(i))
                # origin
                if rseed == 0:
                    origin = 0.0,0.0,0.0
                    origin_x = 0.0
                    origin_y = 0.0
                    origin_z = 0.0
                else:
                # randomise origin
                    noise.seed_set( rseed )
                    origin = noise.random_unit_vector(3)
                    origin_x = ( 0.5 - origin[0] ) * 1000.0
                    origin_y = ( 0.5 - origin[1] ) * 1000.0
                    origin_z = ( 0.5 - origin[2] ) * 1000.

                coords.append((vertices[i] / m_octaves + origin_x, vertices[i] / m_octaves + origin_y, vertices[i] / m_octaves + origin_z))
                print('coords: {0}'.format(coords))
                out.append([turbulence_function(v, m_octaves, m_hard, _noise_type, m_amp, m_freq) for v in coords])

````

the problem is that i get an error and 'coords' values are always the same for every vertices. See the log

````
verts: [[(0.0, 0.0, 0.0), (1.0, 0.0, 0.0), (0.0, 1.0, 0.0), (1.0, 1.0, 0.0)]]
(0.0, 0.0, 0.0)
0.0
index: 0
coords: [(66.15439057350159, 1394.7445154190063, 394.1212296485901)]
verts: [[(0.0, 0.0, 0.0), (1.0, 0.0, 0.0), (0.0, 1.0, 0.0), (1.0, 1.0, 0.0)]]
(1.0, 0.0, 0.0)
1.0
index: 1
coords: [(66.15439057350159, 1394.7445154190063, 394.1212296485901)]
verts: [[(0.0, 0.0, 0.0), (1.0, 0.0, 0.0), (0.0, 1.0, 0.0), (1.0, 1.0, 0.0)]]
(0.0, 1.0, 0.0)
0.0
index: 2
coords: [(66.15439057350159, 1394.7445154190063, 394.1212296485901)]
verts: [[(0.0, 0.0, 0.0), (1.0, 0.0, 0.0), (0.0, 1.0, 0.0), (1.0, 1.0, 0.0)]]
(1.0, 1.0, 0.0)
1.0
index: 3
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/core/update_system.py", line 322, in do_update_general
    node.process()
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/vector/turbulence.py", line 159, in process
    coords.append((vertices[i] / m_octaves + origin_x, vertices[i] / m_octaves + origin_y, vertices[i] / m_octaves + origin_z))
Node Vector Turbulence had exception tuple index out of range

````
At least random seed works (if i change rseed i get different values).

**Comment by ly29** [comment]
```
 coords.append((vertices[i] / m_octaves + origin_x, vertices[i] / m_octaves + origin_y, vertices[i] / m_octaves + origin_z))
```

This should probably be:
``` 
coords.append((vertices[0] / m_octaves + origin_x, vertices[1] / m_octaves + origin_y, vertices[2] / m_octaves + origin_z))
```

The main problem here is that ```vertices``` is badly named. if name it ```vertex``` or ```point``` I think it should become clear that you aren't referring to an array of vertices but a singular vertex.

**Comment by ly29** [comment]
The line afterward should probably be removed from the for loop as well.

**Comment by kalwalt** [comment]
> The main problem here is that vertices is badly named. if name it vertex or point I think it should become clear that you aren't referring to an array of vertices but a singular vertex.

@ly29 yes, i was refering to a single vertex more precisely the x y z component of the vertex. The hint you give me solve the error ( i did it before in that way but i thought wasn't correct) . Now the Node output values but i think the values are not complitely correct. Myabe the math i used is not correct. I need to do some test, and i will post the result.
Thank's!

File: nodes/vector/turbulence.py [hunk 4]
```diff
@@ -108,22 +110,49 @@ def process(self):
...
+                    #origin_x = ( 0.5 - origin[0] ) * 20.0
+                    #origin_y = ( 0.5 - origin[1] ) * 20.0
+                    #origin_z = ( 0.5 - origin[1] ) * 20.0
+                    origin_x = noise.random()  * 20.0
+                    origin_y = noise.random()  * 20.0
+                    origin_z = noise.random()  * 20.0
+
+                coords.append((vertices[0] / m_octaves + origin_x, vertices[1] / m_octaves + origin_y, vertices[2] / m_octaves + origin_z))
+                print('coords: {0}'.format(coords))
+                out.append([turbulence_function(v, m_octaves, m_hard, _noise_type, m_amp, m_freq) for v in coords])
```
ly29: I am quite sure this is a bug and should be outside of the for-loop

ly29: Since you are building up the ```coords``` but for every vertex append a new list of the ```coords``` you already have

kalwalt: i think the problem is here because putting out origin, give me the same result: a flat plane instead of a noisy-turbulence surface.

kalwalt:
turbulencenode-randomseed
i receive scalars but maybe not in the right range?

ly29:
```python

            for vertex in verts[0]:
                coords = []
                if rseed == 0:
                    origin = 0.0,0.0,0.0
                    origin_x = 0.0
                    origin_y = 0.0
                    origin_z = 0.0
                else:
                    noise.seed_set( rseed )

                    origin_x = noise.random()  * 20.0
                    origin_y = noise.random()  * 20.0
                    origin_z = noise.random()  * 20.0
                coords.append((vertex[0] + origin_x, vertex[1] + origin_y, vertex[2] + origin_z))
            # this MUST be outside of the forloop
            out.append([turbulence_function(v, m_octaves, m_hard, _noise_type, m_amp, m_freq) for v in coords])
```

ly29:
Also note ```coords``` should be created outside of the loop

kalwalt:
sorry @ly29  i misunderstood, i will fix it now!

kalwalt:
yes!! it seems to work! :smile: 
turbulencenode-randomseed_ok

kalwalt:
In your opininon better this:

````python
                    noise.seed_set( rseed )
                    
                    origin_x = noise.random()  * 20.0
                    origin_y = noise.random()  * 20.0
                    origin_z = noise.random()  * 20.0

                coords.append((vertex[0] + origin_x, vertex[1] + origin_y, vertex[2] + origin_z))

        
````

or this:

````python
                    noise.seed_set( rseed )
                    
                    origin = random_unit_vector()
                    origin_x = ( 0.5 - origin[0] ) * 1000.0
                    origin_y = ( 0.5 - origin[1] ) * 1000.0
                    origin_z = ( 0.5 - origin[2] ) * 1000.0


                coords.append((vertex[0]/m_octaves + origin_x, vertex[1]/m_octaves + origin_y, vertex[2]/m_octaves + origin_z))

````

**Comment by ly29** [comment]
Something like:
```
noise.seed_set(rseed)
offset = noise.random_unit_vector() * 10 # factor then is just random, to be sufficient distance away to make each seed different...
for vertex ....:
    new_vertex = [ x+y for x, y in zip(vertex, offset)]
```

**Comment by kalwalt** [comment]
In my opinion dividing for the octaves give the resulting noise(turbulence) as dilating, But in the ant_landscape addon there is also an offset values that has the effect to stretch the noise, very useful indeed. We could add it if we want, look:

```python
 # origin    
    if rseed == 0:
        origin = 0.0 + x_offset, 0.0 + y_offset, 0.0
        origin_x = x_offset
        origin_y = y_offset
        origin_z = 0.0
    else:
        # randomise origin
        seed_set(rseed)
        origin = random_unit_vector()
        origin[0] += x_offset
        origin[1] += y_offset
        origin_x = ((0.5 - origin[0]) * 1000.0) + x_offset
        origin_y = ((0.5 - origin[1]) * 1000.0) + y_offset
        origin_z = (0.5 - origin[2]) * 1000.0
```

nodes/vector/turbulence.py
```diff
@@ -120,27 +120,18 @@ def process(self):
...
             for vertex in verts[0]:
 
-                # origin
+                # set the offset origin for random seed
                 if rseed == 0:
-                    origin = 0.0,0.0,0.0
-                    origin_x = 0.0
-                    origin_y = 0.0
-                    origin_z = 0.0
+                #we set offset value to 0.0
```
zeffii: a comment directly under an `if` should be indented by 4 spaces.. like the next line

nodes/vector/turbulence.py
```diff
@@ -120,27 +120,18 @@ def process(self):
...
+                    offset = [0.0,0.0,0.0]
                 else:
                 # randomise origin
```
zeffii: same here, indent it,

nodes/vector/turbulence.py
kalwalt: i will do it, thank's,

kalwalt:
@zeffii  Should I also compact this [line](https://github.com/kalwalt/sverchok/blob/30854ac1b4348ed439aaa072239936ae872bf8af/nodes/vector/turbulence.py#L81) and after? I saw you was refering about in another pull request.

zeffii:
> @zeffii Should I also compact this line and after?

We don't enforce the 80 char line-limit from pep8. In the Sverchok codebase you'll see the majority of properties span about 2 or 3 lines..  I start to get the mild shakes if there are 10+ properties with all arguments on a separate line, then grouping and compacting should be encouraged.
I'm OK with that being on one line.  or

```python
octaves = IntProperty(
    default=3, min=0, max=6, description='Octaves', name='Octaves', update=updateNode
)
```
don't worry about what I think is right, be consistent and you'll be fine.

kalwalt:
random seed with 0 values works fine in Cell and Turbulence node, without the issue [here](https://github.com/nortikin/sverchok/issues/1095#issuecomment-271261600)

kalwalt:
needs an image example for cell node

nodes/vector/turbulence.py
```diff
@@ -121,22 +132,24 @@ def process(self):
...
+                    noise.seed_set(rseed)
                     offset = noise.random_unit_vector() * 10
 
-                new_vertex = [ x+y for x, y in zip(vertex, offset)]
+                new_vertex = [x + y for x, y in zip(vertex, offset)]
                 coords.append(new_vertex)
 
-            out.append([turbulence_function(v, m_octaves, m_hard, _noise_type, m_amp, m_freq) for v in coords])
+            out.append([turbulence_function(v, m_octaves, m_hard,
+                                            _noise_type, m_amp, m_freq) for v in coords])
```
zeffii:
```diff
-            out.append([turbulence_function(v, m_octaves, m_hard,
-                                            _noise_type, m_amp, m_freq) for v in coords])
```
awkward vs
```diff
+            constant_params = m_octaves, m_hard, _noise_type, m_amp, m_freq
+            out.append([turbulence_function(v, *constant_params) for v in coords])

```

zeffii: unless you planned to make the params different per inputted set of coords too, (_fully vectorized_).

kalwalt: i had to say: more elegant!

kalwalt: I will change it later now back in the texture viewer stuff.

ly29:
I just have to ask what is the difference between the cell node and the cell mode in the vector noise node?

kalwalt:
> I just have to ask what is the difference between the cell node and the cell mode in the vector noise node?

There is little difference i think, i need to look in the source C++ implementation if there is a real difference, Not haved the time to look at, though.

ly29:
Okay but if there is any difference in the C wouldn't it be better to just add as an mode to the vector noise instead off another node?

zeffii:
interesting read: https://medium.com/coding-with-clarity/speeding-up-python-and-numpy-c-ing-the-way-3b9658ed78f4#.uteqyjdjg

ly29:
@zeffii 
interesting read, however a shame that he didn't investigate where the overhead was located. I think the python list -> np.array functions could be sped up a lot. it would be interesting to see the performance if the data was already in np format even if it probably didn't fit his problem domain.

kalwalt:
i read in the C code and i did ths simple test:

cell_noise-func-compared

my conclusion is that you can do the same thing with the fractal node with octaves=1

main difference:

- `cell node ` output can be scalar or vectors

- `fractal node` output is only scalar

zeffii:
@ly29 
> interesting read, however a shame that he didn't investigate where the overhead was located. 

yes!  i'm interested in doing some of this in C. using distutils (completely alien to me)

ly29:
And Cell node compared to vector noise node in cell mode?

kalwalt:
@ly29 they are compared in the previous picture, look this comparison of the scalar values:
cell-noidse comparison-values

scalar values from Fractal an Cell node are the same.
reminder:

 - `Fractal node` with Cell noise basis range: -1, 1

- `Vector Noise node` with Cell noise basis range: 0, 1 

- `Vector Cell node` range: -1, 1

this is Vector out:

cell-noise comparison-vector values

Vector values from Fractal an Cell node are Not the same. ( i know fractal does not output Vector, but maybe we can compare the .z value) 

i will do some practical test to see what we could do with this Cell node and comparing with the others.

ly29:
Test with a seed different to zero. 
Also if yes, why not a mode on the vector noise node if there is any significant difference.

kalwalt:
i would but fractal node has not the seed option at the moment, i should do something like for the new turbulence node...

kalwalt:
cell-noidse comparison-wiyh_color_red

cell-noidse comparison-wiyh_color_red _3d

simple test with color output only red channel

kalwalt:
cell-noise comparison-with_mixed_colors
with all RGB channels

ly29:
So add it as a mode to the vector noise function.

kalwalt:
> So add it as a mode to the vector noise function.
@ly29  but this not will brake retro compatibility?

ly29:
No as long as you respect the identifier in enum

kalwalt:
> No as long as you respect the identifier in enum
ok i will proceed in this direction.

kalwalt:
turbulence-finaltest-comparison
Test comparison between my `Turbulence node` and the script with `turbulence function` It seems that my node does the same of this simple script:

````python

from mathutils import noise

def sv_main( verts=[], octaves=3, hard=1, noise_basis=1, amp=0.5, freq=2.0 ):

    data = []

    in_sockets = [
        ['v','Vector', verts],
        ['s', 'octaves', octaves],
        ['s', 'Hard ', hard],
        ['s', 'Noise Basis', noise_basis],
        ['s', 'Amplitude', amp],
        ['s', 'Frequency', freq]
        
    ]

    out_sockets = [
       ['s', 'Float Data', [data]]
    ]
    
    if hard == 0:
        hard = False
    else:
        hard = True
        
    if verts and verts[0]:    
        for v in verts[0]:
             out = noise.turbulence(v, octaves, hard, noise_basis, amp, freq )
             data.append(out)
                     
    return in_sockets, out_sockets
````
,but with my node we add random seed!

**Comment by enzyme69** [comment]
Is this Vector Turbulance node still coming soon?

**Comment by kalwalt** [comment]
@ly29 @zeffii may we think to merge this?

**Comment by zeffii** [comment]
I think it should be vectorized to accept multiple input vertex lists. You'll get that request immediately when it hits master.

**Comment by zeffii** [comment]
therefore, instead of just reading `vertex in verts[0]` you'd iterate over all incoming vert lists. and instead of `get()[0][0]` on the properties, you'd `get()[0]` and duplicate the found value if there aren't enough values to match the number of vertex_lists. 

if this doesn't make sense, have a look at Random Vector MK2

**Comment by kalwalt** [comment]
> therefore, instead of just reading vertex in verts[0] you'd iterate over all incoming vert lists. and instead of get()[0][0] on the properties, you'd get()[0] and duplicate the found value if there aren't enough values to match the number of vertex_lists.

> if this doesn't make sense, have a look at Random Vector MK2

Now it's clear, the comment  above less for me. I will look into Random Vector MK2.

**Comment by kalwalt** [comment]
@zeffii verts in properties is already:
`verts = inputs['Vertices'].sv_get()`

maybe you are refering to the other properties?

```python
        m_octaves = inputs['Octaves'].sv_get()[0][0]
        m_hard = inputs['Hard'].sv_get()[0][0]
        m_amp = inputs['Amplitude'].sv_get()[0][0]
        m_freq = inputs['Frequency'].sv_get()[0][0]
        rseed = inputs['Random seed'].sv_get()[0][0]
```

and the use `match_long_repeat` function?

zeffii:
yes, but you are always only taking the first set of verts by doing 
```
    if verts and verts[0]:
        for vertex in verts[0]:
            ...
```
there's a decorator to automatically `vectorize` stuff. let me find it..

zeffii:
hmmm..  `from sverchok.utils.geom import vectorize`  but that except the function being wrapped to accept kw arguments... i don't know if that helps you.. the turbulence function doesn't accept kw args, unless you wrap the function first... haha.. `wrapping^2`

zeffii:
we could make a non kw dependant version too..

kalwalt:
@zeffii kw args means keywords argument, right?  well, i can try to see if i can turn that function in a wrapped version., athough it appear to me something not very clear.

kalwalt:
@zeffii where i can found an example of this `vectorize` func? where is used? Do you remember? this will help me a lot...

kalwalt:
found this old issue: https://github.com/nortikin/sverchok/issues/105 to understand the concept

zeffii:
forgive me for not having much time to answer this now. it's harder to explain than do :)

kalwalt:
@zeffii ok, no problem. Anyway found different issues on this in the issue tracker. for example: https://github.com/nortikin/sverchok/pull/1209 even if not esplictly mention of that `vectorize` function.

zeffii:
@kalwalt  ..same advice then.. struggle a little, learn a lot.

kalwalt:
i will try with a simple script

kalwalt:
@zeffii I made this script, it doesn't work, but just to get the idea:

```python
from sverchok.utils.geom import vectorize

def sv_main(verts=[]):
    data=[]

    in_sockets = [
        ['v', 'verts', verts],
        ['s', 'factorX', factorX]
        ]
        

    out_sockets = [
        ['v', 'data', data]
    ]
    
    list_params = {
        'verts': 0, 'factorX':1
        }
        
    def addiction(list_params):
        print("i'm here")
        return verts + factorX + 2.0
    
    data = [vectorize(addiction(**args) for args in list_params] 

    return in_sockets, out_sockets

```

kalwalt:
Also i think fractal node is helpful

zeffii:
you should probably wrap the turbulence function anyway, to accept a list of verts.

there's several imlementations of a function called `vectorize`. 

1.  utils/snlite_utils.py  (vectorize, repeats all incoming data to match up, not a decorator)
2.  utils/geom.py         (this does decorate the uncoming function)

zeffii:
something crazy like (untested)
```python
    def process(self):
        inputs, outputs = self.inputs, self.outputs

        if not outputs[0].is_linked:
            return

        out = []
        verts = inputs['Vertices'].sv_get(deepcopy=False)

        m_octaves = inputs['Octaves'].sv_get()[0]
        m_hard = inputs['Hard'].sv_get()[0]
        m_amp = inputs['Amplitude'].sv_get()[0]
        m_freq = inputs['Frequency'].sv_get()[0]
        m_seed = inputs['Random seed'].sv_get()[0]

        @vectorize
        def wrapped_turb_func(verts=None, octaves=None, h=None, a=None, f=None):
            _noise_type = noise_dict[self.noise_type]
            return [turbulence_f[self.out_mode](v, m_octaves, h, _noise_type, a, f) for v in verts]

        def get_offset(seed):
            if seed == 0:
                offset = [0.0, 0.0, 0.0]
            else:
                noise.seed_set(seed)
                offset = noise.random_unit_vector() * 10.0
            return offset

        def adjust_verts_for_seed():
            new_verts = []
            if idx, vert_list in enumerate(verts):
                seed = m_seed[idx] if idx < len(m_seed) else m_seed[-1]

                if not seed == 0.0:
                    offset = get_offset(seed)
                    for vertex in vert_list:
                        new_vertex = [x + y for x, y in zip(vertex, offset)]
                        new_verts.append(new_vertex)
                else:
                    new_verts.append(vert_list)
            return new_verts

        new_verts = adjust_verts_for_seed()
        out.append(wrapped_turb_func(verts=new_verts, octaves=m_octaves, h=m_hard, a=m_amp, f=m_freq))

        if 'Noise V' in outputs:
            out = Vector_degenerate(out)
        outputs[0].sv_set(out)
```

kalwalt:
ah now i undertand the vectorize function !  my script was wrong!

zeffii:
there's several ways to do this, but the concept is always the same. the complexity here is that you are manually adding the seed/offset,

kalwalt:
i get this error:
`if idx, vert_list in enumerate(verts):
          ^
SyntaxError: invalid syntax
`

don't know why... maybe indent error?

kalwalt:
missed parenthesis

zeffii:
oh.. it should be 'for idx ....`

kalwalt:
yes it was that, i will push the new changes, but now the node output list of Vector, and see here:
vectorize_turbulence

zeffii:
ok, i think that means the output of the wrapped function is a generator.. (is evaluated lazy)
```diff
- ` out.append(wrapped_turb_func(ve`
+ out.append(list(wrapped_turb_func(ve
```

kalwalt:
better, but now i get only zeros from scalar output

zeffii:
no idea : )

zeffii:
```
  @vectorize
        def wrapped_turb_func(verts=None, octaves=None, h=None, a=None, f=None):
            _noise_type = noise_dict[self.noise_type]
            print('----', _noise_type)
            print(verts)
            print('args', octaves, h, a, f)
            print('function', turbulence_f[self.out_mode])
            return [turbulence_f[self.out_mode](v, octaves, h, _noise_type, a, f) for v in verts]
```
in that function add a debug print with something like
```python

**Comment by kalwalt** [comment]
it is related to nestedness, but i get semi-correct out for voronoi noise basis, The node also hang for random seed different from zero,

**Comment by zeffii** [comment]
maybe revert and figure out a more sane way to vectorize :)

btw
```python
new_vertex = [x + y for x, y in zip(vertex, offset)]
```
its faster to do it longform
```python
new_vertex = [v[0] + offset[0], v[1] + offset[1], v[2] + offset[2]]
```

kalwalt:
see here the out should one level less:

```
---- 0
[(0.0, 0.0, 0.0), (0.004999999888241291, 0.0, 0.0), (0.009999999776482582, 0.0, 0.0), (0.014999999664723873, 0.0, 0.0), (0.0, 0.004999999888241291, 0.0), (0.004999999888241291, 0.004999999888241291, 0.0), (0.009999999776482582, 0.004999999888241291, 0.0), (0.014999999664723873, 0.004999999888241291, 0.0), (0.0, 0.009999999776482582, 0.0), (0.004999999888241291, 0.009999999776482582, 0.0), (0.009999999776482582, 0.009999999776482582, 0.0), (0.014999999664723873, 0.009999999776482582, 0.0), (0.0, 0.014999999664723873, 0.0), (0.004999999888241291, 0.014999999664723873, 0.0), (0.009999999776482582, 0.014999999664723873, 0.0), (0.014999999664723873, 0.014999999664723873, 0.0)]
args 4 True 1.2000000476837158 0.029999999329447746
function <built-in function turbulence>
[[[0.0, 0.002267966279760003, 0.004461126867681742, 0.006580439396202564, 0.008359326049685478, 0.006090768147259951, 0.0038970881141722202, 0.0017774563748389482, 0.01691337488591671, 0.014643671922385693, 0.012448914349079132, 0.010327829048037529, 0.025653915479779243, 0.02338230609893799, 0.021185416728258133, 0.01906224898993969]]]

```

zeffii:
then something like
```
        @vectorize
        def wrapped_turb_func(verts=None, octaves=None, h=None, a=None, f=None):
            _noise_type = noise_dict[self.noise_type]
            values = [turbulence_f[self.out_mode](v, octaves, h, _noise_type, a, f) for v in verts]
            if self.noise_type == 'scalar':
                  values = values[0]
            return values
```
i don't know :)

zeffii:
i hope you have gained some appreciation for the effort that goes into this dense under-the-hood stuff.

kalwalt:
absolutely! at least for me it's a lot of effort..

zeffii:
but the hardest part is writing the non-vectorized version, making it vectorized is often much easier.

zeffii:
here there's two bits of extra complexity.  1. changing output vec/scalar and 2. the offset in the verts using seed because the function doesn't natively support seed.

kalwalt:
modified your tip into this:


```python
            values = [turbulence_f[self.out_mode](v, octaves, h, _noise_type, a, f) for v in verts]
            if self.out_mode == 'SCALAR':
                values = values[0]
            return values
```

but now i get out correct level nestedness but only 1 value (note i have a grid 64x64) kind of crazyness!

zeffii:
looks like scalar output needs to be `values[0][0]`  it's nested twice?

kalwalt:
i think the problem is in:
`out.append(list(wrapped_turb_func(verts=new_verts, octaves=m_octaves, h=m_hard, a=m_amp, f=m_freq)))`

zeffii:
an alternative.. ( i think i prefer this )

```python
    def process(self):
        inputs, outputs = self.inputs, self.outputs

        if not outputs[0].is_linked:
            return

        out = []
        verts = inputs['Vertices'].sv_get(deepcopy=False)

        m_octaves = inputs['Octaves'].sv_get()[0]
        m_hard = inputs['Hard'].sv_get()[0]
        m_amp = inputs['Amplitude'].sv_get()[0]
        m_freq = inputs['Frequency'].sv_get()[0]
        m_seed = inputs['Random seed'].sv_get()[0]

        max_len = len(verts)
        fullList(m_octaves, maxlen)
        fullList(m_hard, maxlen)
        fullList(m_amp, maxlen)
        fullList(m_freq, maxlen)
        fullList(m_seed, maxlen)


        def get_offset(seed):
            if seed == 0:
                offset = [0.0, 0.0, 0.0]
            else:
                noise.seed_set(seed)
                offset = noise.random_unit_vector() * 10.0
            return offset


        for idx, (args) in enumerate(verts, m_octaves, m_hard, m_amp, m_freq, m_seed):
            seed = args[5]
            vert_list = args[0]

            if not seed == 0.0:
                offset = get_offset(seed)
                vert_list = [[v[0] + offset[0], v[1] + offset[1], v[2] + offset[2]] for v in vert_list]:

            _noise_type = noise_dict[self.noise_type]
            octaves, hard, amp, freq = args[1:5]
            out.append([turbulence_f[self.out_mode](v, octaves, hard, _noise_type, amp, freq) for v in verts])


        if 'Noise V' in outputs:
            out = Vector_degenerate(out)
        outputs[0].sv_set(out)        
```

zeffii:
(update that last snippet)

kalwalt:
i get!
```
out.append(list(wrapped_turb_func(verts=new_verts, octaves=m_octaves, h=m_hard, a=m_amp, f=m_freq)))

        out = list(wrapped_turb_func(verts=new_verts, octaves=m_octaves, h=m_hard, a=m_amp, f=m_freq)) ## this will output [None]
```

zeffii:
i hope i'm not confusing you :)  it's dense stuff. take a walk :)

kalwalt:
we are close to solve it! i hope... :smile:

kalwalt:
without the vectorize function?

zeffii:
yeah do it by hand :)

kalwalt:
`fullList` is  from sverchok.data_structure, right?

kalwalt:
It works fine! i will commit as i clean the code, also have to test if everything is correct.

zeffii:
yep. many nodes use it :)

kalwalt:
i think it is enough for today, i will check again tomorrow, Many thank's @zeffii for the help!

zeffii:
test it.. then it can be accepted tomorrow. 

you are welcome @kalwalt . a learning experience all round :)

kalwalt:
@zeffii Yesterday i was wrong, it doesn't works. Again needs some fixes, soon i will post a reviewed commit in which i show what is the issue.

zeffii:
```python
        maxlen = len(verts)
        fullList(m_octaves, maxlen)
        fullList(m_hard, maxlen)
        fullList(m_amp, maxlen)
        fullList(m_freq, maxlen)
        fullList(m_seed, maxlen)

        ... print here ...

```
feed it 3 circles in verts (with 3 points each at different radius)
and print out the result of `m_octaves .... m_seed` , those need to be a list of 3 values each `[val1, val2, val3]`.

zeffii:
image

zeffii:
oh..  `enumerate(zip([ .....`

kalwalt:
the out in the console is: 
```
maxlen 3
m_octaves : [3, 3, 3]
m_hard : [True, True, True]
m_amp : [0.5, 0.5, 0.5]
m_freq : [2.0, 2.0, 2.0]
m_seed : [0, 0, 0]
from args[0]:  [(0.7199999690055847, 0.0, 0), (-0.36000002084375743, 0.6235382429013775, 0), (-0.3599999118208586, -0.6235383058457732, 0)]
```

i think the problem is in this line:

   ````python
for idx, (args) in enumerate([verts, m_octaves, m_hard, m_amp, m_freq, m_seed]):
````

**Comment by zeffii** [comment]
yep
```python
verts = [[(1.2300000190734863, 0.0, 0), (-0.6150000716192286, 1.065211227329641, 0), (-0.6149998853717658, -1.0652113348596566, 0)], [(3.1700000762939453, 0.0, 0), (-1.5850001981481765, 2.745300503692421, 0), (-1.5849997181445485, -2.745300780822626, 0)], [(5.110000133514404, 0.0, 0), (-2.5550003246771245, 4.425389780055201, 0), (-2.5549995509173313, -4.425390226785596, 0)]]
m_val1 = [1,2,3]
m_val2 = [23, 24, 25]

for idx, (args) in enumerate(zip(verts, m_val1, m_val2)):
    print('-----', idx)
    print(args[0])
    print(args[1])
    print(args[2])        
```

kalwalt:
yes exactly, i have some progress, now it hang at this line see:

```
maxlen 3
m_octaves : [3, 3, 3]
m_hard : [True, True, True]
m_amp : [0.5, 0.5, 0.5]
m_freq : [2.0, 2.0, 2.0]
m_seed : [0, 0, 0]
from args[0]:  [(0.7199999690055847, 0.0, 0), (-0.36000002084375743, 0.6235382429013775, 0), (-0.3599999118208586, -0.6235383058457732, 0)]
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/core/update_system.py", line 322, in do_update_general
    node.process()
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/vector/turbulence.py", line 162, in process
    out.append([turbulence_f[self.out_mode](v, octaves, hard, _noise_type, amp, freq) for v in verts])
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/vector/turbulence.py", line 162, in <listcomp>
    out.append([turbulence_f[self.out_mode](v, octaves, hard, _noise_type, amp, freq) for v in verts])
Node Vector Turbulence had exception turbulence: invalid 'position' arg: sequence index 2 expected a number, found 'tuple' type, 



 
```

zeffii:
because args is a little ugly, we tend to
```python
        arguments = verts, m_octaves, m_hard, m_amp, m_freq, m_seed
        for idx, (vert_list, octaves, hard, amp, freq, seed) in enumerate(zip(*arguments)):

            if not seed == 0.0:
                offset = get_offset(seed)
                vert_list = [[v[0] + offset[0], v[1] + offset[1], v[2] + offset[2]] for v in vert_list]:

            _noise_type = noise_dict[self.noise_type]
            out.append([turbulence_f[self.out_mode](v, octaves, hard, _noise_type, amp, freq) for v in vert_list])

```

zeffii:
or...
```python
        _noise_type = noise_dict[self.noise_type]
        arguments = verts, m_octaves, m_hard, m_amp, m_freq, m_seed
        for idx, (vert_list, octaves, hard, amp, freq, seed) in enumerate(zip(*arguments)):
            final_vert_list = vert_list
            if not seed == 0.0:
                offset = get_offset(seed)
                final_vert_list = [[v[0] + offset[0], v[1] + offset[1], v[2] + offset[2]] for v in vert_list]:

            out.append([turbulence_f[self.out_mode](v, octaves, hard, _noise_type, amp, freq) for v in final_vert_list])
```

kalwalt:
i will try the first version for now, thank's @zeffii !

zeffii:
```diff
- thank's
+ thanks
```
: )

kalwalt:
sorry, it's not my mother tongue.... :smile:

zeffii:
or.. my preferences goes to this..
```python
        def seed_adjusted(vert_list, seed):
            if seed == 0.0:
                return vert_list

            ox = get_offset(seed)
            return [[v[0] + ox[0], v[1] + ox[1], v[2] + ox[2]] for v in vert_list]


        _noise_type = noise_dict[self.noise_type]
        arguments = verts, m_octaves, m_hard, m_amp, m_freq, m_seed
        for idx, (vert_list, octaves, hard, amp, freq, seed) in enumerate(zip(*arguments)):
            final_vert_list = seed_adjusted(vert_list, seed)
            out.append([turbulence_f[self.out_mode](v, octaves, hard, _noise_type, amp, freq) for v in final_vert_list])
```

zeffii:
which i think nicely compartmentalises the logic.

nodes/vector/turbulence.py
```diff
@@ -148,18 +148,15 @@ def get_offset(seed):
...
-
-            _noise_type = noise_dict[self.noise_type]
-            octaves, hard, amp, freq = args[1:5]
-            out.append([turbulence_f[self.out_mode](v, octaves, hard, _noise_type, amp, freq) for v in verts])
+            arguments = verts, m_octaves, m_hard, m_amp, m_freq, m_seed
+            for idx, (vert_list, octaves, hard, amp, freq, seed) in enumerate(zip(*arguments)):
+
+                if not seed == 0.0:
+                    offset = get_offset(seed)
+                    verts = [[v[0] + offset[0], v[1] + offset[1], v[2] + offset[2]] for v in vert_list]
```
zeffii: don't reuse the variable name.. verts... unless you call the origin verts `m_verts`.

kalwalt: ok, i'm trying you latest version, i will compare both.

kalwalt:
i say nothing, yet need some testing  but seems to works now.

nodes/vector/turbulence.py
```diff
@@ -124,21 +123,12 @@ def process(self):
         m_freq = inputs['Frequency'].sv_get()[0]
         m_seed = inputs['Random seed'].sv_get()[0]
 
-        #if verts and verts[0]:
-
         maxlen = len(verts[0])
```
zeffii: this should test the length of incoming verts... so if there are three circle vert_lists coming in `len(verts) == 3`

zeffii:
```diff
- maxlen = len(verts[0])
+ maxlen = len(verts)
```

kalwalt: so better to leeave there that if?

zeffii: no.. you'll always be able to do `len(verts)` . All verts data is (by convention) wrapped .
```
verts = [one_circle_list]
# len(verts) == 1
verts = [one_circle_list, two_circle_list]
# len(verts) == 2
verts = [one_circle_list, two_circle_list, three_circle_list]
# len(verts) == 3
```
test with circle node :)

kalwalt: sorry i didn't see your other reply...

zeffii: you can also move the function definitions for `get_offset` and `seed_adjusted` outside of the node, keep the process function shorter.

kalwalt: ok i will do, right observation.

kalwalt:
the last is better... :smile:

zeffii:
edited..
```python
        verts = inputs['Vertices'].sv_get(deepcopy=False)
        maxlen = len(verts)

        arguments = [verts]
        for socket in self.inputs[1:]:
            arguments.append(fullList(socket.sv_get()[0], maxlen))

        ...
```
lol

zeffii:
means the process() function isn't as painful to read
```python
    def process(self):
        inputs, outputs = self.inputs, self.outputs

        if not outputs[0].is_linked:
            return

        out = []
        _noise_type = noise_dict[self.noise_type]

        verts = inputs['Vertices'].sv_get(deepcopy=False)
        maxlen = len(verts)

        arguments = [verts]
        for socket in self.inputs[1:]:
            arguments.append(fullList(socket.sv_get()[0], maxlen))

        for idx, (vert_list, octaves, hard, amp, freq, seed) in enumerate(zip(*arguments)):
            final_vert_list = seed_adjusted(vert_list, seed)
            out.append([turbulence_f[self.out_mode](v, octaves, hard, _noise_type, amp, freq) for v in final_vert_list])

        if 'Noise V' in outputs:
            out = Vector_degenerate(out)
        outputs[0].sv_set(out)
```

zeffii:
hmm.. i could rewrites parts of this endlessly. just make sure it works - we can streamline later.

kalwalt:
with the latest changes i get:

```
File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/core/update_system.py", line 322, in do_update_general
    node.process()
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/vector/turbulence.py", line 141, in process
    for idx, (vert_list, octaves, hard, amp, freq, seed) in enumerate(zip(*arguments)):
Node Vector Turbulence had exception zip argument #2 must support iteration

```

zeffii:
ok.. print the content of `arguments` before it hits the `enumerate(zip(*`  with the simplest 3 circle example.

kalwalt:
this from console;
```
arguments [[[(0.11999999731779099, 0.0, 0), (-0.06000000471572313, 0.1039230426343602, 0), (-0.059999986545239616, -0.10392305312509302, 0)], [(2.915000017732382, 0.0, 0), (-1.4575001559966332, 2.5244639824425255, 0), (-1.457499714605292, -2.524464237279918, 0)], [(5.710000038146973, 0.0, 0), (-2.855000307277543, 4.945004922250691, 0), (-2.8549994426653442, -4.945005421434743, 0)]], None, None, None, None]
arguments [[[(0.11999999731779099, 0.0, 0), (-0.06000000471572313, 0.1039230426343602, 0), (-0.059999986545239616, -0.10392305312509302, 0)], [(2.915000017732382, 0.0, 0), (-1.4575001559966332, 2.5244639824425255, 0), (-1.457499714605292, -2.524464237279918, 0)], [(5.710000038146973, 0.0, 0), (-2.855000307277543, 4.945004922250691, 0), (-2.8549994426653442, -4.945005421434743, 0)]], None, None, None, None, None]

```
zip  needs to access a nested list

zeffii:
my mistake.. `fillList` doesn't return a value.. it only adjusts the data passed into it.
```python
        arguments = [verts]
        for socket in inputs[1:]:
            data = socket.sv_get()[0]
            fullList(data, maxlen)
            arguments.append(data)
```

kalwalt:
we are in the... turbulence but seems to fix this!

zeffii:
i think a `fullList` implementation could return the data too.. but enough for now.

kalwalt:
@zeffii ok, just tesing now, to be sure that it works properly.

kalwalt:
i dont find any issues for now: seed, noisetype,and other settings works as expected.

kalwalt:
@zeffii i will fix tomorrow, i also want improve a bit the documentation, the output range as i wrote in the docs is not always in the same range, for certain noise type goes beyoynd 1.0 value. A table of output values will help if the user want visualize or save the image in a correct way.

zeffii:
don't add this... just think about it 
```python
def wrapped(t_func, vlist, _noise_type, octaves, hard, amp, freq):
    return [tfunc(v, octaves, hard, _noise_type, amp, freq) for v in vlist]


...

            # iterate over vert lists and pass arguments to the turbulence function
            out = []

            for idx, (vert_list, *args, seed) in enumerate(zip(*arguments)):
                final_vert_list = seed_adjusted(vert_list, seed)
                out.append(wrapped(tfunc, final_vert_list, _noise_type, *args))
```

kalwalt:
@zeffii Could be some grammatical errors in the docs, sorry for my english! if you find some please tell me and i will correct. Apart this i'm fine with this node now.

docs/nodes/vector/turbulence.rst
```diff
@@ -46,7 +46,7 @@ Inputs & Parameters
 Range table
 -----------
 Scalar values from turbulence node with size(n.verts)=64x64, step=0.05, octaves=3, amplitude=0.5, frequency=2.0, random seed=0.
-
+Plug a map range node in the scalar output and map it to the desired range (min=0, max=1) as in the image below.
 
 +----------------+----------------------+---------------------+----------------------+
 |  Noise Type    |       media          |  maximum            |   minimum            |
```
zeffii: median?

kalwalt: probably, the average values between min and max, i used numpy.median() in italina we say media (latin words also)

zeffii: then _median_ or _average_ (but i don't mind)

zeffii: _mean_ ?

kalwalt: fixed!

zeffii:
just give the word and i'll merge.

kalwalt:
i'm ok, merge it!  :+1:

kalwalt:
:tada: :champagne:  another baby in the family! thank you @zeffii  for the time and the patience!

zeffii:
my pleasure. I look forward to using it.

Issue #1195, opened 2017-02-02
SV for 2D Plane Tracking, need UV, maybe Adaptive Polygon


enzyme69:
![screen shot 2017-02-02 at 20 33 32](https://cloud.githubusercontent.com/assets/3952687/22543947/f2c2a2b8-e986-11e6-92d4-167a19929519.png)

I tried this Four Corner Points tracking, and I have 4 Empty, I manage to create the Plane in Sverchok, thinking to place 3D monkey head into it.

Polygon Adaptive actually does a very good job, tracking the plane etc. 

Is there another method to do this plane tracking without using Polygon Adaptive?

Btw, if the plane is being generated in Sverchok, it does not have UV and if I create the Smart UV, the next frame, it gets reset and destroyed. The only way is to bake the mesh.

ly29:
I mesh out try with setting the vertices only option, in that case it writes out the mesh in less destructive form of only updating the vertices unless the topology changes. So for a simple plane a lot more data should survive. But this is untested by me...

enzyme69:
@ly29 what is the setting with Vertices only option?

![screen shot 2017-02-03 at 22 40 29](https://cloud.githubusercontent.com/assets/3952687/22590238/49aac052-ea62-11e6-8cac-cc5fe8d8bb7e.png)

I have few ideas, from Sverchok, the Four Point Empty Tracking can become:
- Lattice
- Bones
- Another some kind of Empty or something to Instance object

enzyme69:
UV Data that get reset everytime seems to be a big problem at the moment. Unless I bake it out as Alembic, and reload... will try.

ly29:
In the extra options in the side panel.

enzyme69:
@ly29 which node? bmesh viewer? didn't see... wait

enzyme69:
Fixed vert count?

ly29:
@enzyme69 
Yes exactly.

ly29:
It seems to be working for me.

- 1 Crete plane in Sverchok
- 2 Select fixed vert count
- 3 uv map it.
- 4 update vertices and uv map survives

enzyme69:
@ly29 hmm.. that's unusual and not working on my side. Will try again for proof. However, I think baking as Alembic is ok. 

At later stage, I want to setup pose bone to follow this 4 point plane tracking. I think it should be some kind of workflow. I will make live noding video for this.

Thanks!

enzyme69:
Quick one:
[...]

PR #1204, opened 2017-02-04
Bmesh viewer mk2 mini review

ly29:
Just untested right now, do not merge.

The goal is to get the merging variant to respect the only vertices options.

@zeffii 
Also messed bit with generator code.


ly29:
So this seems to be working fine but would like a another opinion before pushing,

The reason for this is #1195 and to unify the fixed verts option.

nodes/viz/viewer_bmesh_mk2.py
```diff
@@ -122,10 +122,7 @@ def make_bmesh_geometry_merged(node, idx, context, yielder_object):
     big_edges = []
     big_faces = []
 
-    while (True):
```
zeffii: what was I thinking....

ly29:
We all have those moments.

PR #1218, opened 2017-02-11
Formal FlipNormals Node

zeffii:


zeffii:
image

There are two modes, and reverse works on both to flip the output.

```python
def flip_from_mask(mask, geom, reverse):
    """
    this mode expects a mask list with an element corresponding to each polygon
    """
    ...


def flip_to_match_1st(geom, reverse):
    """
    this mode expects all faces to be coplanar, else you need to manually generate a flip mask.
    """
    ...
```

PR #1221, opened 2017-02-11
add svrx nodescript to reload magic

zeffii:


zeffii:
If a SN has been loaded and the values have been adjusted on the sliders, then this won't reset those values, but it will reload the internal code using `node.load_text()`

@ly29  is OK?

ly29:
It makes sense to me.

PR #1242, opened 2017-02-18
Custom icons support

DolphinDream:
Started adding custom icon support.

Currently the implementation expects the icons to be stored in the ui/icons folder and be given in sv_icon_name.png format. An icon filename in this folder is converted/uppercased into icon string ID SV_ICON_NAME, which then can be used in sv modules wherever the nodes/props need to add a custom icon. 

As suggested by @zeffii I added support to nodeview_space_menu.py to handle a new bl_icon2 attribute that a custom node can specify to declare a custom icon. These custom icons will be displayed in the search menu. (see commit comments for more details). If a node declares bl_icon attribute, this icon is used in the search menu, otherwise, if bl_icon2 is given, the custom icon is used instead. 

Note: there’s no need to specify both bl_icon and bl_icon2, but if bl_icon2 is also specified, it takes precedence over the bl_icon.

```python
class SvMixNumbersNode(bpy.types.Node, SverchCustomTreeNode):
    ''' Mix Numbers '''
    bl_idname = 'SvMixNumbersNode'
    bl_label = 'Mix Numbers'
    bl_icon2 = ‘SV_MIX_NUMBER'
````

To add custom icons to various UI properties, this is done by adding the attribute icon_value=customIcon(“CUSTOM_ICON_ID”) to the property:
```python
layout.prop(self, “floatValue”, text=“Float value”, icon_value=customIcon(“SV_FLOAT”))
```

Of course, to use custom icons in a module you need to include the custom icon module:

```python
from sverchok.ui.sv_icons import customIcon
```

sv-customicon-example2


ly29:
Please avoid import *, use from icons import customIcom
Module level dictionary that shouldn't be accessed from outside should have a _ as first character. 

Glad you are taking your time to do this :)

enzyme69:
This is interesting! :) Pretification of Sverchok. We need a proper logo.

DolphinDream:
Thx for the tip @ly29 . I’m kind of new to python and I had no clue what the import * does. I though it only exposes the functions. You’re right, the global vars in that module should not be exposed. I’ll fix that.

zeffii:
No need for global in the icon loader,  see cfg.py in mesh_tiny_cad

ly29:
@DolphinDream 
Great work considering lack of python knowledge.

I would do the following changes:

```
sv_icons.py -> icons/__init__.py
```
 Change in root ```__init__.py``` as well. 
```
"monad", "sv_icons",
# to
"monad", "icons",
```
Then you can do this,
```
from sverchok.ui.icons import customIcon
```
Which looks very clean.

In the ```customIcon``` function I would use the ```dict.get``` method that lets us return a default in the value doesn't exist, default is None.

```python
def customIcon(name): 
    logDebug("customIcon called: ", name) 
 
    global _custom_icons 
    if name in _custom_icons: 
        return _custom_icons[name].icon_id 
    else: 
        logDebug("No custom icon found for name: ", name) 
        return 0 
# to
def customIcon(name): 
    logDebug("customIcon called: ", name) 

    icon = _custom_icons.get(name)
    if icon is None:
        logDebug("No custom icon found for name: ", name)
        return None # or {}
    else:
        return icon.icon_id

```

ly29:
Anyway return ```None``` instead of 0 for missing value

ly29:
Of course the path for loading files will change as well

zeffii:
i think, upon reflection, that the property should be `sv_icon` instead of `bl_icon2` .

ui/sv_icons.py
```diff
@@ -0,0 +1,59 @@
...
+import glob
+import bpy.utils.previews
+
+DEBUG = False
+def logDebug(message, extra=""):
+    if DEBUG:
+        print(message, extra)
+
+# global variable to store icons in
+_custom_icons = None
```
zeffii: make this a `{}` , `_custom_icons = {}`

ui/sv_icons.py
```diff
@@ -0,0 +1,59 @@
...
+
+def customIcon(name):
+    logDebug("customIcon called: ", name)
+    global _custom_icons
```
zeffii: dicts are global to the file, hence this line won't be needed
```diff
-    global _custom_icons
```
if doesn't matter where in Sverchok the customIcon function is called , code inside the function will first look at the file it exists in, and it will find the `_custom_icons` dict. and it will be read/writable without using the `global` keyword.

You can leave it as it is, but i _will_ change it.

ui/sv_icons.py
```diff
@@ -0,0 +1,59 @@
...
+
+    if name in _custom_icons:
+        return _custom_icons[name].icon_id
+    else:
+        logDebug("No custom icon found for name: ", name)
+        return 0
+
+def loadCustomIcons():
+    logDebug("loadIcons called")
+    global _custom_icons
```
zeffii: same story.
```diff
-    global _custom_icons
```

ui/sv_icons.py
```diff
@@ -0,0 +1,59 @@
...
+    for iconFile in iconFiles:
+        iconName = os.path.splitext(iconFile)[0]
+        iconID = iconName.upper()
+        iconIDs.append(iconID)
+        logDebug(iconID)
+        _custom_icons.load(iconID, os.path.join(iconsDir, iconFile), "IMAGE")
+
+def removeCustomIcons():
+    logDebug("unloadIcons called")
+    global _custom_icons
```
zeffii: same story.
```diff
-    global _custom_icons
```

zeffii:
for consistency with the codebase and style, we don't generally use caps anywhere in function names. so `removeCustomIcons()`  -> `remove_custom_icons()` . Get in the habit.

zeffii:
this pesky thing called pep8 https://www.python.org/dev/peps/pep-0008/#id43

if your code editor has pylint / auto pep8 checker, it's time to switch it on and suffer.

ly29:
Well we do have some old ones like ```updateNode``` but we try to avoid them for new stuff.

DolphinDream:
Sure, i’ll make sure to conform to the sverchok_function_style :) I’m used to conforming to camelCase at work, hence the tendency. I like constituency, so whenever I see both style I typically chose the one I’m comfortable with. I set my sublime text editor to have PEP8 auto format to make sure I comply.

DolphinDream:
I’m not sure what you mean by sv_icons.py -> icons/__init__.py
I thought that it’s recommended to prefix modules with sv_ (like the
sv_easing_functions.py for instance). Also, if the custom icon script is
put inside the ui/icons folder it will get mixed up with the icons
themselves and I would recommend to keep only icons/images in the icons
folder.

As to the dict.get() recommendation, I don’t really see the difference. Can
you please elaborate ? Also the reason why I returned 0 when no icon was
available is because I wanted the external code to still work if an icon
was not available without having to add extra conditional to test for the
existence of the icon:

```python
layout.prop(self, “myProp”, text=“some prop”,
icon_value=custom_icon(“non_existent_icon.png”)
```

if custom_icon returns zero the icon_value=0 will work but have no icon to
display. If the custom_icon would return None or {} i’m not sure the same
will be true and I may have to do some additional testing before hand

```python
if custom_icon(“some_icon.png”):
    layout.prop(self, “myProp”, text=“some prop”,
icon_value=customIcon(“some_icon.png”)
else:
    layout.prop(self, “myProp”, text=“some prop”)
```

DolphinDream:
@ly29 I’m not sure what you mean by sv_icons.py -> icons/__init__.py
I thought that it’s recommended to prefix modules with sv_ (like the
sv_easing_functions.py for instance). Also, if the custom icon script is
put inside the ui/icons folder it will get mixed up with the icons
themselves and I would recommend to keep only icons/images in the icons
folder.

As to the dict.get() recommendation, I don’t really see the difference. Can
you please elaborate ? Also the reason why I returned 0 when no icon was
available is because I wanted the external code to still work if an icon
was not available without having to add extra conditional to test for the
existence of the icon:

```python
layout.prop(self, “myProp”, text=“some prop”,
icon_value=custom_icon(“non_existent_icon.png”)
```

if custom_icon returns zero the icon_value=0 will work but have no icon to
display. If the custom_icon would return None or {} i’m not sure the same
will be true and I may have to do some additional testing before hand

```python
if custom_icon(“some_icon.png”):
    layout.prop(self, “myProp”, text=“some prop”,
icon_value=customIcon(“some_icon.png”)
else:
    layout.prop(self, “myProp”, text=“some prop”)
```

ly29:
@DolphinDream 
Admittedly you are right.

ui/sv_icons.py
DolphinDream: I just removed all globals from the file and now the icons are not found.

zeffii:
@DolphinDream 

given a dictionary
```python
some_dict = {'A': 20, 'B': 23}
a = some_dict.get('C')

a
>>> None
```
get automatically returns None, if the key isn't found.

A neat feature of the .get() for dicts is that you can pass a default for when the key isn't found
```python
some_dict = {'A': 20, 'B': 23}
a = some_dict.get('C', 0)

a
>>> 0
```

zeffii:
meaning
```python
def custom_icon(name):
     logDebug("custom_icon called: ", name)

     if name in _custom_icons:
         return _custom_icons[name].icon_id
     else:
         logDebug("No custom icon found for name: ", name)
         return 0
```
becomes:

```python
def custom_icon(name):
     default = lambda: None
     default.icon_id = 0
     return _custom_icons.get(name, default).icon_id
```
but if you want the debug stuff.. then don't do that :)

DolphinDream:
Behold! the EnumProperty / drop-down list custom-iconified ! :) 

enumprop-dropdown-customicons

nortikin:
lovely!!! if it works, please, merge.

DolphinDream:
Huh? did you just merged my master into my customIcons branch? I’m a bit confused. Now my origin/customIcons branch (which i assumed was on my forked repo) has moved. Does the owner of the forked repo have permissions to merge branches (update refs) on my forked repo? Btw..I haven’t pushed my local customIcons commits to the remote (my forked repo) yet which allows for enumProp custom icons yet. 

And removing the global from the sv_icons.py btw, did break the code, so i had to revert it.

zeffii:
i pushed master into your branch using the "Resolve Conflicts button" . all you had to do ( i think ) was git pull, before committing..

zeffii:
> Also i noticed that there is a little lag when displaying the custom icons the first time in the search menu for instance.

if you find a solution to that i'm eager to hear it.

DolphinDream:
sorry. I had no idea that anybody can update my remotes and pushed forced my local changes to the remote customIcons. Why would you merge the official repo branches onto a forked repo branch. Wouldn’t you first create a branch with he same name on the official branch and then do the merges? 

I was under the impression that only collaborators I add to a repo are allowed to merge in my forked repo. (no complains here, just trying to understand what’s going on as I haven’t used GitHub for very long).

zeffii:
i'm surprised that's what happened too. I think resolving conflict is seen as some kind of benevolent correction by the master to a branch to help it merge, therefor requiring no permission.. i don't know.

zeffii:
i'll let you resolve your own conflicts in future :D

DolphinDream:
Were you trying to close the pull request ? If so, i would assume that customIcons branch should have been merged into master, not the other way around, right?

zeffii:
we merge master into the branches so that the branch stays relatively up-to-date, then pushing the branch into master isn't such a painful experience.

zeffii:
press the Resolve Conficts button.. to see what you need to resolve.

DolphinDream:
Cool. I resolved it locally and pushed again. It seems happy now. :D 
I would normally not merge master into a feature branch myself (on a forked repo) since I never know if the feature branch i work on will be merged into master later on (e.g. if a pull request is not going to be accepted), but I understand why you would want to do this kind of merges.

ly29:
This will be merged.

ui/sv_icons.py
```diff
@@ -0,0 +1,76 @@
...
+        iconID = iconName.upper()
+        logDebug(iconID)
+        _custom_icons.load(iconID, os.path.join(iconsDir, iconFile), "IMAGE")
+
+    _custom_icons_loaded = True
+
+
+def remove_custom_icons():
+    logDebug("remove_custom_icons called")
+    global _custom_icons
```
zeffii: dicts are global to the file,  this global statement doesn't do what you think. this line can go.

ui/sv_icons.py
```diff
@@ -0,0 +1,76 @@
...
+    global _custom_icons_loaded
```
zeffii: This one is more subtle,  we hope to keep the usage of the `global` keyword down. While i understand this is a convenient way to do it, and it will work fine, consider using the property of `_custom_icons` having content to mean that custom_icons _are_ loaded..   `len(_custom_icons_loaded)` will return Truthy whenever it has content, and at no point will the content be half-full . (at least.. i don't see that it would be from your code flow )

ui/sv_icons.py
DolphinDream: I dropped the global lines and it seems that without these the dictionary is not being updated.. and every time a custom_icon is being called, the icon has to be reloaded from disk, not from the dictionary. You can correct me on that but that's what I see at my end. I'll double check once again to make sure.

ui/sv_icons.py
DolphinDream: Yeah. that will cleanup the code. Thx for the tip. I'll update.

ui/sv_icons.py
zeffii: well, you have this line `_custom_icons = bpy.utils.previews.new()`   which overwrites the dict.

zeffii: follow what I did here https://github.com/zeffii/mesh_tiny_cad/blob/master/CFG.py

zeffii: i think you're almost there
```

    pcoll = bpy.utils.previews.new()

    ### you have your own pcoll loading code, that's fine.
    # icons_dir = os.path.join(os.path.dirname(__file__), "icons")
    # for icon_name in ICONS:
    #     pcoll.load(icon_name, os.path.join(icons_dir, icon_name + '.png'), 'IMAGE')

    icon_collection["main"] = pcoll   

```

zeffii: and in the above snippet `icon_collection` is directly analogous to what `_custom_icons` should be. you can also drop the `_` prefix. The only thing from that file that should be imported in other places is the `custom_icons` dict.

DolphinDream: Cool. I'll fix. Also i think @ly29 had previously suggested I should add _ prefix to global vars that are not to be used externally. What is the rule with this prefix btw?

zeffii: he said that (iirc ) to prevent the namespace polution when using `import *` , but you shouldn't need to import anything other the `custom_icon(name)` function, `  ( i incorrectly suggested that you import `custom_icons` dict above... but that's not needed since you've made a nice function for that dict to always produce a valid response )

any file that needs a custom icon : 
`from sverchok.ul.sv_icons import custom_icon`

ly29:
The _ prefix is recommended for module level storage, it isn't private in any real way but is a strong don't touch hint.

ly29:
https://www.python.org/dev/peps/pep-0008/#global-variable-names

zeffii:
i'll retract my objection entirely to the `prefix` @DolphinDream  - use it.   it won't be exposed outside of that file anyway.

DolphinDream:
Super. How about now?  it pretty much looks like @zeffii 's tinyCad code :) what's not to like ? ;p

zeffii:
that takes care of `global`, methinks.

ui/sv_icons.py
```diff
@@ -0,0 +1,52 @@
...
+    for iconFile in iconFiles:
+        iconName = os.path.splitext(iconFile)[0]
+        iconID = iconName.upper()
+        custom_icons.load(iconID, os.path.join(iconsDir, iconFile), "IMAGE")
+
+    _icon_collection["main"] = custom_icons
+
+
+def remove_custom_icons():
+    bpy.utils.previews.remove(custom_icons)
```
zeffii: i think this needs to read 
```diff
-    bpy.utils.previews.remove(custom_icons)
+    bpy.utils.previews.remove(_icon_collection["main"])
```
because `custom_icons` won't be available inside `remove_custom_icons`

zeffii: or failing that..

https://github.com/zeffii/mesh_tiny_cad/blob/master/CFG.py#L73-L76

DolphinDream: Oops. Good catch. Fixed again.

zeffii:
i have no more significant issues with the branch atm.

whether i'll be cool with the custom icon _designs_ is another matter entirely :)  but we'll sort out a way to set preferences to icon packs from zips. That can be done separately.

DolphinDream:
Hallelujah 🤓

DolphinDream:
Here’s a little preview of the “icon designer" blend / script :) 

sv-iconmanager-example1

sv-icon-example1

DolphinDream:
Checkout this temporary branch [iconManager](https://github.com/DolphinDream/sverchok/tree/iconManager) in combination with this blend file.

[iconGridScript.blend.zip](https://github.com/nortikin/sverchok/files/795466/iconGridScript.blend.zip)

ui/sv_icons.py
```diff
@@ -0,0 +1,54 @@
...
+    custom_icons = _icon_collection["main"]
+
+    default = lambda: None  # for no icon with given name will return zero
+    default.icon_id = 0
+
+    return custom_icons.get(name, default).icon_id
+
+
+def load_custom_icons():
+    if len(_icon_collection):  # return if custom icons already loaded
```
ly29: There is now need to check ```len(_icon_collection)``` just ```_icon_collection``` works as well.

ly29:
👍 I am for merging this now, the icons themselves are a separate and interesting issue.

ui/sv_icons.py
DolphinDream: noted. I assume this is true for any type (list, dictionary, array)?

zeffii: any non empty collection / iterable will return True if it's not empty, even a string.

ly29: All standard types behave this way. 
Not numpy arrays however... Don't know what they where thinking there. 😢

zeffii: even `generator expressions`
```python
k = [2,3,4,5,6,7]

g = (_ for _ in k)

for n in g:
    pass

if g:
    print('yes!')
    print(g)
```

ly29:
```python
>>> bool(np.array([1,2,3]))
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

zeffii: both empty and non-empty generator expressions will both return True
```python
k = [2,3,4,5,6,7]

g = (_ for _ in k)

for n in g:
    print(n)
    pass

if g:
    print('yes!')
    print(g)
    print(list(g))
```
like
```python
2
3
4
5
6
7
yes!
<generator object <genexpr> at 0x0000022302397468>
[]
```

ly29: @zeffii 
But that should be false since the generator expression is empty? ```next(g) -> StopIteration```

zeffii: that might change in `3.6.x` , but at the moment it doesn't appear be consistent. I would expect an empty generator to evaluate to False..

ly29: You still have the function object, oh well, one should loop with generators not check them.
http://stackoverflow.com/questions/661603/how-do-i-know-if-a-generator-is-empty-from-the-start

zeffii: a potential gotcha moment.

PR #1246, opened 2017-02-18
Iojson bump

zeffii:


zeffii:
This aims to resolve primarily issues with those nodes that use the property 'prop_types` to force Integer/Float inputs on their interface.

PR #1247, opened 2017-02-18
testing boolprop +  code paths

ly29:
Okay so not for merging, but for testing

With our without ```deepcopy=False``` 1000x1000 uv sphere

```
Processed  Sphere in: 2.8220
Processed  Viewer BMeshMK2 in: 5.4762
# deepcopy=False below
Processed  Sphere in: 2.7650
Processed  Viewer BMeshMK2 in: 2.6624
```

With our without ```deepcopy=False``` 32x32 uv sphere

```
Processed  Sphere in: 0.0029
Processed  Viewer BMeshMK2 in: 0.0079
# deepcopy=False below
Processed  Sphere in: 0.0027
Processed  Viewer BMeshMK2 in: 0.0030
```

Based on observation from here
**https://github.com/Sverchok/Sverchok/issues/47#issuecomment-280333879**

### DIFF
--- a/nodes/viz/viewer_bmesh_mk2.py
+++ b/nodes/viz/viewer_bmesh_mk2.py
@@ -256,6 +256,8 @@ class SvBmeshViewerNodeMK2(bpy.types.Node, SverchCustomTreeNode):
         default=False,
         description='Allows mesh.transform(matrix) operation, quite fast!')
 
+    test_deep_copy = BoolProperty(default=False)
+
     def sv_init(self, context):
         gai = bpy.context.scene.SvGreekAlphabet_index
         self.basemesh_name = greek_alphabet[gai]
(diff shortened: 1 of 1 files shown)


ly29:
Not super accurate test my could observe the same trend

ly29:
If you want sverchok to be fast optimize this function. Or better yet make it not to be used.

```python

def sv_deep_copy(lst):
    """return deep copied data of list/tuple structure"""
    if isinstance(lst, (list, tuple)):
        if lst and not isinstance(lst[0], (list, tuple)):
            return lst[:]
        return [sv_deep_copy(l) for l in lst]
    return lst
```

ly29:
I feel like the right solution might be to change the default and put a ```deepcopy=True``` on every ```sv_get``` in every node that doesn't declare ```deepcopy=False``` today...

ly29:
As far as I can see this node doesn't modify incoming data and it should be safe to turn on

PR #1252, opened 2017-02-19
add exception handling to remove_keymap

zeffii:
```python
def remove_keymap():

    for km, kmi in nodeview_keymaps:
        try:
            km.keymap_items.remove(kmi)
        except Exception as e:
            err = repr(e)
            if "cannot be removed from 'Node Editor'" in err:
                print('keymaps for Node Editor already removed by another add-on, sverchok will skip this step in unregister')
                break

    nodeview_keymaps.clear()
```
or the more crude " hair on fire " 
```python
def remove_keymap():

    for km, kmi in nodeview_keymaps:
        if kmi.name in km.keymap_items:
            km.keymap_items.remove(kmi)

    nodeview_keymaps.clear()
```

### DIFF
--- a/ui/nodeview_keymaps.py
+++ b/ui/nodeview_keymaps.py
@@ -48,8 +48,16 @@ def add_keymap():
 
 
 def remove_keymap():
+
     for km, kmi in nodeview_keymaps:
-        km.keymap_items.remove(kmi)
+        try:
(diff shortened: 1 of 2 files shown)


zeffii:
i'm pushing the verbose version now..  but it should probably be a library function... not a priority.

PR #1253, opened 2017-02-19
Colornodes

zeffii:
just make them already

### DIFF
--- a/index.md
+++ b/index.md
@@ -159,8 +159,6 @@
     SvVectorRewire
     svAxisInputNode
     SvAxisInputNodeMK2
-    SvColorsInNode
-    SvColorsOutNode
 
 ## Matrix
     MatrixGenNode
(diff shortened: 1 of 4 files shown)


zeffii:
image

prelim implementation. break it in.

zeffii:
maybe this node set needs a mixRGB node.

portnov:
Isn't "mix RGB" just a "vector linear interpolation"?

zeffii:
there tonnes of ways to mix rgb, like photoshop layer mixing..   
- "Vector Linear Interpolation" would be average
- mix maximum of each component
- mix minimum of each component
- subtract a from b 

etc.

enzyme69:
I think Vector Interpolation is acting a bit like Color Ramp. Will make video about it.

screen shot 2017-02-21 at 12 24 51

[evaluate_lerp_001_2017_02_21_01_24.zip](https://github.com/nortikin/sverchok/files/788841/evaluate_lerp_001_2017_02_21_01_24.zip)

I think the Color In and Out are fine like that. Simple. MixRGB would be cool.

Issue #1248, opened 2017-02-18
Texture viewer node proposal


kalwalt:
I'm trying to implement a simple "texture viewer node" very useful to display data from noise, fractals, ect. nodes. This should help to build complex tree nodes with multiple noise nodes. I started from the easing.py node, not sure if it's a good idea, I added all the code to display the texture but as a result i get displayed only a 64x64 black square , the code is in  [my repo](https://github.com/kalwalt/sverchok/blob/texture_viewer_node/nodes/viz/texture_viewer.py) , but  i dont send yet a pull request. I tryed different combination of type format, and i'm not sure on some bgl OpenGL calls, i think blender treat OpenGl  in a particular manner ( probably i missed something or there is another approach in doing this).
in particular im not sure:

1. how i initialized the `Buffer `,
2. texname = 0 ---> bgl.glBindTexture(bgl.GL_TEXTURE_2D, texname)
                          from Opengl docs:  void glBindTexture( GLenum target, GLuint texture);  you can set texture in this way: static GLuint texName
3. if glTexCoords are correct

Any help is welcome and really appreciated.
Walter

zeffii:
The `easing.py` is a good place to start..
http://blender.stackexchange.com/questions/51465/why-is-texture-colour-changing-in-bgl
http://blender.stackexchange.com/questions/23465/how-to-add-backdrop-images-for-py-nodes

for the time being you could draw 64*64 verts :)

kalwalt:
@zeffii I think the links you posted are crucials!

zeffii:
sadly my familiarity with drawing _texture_ in bgl is superficial at best.

kalwalt:
YESSS! it works!
![text viewer-test](https://cloud.githubusercontent.com/assets/1275858/23096485/e8813c80-f61d-11e6-99b0-c322008e0d95.png)

the issue was here: 

````python

bgl.glBegin(bgl.GL_QUADS)
        '''
        for coord in [(x, y), (x, y+1), (x+1, y+1), (x+1, y)]:
            bgl.glTexCoord2f(*coord)
        for coord in [(x, y), (x+w, y), (w+x, y-h), (x, y-h)]:
            bgl.glVertex2f(*coord)
        '''
        bgl.glNormal3f(0, 0, 1)
        bgl.glTexCoord2d(0, 0); bgl.glVertex3f( 0, 0, 0)
[...]

PR #1255, opened 2017-02-19
Texture viewer node

kalwalt:
as discussed [here](https://github.com/nortikin/sverchok/issues/1248)

Now the node has:

- the ability to change the display size (64,128,256,512,1024)

- hide and unhide of texture

to do: 

- at the moment the data should be in the form of square matrix, as the buffer need this kind of data. 
how can i transform the data from a noise node or fractal node in the form of [[0.0,1.0,0.2,....ect.]] to 
a matrix like data? Does sverchok has some code utility to do this? 
the idea  in pseudocode:

````python
if data: [[....]]

    convert_to_matrix >>>> [[...],[...],[...],........]

else data:  [[...],[...],[...],........]

go straight

````

- i think there is a better way for the tex_list stuff, i'm not so smart yet with .

### DIFF
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -252,8 +252,16 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
     def copy(self, node):
         self.n_id = ''
 
-def save_bitmap(self):
-    pass
+    def save_bitmap(self,):
+       img = bpy.data.images.new(name="depthmap", width=64,height=64,alpha=False, float_buffer=True)
+       img.colorspace_settings.name = 'Linear'
+       # bake() #implemented in separate function
+       img.file_format = 'PNG'
+       # need to set to 16-bit here
+       img.filepath_raw = "/tmp/my_new_bake.png"
+       img.save()
+       print('saved!')
+
 
 def register():
     bpy.utils.register_class(SvTextureViewerNode)
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -59,13 +59,13 @@ def simple_screen(x, y, args):
         bgl.glTexEnvf(bgl.GL_TEXTURE_ENV, bgl.GL_TEXTURE_ENV_MODE, bgl.GL_REPLACE)
         bgl.glBindTexture(bgl.GL_TEXTURE_2D, texname)
 
-        #bgl.glColor4f(*color)
         bgl.glBegin(bgl.GL_QUADS)
 
-        for coord in [(x, y), (x, y+1), (x+1, y+1), (x+1, y)]:
-            bgl.glTexCoord2f(*coord)
-        for coord in [(x, y), (x+w, y), (w+x, y-h), (x, y-h)]:
-            bgl.glVertex2f(*coord)
+        #bgl.glNormal3f(0, 0, 1)
+        bgl.glTexCoord2d(0, 0); bgl.glVertex2f( x, y )
+        bgl.glTexCoord2d(1, 0); bgl.glVertex2f( x + 64 , y )
+        bgl.glTexCoord2d(1, 1); bgl.glVertex2f( x + 64, y + 64 )
+        bgl.glTexCoord2d(0, 1); bgl.glVertex2f( x, y + 64 )
 
         bgl.glEnd()
 
@@ -76,14 +76,14 @@ def simple_screen(x, y, args):
     def init_texture(width,height,texname,data):
 
         #bgl.glClearColor(0.0,0.0,0.0,0.0)
-        bgl.glShadeModel(bgl.GL_FLAT)
+        bgl.glShadeModel(bgl.GL_SMOOTH)
         bgl.glEnable(bgl.GL_DEPTH_TEST)
-        bgl.glEnable(bgl.GL_TEXTURE_2D)
+
         Buffer = bgl.Buffer(bgl.GL_FLOAT, [width,height], data)
         bgl.glPixelStorei(bgl.GL_UNPACK_ALIGNMENT,1)
 
         bgl.glGenTextures(1,Buffer)
-        #bgl.glEnable(bgl.GL_TEXTURE_2D)
+        bgl.glEnable(bgl.GL_TEXTURE_2D)
         #glBindTexture(target, texture): texture (unsigned int) – Specifies the name of a texture.
         bgl.glBindTexture(bgl.GL_TEXTURE_2D, texname)
 
@@ -93,10 +93,6 @@ def simple_screen(x, y, args):
         bgl.glTexParameterf(bgl.GL_TEXTURE_2D, bgl.GL_TEXTURE_WRAP_T, bgl.GL_CLAMP)
         bgl.glTexParameterf(bgl.GL_TEXTURE_2D, bgl.GL_TEXTURE_MAG_FILTER, bgl.GL_LINEAR)
         bgl.glTexParameterf(bgl.GL_TEXTURE_2D, bgl.GL_TEXTURE_MIN_FILTER, bgl.GL_LINEAR)
-
-        #bgl.glEnable(bgl.GL_BLEND)
-        #bgl.glBlendFunc(bgl.GL_SRC_ALPHA, bgl.GL_ONE_MINUS_SRC_ALPHA)
-        #bgl.glShadeModel(bgl.GL_SMOOTH)
 
         bgl.glTexImage2D(
                bgl.GL_TEXTURE_2D, 0, bgl.GL_LUMINANCE, width, height, 0,
@@ -160,7 +156,8 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
             }
             nvBGL2.callback_enable(n_id, draw_data)
 
-
+    def free(self):
+        nvBGL2.callback_disable(node_id(self))
 
     # reset n_id on copy
     def copy(self, node):
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -120,13 +120,7 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
         description="offers bitmap saving",
         default="PNG"
     )
-    '''
-    directory = StringProperty(
-        maxlen=1024,
-        subtype='FILE_PATH',
-        options={'HIDDEN', 'SKIP_SAVE'}
-    )
-    '''
+
     in_float = FloatProperty(
         min=0.0, max=1.0, default=0.0, name='Float Input',
         description='input for texture', update=updateNode
@@ -154,11 +148,10 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
         c.prop(self, 'activate')
 
     def draw_buttons_ext(self, context, l):
-        l.label(text="save texture as bitmap image, choose a format:")
+        l.label(text="save texture as a bitmap image, choose a format:")
         l.separator()
         l.prop(self, "bitmap_save")
         l.separator()
-        l.row()
         l.operator("node.scriptlite_ui_callback", text="S A V E").fn_name="save_bitmap"
 
     def sv_init(self, context):
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -44,7 +44,7 @@ size_tex_dict = {
 
 bitmap_save_list=[
     ('PNG','png format', 'save texture in .png fromat','',0),
-    ('TGA','tga format','save texture in .tga fromat','',1),
+    ('TARGA','tga format','save texture in .tga fromat','',1),
     ('TIFF','tiff format','save texture in .tiff format','',2),
     ('BMP','bmp format','save texture in .tiff format','',3),
     ('JPEG','jpeg format','save texture in .jpeg format','',4)
@@ -235,8 +235,13 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
         np_buff.shape = -1
         img.pixels[:] = np_buff
 
-        img.filepath_raw = "/tmp/" + image_name
-        img.save()
+        scene=bpy.context.scene
+        scene.render.image_settings.file_format= img_format
+
+        path = img.filepath_raw
+        path = "/tmp/" + image_name
+        img.save_render(path,scene)
+
         print('saved!')
 
 def register():
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -64,16 +64,6 @@ bitmap_save_list=[
     ('BMP','bmp format','save texture in .tiff format','',3),
     ('JPEG','jpeg format','save texture in .jpeg format','',4)
 ]
-def assign_image(image_name, size ,buffer):
-    import numpy as np
-    image = bpy.data.images.new(image_name, size, size, alpha=False)
-    np_buff = np.empty(len(image.pixels), dtype=np.float32)
-    np_buff.shape = (-1, 4)
-    np_buff[:,:] = np.array(buffer)[:,np.newaxis]
-    np_buff[:,3] = 1
-    np_buff.shape = -1
-    image.pixels[:] = np_buff
-    return image
 
 def simple_screen(x, y, args):
     #draw a simple scren display for the texture
@@ -161,9 +151,20 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
         items=theme_mode_options, default="default", update=updateNode
     )
 
-    def get_buffer(self,texture):
-        self.texture = texture
-        return self.texture
+    def get_buffer(self):
+        data = self.inputs['Float'].sv_get(deepcopy=False)[0]
+        size_tex = size_tex_dict.get(self.selected_mode)
+        total_size = size_tex * size_tex
+        if len(data) < total_size:
+            default_value = 0
+            new_data = [default_value for j in range(total_size)]
+            new_data[:len(data)] = data[:]
+            data = new_data
+        elif len(data) > total_size:
+            data = data[:total_size]
+
+        texture = bgl.Buffer(bgl.GL_FLOAT, total_size, data)
+        return texture
 
     def draw_buttons(self, context, l):
         c = l.column()
@@ -190,7 +191,7 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
 
     def process(self):
 
-        data = self.inputs['Float'].sv_get(deepcopy=False)[0]
+        #data = self.inputs['Float'].sv_get(deepcopy=False)[0]
         n_id = node_id(self)
 
         # end early
@@ -199,19 +200,9 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
         #print(_data)
         if self.activate:
 
+            texture = self.get_buffer()
+
             size_tex = size_tex_dict.get(self.selected_mode)
-            total_size = size_tex * size_tex
-            if len(data) < total_size:
-                default_value = 0
-                new_data = [default_value for j in range(total_size)]
-                new_data[:len(data)] = data[:]
-                data = new_data
-            elif len(data) > total_size:
-                data = data[:total_size]
-            # and then in init texture
-            texture = bgl.Buffer(bgl.GL_FLOAT, total_size, data)
-            self.get_buffer(texture)
-            #print(tex)
 
             palette = palette_dict.get(self.selected_theme_mode)[:]
             x, y = [int(j) for j in (self.location + Vector((self.width + 20, 0)))[:]]
@@ -225,6 +216,7 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
 
                 #bgl.glGenTextures(1,Buffer)
                 bgl.glGenTextures(1,texture)
+
                 bgl.glEnable(bgl.GL_TEXTURE_2D)
                 #glBindTexture(target, texture): texture (unsigned int) – Specifies the name of a texture.
                 bgl.glBindTexture(bgl.GL_TEXTURE_2D, texname)
@@ -265,7 +257,7 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
 
     def save_bitmap(self, image_name='image_name', filepath_raw='', alpha=False,texture=texture):
         import numpy as np
-        buf = self.get_buffer(texture)
+        buf = self.get_buffer()
         img_format = self.bitmap_save
         image_name = image_name + '.' + img_format.lower()
         dim = size_tex_dict[self.selected_mode]
@@ -273,21 +265,16 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
         img = []
         if image_name in bpy.data.images:
             img = bpy.data.images[image_name]
-            print("A")
         else:
             img = bpy.data.images.new(name=image_name,width=width,height=height,alpha=alpha, float_buffer=True)
-            print("B")
-        print(img)
-        #img = bpy.data.images.new(name=image_name, width=width,height=height,alpha=alpha, float_buffer=True)
+        #print(img)
         np_buff = np.empty(len(img.pixels), dtype=np.float32)
         np_buff.shape = (-1, 4)
         np_buff[:,:] = np.array(buf)[:,np.newaxis]
         np_buff[:,3] = 1
         np_buff.shape = -1
         img.pixels[:] = np_buff
-        print("buffer is:{0}".format(buf))
-        #img = assign_image(image_name,width*height,buf)
-        img.colorspace_settings.name = 'Linear'
+        #print("buffer is:{0}".format(buf))
         img.filepath_raw = "/tmp/" + image_name
         img.save()
         print('saved!')
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -27,20 +27,6 @@ from sverchok.data_structure import updateNode, node_id
 from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.ui import nodeview_bgl_viewer_draw_mk2 as nvBGL2
 
-palette_dict = {
-    "default": (
-        (0.243299, 0.590403, 0.836084, 1.00),  # back_color
-        (0.390805, 0.754022, 1.000000, 1.00),  # grid_color
-        (1.000000, 0.330010, 0.107140, 1.00)   # line_color
-    ),
-    "scope": (
-        (0.274677, 0.366253, 0.386430, 1.00),  # back_color
-        (0.423268, 0.558340, 0.584078, 1.00),  # grid_color
-        (0.304762, 1.000000, 0.062827, 1.00)   # line_color
-    )
-
-}
-
 size_tex_list=[
     ('EXTRA_SMALL','extra_small 64x64px','extra small squared tex: 64px','',64),
     ('SMALL','small 128x128px','small squared tex: 128px','',128),
@@ -67,11 +53,11 @@ bitmap_save_list=[
 
 def simple_screen(x, y, args):
     #draw a simple scren display for the texture
-    back_color, grid_color, line_color = args[0]
+    border_color = (0.390805, 0.754022, 1.000000, 1.00)
 
-    texture = args[1]
-    size = args[2]
-    texname = args[3]
+    texture = args[0]
+    size = args[1]
+    texname = args[2]
     #print('size of tex inside simple screen: {0}'.format(size))
     texture = 1
     width = size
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by kalwalt** [comment]
just to note that i also discovered `int node` do not accept more than 1000 value, but i can set  `List repeater` and 'Range Float count' to 1024, This creating the texture.

**Comment by zeffii** [comment]
we should consider changing the limit to 1024, default.

**Comment by ly29** [comment]
You can edit the limit in the side bar, the n-panel

**Comment by ly29** [comment]
But 1024 is an equally sane max default as 1000.

**Comment by kalwalt** [comment]
> You can edit the limit in the side bar, the n-panel

I didn't know this! thank's @ly29

**Comment by kalwalt** [comment]
[this](https://github.com/nortikin/sverchok/blob/master/nodes/list_struct/split.py#L29) will solve the first point

File: nodes/viz/texture_viewer.py [hunk 1]
```diff
@@ -0,0 +1,214 @@
...
+        c.prop(self, 'activate')
+
+    def draw_buttons_ext(self, context, l):
+        l.prop(self, "selected_theme_mode")
+
+    def sv_init(self, context):
+        self.inputs.new('StringsSocket', "Float").prop_name = 'in_float'
+
+    def process(self):
+        p = self.inputs['Float'].sv_get()
```
ly29: deepcopy=False is a very good idea for this node

kalwalt:
i will add also the bitmap save stuff in the properties panel.

nodes/viz/texture_viewer.py
```diff
@@ -120,6 +128,13 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
         update=updateNode
     )
 
+    bitmap_save = EnumProperty(
+        items=bitmap_save_list,
+        description="offers bitmap saving",
+        default="PNG",
+        update=updateNode
```
zeffii: unnecessary to use `updateNode` here, it will cause `process` to be run again

kalwalt: ok, what should be more appropriate?

zeffii: if you want to trigger a call to `process` you add `update=updateNode`, but in this case you just want to `set` the property for the save operator to us.  Drop `update=updateNode` ,  will be ok.

nodes/viz/texture_viewer.py
```diff
@@ -152,9 +152,17 @@ def draw_buttons(self, context, l):
...
     def draw_buttons_ext(self, context, l):
+        l.label(text="choose a different color for the border:")
         l.prop(self, "selected_theme_mode")
+        l.separator()
         l.label(text="save texture as bitmap image, choose a format:")
         l.prop(self, "bitmap_save")
+        row = l.row()
+        #addon = context.user_preferences.addons.get(sverchok.__name__)
+        #row.scale_y = 4.0 if addon.preferences.over_sized_buttons else 1
+        opera = row.operator('save_bitmap', text="S A V E")
```
kalwalt: This is wrong because the operator neeed an operator, right? 
looking in viewer_mk2:
how should solve this? should i create another class inherited to bpy.types.Operator?

nodes/viz/texture_viewer.py
```diff
@@ -209,11 +217,14 @@ def init_texture(width,height,texname,texture):
                        bgl.GL_TEXTURE_2D, 0, bgl.GL_LUMINANCE, width, height, 0,
                        bgl.GL_LUMINANCE, bgl.GL_FLOAT, texture
                    )
+            def save_bitmap(self):
```
kalwalt: see line 163

DolphinDream:
What’s the MK/MK2 stand for ?

zeffii:
the simplest way is to make an operator, which in turn will call a function on the node. see
https://github.com/nortikin/sverchok/blob/master/nodes/generator/script1_lite.py

for now my advice is to add your own operator for this file.. but we (team sverchok)  should consider including such an operator as a util 
```
class SvScriptNodeLiteCallBack(bpy.types.Operator):

    bl_idname = "node.scriptlite_ui_callback"
    bl_label = "SNLite callback"
    fn_name = bpy.props.StringProperty(default='')

    def execute(self, context):
        getattr(context.node, self.fn_name)()
        return {'FINISHED'}

```

in `layout.operator("node.scriptlite_ui_callback").fn="write_to_disk"`

the callback operator would execute the function on the node called 'write_to_disk`.

zeffii:
> What’s the MK/MK2 stand for ?
Mark / Mark 2   versioning.

kalwalt:
looking in belnder template, i did this not yet committed:

```python
class SvSaveBitmap(bpy.types.Operator):
    '''Save Bitmap operator'''
    bl_idname = 'node.save_bitmap'
    bl_label = "Sverchok mesh baker mk2"
    bl_options = {'REGISTER', 'UNDO'}

    idname = StringProperty(
        name='idname',
        description='name of parent node',
        default='')

    idtree = StringProperty(
        name='idtree',
        description='name of parent tree',
        default='')

    @classmethod
    def poll(cls, context):
        return context.active_object is not None

    def execute(self, context):
        main(context)
        return {'FINISHED'}
```

kalwalt:
@zeffii i want to test the scipt now

ly29:
If you have a ```def save(self)``` in the node do this to test.

in draw_buttons/draw_buttons_ext

```python
layout.operator("node.scriptlite_ui_callback").fn="save"
```

kalwalt:
```python
class SvSaveBitmap(bpy.types.Operator):
    '''Save Bitmap operator'''
    bl_idname = 'node.save_bitmap'
    bl_label = "Sverchok save bitmap utility"
    fn_name = bpy.props.StringProperty(default='')
    bl_options = {'REGISTER', 'UNDO'}
   
    def execute(self, context):
        main(context)
        return {'FINISHED'}
```

and in def draw_button_ext:

`opera = row.operator('node.save_bitmap', text="S A V E").fn="save_bitmap"`

but i get:

```
search for unknown operator 'NODE_OT_save_bitmap', 'node.save_bitmap'
rna_uiItemO: operator missing srna 'node.save_bitmap'
/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py:188
search for unknown operator 'NODE_OT_save_bitmap', 'node.save_bitmap'
rna_uiItemO: operator missing srna 'node.save_bitmap'
/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py:188
search for unknown operator 'NODE_OT_save_bitmap', 'node.save_bitmap'
rna_uiItemO: operator missing srna 'node.save_bitmap'
/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py:188
search for unknown operator 'NODE_OT_save_bitmap', 'node.save_bitmap'
rna_uiItemO: operator missing srna 'node.save_bitmap'
/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py:188
search for unknown operator 'NODE_OT_save_bitmap', 'node.save_bitmap'
rna_uiItemO: operator missing srna 'node.save_bitmap'
/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py:188
search for unknown operator 'NODE_OT_save_bitmap', 'node.save_bitmap'
rna_uiItemO: operator missing srna 'node.save_bitmap'
/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py:188
search for unknown operator 'NODE_OT_save_bitmap', 'node.save_bitmap'
rna_uiItemO: operator missing srna 'node.save_bitmap'
/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py:188
search for unknown operator 'NODE_OT_save_bitmap', 'node.save_bitmap'
rna_uiItemO: operator missing srna 'node.save_bitmap'
/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py:188
search for unknown operator 'NODE_OT_save_bitmap', 'node.save_bitmap'
rna_uiItemO: operator missing srna 'node.save_bitmap'
/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py:188
search for unknown operator 'NODE_OT_save_bitmap', 'node.save_bitmap'
rna_uiItemO: operator missing srna 'node.save_bitmap'
/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py:188
```

zeffii:
see. https://github.com/nortikin/sverchok/issues/1256

but for now, `layout.operator("node.scriptlite_ui_callback").fn="save_bitmap"`

kalwalt:
@zeffii ok now the button is displayed, with node.scriptlite_ui_callback

zeffii:
now @kalwalt   the save_button function needs to be class level, not nested in `process`.

zeffii:
now @kalwalt   the `save_bitmap` function needs to be class level, not nested in `process`.

kalwalt:
Anyway i get this error:

```
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/startup/bl_ui/space_node.py", line 359, in draw
    node.draw_buttons_ext(context, layout)
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py", line 175, in draw_buttons_ext
    l.operator("node.scriptlite_ui_callback", text="S A V E").fn="save_bitmap"
AttributeError: 'NODE_OT_scriptlite_ui_callback' object has no attribute 'fn'


```

zeffii:
use `.fn_name="save_bitmap"`

kalwalt:
@zeffii  already did! thanks, just reading in the code...

zeffii:
my mistake :)  the property (StringProperty.. if you look at the declaration) is `fn_name`  , elsewhere (other callbacks) i think we just use `fn` .but that's a little short :)

kalwalt:
it works,  a pseudo -tex isaved in temp folder!

zeffii:
now you modularize your _texture from socket_  code so it can be called in process or directly from a function.

ly29:
Some simple code for assign the buffer
```python
def assign_image(image_name, size ,buffer):
    import numpy as np
    image = bpy.data.images.new(image_name, size, size, alpha=False)
    np_buff = np.empty(len(image.pixels), dtype=np.float32)
    np_buff.shape = (-1, 4)
    np_buff[:,:] = np.array(buffer)[:,np.newaxis]
    np_buff[:,3] = 1
    np_buff.shape = -1
    image.pixels[:] = np_buff
    return image
```

ly29:
Also I guess it makes sense to continuously update an image as well as having open gl viewer.

kalwalt:
@ly29 @zeffii  i get this error:

```
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/generator/script1_lite.py", line 68, in execute
    getattr(context.node, self.fn_name)()
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py", line 256, in save_bitmap
    img = assign_image(image_name,width*height,self.texture)
NameError: name 'assign_image' is not defined

```

what i'm doing wrong here? i receive the same error no matter where i put the assign function.

kalwalt:
i get: the function need to reside outside the class.

ly29:
Ok please don't take this the wrong way, but you keep having problems in places where python differ a bit from other languages. Like explicit/implict self/this, indentation vs braces.

nodes/viz/texture_viewer.py
```diff
@@ -0,0 +1,286 @@
...
+}
+
+bitmap_save_list=[
+    ('PNG','png format', 'save texture in .png fromat','',0),
+    ('TGA','tga format','save texture in .tga fromat','',1),
+    ('TIFF','tiff format','save texture in .tiff format','',2),
+    ('BMP','bmp format','save texture in .tiff format','',3),
+    ('JPEG','jpeg format','save texture in .jpeg format','',4)
+]
+def assign_image(self,image_name, size ,buffer):
```
ly29: Here remove self and pass the image reference instead of name since you create the image somewhere else.

nodes/viz/texture_viewer.py
```diff
@@ -0,0 +1,286 @@
...
+        l.prop(self, "bitmap_save")
+        l.operator('file.select_all_toggle', text='choose a directory')
+
+        row = l.row()
+        #addon = context.user_preferences.addons.get(sverchok.__name__)
+        #row.scale_y = 4.0 if addon.preferences.over_sized_buttons else 1
+
+        l.operator("node.scriptlite_ui_callback", text="S A V E").fn_name="save_bitmap"
+
+    def invoke(self, context, event):
```
ly29: Does this work? to have a node behave as an operator?
The callback we do something far simpler than a full operator

nodes/viz/texture_viewer.py
```diff
@@ -0,0 +1,286 @@
...
+    def free(self):
+        nvBGL2.callback_disable(node_id(self))
+
+    # reset n_id on copy
+    def copy(self, node):
+        self.n_id = ''
+
+    def save_bitmap(self,image_name='image.png',filepath_raw='',img_format='PNG',width=64,height=64,alpha=False,buf=tex):
+        #print(buf)
+        img = bpy.data.images.new(name="bitmap", width=64,height=64,alpha=False, float_buffer=True)
```
zeffii: i think this function signature can be shorter
```python
    def save_bitmap(self, image_name='image_name', filepath_raw='', alpha=False, buf=tex):
        img_format = self.bitmap_save
        image_name = image_name + '.' + img_format.lower()
        dim = size_tex_dict[self.selected_mode]
        width, height = dim, dim
        ...

```

nodes/viz/texture_viewer.py
kalwalt: yes this a nonsense i didn't realize it, now it works. but the problem is the buffer, how can i acces the data from the buffer? i added a tex variable in the class copy the data into it but in save_bitmap i receive no data, i have the suspect that is not the right way.

nodes/viz/texture_viewer.py
kalwalt: i don't know but i saw i different example that was needed.

ly29:
So the problems you have run into now are about behavior for blender and how operators etc work. Also how sverchok works and what can be done where, it isn't totally straightforward what you are trying to do. 

Unless you mind I think it is better to hand it over to us/me and I can make work quickly and then you can polish it, but if you want to do it yourself I totally respect that.

zeffii:
@kalwalt  you are making a new image in save_bitmap _and_ in assign image.

kalwalt:
@zeffii  @ly29 i will give another try to it, i think i'm close to solve it.

ly29:
@kalwalt 
Ok good luck!

zeffii:
to test if an image exists, the longform is this
```python

...

    if image_name in bpy.data.images:
        return bpy.data.images[image_name]
    else:
        return bpy.data.images.new(name=image_name, size, size, alpha=False)
```
but maybe you need to test the size of the existing image, and if it doesn't match the desired size, then you delete the image and recreate it. (not sure if you can adjust the image size on the fly. )

ly29:
Use this assign image instead
```python
def assign_image(image, size ,buffer):
    import numpy as np # move to top of file
    np_buff = np.empty(len(image.pixels), dtype=np.float32)
    np_buff.shape = (-1, 4)
    np_buff[:,:] = np.array(buffer)[:,np.newaxis]
    np_buff[:,3] = 1
    np_buff.shape = -1
    image.pixels[:] = np_buff
    return image
```

kalwalt:
I think this not change so much: i get another error:

```
B
<bpy_struct, Image("image_name.png")>
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/generator/script1_lite.py", line 68, in execute
    getattr(context.node, self.fn_name)()
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py", line 291, in save_bitmap
    np_buff[:,:] = np.array(buf)[:,np.newaxis]
ValueError: could not broadcast input array from shape (0,1) into shape (65536,4)

```

maybe was better before to merge the functions.  Not sure if i solved the question of the buffer.

ly29:
Because buf is empty.

Sadly what you are trying to do right now won't work, node instance aren't really like normal python classes instances so it won't work out the way you are trying to do right now.

ly29:
Helpful hint, make a method called ```get_buffer(self)``` and use in ```process``` to get the buffer.
Then reuse it in ```save_bitmap``` instead of ```buf=texture```.

ly29:
Also don't override init.

kalwalt:
> Helpful hint, make a method called get_buffer(self) and use in process to get the buffer.
Then reuse it in save_bitmap instead of buf=texture.

this make sense!

> Also don't override init.

ok i will delete it.

nodes/viz/texture_viewer.py
```diff
@@ -165,6 +161,10 @@ def __init__(self):
         items=theme_mode_options, default="default", update=updateNode
     )
 
+    def get_buffer(self,texture):
+        self.texture = texture
+        return self.texture
+
```
kalwalt: @ly29 Does it correct?  or maybe you mean another thing? but i suppose this is not right...

ly29: No that is very far from what I meant.

```python
    def get_buffer(self):
        data = self.inputs['Float'].sv_get(deepcopy=False)[0]
        size_tex = size_tex_dict.get(self.selected_mode)
        total_size = size_tex * size_tex
        if len(data) < total_size:
            default_value = 0
            new_data = [default_value for j in range(total_size)]
            new_data[:len(data)] = data[:]
            data = new_data
        elif len(data) > total_size:
            data = data[:total_size]
        # and then in init texture
        texture = bgl.Buffer(bgl.GL_FLOAT, total_size, data)
        return texture

# in process
    def process(self):

        n_id = node_id(self)
        nvBGL2.callback_disable(n_id)

        if self.activate:
            texture = self.get_buffer()
```
```

Comment (ly29) [inline nodes/viz/texture_viewer.py]: Then you can use ```get_buffer``` in save_bitmap

Comment (kalwalt) [inline nodes/viz/texture_viewer.py]: i get it a image_name.png in /tmp folder, it is completely black but it is a progress!

Comment (kalwalt) [inline nodes/viz/texture_viewer.py]: `img.colorspace_settings.name = 'Linear'
`
Uncommenting this solve this issue, now i have an image as expected!

**Comment by kalwalt** [comment]
image are saved in tmp folder, but .bmp, .jpeg, .tga .tiff do not open in krita software and other viewer, only png image open correctly. 
bmp, tga, tiff, , jpeg need addictionally settings? it seems that is missed an header in the file I saw some posts in blenderexchange saying that a header is required at the beggining of the bitmap in that cases but i thought that maybe was old infos...

**Comment by ly29** [comment]
try to save the image from image editor in some other format and see what happens

**Comment by zeffii** [comment]
@kalwalt   could you make a simple bpy script that correctly saves a png , but also makes a .bmp that doesn't load in another program.

i bet a 2*4 px image would already be enough to show what's wrong.

**Comment by ly29** [comment]
https://docs.blender.org/api/blender_python_api_2_78_release/bpy.types.Image.html#bpy.types.Image

**Comment by kalwalt** [comment]
>  could you make a simple bpy script that correctly saves a png , but also makes a .bmp that doesn't load in another program.

@zeffii @ly29 i think it's the best approach to figure out this issue.

@ly29 i red the api doc, but most of these attributes are readonly. So for example if you want to assign to the new image obj a depth value you will receive an  AttributeError:

```
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/generator/script1_lite.py", line 68, in execute
    getattr(context.node, self.fn_name)()
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py", line 292, in save_bitmap
    img.depth='8'
AttributeError: bpy_struct: attribute "depth" from "Image" is read-only


```

So i think such kind of settings must be added before `images.new` or in another way. Need deep investigations on this.

**Comment by ly29** [comment]
@kalwalt
Can you save as other image format from the image editor?

**Comment by zeffii** [comment]
i think you should do bpy.data.images.new(     )  with the right format, then it's already got all the info. then do .save()

**Comment by zeffii** [comment]
http://blender.stackexchange.com/questions/6865/save-image-as-for-example-png-by-a-script

**Comment by kalwalt** [comment]
made this simple test script:

```
import bpy
image_name = 'image-test.jpeg'
img = bpy.data.images.new(name= image_name,width=4,height=4,alpha=False,float_buffer=True)
pix=[]
for p in range(16):
    pix.append(p)
img.pixels = pix
img.filepath_raw = '/tmp/' + image_name
img.save()
print('saved!')
```
image saved but shotwell will not open this is the error message:

`Error interpreting JPEG image file (Not a JPEG file: starts with 0x89 0x50)`

**Comment by ly29** [comment]
Set ```file_format``` perhaps...

**Comment by kalwalt** [comment]
@zeffii interesting link, need to undestand if i can use it or not in this case.

**Comment by kalwalt** [comment]
> oh you definitely need to set .file_format as @ly29 suggests.

i have already set but that setting not solve the issue. I m looking into the `image` instead of `images` method.

**Comment by zeffii** [comment]
not sure that will work..
http://blender.stackexchange.com/questions/643/is-it-possible-to-create-image-data-and-save-to-a-file-from-a-script

There's not a whole lot written about this on blenderstackexchange, and i've never needed to save anything other than a png :/   i'll bow out.

**Comment by ly29** [comment]
Though desirable it isn't a show stopper. I think you should focus on getting everything thing else working and then we can deal with file formats later.

**Comment by kalwalt** [comment]
> Though desirable it isn't a show stopper. I think you should focus on getting everything thing else working and then we can deal with file formats later.

Probably you are right, i will  some other try and if i fail i will come back to focus on getting  everything working.

**Comment by zeffii** [comment]
you can strip out the palette stuff too, add it back later if you need to customize some color stuff

**Comment by kalwalt** [comment]
ok i think this [link](https://www.blender.org/forum/viewtopic.php?t=24348) will help a lot: i have tried the tga export and seems to work. But i think it's better do this in another step i will focus on the whole node.

**Comment by kalwalt** [comment]
back to the svae image. this script works, it save a.bmp as expected:

```python
import bpy
image_name = 'image-test.jpeg'
size = 4, 4
img = bpy.data.images.new(name= image_name,width=size[0],height=size[1],alpha=False,float_buffer=True)

#pix=[]

pixels = [1.0] * (4 * size[0] * size[1])
pixels = [None] * size[0] * size[1]
for x in range(size[0]):
    for y in range(size[1]):
        # assign RGBA to something useful
        r = x / size[0]
        g = y / size[1]
        b = (1 - r) * g
        a = 1.0

        pixels[(y * size[0]) + x] = [r, g, b, a]

# flatten list
pixels = [chan for px in pixels for chan in px]
#img.file_format = 'JPEG'
img.pixels = pixels
#img.filepath_raw = '/tmp/' + image_name
#img.save()
scene=bpy.context.scene
scene.render.image_settings.file_format='BMP'

img.save_render('/tmp/img.bmp',scene)
print('saved!')

```

the most important thing is `img.save_render(..)` and the scene settings

kalwalt:
result with jpeg:
img

kalwalt:
@zeffii @ly29 When you have time, test please It should work now!
The only thing to refine is the fact that the .tga format is saved with extension .targa

ly29:
I will have look tomorrow. 
Brief notes:
- import numpy to top of file
- some pep8

zeffii:
>   .tga format is saved with extension .targa

```diff
- image_name = image_name + '.' + img_format.lower()
+ image_name = image_name + '.' + img_format.lower().replace('targa', 'tga')
```

zeffii:
be sure to add the parenthesis (i updated the diff ) @kalwalt

zeffii:
they might not be strictly necessary, but it removes ambiguity.

zeffii:
also you can drop the `from mathutils import Vector`, by adding this _property-decorated_ function to your Node's class

```python
    # class level function
    @property
    def xy_offset(self):
        a = self.location[:]
        b = int(self.width) + 20
        return int(a[0] + b), int(a[1])
```

```diff
- x, y = [int(j) for j in (self.location + Vector((self.width + 20, 0)))[:]]
+ x, y = self.xy_offset()
```

kalwalt:
@zeffii the decorated function give me this error:

```
File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/core/update_system.py", line 322, in do_update_general
    node.process()
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py", line 178, in process
    x, y = self.xy_offset()
Node Texture viewer had exception 'tuple' object is not callable

```

ly29:
Drop the ()

zeffii:
oops.

kalwalt:
Trying to save a 1024X1024 i get this error:

```
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/generator/script1_lite.py", line 68, in execute
    getattr(context.node, self.fn_name)()
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/viz/texture_viewer.py", line 237, in save_bitmap
    np_buff.shape = -1
ValueError: could not broadcast input array from shape (1048576,1) into shape (262144,4)

```

kalwalt:
I will check tomorrow for other issues

ly29:
The image is size is wrong

kalwalt:
> The image is size is wrong

this happens only for .png format, i can save a 1024x1024 px in the other(bmp,tga,tiff,jpeg), that's strange... i will re-check tomorrow.

zeffii:
image
@kalwalt  that's gotta be one of the happiest pixel images ever :)

kalwalt:
> that's gotta be one of the happiest pixel images ever :)

@zeffii yeah, years of hard study for making this!

anyway the png issue disappear restarting blender, maybe was a temp issue...

ly29:
@kalwalt 
There small, stylistic issues to sort out:
First should install a pep8 linter in your text editor and see how it lights up the file. 

Function signature of  ```draw_buttons```, I really want to keep this consistent, the little time you save by typing ```l``` instead of layout isn't worth the tradeoff, especially since ```l``` is extra bad variable name.  Also I think general  ```draw_buttons_ext``` should draw the same options as the ```draw_buttons``` + extra options. (*)
```python
```

In general I am very happy with this node I want it merged asap.

(*) I know I am myself guilty of some of these sins in older code I have been hurt by not following these recommendations.

kalwalt:
@ly29 i use most of the time Atom editor and a bit the blender editor with an addon extension, but maybe better eclipse editor. 
i've made the changes you suggested, in `draw_buttons` and `draw_buttons_ext` i will re-check the PEP8 in another editor. Also some lines are longer than others, Should i break them and extend in a vertical manner?

ly29:
Atom is what I use, just install ```linter-pycodestyle```

kalwalt:
Good to know, i'm looking into it!

kalwalt:
ok, now with `linter-pycodestyle` i have a list of PEP8 warning: i'm fixing them.

zeffii:
add lines like 
`# pylint: disable=C0326` 

if you find the fix to the warning is getting in the way of readable code.

kalwalt:
i have no warnings now, should be ok?  in some cases was lines too long i had to break them. not sure if was a good or bad idea.

ly29:
It looks mostly easier on the idea now.

This I would restore as one line.
```python
        image_name = image_name + '.'
        + (img_format.lower().replace('targa', 'tga'))
```

zeffii:
`size_tex_list`  looks worse with char 80 limit, 

in the pylinter preferences you should be able to add a list of warnings to ignore, mine looks like this
``` 

    "disable": ["C0103", "C0111", "R0914", "C0301", "R0903", "W0703"],   

```

kalwalt:
@ly29 i agre with you, that was one of my doubts. I'm going to restore it.

p.s. `linter-pycodestyle` great discovery

kalwalt:
the node need also the docs, i'm going to create it.

ly29:
@kalwalt 
Hmm noticed now that I don't use ```linter-pycodestyle``` but ```linter-flake8```.
Recommend you switch as well.

Main plus 
- its easier to configure
- it tracks unused symbols etc as well

ly29:
@kalwalt 
I feel comfortable with merging this, there are some issues but I feel they be easily sorted out with testing.

@zeffii ?

zeffii:
yeah. merge it. it's had plenty of 'eye-time' :)

kalwalt:
wow! It was worth the time spent to develop it! :smile:

PR #1267, opened 2017-02-24
Texture Viewer Node upgrades

kalwalt:
Bitmap image is saved in the desired folder now, should i fix also some other things
wait please

### DIFF
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -37,11 +37,11 @@ class SvTextureViewerOperator(bpy.types.Operator, SvGenericCallbackWithParams):
     bl_idname = "node.sv_texview_callback"
     bl_label = "Execute a function on the calling node"
 
+
 class SvTextureViewerDirSelect(bpy.types.Operator, SvGenericDirectorySelector):
     """ Pick the directory to store images in """
     bl_idname = "node.sv_texview_dirselect"
     bl_label = "Pick directory"
-
 
 
 size_tex_list = [
@@ -61,27 +61,24 @@ size_tex_dict = {
 }
 
 bitmap_format_list = [
-    ('PNG', 'png format', 'save texture in .png format', '', 0),
-    ('TARGA', 'tga format', 'save texture in .tga format', '', 1),
-    ('TARGA_RAW', 'tga_raw format', 'save texture in .tga(raw) format', '', 2),
-    ('TIFF', 'tiff format', 'save texture in .tiff format', '', 3),
-    ('BMP', 'bmp format', 'save texture in .tiff format', '', 4),
-    ('JPEG', 'jpeg format', 'save texture in .jpeg format', '', 5),
-    ('JPEG2000', 'jpeg2000 format', 'save texture in .jpeg2000 format', '', 6),
-    ('OPEN_EXR_MULTILAYER', 'exr multilayer format', 'save texture in .exr multilayer format', '', 7),
-    ('OPEN_EXR', 'exr format', 'save texture in .exr format', '', 8),
-    ('IRIS', 'iris format', 'save texture in .rgb format', '', 9)
+    ('PNG', '.png', 'save texture in .png format', '', 0),
+    ('TARGA', '.tga', 'save texture in .tga format', '', 1),
+    ('TARGA_RAW', '.tga (raw)', 'save texture in .tga(raw) format', '', 2),
+    ('TIFF', '.tiff', 'save texture in .tiff format', '', 3),
+    ('BMP', '.bmp', 'save texture in .tiff format', '', 4),
+    ('JPEG', '.jpeg', 'save texture in .jpeg format', '', 5),
+    ('JPEG2000', '.jp2', 'save texture in .jpeg (2000) format', '', 6),
+    ('OPEN_EXR_MULTILAYER', '.exr (multilayer)', 'save texture in .exr (multilayer) format', '', 7),
+    ('OPEN_EXR', '.exr', 'save texture in .exr format', '', 8),
     # ('HDR','hdr format', 'save texture in .hdr format', '', 9),
 ]
 
 format_mapping = {
-    'IRIS': 'rgb',
     'TARGA': 'tga',
     'TARGA_RAW': 'tga',
     'JPEG2000': 'jp2',
     'OPEN_EXR_MULTILAYER': 'exr',
     'OPEN_EXR': 'exr',
-    'IRIS': 'rgb'
 }
 
 
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -247,10 +247,9 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
     def copy(self, node):
         self.n_id = ''
 
-    # def save_bitmap(self, image_name='image_name', alpha=False):
     def save_bitmap(self, alpha=False):
-        image_name = self.image_name or 'image_name'    # if self.image_name was empty it will give a default
-
+        # if self.image_name was empty it will give a default
+        image_name = self.image_name or 'image_name
         # save a texture in a bitmap image
         # in different formats supported by blender
         buf = self.get_buffer()
@@ -279,10 +278,10 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
         scene.render.image_settings.file_format = img_format
         # get the path for the file and save the image
 
-        desired_path = os.path.join(self.base_dir, self.image_name + extension)   #####
+        desired_path = os.path.join(self.base_dir, self.image_name + extension)
 
         img.save_render(desired_path, scene)
-        print('Saved!  path is:', desired_path)
+        print('Bitmap saved!  path is:', desired_path)
 
 
 def register():
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -281,7 +281,16 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
         # in different formats supported by blender
         buf = self.get_buffer()
         img_format = self.bitmap_format
-        extension = '.' + img_format.lower().replace('targa', 'tga')
+        print(img_format)
+        if img_format == 'TARGA':
+            extension = '.' + img_format.lower().replace('targa', 'tga')
+        elif img_format == 'JPEG2000':
+            extension = '.' + img_format.lower().replace('jpeg2000', 'jp2')
+        elif img_format == 'OPEN_EXR':
+            extension = '.' + img_format.lower().replace('open_exr', 'exr')
+        else:
+            extension = '.' + img_format.lower()
+        # extension = '.' + img_format.lower().replace('targa', 'tga')
         image_name = image_name + extension
         dim = size_tex_dict[self.selected_mode]
         width, height = dim, dim
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -45,20 +45,14 @@ class SvTextureViewerDirSelect(bpy.types.Operator, SvGenericDirectorySelector):
 
 
 size_tex_list = [
-    ('EXTRA_SMALL', 'extra_small 64x64px', 'extra small squared tex: 64px', '', 64),
-    ('SMALL', 'small 128x128px', 'small squared tex: 128px', '', 128),
-    ('MEDIUM', 'medium 256x256px', 'medium squared tex: 256px', '', 256),
-    ('LARGE', 'large 512x512px', 'large squared tex: 512px', '', 512),
-    ('EXTRA_LARGE', 'extra_large 1024x1024px', 'extra large squared tex: 1024px', '', 1024)
+    ('XS', 'XS', 'extra small squared tex: 64px', '', 64),
+    ('S', 'S', 'small squared tex: 128px', '', 128),
+    ('M', 'M', 'medium squared tex: 256px', '', 256),
+    ('L', 'L', 'large squared tex: 512px', '', 512),
+    ('XL', 'XL', 'extra large squared tex: 1024px', '', 1024)
 ]
 
-size_tex_dict = {
-    'EXTRA_SMALL': 64,
-    'SMALL': 128,
-    'MEDIUM': 256,
-    'LARGE': 512,
-    'EXTRA_LARGE': 1024
-}
+size_tex_dict = {item[0]: item[4] for item in size_tex_list}
 
 bitmap_format_list = [
     ('PNG', '.png', 'save texture in .png format', '', 0),
@@ -175,7 +169,7 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
     selected_mode = EnumProperty(
         items=size_tex_list,
         description="Offers display sizing",
-        default="SMALL",
+        default="S",
         update=updateNode
     )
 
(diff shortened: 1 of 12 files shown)


kalwalt:
Shall i had also other bitmap format? For example .hdr .exr even if i'm not sure if it make sense now for a grayscale image.
other possible format are:

 TARGA_RAW , CINEON, DPX, OPEN_EXR_MULTILAYER, IRIS(this very old)

zeffii:
give me like an hour.. and then pull from master.

zeffii:
some variable renaming must happen..
you have a function called `save_bitmap` and a enum called `bitmap_save` - and it's giving me palpations.

zeffii:
you should rebase, or pull merge now.

kalwalt:
@zeffii ok will pull now

kalwalt:
i added .exr and .jpeg2000 format eventually .hdr ( now is commented) it can be saved but since you can save only a grayscale image will not saved correctly. Also i haven't a software for displaying .jpeg2000 (maybe Gimp but i haven't it installed now, i use krita)

kalwalt:
ok, Gimp open jpeg2000 but should be in the extension `.jp2`  or `.jpc` `.jpx`. Jpeg2000 format described [here](https://jpeg.org/jpeg2000/) better than standard jpeg.  irenamed the extension to .jp2 and gimp will open it. i will change the code.

kalwalt:
@zeffii @ly29 Should we add now a slider for compression level and compression type buttons?

nodes/viz/texture_viewer.py
```diff
@@ -284,10 +286,16 @@ def save_bitmap(self, operator):
...
+        elif img_format == 'TARGA_RAW':
+            extension = '.' + img_format.lower().replace('targa_raw', 'tga')
         elif img_format == 'JPEG2000':
             extension = '.' + img_format.lower().replace('jpeg2000', 'jp2')
         elif img_format == 'OPEN_EXR':
             extension = '.' + img_format.lower().replace('open_exr', 'exr')
+        elif img_format == 'OPEN_EXR_MULTILAYER':
+            extension = '.' + img_format.lower().replace('open_exr_multilayer', 'exr')
+        elif img_format == 'IRIS':
+            extension = '.' + img_format.lower().replace('iris', 'rgb')
```
zeffii: make a mapping dict.

kalwalt: you are right, but i started with one more format added and ended with a lot of other...! :smile:

zeffii:
```
format_mapping = {'IRIS': 'rgb', ........}
extension = format_mapping.get(img_format, img_format.lower())
```
if img_format isn't in format_mapping, it will return `img_format.lower()`  :)   only a hint.

kalwalt: I will do tomorrow!

zeffii:
```diff
+ format_mapping = {'IRIS': 'rgb', ........}
+ extension = '.' + format_mapping.get(img_format, img_format.lower())
```

kalwalt:
Some format type are saved but  probably with wrong depth or lack of channels, need to be fixed this probably when we have color modes (RGB. RGBA). I need to investigate on that .

ly29:
Honestly have anybody ever used JPEG2000??? For odd format like that export to blender image editor and them por souls save from there. 
Tiff and png and exr. Okay but seriously what formats make sense. Don't go overboard. Be conservative

kalwalt:
ok, so i will not add CINEON and DPX format... :smile:

kalwalt:
But Jpeg2000 even if is not used is better than standard jpeg, don't know why it's not used...

ly29:
Priorities like this
- save as packed image
- color rgba/rgb

kalwalt:
> Priorities like this

> - save as packed image
> - color rgba/rgb

may i do in this pull request? or it is better a separate request?

and so if we can output to the editor how many format we let to save? 

@zeffii @ly29

zeffii:
@kalwalt  tricky..  try to not pre-empt every feature request you might get in the future. handle the formats you intend to use. keep notes on potential formats, implement them if you need..

kalwalt:
texture_viwer_color_mode
working also on color mode layout. But maybe better buttons?

kalwalt:
texture_viwer_color_mode01
with buttons

zeffii:
`row.prop(self, 'whatever_enum', expand=True)`  would look fine.

kalwalt:
@zeffii much better!

kalwalt:
texture_viz_rgb_test

Texture viewer color RGB test
but color buffers need adequate size,  the next step to do.

kalwalt:
texture_viz_rgb_test01

Almost done!

kalwalt:
i' m trying to save a png as RGB or RGBA, look the message from terminal:

````
buffer  tex buff length: 16384
buffer  tex buff length: 16384
buffer  tex buff length: 16384
PNG
width is: 64
length img pixels: 16384
pass to...
ValueError('could not broadcast input array from shape (16384,1) into shape (4096,4)',)
````
the problem is in these lines: 

````python
np_buff.shape = (-1, 4)
np_buff[:, :] = np.array(buf)[:, np.newaxis]
````

**Comment by ly29** [comment]
Yeah, that will have to be redone...

**Comment by ly29** [comment]
Untested
```python
# rgb
np_buff[:,:3] = np.array(buf)
# rgba
images.pixels[:] = buf
```

kalwalt:
@ly29 i will try tomorrow, thank's!

kalwalt:
This code seems to work, i can save in BW, RGB, RGBA color mode for all the bitmap format. I have not yet committed, Still need some testing and reviewing. The images saved seems with less color, i don't undertand why, i will show a comparison later.

````python

        img.scale(width, height)
        print('width is: {0}'.format(width))
        print('length img pixels: {0}'.format(len(img.pixels)))
        if col_mod == 'BW':
            np_buff = np.empty(len(img.pixels), dtype=np.float32)
            print('pass to...')
            np_buff.shape = (-1, 4)
            np_buff[:, :] = np.array(buf)[:, np.newaxis]
            if col_mod == 'RGB':
                print("passing data from buf to pixels RGB")
                np_buff[:, :3] = np.array(buf)
            np_buff[:, 3] = 1
            np_buff.shape = -1
            img.pixels[:] = np_buff
        elif col_mod == 'RGBA':
            print("passing data from buf to pixels RGBA")
            # this works for RGBA!
            img.pixels[:] = buf

        # get the scene context
        scene = bpy.context.scene
        # set the scene quality to the maximum
        scene.render.image_settings.quality = 100
        # set different color depth
        if img_format == 'JPEG2000' or 'TIFF':
            scene.render.image_settings.color_depth = '16'
        else:
            scene.render.image_settings.color_depth = '8'
        # set compression level to no compression(0)
        scene.render.image_settings.compression = 0
        print('settings done!')
        # scene.render.image_settings.color_mode = col_mod
        scene.render.image_settings.file_format = img_format

        # get the path for the file and save the image
        desired_path = os.path.join(self.base_dir, self.image_name + extension)

        img.save_render(desired_path, scene)

        print('Bitmap saved!  path is:', desired_path)
````

Apart this i have some small issues with the texture_viz, the fonts sometimes became wierd and also icons, as alpha blending, disappear. Did you encounter this issue?

**Comment by kalwalt** [comment]
texture_viz_freeze
@zeffii @ly29 issue like this, see the fonts in the Text editor.

**Comment by zeffii** [comment]
@kalwalt  thats bgl settings bleeding over into the next context..  anything you do in bgl.enable() needs to be disabled.

**Comment by zeffii** [comment]
we should maybe look into a way to collect the current state of the bgl machine before going into our draw calls, and restore back to that state when leaving it.  ( perfect for a context manager)

**Comment by zeffii** [comment]
(At least... that's what it looks like)

**Comment by zeffii** [comment]
https://github.com/nortikin/sverchok/pull/1273

hovering over the mode shows info.. also uses the draw_label function to reflect current mode

texview_gif

**Comment by kalwalt** [comment]
> thats bgl settings bleeding over into the next context.. anything you do in bgl.enable() needs to be disabled.

or re-enabled? i will look into it again..

> we should maybe look into a way to collect the current state of the bgl machine before going into our draw calls, and restore back to that state when leaving it. ( perfect for a context manager)

do you refer to the drawcallbacks right?

how can i get this latest layout? git pull or other command?

**Comment by kalwalt** [comment]
Could be this [line](https://github.com/kalwalt/sverchok/blob/d6451a6a99b076e4a046689a22c811a385837dcd/nodes/viz/texture_viewer.py#L132) the reason of the issue? Maybe wrong parameters...
[glTexEnv](https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glTexEnv.xml)

**Comment by zeffii** [comment]
see  https://github.com/nortikin/sverchok/compare/texview_ui_mod...kalwalt:feature_tex_viz_saving

**Comment by zeffii** [comment]
probably want the comparison in reverse..

**Comment by zeffii** [comment]
anyway, the significant changes are here
```
https://github.com/nortikin/sverchok/pull/1273/files
```
less verbose is better for stuff that you only have to read once.

**Comment by kalwalt** [comment]
Later I will add  your changes too @zeffii

**Comment by kalwalt** [comment]
Cool!

texture_viz_new_layout

File: nodes/viz/texture_viewer.py [hunk 2]
```diff
@@ -236,7 +257,21 @@ def draw_buttons_ext(self, context, layout):
...
         layout.separator()
-
+        row = layout.row()
+        row.prop(self, 'color_mode_save', expand=True)
+        layout.separator()
+        if img_format == 'PNG':
+            row = layout.row()
+            row.prop(self, 'compression_level', text='set compression')
+            layout.separator()
+        if img_format == 'JPEG':
```
kalwalt: Jpeg and jpeg2000 share same quality settings but if i did so:
 if img_format == 'JPEG' or 'JPEG2000': the condition will ever respected and quality button will show for every bitmap format. How we compact them?

nodes/viz/texture_viewer.py
```diff
@@ -348,16 +390,20 @@ def save_bitmap(self, operator):
...
-            if col_mod == 'RGB':
-                print("passing data from buf to pixels RGB")
-                np_buff[:, :3] = np.array(buf)
             np_buff[:, 3] = 1
             np_buff.shape = -1
             img.pixels[:] = np_buff
+        elif col_mod == 'RGB':
+            print("passing data from buf to pixels RGB")
+            np_buff = np.empty(len(img.pixels), dtype=np.float32)
+            np_buff[:, :3] = np.array(buf)
```
kalwalt: this not works we should find the solution:
```
passing data from buf to pixels RGB
IndexError('too many indices for array',)
```

nodes/viz/texture_viewer.py
zeffii: `if img_format in {'JPEG','JPEG2000'}:`      not sure I understand the question.

nodes/viz/texture_viewer.py
zeffii: doesn't it need to be reshaped first?
```python
np_buff = np.empty(len(img.pixels), dtype=np.float32)
np_buff.shape = (-1, 4)
np_buff[:, :3] = np.array(buf)
```

i haven't tried.

nodes/viz/texture_viewer.py
kalwalt: exactly!

nodes/viz/texture_viewer.py
kalwalt: going to try

kalwalt: i get:

```
buffer  tex buff length: 12288
col_mod is: RGB
col_mod_s is: RGBA
img_format is: PNG
width is: 64
length img pixels: 16384
passing data from buf to pixels RGB
ValueError('could not broadcast input array from shape (12288) into shape (4096,3)',)

```

zeffii:
I renamed the title of this Pull request to correspond more closely to what's happening.

nodes/viz/texture_viewer.py
```diff
@@ -277,44 +362,80 @@ def save_bitmap(self, operator):
...
+        desired_path = os.path.join(self.base_dir, self.image_name + extension)
 
         img.save_render(desired_path, scene)
-        print('Saved!  path is:', desired_path)
+
+        print('Bitmap saved!  path is:', desired_path)
 
 
 def register():
+    bpy.utils.register_class(SvTextureVieverSliderCompression)
```
zeffii: what?

kalwalt: i must delete i was tryng to add a slider with an operator, that is not  the rigth approach.

zeffii:
use a wrapper for `updateNode` like in the Turbulence node. `def changeMode(self, context):`

call `self.process()` at the end of your wrapper if the code inside the wrapper didn't already cause a process call.

kalwalt:
This the starting point rgb image from a `noise` node:
texture_viz_rgb_test_rgb

this is the result as rgb(a) bitmap:
sv_test

kalwalt:
i will delete the switch (bw, rgb, rgba) in the properties panel. The idea was to save in a different format while saving. I think it's a feature that we can add in the future if we desire, not for the moment. I will focus now on custom texture size display and the bug with the viz.

kalwalt:
custom_texture_format
Custom texture size display, not yet with button layout but just to have an idea...

kalwalt:
custom_texture_buttons_rgb
With width and height buttons, will accept input also... this to do... :smile:

nodes/viz/texture_viewer.py
```diff
@@ -220,12 +241,26 @@ def xy_offset(self):
         return int(a[0] + b), int(a[1])
 
     def get_buffer(self):
+        # width=0
+        # height=0
+        size_tex = 0
         # data = self.inputs['Float'].sv_get(deepcopy=False)[0]
         data = np.array(self.inputs['Float'].sv_get(deepcopy=False)).flatten()
+        #### This code below stall the machine ####
```
kalwalt: Does it the right way to access to the width and heigth data of the custom texture?

kalwalt:
> use a wrapper for updateNode like in the Turbulence node. def changeMode(self, context):

> call self.process() at the end of your wrapper if the code inside the wrapper didn't already cause a process call.

i have to do, haven't had time for this.

nodes/viz/texture_viewer.py
```diff
@@ -101,8 +102,8 @@ def simple_screen(x, y, args):
     size = args[1]
     texname = args[2]
 
-    width = size
-    height = size
+    width = args[3]
+    height = args[4]
```
zeffii:
```diff
-    texture = args[0]
-    size = args[1]
-    texname = args[2]

-    width = args[3]
-    height = args[4]
+    texture, size, texname, width, height = args
```

simple unpacking.

nodes/viz/texture_viewer.py
kalwalt: this code is wrong do NOT use it:  the `total_size` of the Buffer with the custom texture size became: 

```
206  *  124 = 
25544

 25544 * 25544 =

652495936

652495936 * 3=
1957487808  Pixels!!!
```
how to kill a computer.... :disappointed: 

I will fix it.

nodes/viz/texture_viewer.py
kalwalt: ok!

kalwalt:
custom_texture_works_fine

Now we can resize the texture on the fly!

nodes/viz/texture_viewer.py
kalwalt: it was that, solved!

nodes/viz/texture_viewer.py
kalwalt: done!

kalwalt:
I think this  [link](https://docs.blender.org/api/blender_python_api_2_78_3/gpu.offscreen.html) may help for the viz bug.

PR #1269, opened 2017-02-25
Custom Icons - Icon Manager node - WIP

DolphinDream:
I added a new icon grid node to this branch to assist with managing SV custom icon design, layout. The previous code written for the script node was converted into a node (alpha category for now) and added new features, cleaned up the code, fixed some bugs and beautify the node with some custom icons.

Attached are some blend files to use with this node. 

As you adjust the iconGrid layout orientation, wrapping and scale, the grid cells readjust their location and the empties already “locked” onto a grid cell follow the cell. As you create new empties with the predefined SV icon name (SV_ICON_NAME) and scrub the timeline the new empty will snap to its corresponding location in the icon grid. (Assuming that icon name is listed in the sverchok/iconList.json file. This file will be updated to include all the icons we need to support in SV nodes).

sv-iconmanager-example2
sv-iconmanager-example3

icongridnode

[IconGridLayout.blend.zip](https://github.com/nortikin/sverchok/files/800998/IconGridLayout.blend.zip)
[iconGridNodeDev.blend.zip](https://github.com/nortikin/sverchok/files/800999/iconGridNodeDev.blend.zip)


ly29:
We should allow a way for a separate add on to plugin to the menu for things like this

DolphinDream:
@ly29  you mean have a separate blender add-on that somehow uses sverchock nodes and injects itself into the sverchock node tree and the SV search menu?

DolphinDream:
Btw, what does Sverchock mean? I see that verchock (сверчок) means cricket (not sure if that is the insect or the game) :) as for the S (C) i have no clue.

zeffii:
i like the Vector in/out icons btw. those can't come soon enough.

ly29:
The system has no knowledge about the nodes so to speak( some execptions). 
If you just write a node in a separate name space that poll true and can be added everything works out. 
Basically if the appear in the menu...

DolphinDream:
@ly29 And you are pondering the idea o a separate add-on because you don’t think such icon grid node should not be included in sverchock ?

ly29:
Not this specific case perhaps, I haven't looked into it. But it feels like low overhead for thing that would be better for some nodes already in the system.

DolphinDream:
@ly29  you mean it would be easy to do what the icon grid node does using nodes already in the system? I think I may be possible though I’m not sure how easy. (personally I don’t know SV well enough to do the sort of things this icon grid node does. Perhaps someone can show me an example ? :)

Here are some design inputs that I wanted an icon layout setup to be able to do, whether it’s via an iconGrid node or via a set of existing system nodes:

- list the icons belonging to a certain category together and in a certain order (for generators, transforms, etc. or for other categories we may come up with that are not necessarily part of the nodosphere).  The JSON file was the way I got around solving this. Not only I can group the icons in categories this way, but I can also preserve the order in which they are listed, matching the search menu order. (I referred to blender’s icon wiki for an example of how blender organizes its icons: https://wiki.blender.org/index.php/Dev:2.5/Doc/How_to/Add_an_icon). DONE
- list the icons on a grid with a certain width (wrapping longer sequence of icons in a category on separate grid lines). DONE
- be able to identify and snap (locate) empties in the scene having a given name (SV_ICON_NAME) to the grid. These are the icon parent recipients to keep all mesh related to an icon together. DONE
- be able to change the size of the grid and have the icon cells reorganize and the locked empties locations follow the cells they are “locked” to. DONE
- be able to change the direction in which the icons are listed on the grid (left->right & down->up, and all the other 7 directions/orders). DONE
- be able to identify the locations of the icons currently locked on the grid (e.g. locations of the locked empties with given icon name pattern SV_ICON_NAME that exist in the scene) so that a rendering camera can switch from one icon to the next in order to do the rendering. This way you only render the icons existing in the scene and avoid having to render empty cells. DONE
- be able to traverse the icons existing in the scene and render them one at a time, save them to disk with the name given by the icon id (lowercased) SV_TORUS -> sv_torus.png.  (to be done)
- be able to separate the icons apart (the scale factor)  DONE
- be able to add names next to each icon cell (the operator name and/or the SV_ICON_NAME) DONE
- be able to center the grid to the origin DONE
- have compact mode (leave no empty cell spaces between icons due to category separation) (to be done)
- sort the icons alphabetically on the grid (to be done)
- add category cells of different colors behind the icons cells belonging to one category or add labels next to the colums/rows hosting nodes belonging to a given category (to be done)
- be able to provide a subset of icons (names) to be re-rendered from the grid (e.g. render only the newly added icons or re-render a subset of icons in case their design changed). This would allow to render only the icons we need and not all of them. (to be done) 

These are some of the design inputs I used to guide the icon grid node development. You can argue that not all of them are necessary (but are nice to have), though even with the essential features in place I can’t imagine how easy would it be to do all these using the existing system nodes.

ly29:
No I don't think it is easy to replicate what it does. 
But I see it more as a library/devel tool than a core node. At a glance. As I said I haven't looked into it. 

What would like is to make it easy to make nodes that work with system without expanding the core node set too much.

nortikin:
@DolphinDream hi!
i imagin tumbler in properties for this node. we can use it as developers, it is cool.
SVERCHOK means СВеРЧОК - Система Векторного Рабочего Черчения и Объктного Конструирования
that means Vector Work Drawing and Object Constructing System (VWDOCS)
Initially...

DolphinDream:
Thanks @nortikin . Could you please elaborate on the “tumbler in properties” idea? I’m not sure what you mean by it. I’ll continue to develop this node to make it more useful for developers to create and manage SV icons, so any suggestions are very welcomed.  😄

DolphinDream:
Btw, I did know what the acronym stand for from sverchok’s site, however I was wondering if the word itself means something else in ukrainian 😀 (like cricket? which is what comes up as google translation)

nortikin:
Russian cricket, yes. I'm russian ukrainian.
Properties mean ctrl+alt+U addons tab sverchok box
SverchokPreferences defined in sverchok-master/settings.py

DolphinDream:
It’s been a while since the last update on this PR 😄 so.. here is a revamped version of the node (still under developing).

Aside from code cleanup, optimizations and socket name renaming .. some of the new features include:
    - select icon from dropdown list
    - navigate icons in the list (prev/next buttons)
    - render current/selected icon
    - render all icons (currently freezes UI until all icons are rendered)
    - generate empties (the parents of the icon meshes)
    - generate and configure camera (used to follow selected icon)
    - add output of the camera location for the selected icon
    - add grid/icon/category meshes to be used to render the icon frames

sv-iconmanager-2

DolphinDream:
Should I be concerned with the “Travis-CI” giving “All checks have failed” error ? I’m not familiar with this new github tool. The branch (rebased on master) seems to be OK to me.

DolphinDream:
Anybody has any suggestions how to make the “Render All Icon” feature work interactively instead of freezing the UI? It would be nice to get a progress indicator of some sort showing what icon is currently rendered, the percentage of the whole rendering completed, and update the camera, object selection etc in the viewports, the render result in the UVImage views etc . Perhaps the “Render All” operation could enter a “render all mode” (instead of rendering in a loop) and have the node somehow trigger another update (without user interaction) and render one icon per update cycle until all the icons are rendered, at which point it will exit the “render all mode” and stop triggering the node updates. Any feedback appreciated.

DolphinDream:
What is the proper way to have this icon only show / be registered when the preference debugging mode is enabled ?

DolphinDream:
@nortikin Do you want this feature as a node or as a panel (as @zeffii suggested) ? If displaying the icon grid/cell frames is not a desired feature then perhaps a panel implementation could take care of everything else the current node implementation does.

zeffii:
i reall really think this should be a separate add-on. really.

DolphinDream:
> i reall really think this should be a separate add-on. really.

I agree. I already started on making this an add-on. I'll revisit this.

Issue #1270, opened 2017-02-25
Brainstorming about a Hexagonal Grid node 


DolphinDream:
Let’s talk hexa ! :) 

I’ve been playing around with various nodes to create hexagonal grids and I am a bit frustrated I can’t create a hexagonal grid to my linking, preferably using a simple setup. Jimmy’s videos inspired me but only to leave me even more frustrated that we can’t create a hexagonal grid with a simple node setup. I know the brick node can do some magic with getting this done, but it’s not quite doing it for me, since the edges of the brick layout are cropped. Maybe I could get rid of those vertices at the edge of the brick grid with some list magic nodes, but why going through so much trouble just to create a simple hexagonal grid?   

There may be other node setups that can do this, other than the brick node, like creating a staggered set of points using the points on the plane node, with every other line shifted by some amount to get the triangular stagger pattern.. which then can be used to place a properly scaled circle (6 points) to create the hexagonal grid. Yeah.. this is how my node setup looks like, using two lines + circle generators: 

![hexa grid setup](https://cloud.githubusercontent.com/assets/2083719/23334724/06d9fe74-fb73-11e6-82d5-6d6b523cd79e.png)

Not pretty, but it does the job. I can even put it in a monad and call it a day. But i’m not quite satisfied with that either. I want a magical hexagonal grid node that can do something like this out of the box:

http://www.redblobgames.com/grids/hexagons/implementation.html

Checkout the section 4.4 (map shape). where the hexagons can be arranged to form triangle shapes, hexagonal shapes, rectangular shapes, rhombic shapes etc. 

Anyway, before I dive into node coding I figured maybe I can brainstorm around this a bit with y’all.

DolphinDream:
Having fun with hexa grids ☺️  No SV node was hurt in the making of this grid. 😉 

![hexagridlevels-1k](https://cloud.githubusercontent.com/assets/2083719/23335596/467070aa-fb86-11e6-93f1-793c6fb17655.jpg)

enzyme69:
This excites my brain, because suddenly something make sense a bit more! This is like Kaleidoscope Puzzle.

DolphinDream:
And now let’s abuse the SV nodes to make the above :) This is based on a series of triangle meshes subdivided, to use their vertices to replicate a hexagon. 

![hexagridlevels-1](https://cloud.githubusercontent.com/assets/2083719/23335909/522befda-fb8d-11e6-9c9f-05f26bbc2a06.png)

[...]

PR #1271, opened 2017-02-27
Hexa grid node 

DolphinDream:
Based on the [Hexagonal grid brainstorming](https://github.com/nortikin/sverchok/issues/1270) I borrowed @zeffii’s code as a starting point to implement a hexa grid node. For the time being I added this as a generator and made it an “alpha” node in the search menu. The node generates vertices in a staggered pattern suitable to create hexagonal grid patterns (e.g. in combination with a hexagon / 6-point circle generated by the circle node). The node has a layout type feature (inspired by this [site](http://www.redblobgames.com/grids/hexagons/implementation.html)) that allows the generation of the grid points within one of four layout types: RECTANGULAR (zeffii’s script code example) as well as TRIANGULAR, DIAMOND and HEXAGON layout. The node inputs set vary based on the layout type: a Nx, Ny, Radius inputs for the rectangular layout and Level, Radius for the other layouts. I also added a centering option to center the grid around the origin. The node is vectorized.

hexagridnode-layouts-example1

hexagrid-layouts

See the attached blend file used in conjunction with the hexaGrid node to generate the above screenshots.

[hexaGridNodeDev.blend.zip](https://github.com/nortikin/sverchok/files/802500/hexaGridNodeDev.blend.zip)


DolphinDream:
Replicating a circle instead of a hexagon on a hexa grid.. ta-da : the Flower of Life (sacred geometry) 

hexagrid-floweroflife

hexagrid-floweroflife-trim

DolphinDream:
Some demos GIFs 
hexagrid-demo2

enzyme69:
This is BEAUTIFUL!!!

zeffii:
stick it in `generators_extended` please, because that's where it will end up.

DolphinDream:
If there are no other suggestions to further improve the node/code, I’m done with implementing this node for now.  Thanks for all the feedback so far.

portnov:
This is nice; but maybe this node could also generate edges and faces?

DolphinDream:
@Portnov I was actually thinking about that but I wasn't sure if that would be ok.

zeffii:
```python
- centers
- verts
- edges
- pols
```
but only process verts/edges/pols if verts is connected.

DolphinDream:
@Zeffii would it be ok to invoke the circle code internally to get the V:E:P ?

zeffii:
i encourage it. any reuse.. checkout 
https://github.com/nortikin/sverchok/blob/master/utils/geom.py#L89-L146
for exactly this.

zeffii:
whether you choose to offer a 'separate' or 'join' toggle (an internal mesh join) is up to you. I have no preference. If you need library code for that i'm sure we have it somewhere.

zeffii:
mesh join : https://github.com/nortikin/sverchok/blob/master/utils/sv_mesh_utils.py

DolphinDream:
Thx @zeffii . Can you demystify the socket output format for me? Currently the hexa grid node output values are [[ [x,y,z] ]]. What are the rules for the output format for the verts edges and polys as far as the nestedness goes. I see the circle node’s output (non vectorized) is [[(x,y,z),(x,y,z) ]] and vectorized [ [(x,y,z), (x,y,z)], [(x,y,z), (x,y,z)] ].

DolphinDream:
And what is the best way to replicate mesh at some vertex location? I’m assuming that once I have the list of hexa grid center I would have to replicate a hexagon (VEPs) to the location of these centers and either join them or keep them separate before feeding them into the VEP outputs.

DolphinDream:
Btw, should we be consistent with the output/input socket naming? E.g. Faces / Polygons, Verts/Vertices ? What is the preferred notation?  Vertices/Edges/Faces ? Vertices/Edges/Polygons ? Verts/Edges/Polys ?

DolphinDream:
HexaGrid with mesh preview. Not quite fully vectorized yet. 
I was also considering adding the rotation input (angle) into the node to allow the grid and the tiles to be rotated by some arbitrary angle (or multiple of 30). I need some hints as to how to rotate the grid of points [[(x,y,z),...]]. (note: the angle input in the node currently only rotates the tiles not the grid itself).

hexagrid-demo4

zeffii:
@DolphinDream our vertex output lists is always wrapped, a generator either outputs a single mesh or multiple meshes 
```python
# vN = (x, y, z)   or  [x, y, z] ...but preferably tuple
vert_list = [v0, v1, v2, v3....]
single_vert_list = [vert_list]
multiple_vert_list = [vert_list, vert_list, vert_list, ....]
```
oh.. looks like you've figured it out.

zeffii:
we _should_ be consistent with output naming but there's heterogeneity among nodes. The output names you picked are sensible.

zeffii:
>  And what is the best way to replicate mesh at some vertex location?

'best', is a battle between what's easiest to read.. and what's fastest to execute :) As soon as you use 'Vector()` you'll get a speed penalty (more noticeable on larger output counts).

if the scale is 1.0 and the mesh is 'joined' you aren't merging verts/edges (like doing a remove doubles) ? -- just asking you don't have to, it's a rather slow operation best left to the `remove doubles` node.

nodes/generators_extended/hexa_grid.py
```diff
@@ -0,0 +1,321 @@
...
+
+def generate_tiles(radius, angle, join, gridList):
+    verts, edges, polys = circle(radius, radians(angle), 6, None, 'pydata')
+
+    vertList = []
+    edgeList = []
+    polyList = []
+    for grid in gridList:
+        for cx, cy, cz in grid:
+            verts2 = [(x + cx, y + cy, z + cz) for x, y, z in verts]
```
zeffii:
```python
verts2 = [(x + cx, y + cy, 0.0) for x, y, _ in verts]
```

nodes/generators_extended/hexa_grid.py
```diff
@@ -0,0 +1,321 @@
...
+    return vertList, edgeList, polyList
+
+
+class SvHexaGridNode(bpy.types.Node, SverchCustomTreeNode):
+    ''' Hexa Grid '''
+    bl_idname = 'SvHexaGridNode'
+    bl_label = 'Hexa Grid'
+    sv_icon = 'SV_HEXA_GRID'
+
+    def update_type(self, context):
```
zeffii: we'll need to talk about this.

zeffii: very few nodes use this, there's some side effects to using `.hide`. Those nodes that do have not been scrutinised very closely.
https://github.com/nortikin/sverchok/search?l=Python&p=3&q=.hide&utf8=%E2%9C%93

noodles aren't broken when a `socket.hide = True`

DolphinDream:
@Zeffii I am not removing doubles now. I thought it would be better off left to the remove doubles node. Besides user may want to separate the tiles anyway so no need to do to extra work to remove them. (I thought)

nodes/generators_extended/hexa_grid.py
DolphinDream: What is the proper way to switch new sockets in/out based on the parameters selection if hide is not ok?

nodes/generators_extended/hexa_grid.py
DolphinDream: Yeah, I stayed away from using vectors since the tiling is just a simple xy translation and I figured it would be faster to just do cx/cy addition.

nodes/generators_extended/hexa_grid.py
zeffii: i'll be the first to admit that this sucks... but we 
```python
self.inputs.remove(socket)
self.inputs.new(...)
```
https://github.com/nortikin/sverchok/blob/master/nodes/vector/math_mk2.py#L123-L141

zeffii: meaning you'd place the sockets that do _change_ at the bottom of the node.

zeffii: and do something like
```python
    def update_type(self, context):
        if self.gridType == "RECTANGLE":
            self.inputs["Level"].hide = True
            self.inputs["NumX"].hide = False
            self.inputs["NumY"].hide = False
        elif self.gridType == "TRIANGLE":
            self.inputs["Level"].hide = False
            self.inputs["NumX"].hide = True
            self.inputs["NumY"].hide = True
        elif self.gridType == "DIAMOND":
            self.inputs["Level"].hide = False
            self.inputs["NumX"].hide = True
            self.inputs["NumY"].hide = True
        elif self.gridType == "HEXAGON":
            self.inputs["Level"].hide = False
            self.inputs["NumX"].hide = True
            self.inputs["NumY"].hide = True

        updateNode(self, context)

--------- vs ------------------

    def update_type(self, context):
        if self.gridType == "RECTANGLE":
            self.inputs["Level"].hide = True
            self.inputs["NumX"].hide = False
            self.inputs["NumY"].hide = False
        elif self.gridType in {"TRIANGLE", "DIAMOND", "HEXAGON"}:
            self.inputs["Level"].hide = False
            self.inputs["NumX"].hide = True
            self.inputs["NumY"].hide = True

        updateNode(self, context)

--------- vs ------------------

    def update_type(self, context):
        inputs = self.inputs
        named_sockets = ['NumX', 'NumY']

        if self.gridType == "RECTANGLE":
            if "Level" in inputs:
                inputs.remove("Level")
            for socket_name in named_sockets:
                if not socket_name in inputs:
                    inputs.new("StringsSocket, socket_name).prop_name = socket_name

        elif self.gridType in {"TRIANGLE", "DIAMOND", "HEXAGON"}:
            if not "Level" in inputs:
                inputs.new("StringsSocket, "Level").prop_name = "Level"
            for socket_name in named_sockets:
                if socket_name in inputs:
                    inputs.remove(inputs[socket_name])



```

zeffii: i'm trying to find the thread where we first discussed not using `.hide` a year or two or more ago..

hide might be OK if you also add code that forces disconnecting .links on hidden sockets.

DolphinDream:
@zeffii if you remove input sockets you also cannot do this in process: input_level = self.inputs[“Level”].sv_get()[0] (if the levels are not available) so this code will have to be sanitized too.

DolphinDream:
And there is another problem.. the socket names are not the same as the props names .. e.g. NumX vs numx :) So, I’ll have to code around this to be able to use your socket recreation. Hide was so much easier :) I’m curious to know why was hide NOT a good idea.

zeffii:
The code I suggested indeed wasn't complete, it was offered only as a starting point. Yes process requires extra sanitizing to see if a socket exists.

- hidden sockets don't drop their links automatically, meaning the node can be getting more calls to process from upstream nodes than a user might realize ( testable )
- (bad reason but a reason nonetheless) iojson doesn't store socket properties , gists won't work flawlessly out of the box.

@ly29 anything to add or is the restriction on .hide cargocult, it's been a while since I looked the DAG

It would be worth adding socket properties to iojson to keep code like this simpler

zeffii:
Maybe we add a .hide_safe to the socket class, that auto unlinks hidden socket.links.

zeffii:
The usage of `.hide` should be its own issue-topic, and lead to some resolution.
https://github.com/nortikin/sverchok/issues/1285  ( please continue `.hide` related questions there  - i will fill it out)

zeffii:
>  I’m curious to know why was hide NOT a good idea.

it's an obviously sane solution, but a discussion is needed before that hits master. I don't (despite what you might think..hehe) target .hide to mess with you.

DolphinDream:
Here’s a vectorization preview. 

Joining all the replicated tiles with the straight forward vectorization did not make a whole lot of sense so I decided to join the hexagonal tiles per vectorized grid, meaning the hexa grid parameters (level, radius, numx, numy) are vectorized to generate different grids, then the scale input is vectorized in combination with the list of grids generated via vectorizing the grid inputs and the generated  mesh for each grid+scaled hexagon is then joined.  

I’m not at all good at using the vectorization to its full potential so I can’t say what’s the best way to vectorize this node and what kind of join option to offer but this scheme seems to allow some sort of cool effects. Any ideas how to better vectorize this are welcomed. Perhaps I can offer different ways to join the vectorized meshes? Join ALL of them ? Join them along the “scale dimension” ? 

hexagrid-demo6

DolphinDream:
replicating my hand drawn hexa tile pattern using SV 

hexatiles3
hexatiles6

portnov:
That. is. amazing.
(:

DolphinDream:
Thank you @portnov :smile: I'd like to see what more advanced SV users could do with this vectorized node node.

DolphinDream:
Some thoughts on vectorization. Here’s what I did so far:

The parameters that define a particular hexa grid (radius, N1, N2, level), if any have multiple value they will generate a set of distinct hexa grids. The tile parameters (scale, angle) if they have single values they apply to all vectorized grids. If the tile parameters have multiple inputs the vectorization is done by combining tile values with a grid: (scale, angle, grid). This way you can adjust the tile size (scale, angle) for all the tiles within one grid.

And as for the joining option:

The non-vectorized centers (hexa grid points) are typically created as:
`[c1, c2, ... , c{N}] (where c=(x,y,z)) `

and the vectorized centers are:
`[ [c1, c2, ... , c{N1}], ... , [c1, c2, ..., c{Nm}] ]`
Each inner list is a set of centers in one grid out of multiple grids (when the **grid** parameters are vectorized, multiple grids are created).

The joined vectorized centers are:
`[ c1, c2, ... , c{sum(Nm)} ]`  
Note: not sure how useful this would be for but it can be done this way.

For the hexa tiles, the vectorized verts (in the hexa tiles) are generated as:
`[ [[v1...v6], ... ,[v1...v6]], ... , [[v1...v6], ... ,[v1...v6]] ]` <= a set of grids of tiles of verts

with a per grid join option on:
`[ [v1...v{6*N1}], ... , [v1...v{6*Nm}] ]` <= a set of grids with joint tile verts
This gives a set of meshes with the tiles verts in each grid joined.

with a join all option on:
`[ v1...v{6*sum(Nm)} ]` <= a set of joint grids with joint tile verts
Note: I’m not sure how useful this last one would be, but this gives the completely joint mesh. 

Let me know if this is in line with what (useful) vectorization should look like.

DolphinDream:
Angle vectorization demo: The birth of the Hexa Gallaxy :)

https://gist.github.com/06b334335fb488d956db50975ab65d32

hexagrid-hexagallaxy-setup

hexagrid-hexagallaxy

portnov:
@DolphinDream i believe the conflict in index.md is the only reason why this is not merged yet :)

zeffii:
3 merges in 15 minutes.

*runs away*

DolphinDream:
Wow! Victory is mine! thx @zeffii  :wink:  Sorry, @portnov, it looks like I missed your note about resolving conflicts. I assume the index.md conflict was a pretty easy to fix. :+1:

kalwalt:
good job @DolphinDream! thanks for this node!

PR #1280, opened 2017-02-28
add auto attr node and connect to new material

zeffii:
image

will connect if render engine is set to Cycles.


zeffii:
have fun with this one :)

zeffii:
@enzyme69  you mentioned this once a while back and i have been too swamped to commit it.

enzyme69:
@zeffii Thanks again man, another additional streamline process, I actually can make an animated GIF that shows this process. Will try!

PR #1286, opened 2017-03-02
Polyline mk1 rev3

zeffii:
image

- [x] add show wires.
- [ ] interpret multiple input meshes as one Curve object.. and Matrices to transform the verts and not the object (this is a new mode... no ETA)


zeffii:
expect breaking changes to this node within the next 2 days, layouts using the existing version of the node may produce unexpected results.

PR #1291, opened 2017-03-02
polyline mk1 rev5

zeffii:
image

- [x] add show wires.
- [ ] interpret multiple input meshes as one Curve object.. and Matrices to transform the verts and not the object (this is a new mode... no ETA)


zeffii:
image

Vector Interpolate MK3 is fantastic btw.

PR #1293, opened 2017-03-02
Polyline mk1b code reuse

zeffii:
sum of the current changes

- [x] better naming (zfill 4 )  `name.0000`
- [x] now and children are found using ID properties (and reuses code from sv_viewer_utils).
- [x] vert_list is extended if `len(matrices) >  len(vert_list)`
- [x] has radii, radius per vertex
- [x] has twist, twist per vertex
- [x] allows bevel object selection
- [x] has bevel objects-in socket
- [x] has generated curve objects-out socket
- [x] has `show wires` toggle
- [x] has add caps toggle for when bevel object is used.
- [x] merge mode: Matrices no longer are pushed to multiple objects, but instead transform each set of verts before joining them into one Curve Object.
- [x] handle no matrix in Single mode
- [x] shorter functions


zeffii:
image

enzyme69:
@zeffii If the line has branching, I normally use Skin Modifier. But in case such thing happen, does Sverchok has "split line for PolyLine when there is branching" node?

zeffii:
> does Sverchok has "split line for PolyLine when there is branching" node? 

of course not. (yet) but most of the needed code exists already. in sudocode
```
# separate branches, but keep in one mesh (means doubles)
for v in verts:
    if num_edges_that_use_this_v(v) >=3:
        make a new edge, with a new vertex for every extra branch
for entire mesh of Verts,Edges separate all disjoint edge loops
```

zeffii:
btw, that should be its own node first, at least.

zeffii:
not massively tested

zeffii:
image

zeffii:
image

zeffii:
image
anyway.. enough of these.

zeffii:
image

zeffii:
test file 
[bluegold4_2017_03_05_18_42.zip](https://github.com/nortikin/sverchok/files/819643/bluegold4_2017_03_05_18_42.zip)

zeffii:
i forgot to set the default `use_smooth=True`.

zeffii:
capture_vaschnikt333_edit

zeffii:
found a but where the vert count must be either identical for each incoming set.. or increasing... but can't be varied. This needs a fix ;)

PR #1300, opened 2017-03-03
Add new Torus 2D node as generator (extended)

DolphinDream:
This node will basically generate a 2D ring (aka 2D torus) alowing for the same kind of settings the torus 3D offers, albeit adapted to the 2D space.

Minor/Major or Exterior/Interior radii could be adjusted to control the size and the width of the torus (ring) as well as the radial and circular sections. The radial sections can also be phased.

The node is pretty much a stripped down version of the Torus 3D node so it is fully vectorized.

Note: The property list was compacted a little more (compared to Torus 3D node) to appease zeffii's quest to squash the vertical scroll as much as possible without compromising readability or stepping on the pep8 nerves ;)

torus2dnode

### DIFF
--- a/nodes/generators_extended/torus2D.py
+++ b/nodes/generators_extended/torus2D.py
@@ -32,6 +32,7 @@ def torus_verts(R, r, N1, N2, p, Separate):
         N1  : major sections - number of RADIAL sections
         N2  : minor sections - number of CIRCULAR sections
         p   : radial section phase
+        Separate : separate the radial sections into separate vertex lists
     '''
     listVerts = []
 
@@ -236,13 +237,16 @@ class SvTorus2DNode(bpy.types.Node, SverchCustomTreeNode):
             input_R = input_RR
             input_r = input_rr
 
-        parameters = match_long_repeat([input_R, input_r, input_n1, input_n2, input_rp])
+        params = match_long_repeat([input_R, input_r, input_n1, input_n2, input_rp])
 
-        V, E, P = [self.outputs[n] for n in ['Vertices', 'Edges', 'Polygons']]
+        V, E, P = self.outputs[:]
         for s, f in [(V, torus_verts), (E, torus_edges), (P, torus_polygons)]:
-            if s.is_linked:
-                s.sv_set([f(n1, n2) if s != V else f(R, r, n1, n2, p, self.Separate)
-                          for R, r, n1, n2, p in zip(*parameters)])
+            if not s.is_linked:
+                continue
+            if s == V:
+                s.sv_set([f(R, r, n1, n2, p, self.Separate) for R, r, n1, n2, p in zip(*params)])
+            else:
+                s.sv_set([f(n1, n2) for _, _, n1, n2, _ in zip(*params)])
 
 
 def register():
--- a/nodes/generators_extended/torus2D.py
+++ b/nodes/generators_extended/torus2D.py
@@ -1,0 +1,273 @@
+# ##### BEGIN GPL LICENSE BLOCK #####
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software Foundation,
+#  Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# ##### END GPL LICENSE BLOCK #####
+
+import bpy
+from bpy.props import IntProperty, FloatProperty, BoolProperty, EnumProperty
+
+from math import sin, cos, pi
+
+from sverchok.node_tree import SverchCustomTreeNode
+from sverchok.data_structure import updateNode, match_long_repeat
+
+
+def torus_verts(R, r, N1, N2, p, Separate):
+    '''
+        R   : major radius
+        r   : minor radius
+        N1  : major sections - number of RADIAL sections
+        N2  : minor sections - number of CIRCULAR sections
+        p   : radial section phase
+    '''
+    listVerts = []
+
+    # angle increments (cached outside of the loop for performance)
+    da = 2*pi/N1
+
+    for n1 in range(N1):
+        theta = n1 * da + p     # radial section angle
+        sin_theta = sin(theta)  # caching
+        cos_theta = cos(theta)  # caching
+
+        loopVerts = []
+        s = 2 / (N2-1)  # caching
+        for n2 in range(N2):
+            rr = R + (n2*s - 1) * r
+            x = rr * cos_theta
+            y = rr * sin_theta
+
+            # append vertex to loop
+            loopVerts.append([x, y, 0.0])
+
+        if Separate:
+            listVerts.append(loopVerts)
+        else:
+            listVerts.extend(loopVerts)
+
+    return listVerts
+
+
+def torus_edges(N1, N2):
+    '''
+        N1 : major sections - number of RADIAL sections
+        N2 : minor sections - number of CIRCULAR sections
+    '''
+    listEdges = []
+
+    # radial EDGES
+    for n1 in range(N1):
+        for n2 in range(N2-1):
+            listEdges.append([N2*n1 + n2, N2*n1 + n2+1])
+
+    # circular EDGES
+    for n1 in range(N1-1):
+        for n2 in range(N2):
+            listEdges.append([N2*n1 + n2, N2*(n1+1) + n2])
+    for n2 in range(N2):
+        listEdges.append([N2*(N1-1) + n2, N2*0 + n2])
+
+    return listEdges
+
+
+def torus_polygons(N1, N2):
+    '''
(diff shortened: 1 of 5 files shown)


nodes/generators_extended/torus2D.py
```diff
@@ -0,0 +1,273 @@
...
+            input_r = input_rr
+
+        parameters = match_long_repeat([input_R, input_r, input_n1, input_n2, input_rp])
+
+        if self.outputs['Vertices'].is_linked:
+            vertList = []
+            for R, r, n1, n2, p in zip(*parameters):
+                verts = torus_verts(R, r, n1, n2, p, self.Separate)
+                vertList.append(verts)
+            self.outputs['Vertices'].sv_set(vertList)
```
zeffii: i want to point out that if you move the Separate parameter to the first position in torus_verts, this gets a bit lighter.
```python
        if self.outputs['Vertices'].is_linked:
            vertList = [torus_verts(self.Separate, *args) for args in zip(*parameters)]
            self.outputs['Vertices'].sv_set(vertList)
```
in python 3.6 you'd be able to write `torus_verts(self.Separate, *params)` or `torus_verts(*params, self.Separate)` , but in 3.5 the starred argumen must be last.
```python
        V = self.outputs['Vertices']
        if V.is_linked:
            V.sv_set([torus_verts(self.Separate, *args) for args in zip(*parameters)])
```

nodes/generators_extended/torus2D.py
```diff
@@ -0,0 +1,273 @@
...
+    listPolys = []
+    for n1 in range(N1-1):
+        for n2 in range(N2-1):
+            listPolys.append(
+                [N2*n1 + n2, N2*(n1+1) + n2, N2*(n1+1) + n2+1, N2*n1 + n2+1])
+
+    for n2 in range(N2-1):
+        listPolys.append([N2*(N1-1) + n2, N2*0 + n2, N2*0 + n2+1, N2*(N1-1) + n2+1])
+    listPolys.append([N2*(N1-1) + N2-1, N2*0 + N2-1, N2*0 + 0, N2*(N1-1) + 0])
+
```
zeffii: what's with the no-ops..`N2*0` ?

nodes/generators_extended/torus2D.py
```diff
@@ -0,0 +1,273 @@
...
+                edgeList.append(edges)
+            self.outputs['Edges'].sv_set(edgeList)
+
+        if self.outputs['Polygons'].is_linked:
+            polyList = []
+            for R, r, n1, n2, _ in zip(*parameters):
+                polys = torus_polygons(n1, n2)
+                polyList.append(polys)
+            self.outputs['Polygons'].sv_set(polyList)
+
```
zeffii:
```python
        E, P = self.outputs['Edges'], self.outputs['Polygons']
        for socket, func in [(E, torus_edges),(P, torus_polygons)]:
            if socket.is_linked:
                socket.sv_set([func(n1, n2) for _, _, n1, n2, _ in zip(*parameters)])
```
Maybe that's taking it too far, but that's how I would write these..because they are identical in signature.

enzyme69:
This can be handy and definitely a fun exercise.

I would use 3D torus and flatten it using Zeffii Rewire and remove doubles.

But your one will be more efficient.

Thanks!
-- 


- Jimmy Gunawan

nodes/generators_extended/torus2D.py
DolphinDream: I can clean these out. I wrote it this way to better understand the code as it shows that that is actually the edges at the zero index.  These no-ops are in the regular torus code btw. :) you missed those.

nodes/generators_extended/torus2D.py
DolphinDream: That buys you one line of code and sacrifices somewhat the readability  :) Also usually tent to arrange the arguments to a function in the order of their significance. The Radius is first in the list because that defines primarily what the function will do: create a band of some radius. The Separate is more of a “extra” parameters we provide to the function so that it can can arrange the generated vertices in a certain way for other type of processing later on using other nodes, which is why I place it last. This makes more sense to me, but, if you insist I can implement your way, no problem.

DolphinDream: Besides.. I’m not even sure the Separate is actually needed here. I ported it over from the Torus (3D) node where separating the slices makes sense, but for a 2D shape like this I don’t see where the use for Separate would be.

nodes/generators_extended/torus2D.py
DolphinDream: This is more of a distillation art than optimizing anything in my opinion. :) I personally like the way the previous code reads. The more condensed version is a bit obscure and it takes me a bit to translate it in my head.  However, knowing you’re on the quest to squeeze the vertical scroll as much as you can, your suggestion of course does the job. Btw, I’m curious, why the vertical scroll is such a big issue? Unless you’re programming on a 13” monitor, I can’t see the need to condense the code that much. Horizontal spread I understand, but vertical, not as much. ;)

DolphinDream: And how about this? :) I know i’m gonna regret it now, but I couldn’t help it.

V, E, P = [self.outputs['n'] for n in 'Vertices', 'Edges', 'Polygons']

DolphinDream: Or if you really want to nuke it.. I expect this should probably excite you even more :)

```python
V, E, P = [self.outputs[n] for n in ['Vertices', 'Edges', 'Polygons']]
        fv, fe, fp = torus_verts, torus_edges, torus_polygons
        for s, f in [(V,fv), (E,fe), (P,fp)]:
            if s.is_linked:
                s.sv_set([f(n1,n2) if s != V else f(R, r, n1, n2, p, self.Separate) for R, r, n1, n2, p in zip(*parameters)])
```

DolphinDream:
And now, ladies and gentlemen: The Hexa Grid and Torus 2D duet :)

hexagrid-torus2d-fol-setup
hexagrid-torus2d-fol

DolphinDream:
node vectorization demo.. 

https://gist.github.com/e88878b8aa9d663d1d077fcfb672e305

torus2dnode-vectorizationdemo1

nodes/generators_extended/torus2D.py
```diff
@@ -0,0 +1,256 @@
...
+            input_R = input_RR
+            input_r = input_rr
+
+        parameters = match_long_repeat([input_R, input_r, input_n1, input_n2, input_rp])
+
+        V, E, P = [self.outputs[n] for n in ['Vertices', 'Edges', 'Polygons']]
+        for s, f in [(V, torus_verts), (E, torus_edges), (P, torus_polygons)]:
+            if s.is_linked:
+                s.sv_set([f(n1, n2) if s != V else f(R, r, n1, n2, p, self.Separate)
+                          for R, r, n1, n2, p in zip(*parameters)])
```
zeffii: i do like iterating over VEP like that. But linesplitting the list comprehension is murky territory (especially with a nested if/else), It's a sign we're forcing things. We can still be explicit without being verbose .

I favour less nested over vertical compression :)

```python
        V, E, P = self.outputs[:]
        params = match_long_repeat([input_R, input_r, input_n1, input_n2, input_rp])

        for s, f in [(V, torus_verts), (E, torus_edges), (P, torus_polygons)]:
            if not s.is_linked:
                continue
            if s == V:
                s.sv_set([f(self.Separate, *args) for args in zip(*params)])
                # s.sv_set([f(R, r, n1, n2, p, self.Separate) for R, r, n1, n2, p in zip(*params)])
            else:
                s.sv_set([f(n1, n2) for _, _, n1, n2, _ in zip(*params)])
```

DolphinDream:
```python
V, E, P = self.outputs[:]
```
sure, but this assumes the outputs are necessarily given in THIS order and that no other outputs come before the VEP (both true for the moment). If we ever want to add another output before V this will break, while the explicit name version will not.

DolphinDream:
Alright.. Latest requests done. 

Note: I’m not particularly  proud of the separate moving to the beginning of the argument list. To be honest, this is the kind of constraints that makes the code in my view a little bit ugly and makes my coding experience less enjoyable. I like to see what parameters i’m sending to a function rather than package them into some obscure *args. I think aiming to distill the code to its finest should not be a goal in itself. If that’ the case, I would be better off writing in assembly :)

Either way, if there are no more suggestions for this node I think I'm done with it at my end.

zeffii:
> I like to see what parameters i’m sending to a function rather than package them into some obscure *args.

The function's own signature tells you what it _should_ be receiving. and match_long_repeat tells you the order of the parameters. Getting rid of certain boilerplate in the process function is fine, i only focus on that because it's about the only thing i regularly read. The functions called from process are assumed to be correct. 

> Moving the `separate' parameter, 

When blender uses python 3.6 you can move it back :)

If this makes the process less enjoyable, feel free to let it go in one ear and out the other.

DolphinDream:
I wonder if it would make more sense to convert the phase input to be degrees instead of radians.

portnov:
Maybe to call this "Ring" is more correct?

DolphinDream:
@portonov Certainly. That makes sense. I kept the “torus” name since the UI is similar to the 3D torus node. Ring, Donut, Torus .. either one is fine with me.

zeffii:
settle on a name for this node, and i'll merge it.

DolphinDream:
Hows Ring ?

zeffii:
yeah, ring's OK.

DolphinDream:
How about the phase input in degrees ? (instead of radians)

zeffii:
Degrees is fine.  lathe node uses that and matrix in too.

My preference is Tau `(range (0.0....1.0)`) . but it's not used anywhere. and i'm not a dictator.

DolphinDream:
Behold the power of the Ring.. node :)

PR #1309, opened 2017-03-06
small upgrades

zeffii:
fixing some things that annoy me most

- [x] add `bpy` to locals
- [x] add `self.user_filename`  (the ui will show `filename: [.........]` )
- [x] add new directive `filehandler / fh` to allow convenient user driven text-file loading.
- [x] auto inject output list variable names
- [ ] resolve the .blend reload resetting the node's parameters.

### DIFF
--- a/nodes/generator/script1_lite.py
+++ b/nodes/generator/script1_lite.py
@@ -320,7 +320,7 @@ class SvScriptNodeLite(bpy.types.Node, SverchCustomTreeNode):
             # snlite supplied custom file handler solution
             fh = snlite_info.get('display_file_handler')
             if fh:
-                layout.prop_search(self, 'user_filename', bpy.data, 'texts')
+                layout.prop_search(self, 'user_filename', bpy.data, 'texts', text='filename')
 
             # user supplied custom draw function
             f = snlite_info.get('drawfunc')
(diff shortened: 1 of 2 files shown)


zeffii:
cool, this works now..
```
"""
in param1 s d=0.2 n=2
in param2 s d=0.2 n=2
out verts v
fh
"""
verts.append([])

print(self.user_filename)
verts[0].append([(param1*i, param2*i, i) for i in range(10)])

```

kalwalt:
@zeffii Cool! I want to try this!

zeffii:
i'll fix the slider values issue in a new branch..   maybe

kalwalt:
so `inject` is not needed anymore in the header of the script for the `vectorize` func? we can just add `fh`?

zeffii:
directives:
```text
fh:    makes a filehandler / stringproperty / prop_search
inject:    inserts the variable 'parameters'
```
they're separate directives @kalwalt . vectorize doesn't need to be imported anymore, but you do need to state `vectorize(parameters)`.

kalwalt:
@zeffii now i get the difference. i think ScriptNode lite is becoming better.

zeffii:
I haven't used SN1 since SNLite hit master :)

zeffii:
https://github.com/nortikin/sverchok/pull/1310

kalwalt:
self_filename_test
It works, tested with a script that i'm developing for importing .ter (terragen) files into blender and sverchok.

zeffii:
: D

PR #1310, opened 2017-03-06
pstate fix

zeffii:
- [x] restore value from slider if it is not the same as the user provided default in the directive. ( not massively tested.. but i think this is how it's going to be)
- [x] don't use eval, sucker.


zeffii:
am wondering also about injecting `parameters` differently..
```python
                if self.inject_params:
                    parameters = eval("[" + ", ".join([i.name for i in self.inputs]) + "]")
```
blows chunks..   maybe
```
                if self.inject_params:
                    locals().update({'parameters': [locals().get(i.name) i in self.inputs ]})
```

zeffii:
i consider this done for the time being.. unless there are real bugs.

Issue #1147, opened 2017-01-17
Line MK2


zeffii:
![image](https://cloud.githubusercontent.com/assets/619340/22034239/eabff626-dceb-11e6-8d9d-c74345428784.png)

slightly tricky to compose the input lists for this, because unlike with the Circle node, the steps had been allowed to be non-constant. (With circle each segment is a constant angle away from the previous..)

I'm not saying this is the best approach to vectorizing a Line node, but it makes the most basic need a lot simple than the current one.

https://github.com/nortikin/sverchok/pull/1146

ly29:
Plane has the same issue. For both it would be nice with a start option instead of only positive...

zeffii:
![image](https://cloud.githubusercontent.com/assets/619340/22034524/e18579c2-dcec-11e6-9e47-841ce4fd7ddc.png)

zeffii:
like a start Vector?

ly29:
The thing is also that line is so easy to do with Vector in/UvConnect that I rarely get annoyed with it...
But I wish we had done the change a long time ago and standardised the generators.

zeffii:
uv connect feels incredibly bloaty for making edges for simple lines..

zeffii:
i almost want my own node folder with Lite versions of everything  :)

zeffii:
the current Node List needs a red wedding.

ly29:
Soon @zeffii, soon. Future nodes will be much lighter.

ly29:
But totally agree with the red wedding...

nortikin:
is it support different steps inside one node?
i propose vectorisation as is for line for everything.
Concept is simple at all
1 level of data - lines
2 level of data - data inside lines
3 level of data - more for sublists as sphere or plane when separated case.

I strongly propose to keep separated principle. I need it anyway.
Simplify? yes, but not loose this nestyness.
Especially for generators

ly29:
The concept might be real simple but if isn't uniform it gets unpredictable real soon which makes things very hard to use. Anyway...

#1112 FTW long term

zeffii:
@nortikin  it's never the intention to regress, but alternative versions (or modes) of the node will help other people who have different needs from a line node. Not a big deal for me either, i would rather have a Lite 'Edges from Verts sequence' node :) Float Range -> Vector In is a fine way to make lines.

PR #1312, opened 2017-03-07
Line Node MK2

DolphinDream:
- cleanup the code
- optimize the calculation
- sanitize the inputs
- and new X,Y,Z direction option to create line along X,Y or Z (no need to do rewire just to swich directions)

linenodemk2-example2

linenodemk2-demo1

linenodemk2-demo3

### DIFF
--- a/index.md
+++ b/index.md
@@ -38,6 +38,7 @@
     SvHexaGridNode
     SvRingNode
     SvPlaneNodeMK2
+    SvLineNodeMK2
 
 ## Analyzers
     SvBBoxNode
--- a/nodes/generator/line_mk2.py
+++ b/nodes/generator/line_mk2.py
@@ -20,7 +20,7 @@ import bpy
 from bpy.props import IntProperty, FloatProperty, BoolProperty, EnumProperty
 
 from sverchok.node_tree import SverchCustomTreeNode
-from sverchok.data_structure import updateNode, fullList
+from sverchok.data_structure import updateNode, fullList, match_long_repeat
 
 directionItems = [("X", "X", ""), ("Y", "Y", ""), ("Z", "Z", "")]
 
@@ -33,7 +33,7 @@ def make_line(steps, center, direction):
     elif direction == "Z":
         v = lambda l: (0.0, 0.0, l)
 
-    ss = -sum(steps) / 2 if center else 0
+    ss = - sum(steps) / 2 if center else 0
     verts = [v(ss)]  # starting vertex
     for s in steps:
         ss = ss + s
@@ -45,6 +45,7 @@ def make_line(steps, center, direction):
 
 
 class SvLineNodeMK2(bpy.types.Node, SverchCustomTreeNode):
+
     ''' Line MK2'''
     bl_idname = 'SvLineNodeMK2'
     bl_label = 'Line MK2'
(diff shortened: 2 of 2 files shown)


zeffii:
I did something similar.. https://github.com/nortikin/sverchok/issues/1147  but the orientation was a socket

enzyme69:
This Line MK2 thing has some kind of "Ratio" happening? Ask Nikitron about
it. He has a thing with LINE node. It can do some SV magic underneath, that
not many people know.

DolphinDream:
@enzyme69 the new LineMK2 is identical in behavior with the current Line node.. except I added the X/Y/Z orientation .. and cleaned up the code. What “ratio” are you referring to ?

DolphinDream:
Trying out some new ideas.
One thing I think i miss in the line node is the ability to define the size of the line. Adjusting the number of vertices and steps always grows the line in an unpredictable way. I added a normalize (just to check it out) option to allow the line to use the same inputs as before but normalize the length to a given size. 

linenodemk2-demo3

portnov:
Maybe add modes similar to RangeFloat node?
Line can be defined by:
a) number of vertices and step
b) line length and number of vertices

enzyme69:
screen shot 2017-03-08 at 21 14 30

enzyme69:
screen shot 2017-03-08 at 21 19 17
screen shot 2017-03-08 at 21 19 08

Line somewhat has accumulate function built in.

DolphinDream:
yeah, I see that. But it’s not accumulation.. it’s a different way of vectorizing (i guess). It seems that if you split the “step” list into separate lists, each list in the list of lists creates a separate mesh (line) which is what you see when you replicate this with the other line along Y. The first line is the one having the step equal to the START value in the range of float.. while the last replicated line (if you match the replications with the number of elements in the range of floats) you’ll get a line with the step equal to the STOP value of the range of floats.  So the line node knows how to vectorize this way when it gets a lists of lists. It’s a bit of a non conventional way of using the generators  based on what I know so far about SV, but maybe this is what the generators were intended to do from the very beginning? @nortikin may testify to that. 

Sure, I can make the Line MK2 handle the list of lists the same way.

DolphinDream:
Original Line node vectorization added to Line MK2 .. side by side comparison.

Note: the Normalization/Size is not yet implemented for this version. Besides, the normalization/size only makes sense for a single line as it can keep a variable steps/verts line within a given size. With split list vectorization each line is generated with one N and one S (equal size step) thus normalization would invalidate the step size, while the number of verts per line will still vary, thus resulting in equal size lines with different numbers of verts. 

linenodemk2-demo4

DolphinDream:
Now.. to take @nortikin vectorization madness a step further :) we could feed the step input a list of list of lists which will allow each replicated line to have a variable step division. :)

And that could spiral fast into a mind boggling philosophical question  :)

“There was a young man who said though, it seems that I know that I know, but what I would like to see is the I that knows me when I know that I know that I know.”
― Alan W. Watts

DolphinDream:
And with normalization (fixed size, variable verts per replication):

linenodemk2-demo5

DolphinDream:
Having fun with some RingLine replication :) 

https://gist.github.com/af5a5f607e80a68c604e7602b046801b

linenodemk2-ringlinereplication

enzyme69:
That's neat, I just remember there is this new RING node! Cool :) Great
addition!

DolphinDream:
With the updated, more compact UI:

linenodemk2-demo7

zeffii:
please resolve conflict and indentation and let's merge this sucker.

DolphinDream:
Will do shortly. I was already working on it 🤓

DolphinDream:
@zeffii resolved. ready for merge.

Issue #1320, opened 2017-03-08
Working in edit mode - Manual selection of vertices/edges/faces in 3D view


portnov:
General idea.
When creating useful meshes, you often want to select "these" faces (or vertices, or edges) and do something with them: extrude or delete or whatever.
Currently you can do it with masks; but the main problem is what numbers to write. Index viewer is of use, but the problem is: when you looked at indices and wrote them to List Input node to be a mask; then next time someone opens blend file, he will have no idea what these numbers in List Input mean, why these numbers were entered, not other.

Looking for competitors: in Grasshopper, you can select parts of created mesh interactively in 3D view at any time.
It would be cool to have similar behaviour in Sverchok, but I have no idea if it is possible in Blender at all.

So, currently suggested solution is:
* Add "Selected verts mask", "Selected edges mask", "selected faces mask" outputs to BMesh Viewer node.
* Add "Get selection" button to it. Node should remember selected verts/edges/faces when you press this button.
* (not implemented yet): BMesh Viewer should use remembered selection when generating new meshes. So it will be easy to slightly modify previously remembered selection.

So the workflow will be like:
* Generate something with BMesh Viewer
* Select that mesh and go to Edit mode
* Select some verts/edges/faces
* In BMesh Viewer node, press "Get Selection". It will remember indicies of verts/edges/faces you selected, and pass them to output.
* Now you can switch UPD in this BMesh Viewer off, and delete generated mesh, if you do not need it. Node will still remember what verts/edges/faces you selected.
* Use Mask outputs of BMesh Viewer somewhere.
* It is possible to use several BMesh Viewers in one node tree. For example: extrude these, then extrude one of extruded faces...

If at some point you decided that you need to change selection:
* enable UPD for corresponding BMesh Viewer (it will generate new mesh)
* select that new mesh and go to edit mode
* (not implemented yet): selection will be the same as you selected first time
* Adjust selection
* Press "Get Selection" button in BMesh Viewer again.
* Now you can switch UPD off and remove mesh.

This workflow is not super-nice, but i think it is better than nothing.

Current implementation is in: https://github.com/portnov/sverchok/tree/viewer_bmesh_select

portnov:
![output1](https://cloud.githubusercontent.com/assets/284644/23697163/da004894-0410-11e7-907e-4501681efa9d.gif)

portnov:
Open questions:
* Do I need to create "BMesh viewer mk3" to add new outputs?
[...]

PR #1322, opened 2017-03-08
Select vertices/edges/faces parametrically

portnov:
Most obvious ways of "parametrically select" something are:

*   Select "leftmost" or "rightmost" vertices. More exactly, select vertices that have (direction, Vertex) >= X, where direction is specified vector and X is some number. And also select edges and faces that include selected vertices.
*    Select by normal direction. Select faces (and faces?), that have (Normal(Face), direction) >= X. And also select vertices that are included into selected faces.
*    Select "in range". Select vertices, that have distance(Vertex, Center) <= Radius. And also select edges and faces that include selected vertices.

See also #1320 .


portnov:
output2

kalwalt:
@portnov I see that is a select by normal option: with this node should be possible also add a displacement along normals?

portnov:
@kalwalt 
This node only selects edges/faces/verts by some criteria. What to do with them you decide with other nodes. For example, you can use "Calculate normals" + "Move" to move along normals.

Btw, I think I can add yet another mode: select by cylinder. (Endless) cylinder can be defined by a point, a vector (direction), and radius.

enzyme69:
@portnov And don't forget Dynamic Paint selector :D It's quite LIVE to select Vertices. It's weird, but works on Frame 1./

portnov:
@enzyme69 
Interactive selections are discussed in #1320. There @zeffii said he is in general against it. So here we are discussing non-interactive aka parametric selections.

zeffii:
> There @zeffii said he is in general against it

not against... and i won't even resist nice solutions,  just.. i hesitate to change BMeshViewer MK2 for features I don't need. BMViewerMK3 should be the goal..not changing MK2.

portnov:
Added to new modes: "by plane" and "by cylinder".

By plane:

by-plane-1
by-plane-2
by-plane-3

portnov:
By cylinder:

by-cylinder-1
by-cylinder-2

portnov:
Also do not forget that it is possible to combine masks by logic functions:

by-sphere-not

portnov:
Well, I accidentally made a tire... :)

tire-1

portnov:
@zeffii 
I believe this is more or less ready for merge.

zeffii:
Thank you @portnov  this covers us in most selection needs..

DolphinDream:
Neato :) What about being able to select random vertices? or every other vertex (1010101) ? or every 3rd vertex (001001001)? or two vertices spaced by 3 vertices (11 000 11 000)? :) This looks like some sort mask value type of selection. I know this i possible with some list node magic, but if we are to have a vertex selection node it would be nice if we could select based on these kind of mask patterns. Just my 2¢

portnov:
@DolphinDream 
I think it is possible right now if you create mask with "List input" node, and maybe apply also "List Match" node with "Cycle" mode.

DolphinDream:
What would be the node setup where I have, let's say, a line of 12 verts and I want to offset the vertices at the index location given by this pattern: 0000 1 00 1 0000 ? (which is: offset the 5th an 8th verts along y, let's say).

portnov:
@DolphinDream 
line-mask

https://gist.github.com/anonymous/d60325695bfb20192a5d41f504d44662

DolphinDream:
@portnov neat. thx for the hint. That’ll do. 

However... I thought I’d take this to the next level and made up a concoction I call the “Transform Select Node” :) for lack of a better name. It sort of expands on the “mask vertices” node but it adds more to it to make it easier not only to separate the vertices but also to transform/group them (see the demo):

**Inputs**:

- mask # mask 0/1 with default repeat 0,1,0,1 if mask input is not linked
- vertices
- polyedges
- matrix True # matrix to transform the True (mask=1) verts
- matrix False # matrix to transform the False (mask=0) verts

**Outputs**:

- Vertices # Original Vertices but transformed by MT and MF
- PolyEdges # original polyEdges (pass through)
- PolyEdges O # polyEdges in the original set connecting the T set with the F set
- Vertices T # vertices transformed by MT
- PolyEdges T # polyEdges in original set that are also in T set
- Vertices F # vertices transformed by MF
- PolyEdges F # polyEdges in original set that are also in F set

transformselectnode-demo1

DolphinDream:
And a circle demo with scale transform:

transformselectnode-demo2

DolphinDream:
And a mesh-select node + transform-select node combo :)

transformselectnode-demo3

portnov:
Cool :)

PR #1339, opened 2017-03-10
Framenode mk2

zeffii:


zeffii:
moving frame_info.py to old should happen soon.

PR #1340, opened 2017-03-10
Stateful snlite

zeffii:


zeffii:
- I don't want to force a try/except there
- maybe a single preprocessing step to acquire locals from calling setup() if found.

PR #1343, opened 2017-03-10
Stateful and uidraw

zeffii:
various function can be refactored, but I want to be certain this works OK first.

### DIFF
--- a/node_scripts/SNLite_templates/Empties_out.py
+++ b/node_scripts/SNLite_templates/Empties_out.py
@@ -3,7 +3,7 @@ in draw_size        s   d=2.0    n=2
 in matrices         m   d=[[]]   n=0
 out obj_generated   o
 """
-import bpy
+
 from mathutils import Matrix
 
 scene = bpy.context.scene
--- a/node_scripts/SNLite_templates/UI_custom_props2.py
+++ b/node_scripts/SNLite_templates/UI_custom_props2.py
@@ -1,12 +1,9 @@
 """
 in objects_in o d=[[]] n=0
 out stuff     v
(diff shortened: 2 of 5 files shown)


zeffii:
i think this breaks the RGB curve example, but to be frank... i've not used that thing ..and the solution is too hacky anyway.

zeffii:
1. there is no longer a `draw` directive for append content to draw_buttons. Just make a function called `ui`. For example:
```python
"""
in objects_in o d=[[]] n=0
out stuff     v
"""

    scn = context.scene
    if not hasattr(scn, 'some_scriptnode_props'):
        return
    layout.prop(scn.some_scriptnode_props, 'custom_1')
        
scn = bpy.context.scene
if hasattr(scn, 'some_scriptnode_props'):
    print(scn.some_scriptnode_props.custom_1)

stuff = []
```
2. Technically you don't need to define output lists anymore either. in the example above 'stuff` is already created for you.
3. There's a new 'setup` feature, just make a function called `setup()` and much like processing anything you want to have state.. you stick in that function.. for example
```python
"""
in some_value s d=2 n=2
out some_dummy s
out curframe s
"""

    layout.prop(context.scene, 'frame_current')

def setup():

    class Temp():
        def __init__(self):
            self.logic = []
        def more(self, num):
            self.logic.append(num)
            if len(self.logic) > 20:
                self.logic = []

    maja = Temp()
    maja.more('two')
    maja.more('five')
    maja.more('twelf')

maja.more(some_value)
some_dummy.append(maja.logic)
curframe.append([bpy.context.scene.frame_current])
```
4. various internal optimizations. auto imports `bpy`

PR #1349, opened 2017-03-12
Update of "mesh select" node

portnov:
Remove unneeded checks.
 Update documentation.
 Add `by normals pointing outside' mode.
 Mesh select: add "by edge direction" mode.


zeffii:
bmesh.ops.similar_faces
bmesh.ops.similar_edges
bmesh.ops.similar_verts

maybe use these?

portnov:
@zeffii 
Interesting. But, these operators get some selection and extend it by some rule. But how to specify initial selection?
Maybe it should be separate node that has input "MaskIn" and output "MaskOut" ?

zeffii:
yeah maybe.. i'll leave it to you guys :)

PR #1351, opened 2017-03-12
raw pep8 and redux

zeffii:


zeffii:
untested... but less headwrecking.

zeffii:
seems to work.. second pair of eyes testing welcomed.

ly29:
Readability is so greatly improved.

PR #1354, opened 2017-03-12
Add separate node color for the extended generator nodes

DolphinDream:
The extended generator nodes have a color slightly off the color of the main generators (greenish for the default theme). This way we can distinguish the extended generators apart from the rest of the nodes including the main generators.

genx-vs-gen-nodecolors


nortikin:
i agree, more colors (nuanced) more fun.
Please, tell me your name to add to contributors list.

PR #1357, opened 2017-03-13
Plane node MK2 

DolphinDream:
Revamped plane node (MK2 currently in the extended generators). 
This version includes:
- clean up code 
- sanitize inputs
- add XY/YZ/ZX orientation 
- add normalize/size options (with size syncing option)
- preserves the previous plane node vectorization (list of list vectorization works)

### DIFF
--- a/docs/nodes/generators_extended/plane_mk2.rst
+++ b/docs/nodes/generators_extended/plane_mk2.rst
@@ -1,0 +1,67 @@
+Plane MK2
+==========
+
+Functionality
+-------------
+
+Plane generator creates a grid in the plane XY/YZ or ZX, based on the number of vertices and the length between them in X and Y directions. It works in a similar way than Line, but creating a grid instead of a line.
+
+Inputs
+------
+
+Just like in Line Node, all inputs are vectorized and they will accept single or multiple values.
+There is two basic inputs **N Verts** and **Step**, but referenced to both X and Y directions, so it results in 4 inputs:
+
+- **N Verts X**
+- **N Verts Y**
+- **Step X**
+- **Step Y**
+
+Same as Line, all inputs will accept a single number or an array of them or even an array of arrays::
+
+    [2]
+    [2, 4, 6]
+    [[2], [4]]
+
+Parameters
+----------
+
+All parameters except **Separate**, **Direction**, **Center**, **Normalize**, **Size X** and **Size Y** can be given by the node or an external input.
+
++---------------+------------+-----------+----------------------------------------------------+
+| Param         | Type       | Default   | Description                                        |
++===============+============+===========+====================================================+
+| **N Verts X** | Int        | 2         | number of vertices in X. The minimum is 2.         |
++---------------+------------+-----------+----------------------------------------------------+
+| **N Verts Y** | Int        | 2         | number of vertices in X. The minimum is 2.         |
++---------------+------------+-----------+----------------------------------------------------+
+| **Step X**    | Float      | 1.00      | length between vertices in X axis                  |
++---------------+------------+-----------+----------------------------------------------------+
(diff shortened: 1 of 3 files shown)


DolphinDream:
planemk2-ui

planenodemk2-demo1

DolphinDream:
A few notes:

- The size sync is currently implemented so that every delta in one size is added to the other size. This may or may not be the desired effect since the ratio of the sizex/sizey is not preserved. If the ratio of the sizes should be preserved when the sizes are synced, then I change the sync code to behave this way.

- The outputs of the node seems to be embedded in parentheses (this is how the original plane output is btw). I wonder if the outputs should have outer [] instead of ().  I don’t fully understand how the () ended up in the outputs during the lists constructions using zips and all. Maybe it has something to do with this list of list vectorization of this node. Anybody have an idea why?

- And talking about output formats.. what is the recommended format? 

**verts**:

- [(x,y,z)] 
- [[x,y,z]]
- ([x,y,z])
- ((x,y,z))

**edges**:

- [[i1,i2]]
- [(i1,i2)]
- ([i1,i2])
- ((i1,i2))

**polys**:

- [[i1,i2,i3,i4]]
- [(i1,i2,i3,i4)]
- ([i1,i2,i3,i4])
- ((i1,i2,i3,i4))

portnov:
Cool.

Though usually such "lock" button means "preserve aspect ratio".

DolphinDream:
@portnov Done :+1: Yeah, it makes sense when you play with it, it's odd not to have it resize proportionally.

zeffii:
good work @DolphinDream

DolphinDream:
Thanks for merging this @zeffii  .. and for the compliments :grin:

zeffii:
it's fun to use.  so thank _you_.

on a style related note, i want to encourage you to do indentation like
```diff
-    direction = EnumProperty(name="Direction", items=directionItems,
-                             default="XY", update=updateNode)

+    direction = EnumProperty(
+        name="Direction", items=directionItems,  default="XY", update=updateNode)
```
or
```diff
+    direction = EnumProperty(
+        name="Direction", items=directionItems,  
+        default="XY", update=updateNode)
```

else all the props have such a weird indentation.

DolphinDream:
Roger that, @zeffii . I actually prefer your second suggestion. I only tried to reduce the vertical spread based on what I was told in the past. I think the consistent prop indentation is better :) I'll update that for this node and keep that in mind moving forward.

zeffii:
>  I only tried to reduce the vertical spread based on what I was told in the past

Then that's a miscommunication and my fault. The eye glides easier over a constant indentation, method 1 or 2. Not a serious issue and some pep8 configurations will prefer to indent the less visually pleasing way.

DolphinDream:
Cool. Also, I have some variables I used internally (e.g. sizeRatio and synching) Do these have to be made properties? I'm not sure what's the proper way to define class variables that are not properties the user can interact with via UI. I think I originally just defined them in the class as sizeRatio =1.0 and synching = False and for some reason they did not work (the methods using them using self.sizeRatio, self.synching did not seem to update them as class variables which is why i ended up creating them as properties).

zeffii:
- within functions you can't define new `self.foo` , (well you can.. but it won't work like it does in a regular class )
- if you want to access a variable from anywhere that `self` is available it has to be made a proper `bpy.props....Property`  in the class definition.  <-- works fine, use it.

- or use a `self.node_dict = {}` for more elaborate data types ( do a search for that in the repo for some insight )

DolphinDream:
When the Normalized is enabled obviously the step size entered directly in the node does not affect the output. Should these inputs be grayed out like other blender UI elements. You can modify them but they don’t affect the output.

 Note: if the step size is provided as list of sizes, it will affect the distribution of the lines along x/y but not the overall size x/y of course.

PR #1365, opened 2017-03-14
"Attraction vectors" node

portnov:
This node calculates vectors directed from given vertices to the attractor. Vector lengths are calculated by one of physics-like falloff laws (like 1/R^2), so it looks like attractor attracts vertices. 
Attractor can be specified as:
* single point or set of points
* straight line (specified by point lying on the line and directing vector)
* plane (specified by point lying on the plane and normal vector).

Documentation is in progress, will push it in near time.

### DIFF
--- a/docs/nodes/vector/attractor.rst
+++ b/docs/nodes/vector/attractor.rst
@@ -1,0 +1,68 @@
+Attraction Vectors
+==================
+
+Functionality
+-------------
+
+This node calculates vectors directed from input vertices to specified attractor. Vector lengths are calculated by one of physics-like falloff laws (like 1/R^2), so it looks like attractor attracts vertices, similar to gravity force, for example.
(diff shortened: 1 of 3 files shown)


portnov:
plane-to-circle
box-to-two-points
torus-to-line
sphere-to-plane
plane-to-point
plane-extrude

zeffii:
Awesome.

portnov:
This

glass_cup_1

is now as simple as

cup-nodes

:)

portnov:
output

kalwalt:
Indeed very interesting !

Kosvor2:
Vert socket type has build in vector prop. Use `self.inputs.new('VerticesSocket', 'name').use_prop = True`

This way no need to create default values like `directions_s = self.inputs['Direction'].sv_get(default=[[(0,0,1)]])`

PR #1374, opened 2017-03-15
this is less code and seems to work

zeffii:
this addresses two problems that manifest in _sv_update_handler_

- suppresses double update upon timeline scrubbing
- allows arrow keys to function as normal (nudging current keyframe)

how? by tracking if the _sv_update_handler_ is being triggered on a different 'current_frame' than the previous trigger. The mechanism is currently a dict, but could be switched out for a scene property,, something like 'scene.sv_current_frame`.  If we go with this. 

I'll leave this for a while before pushing.

### DIFF
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -18,10 +18,19 @@ from sverchok.ui import (
 )
 
 
+_state = {'frame': None}
+
 def sverchok_trees():
     for ng in bpy.data.node_groups:
         if ng.bl_idname == 'SverchCustomTreeType':
             yield ng
+
+
+def has_frame_changed(scene):
+    last_frame = _state['frame']
+    _state['frame'] = scene.frame_current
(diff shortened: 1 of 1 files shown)


zeffii:
scrubbing the timeline will otherwise trigger the `sv_update_handler` twice, i don't like this solution.. but don't see any other route.

zeffii:
@ly29  check the leading comment above, see what you think.

kalwalt:
@zeffii Checked now, seems to works. I can start to think for an image sequence saving option (or something similar)

zeffii:
i can image a scenario if one is rendering, and the timeline was previously at frame 1 , and the user renders an animation starting from 1 it may not update anything on the first frame.

if that fails.. i prefer to revert, and stick with the double update per manual scrub. "tough shit until someone brighter takes a look".

PR #1412, opened 2017-03-24
Node props

zeffii:
not for merging.

todo (non exhaustive list)
- [x] first 'cache current state' if it's showing values already, before re-aquiring values.. else bools are lost.
- [x] convert collection to dict.
- [ ] write a function that converts any of the bpy.props types to a serialize-able type
- [ ] generate json storage and recall/update
- [ ] do first tests of prop injections..
- [ ] option to restore to node's original defaults
- [ ] interface tweaks (UI list  etc )
   - [ ] deal with switching between nodes with same bl_idname, ..auto hide deviations? or auto re-obtain..


zeffii:
prelim screenshot: 
image

zeffii:
image

zeffii:
image

zeffii:
logically that should be displayed in a UIList, but i'll leave that to tweaking..  

todo 
- first 'cache current state' if it's showing values already, before re-aquiring values.. else bools are lost.
- generate json storage and recall/update
- do first tests of prop injections..

DolphinDream:
Wow, that looks nice @zeffii . 

Other things that would be nice to include are:

- a restore node “factory” defaults button to restore the the original default values that are hardcoded in the node. This would be essentially clearing the user defined defaults ? This would remove the user defaults from the json file as well I imagine. Another name for it would be “Clear User Defaults” ? 
- an option to “store all” (toggle all/none store flags) so i don’t have to enable every single param i want to have saved as user defaults.
- show some indication that the param deviated from the hard coded value ? or is this already implicit as not showing that property at all if it not deviate from the original? Btw, if you decided to add a store bool, do you plan to show all props (regardless whether they deviated or not) and have the store bool decide whether to save that default or not? 

Btw, what is the “get deviation” button do? Shouldn’t this setting area simply show all the params deviated or not and have the store flag decide which ones to save as user defaults? A visual cue could be added to each param to indicate whether it has deviated from the original (like a * in front of the prop name maybe? (similar to the text editors indicating if the file has been modified and not saved)

DolphinDream:
Also, perhaps as a side note: if we are to offer a randomize colors to some nodes (like the viewer draw perhaps) they could probably be added to the property panel as an extra setting as well (though this would likely be separate from the defaults mechanism). Come to think about it perhaps such feature should get its own issue#  thread. :)

zeffii:
>  This would be essentially clearing the user defined defaults ?

yeah - "Clear User Defaults”  would be OK. The exact mechanism is TBD.

> i don’t have to enable every single param i want to have saved as user defaults.

we'll see, but I think you'll find there aren't that many genuinely useful deviations from the defaults to begin with. ( i could be mistaken.. -- for discussion at a later stage )

>  already implicit   ? 

yep..

> do you plan to show all props (regardless whether they deviated or not) and have the store bool decide whether to save that default or not?

nope.. for the sake of brevity.. if a user wants to update their stored defaults at a later stage you just add the node and tweak the param, then "get deviations" then store that value.


> Shouldn’t this setting area simply show all the params deviated or not and have the store flag decide which ones to save as user defaults?

i'm doing it this way to keep the code sane. One can't update a collection propertygroup from within the draw function of a panel (nor would it be sane to do so as it gets updated many times a second sometimes).. The other option would be to tie this into `updateNode` , but i'll let you figure out why that's not a super route to go.

In my experience of setting defaults for programs, I think this will be right up there with convenience and "function over form".

---
this is too many questions.

zeffii:
`*problems`

- some nodes already modify the default values of a property during sv_init and beyond -- _before_ a user even touches the node, 
- trivial but not all properties make sense to have modified defaults, as they are internal and used only for proper functioning of the node.
- not all props use the `name` attr, (usually non-user facing props like above)

I still like dealing with those better than adding prop overrides by hand into nodes individually..

zeffii:
maybe beside `store` there should be a `restore` button.

DolphinDream:
Btw, If a property has the “store” set that should probably not mean that if the property is updated in the node itself it should be auto-saved as default. Only the user triggered “Save as Default” should do that. 

Also, the property value in the default section still shows the value currently in the node itself even after I save as default with store, which is confusing. To elaborate: say in the viewer draw node I change the vertex color to say RED and then chose to store it and hit “Save as Defaults". Then change the color of the vertex again to, say BLUE. The RED should still be the color that is saved as default. However, as it is implemented right now, the default setting shows the current color BLUE which is a bit confusing as to which is the saved default color. E.g how do know RED is the default value and not blue? Perhaps both colors should be shown here in the default section? i.e. the current color and the saved/default color. What is problematic I think is that there’s no indication when you look at the default values whether the colors (or any settings) were saved as defaults or not. The “store” option only indicates that a default is to be stored or perhaps was stored.  An indicator that a property was saved already or not would be nice to have. There are really 3 values: original, user stored and current :) Showing all three perhaps would be even more confusing, but you get my point. We need more clarity about what is saved, what was saved or not etc.

DolphinDream:
There seems to be an update bug here.. restoring the color does not update the node itself unless i move the mouse over it.

sv-node-prop-demo1

zeffii:
look at the code you see the save operator is merely a shell, there's quite a bit of plumbing to do first. I didn't feel like enumerating all of the todo, " ain't nobody got time for that " ..

I will ask for a review in due course, but it's too early for that.

zeffii:
> Btw, If a property has the “store” set that should probably not mean that if the property is updated in the node itself it should be auto-saved as default. Only the user triggered “Save as Default” should do that.

nothing will be autosaved, this is also facilitated by the manual acquisition of the current state, possible i might name that button `get snapshot as defaults` or something

For the rest of that question, i ask you to refactor the paragraph.. one side of my brain can't convince the other side to read it - it's too dense.

zeffii:
and ..don't worry if some interactions disappear - i edit these threads mercilessly.

zeffii:
image

zeffii:
https://github.com/nortikin/sverchok/blob/master/node_tree.py#L452 will need a test for the presence of the node in _node_defaults.json_ and do an alternate route if it is..
```
class SverchCustomTreeNode:

    ...

    def sv_init(self, context):
        self.create_sockets()

    def init(self, context):
        ng = self.id_data
        ng.freeze()
        if hasattr(self, "sv_init"):
            self.sv_init(context)
        self.set_color()
        ng.unfreeze()
```

zeffii:
pausing this for the time being

PR #1413, opened 2017-03-25
Preferences/Settings revamping + theme presets - WIP

DolphinDream:
Start revamping the settings / preferences. 

Preferences are structured into tabs (for now: general, themes and defaults).
Added a Centering ON setting to allow certain nodes that have centering option to have the centering ON by default (e.g. the line and plane MK2 are tested for now with this new preference settings)

More things may be added, moved, removed. This is just a starting point for revamping.

sv-preferences-demo3


DolphinDream:
Any ideas about how would i go about saving theme presets ? (similar to the blender theme presets) ?

theme-presets

zeffii:
having it independent from the Blender theme gets my vote.

also for generating `datafiles/sverchok/themes/default.json`
```python
import os
import bpy

def generate_themes_folder():

    dirpath = os.path.join(bpy.utils.user_resource('DATAFILES', path='sverchok', create=True))
    themepath = os.path.join(dirpath, 'themes')
    fullpath = os.path.join(themepath, 'default.json')
    
    # create fullpath if it doesn't exist
    if not os.path.exists(themepath):
        os.mkdir(themepath)
    
    if not os.path.exists(fullpath):
        with open(fullpath, 'w') as _:
            pass

generate_themes_folder()
```

DolphinDream:
I was looking into blender’s user preferences presets code since I want to add code to add/remove theme presets and I have some questions / need some hints. sorry for the many questions .. hope your brain is not going to explode @zeffii :) 

1. Do you need to create a new class for a button callback? I thought that I could define a method in the SV preferences class and have the Add/Remove preset buttons call these methods as callbacks. Looking at other blender/sv code it seems that an operators have their own class and be registered as well. Is this true? (I don’t know much about operators unfortunately). 

2. Should the SV preset code be derived from blender presets or borrowed code/functionality and adapt it to SV? I see blender preset code has some preset subclasses which i suspect are to handle things like “Pop up a menu to ask the user to enter the name of the preset” etc. If there is SV code already to handle presets let me know.

3. Is it ok to put all SV themes in one JSON file (separated by theme names) or should each theme be placed in its own file?

4. I’m thinking of adding the default presets (default, nipon blossom.. and perhaps others) into SV  folders (as predefined SV presets), while the user presets get saved in the datafile location in the blender preference area as you suggested. 

5. Should we build a preset infrastructure/classes to allow various other parts of SV (not just the preferences) to save presets? E.g. Perhaps some nodes could have a particular set of settings that can be saved as presets. This is different than just saving one set of settings a defaults.. presets would allow you to save several configurations of settings that generate a unique output for that node. Perhaps the node defaults (you are working on in the node_prop branch) could be saved as “default”, but you could also save them as a presets with different names?

zeffii:
1. simplest thing is to write one operator that takes a property, the property will then determine the operator's behaviour
```
class SvAddRemoveTheme(bpy.types.Operator):
    """ add current settings as new theme or remove currently selected theme
     (doesn't work on hardcoded themes: default, nippon_blossom) 
    """

    bl_idname = "node.sv_add_remove_theme"
    bl_label = "Add Remove Theme"
    # bl_options = {'REGISTER', 'UNDO'}

    behaviour = bpy.props.StringProperty(default='')

    def execute(self, context):
        n = context.node

        if behaviour == 'some_named_function':
            pass

        elif behaviour == 'some_named_other_function':
            pass

        return {'FINISHED'}
```
and accessed in the ui this way
```python
    foo = "node.sv_add_remove_theme"
    row = layout.row()
    row.operator(foo, icon="PLUS").behaviour = 'store current theme'
    row.operator(foo, icon="MINUS").behaviour = 'remove current theme'
```

2. Not an area i've ever needed to dive into, so can't answer with any authority. For now i suggest checking out the monad's code that pops up when you set a property's defaults. Additionally the [UI cookbook](https://wiki.blender.org/index.php/Dev:Py/Scripts/Cookbook/Code_snippets/Interface) is still the best reference for popups and ui stuff.

3. no.  please don't.

4. the themes: i've thought about it and I think they should be hardcoded (dicts) and written to datafiles once as `json` at an appropriate time. (not automatically.. unless absolutely it can't be done any other way.)

5. Presets for a node.. I think not enough nodes have an overwhelming number of properties to warrant that.

zeffii:
```python
class SvLoadZippedBlendDialog(bpy.types.Operator):
    bl_idname = "node.sv_load_zipped_blend_dialog"
    bl_label = "Load Zipped (Sverchok Dialog)"

    def draw(self, context):
        row = self.layout.row()
        row.label(context.window_manager.clipboard or "No url in clipboard")

    def execute(self, context):
        return bpy.ops.node.sv_load_zipped_blend_url(download_url=context.window_manager.clipboard)
 
    def invoke(self, context, event):
        return context.window_manager.invoke_props_dialog(self)

```

- here the content of execute is performed once the user hits the OK button the dialogue
- the draw function is just like any Node ui's draw..
- The operator can have properties, and the operator can call other operators.

See https://github.com/nortikin/sverchok/issues/1434
https://github.com/nortikin/sverchok/blob/ea0c56660554e8cabe2dd8ac1e0b0ffdbfa152c6/utils/sv_load_zipped_blend.py

DolphinDream:
Btw.. I implemented the SvAddRemoveTheme the way you suggested @zeffii , but since I added a lot of things to it I am considering actually breaking it into two classes one for Add and one for Remove since I have to do a lot of if behaviour=“add” / else conditionals. Are you on board with that?

DolphinDream:
@Zeffi thanks for the dialogue example. I actually figured it out eventually 👌🏼🤓

zeffii:
Yes, two ops is fine.

DolphinDream:
Add various theme preset related settings (see last commit for details about the latest features).

Summary:
- Add new sv_theme module to handle all theme related features
- Consolidate color related code into the new theme module
- Update Settings UI and include add/remove/update theme 

Still left to be done:
- [ ] Add hints about unsaved colors 
- [ ] Create color attributes in settings based on theme colors (hints needed)
- [ ] Cleanup code / remove unneeded tabs/settings and Debug info
- [ ] Add feature to upgrade saved themes in case new colors defined/undefined
- [ ] Add feature to allow to import/export (gist?) themes 

sv-preferencedev-demo1

nortikin:
why not to prepare it for 2.80?

zeffii:
what's the point :)  2.80 is still a long way off, and certain preference customizations are mostly distractions (just because we can do it , doesn't mean we should )

nortikin:
> .80 is still a long way

ok, sorry

Issue #1403, opened 2017-03-22
Exploring Extrude Region By Matrix


enzyme69:
Quite unusual, not exactly sure how to use. If there is a way to do multiple extrusion via Vectorized Matrix I would like to know @portnov 

![screen shot 2017-03-22 at 12 07 26](https://cloud.githubusercontent.com/assets/3952687/24177667/2e168b54-0ef8-11e7-860f-0b175d6a9d17.png)
https://gist.github.com/eba7a98777107dec3c6079ad9f6ec1cf

portnov:
What do you mean by "multiple extrusion"? Current behaviour when many matrices are passed is to apply each matrix to each extruded vertex.

enzyme69:
@portnov 

Something like below:
![screen shot 2017-03-22 at 16 24 08](https://cloud.githubusercontent.com/assets/3952687/24183371/0b79db18-0f1c-11e7-9730-6aeb341ad5bc.png)

Single extrude, resulting in multiple subdivided in between, which would be nice if can be affected by Curve.

Will wait how Sverchok implement this.

enzyme69:
![screen shot 2017-03-22 at 16 27 46](https://cloud.githubusercontent.com/assets/3952687/24183447/8157b972-0f1c-11e7-91d1-121e46e04e7c.png)

portnov:
Maybe we could add an option to Extrude Region node, name it like "multiple extrude". If it is checked, then interpretation of multiple matrices is like one matrix for one extrude operation (all vertices transformed by the same matrix in one operation); otherwise, keep current behaivour, one matrix for one vertex.

enzyme69:
@portnov sounds great and the each matrix extrude can be controlled per matrix like rotation, scale, and translated. definitely a future enhancement request.

enzyme69:
Another possibilities is to combo the nodes by programming and then totally hides it.

portnov:
![parabola](https://cloud.githubusercontent.com/assets/284644/24330825/c914a3fe-1248-11e7-9694-3068659feaf7.png)

zeffii:
the recent commit addresses this question. reopen if not.

nortikin:
this multy extrude makes me think we need two conditions of vectorization for each node. or indeed to act as level-corresponding system. masking objects in objects level, vertices on verts level etc. what way is better don't know till now.

PR #1418, opened 2017-03-26
Multiple extrude mode

portnov:
Extrude Region node updated:

* Add possibility to do several extrusion operation at once
* Move "keep original" parameter to N panel, since I believe it is rarely used
* Update documentation

See also #1403.

Please review and test.


zeffii:
please update the `.hide` here too :)

Issue #1285, opened 2017-03-02
using .hide on sockets (pros/cons/possible resolution)


zeffii:
sparked by https://github.com/nortikin/sverchok/pull/1271 but used in several nodes
- [create_bvh_tree.py](https://github.com/nortikin/sverchok/blob/d37532f5777292a12872e4e8115848d6371b85fc/nodes/scene/create_bvh_tree.py)
- [vector_polar_out.py](https://github.com/nortikin/sverchok/blob/afef289579753820008e2749a670545ead24b402/nodes/vector/vector_polar_out.py)
- [objects_along_edge.py](https://github.com/nortikin/sverchok/blob/021979e956ffd5390fd658bd1caa1af8ca299a36/nodes/modifier_change/objects_along_edge.py)
- [mesh_filter.py](https://github.com/nortikin/sverchok/blob/021979e956ffd5390fd658bd1caa1af8ca299a36/nodes/analyzer/mesh_filter.py)

and 3 or 4 more maybe.

the use of `.hide` on sockets has been very sparse and mostly has crept in under the radar. The point of this issue is to discuss and resolve the issue preventing us from using this feature (or similar custom `.hide_safely` implementation) .

### cons on `.hide`

- hidden sockets don't drop their links automatically, meaning the node can be getting more calls to process from upstream nodes than a user might realize ( testable )
- (bad reason but a reason nonetheless) iojson doesn't store socket properties , gists won't work flawlessly out of the box.

### cons on `.remove`

- it's quite the headache to implement remove/add upon mode switch, certainly compared to simply .hide.
- depending on how `process` is written can require extra sanitizing to see if a socket exists.

### Resolution

There are any number of routes, and one, some or _all_ might be needed.
-  to solve the iojson issue, 
    - It would be worth adding socket properties to iojson to keep code like this simpler
        - also some sockets use `.prop_type` (or something) which we don't track yet either.
    - don't track links between hidden sockets
- to solve calls to process, links on a hidden socket could be ignored and not cause triggers to flow to that node (unless it has other connected sockets of course)
- Maybe we add a `.hide_safe` to the socket class, that auto unlinks hidden socket.links.

@ly29 anything to add or is the restriction on .hide cargocult, it's been a while since I looked the DAG

ly29:
@zeffii 
I don't remember actually if what will happen. I prefer not to add complexity or spend time messing with the overly complicated update system...
From a conservative standpoint I vote no.

zeffii:
[...]

PR #1419, opened 2017-03-26
"Mesh select" by bounding box

portnov:
Add "By bounding box" mode for "select mesh elements by location" node:

* If one point `(0, 0, 0)` is passed, and Radius = 1, then the node will select all vertices that have `-1 <= X <= 1`, `-1 <= Y <= 1`, `-1 <= Z <= 1`.
* If points `(0, 0, 0)`, `(1, 2, 3)` are passed, and Radius = 0.5, then the node will select all vertices that have `-0.5 <= X <= 1.5`, `-0.5 <= Y <= 2.5`, `-0.5 <= Z <= 3.5`.

### DIFF
--- a/docs/nodes/analyzers/mesh_select.rst
+++ b/docs/nodes/analyzers/mesh_select.rst
@@ -17,9 +17,9 @@ This node has the following inputs:
 - **Edges**
 - **Faces**
 - **Direction**. Direction vector. Used in modes: **By side**, **By normal**, **By plane**, **By cylinder**. Exact meaning depends on selected mode.
-- **Center**. Center or base point. Used in modes: **By sphere**, **By plane**, **By cylinder**.
+- **Center**. Center or base point. Used in modes: **By sphere**, **By plane**, **By cylinder**, **By bounding box**.
 - **Percent**. How many vertices to select. Used in modes: **By side**, **By normal**.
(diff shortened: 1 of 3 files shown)


zeffii:
sweeeet :)

zeffii:
i see you are using `.hide` ,  i'm considering adding '.hide_safe`  which first disconnects any connected link to a socket that's being hidden. 

Could you try :  https://github.com/nortikin/sverchok/issues/1285#issuecomment-283674312

zeffii:
confirm that it works OK with gist :)

portnov:
@zeffii 
https://gist.github.com/d2f54cc9884f4015de08ae2f1f4126ae

I tried to change .hide with .hide_safe, but I did not notice any change: when I change the mode, the inputs which are connected do not disappear...

zeffii:
@portnov  did you add hide_safe to SvSocketCommon?

portnov:
ah
I thought it already done in master :)

portnov:
@zeffii 
So, now  I tried that thing with .hide_safe and it works for me. Should I add this change to this pull request?

zeffii:
yes please!

portnov:
@zeffii 
Done.

zeffii:
Thanks for adding that :)

PR #1424, opened 2017-03-27
Various improvements for documentation

kalwalt:
This is a work in progress, I want to make some little addition if i can. There are some node that still hasn't any docs.  Consider that i don't know all the nodes, actually i'm fine with 1/3 of them, 1/3 i have a vague idea and the other simple i don't know.  I will make a list here of nodes that i can write some minimal doc, Maybe i will ask help for clarifications. Do not expect a massive pull request i think it's better some little chunk at a time.

##### Vector
- [x] Vertices sort
- [ ] Vector X Doubles
- [x] Vector math (only to update)
##### Number
- [ ] Map Range
- [x] Range Float
- [ ] Math

### DIFF
--- a/docs/nodes/generator/basic_3pt_arc.rst
+++ b/docs/nodes/generator/basic_3pt_arc.rst
@@ -13,7 +13,7 @@ Inputs
 
     - must be (len % 3 == 0 )
 
-- num verts is either  
+- num verts is either
 
     - constant
     - unique
@@ -23,14 +23,14 @@ Inputs
 Parameters
 ----------
 
-The UI is quite minimal. 
+The UI is quite minimal.
 
 
 - **num verts** can be changed via Slider input on the UI or as described above, it can be fed multiple values through the input sockets.
 
 
 Output
------- 
+------
 
 - (verts, edges) : A set of each of these that correspond with a packet of commands like 'start, through, end, num_verts'
 - verts needs to be connected to get output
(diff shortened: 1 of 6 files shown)


kalwalt:
in Vector math:
```
Noise Vector         u        w         [see mathutils]()
Noise Scalar         u        s         [see mathutils]()
Scalar Cell noise    u        s         [see mathutils]()
Vector Cell noise    u        w         [see mathutils]()
```
I think i can delete this, since `Noise` as his node and `Cell noise` was included in the past? 
Would be super fine an example for each of the options, but i will see this.

zeffii:
correct i stripped out the noise functions from the node a while back, but failed to update docs. 

Thanks for tackling this!  best way to get to know more nodes!

kalwalt:
>  correct i stripped out the noise functions from the node a while back, but failed to update docs.

i will do the update.

> Thanks for tackling this! best way to get to know more nodes!

yes! i agree with you!

zeffii:
two things..
- `math.rst` can be renamed to `math_mk2.rst` and there are a few new _modes_ 
- you will need to update `vector_index.rst` to reflect the rename

kalwalt:
ok! i will fix it!

kalwalt:
i forgot to add variable lacunarity node in the `vector_index.rst` list, i will add it now.

zeffii:
you can continue to push commits to your branch, and do a new PR.

kalwalt:
ok i didn't know this option!

PR #1426, opened 2017-03-27
"Simple Deformation" node

portnov:
This node is more or less an analog of Blender's "Simple Deform" modifier. Only Twist, Bend and Taper deformations are supported at the moment.

Please review and test.

### DIFF
--- a/nodes/transforms/deform.py
+++ b/nodes/transforms/deform.py
@@ -26,9 +26,21 @@ import bmesh.ops
 from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode, match_long_repeat, fullList, Matrix_generate
 
+def limit(src_min, src_max, low_limit, hi_limit):
+    if src_min == src_max:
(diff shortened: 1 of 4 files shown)


nodes/transforms/deform.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def process(self):
+
+        if not any(output.is_linked for output in self.outputs):
+            return
+
+        vertices_s = self.inputs['Vertices'].sv_get(default=[[]])
+        origins = self.inputs['Origin'].sv_get(default=[Matrix()])
+        angles_s = self.inputs['Angle'].sv_get(default=[[pi/4]])
+        factors_s = self.inputs['Factor'].sv_get(default=[[0.785]])
```
zeffii: defaults aren't necessary to set explicitly, when the socket has a `prop_name` pointing at a `self.<property>`

portnov: fixed.

PR #1438, opened 2017-03-29
Docs_improvements_01  part2

kalwalt:
I continue here  pull request https://github.com/nortikin/sverchok/pull/1424.

##### Vector
- [x] Vertices sort
- [x] Vector X Doubles
- [x] Vector math (only to update)
##### Number
- [x] Map Range
- [x] Range Float
- [x] Math


kalwalt:
May i found something about Vector X Doubles? pull requests, issue or other please i can't find anything.

kalwalt:
i did this [example](https://cloud.githubusercontent.com/assets/1275858/24463951/98383630-14a8-11e7-8653-178335ec576b.png) i hope that is correct.

zeffii:
essentially this

##### Vector X doubles
```python
    def remdou(self, vers, levs):
        out = []
        if levs >= 3:
            levs -= 1
            for x in vers:
                out.append(self.remdou(x, levs))
        else:
            for x in vers:
                if x not in out:
                    out.append(x)
        return out
```

zeffii:
Think of it like a Set of the verts, sorted by the index of the vert in the incoming list.
```
# if each letter was a vector, and all A's where the same vector...B's the same vector etc...
AAAABBBBBCJCKCDDDEEEEEFFFFGGGGAAAABBBBCCCCCHHHHIIIIII
```
then the output
```
ABCJKDEFGHI
```

kalwalt:
Interesting, i will explore this node.

zohozer:
I just need to report that the images from the documentation are missing.

Just downloaded the .epub version of the documentation from
[https://readthedocs.org/projects/sverchok/downloads/](https://readthedocs.org/projects/sverchok/downloads/) and seems that the images are missing. It is the same situation with all the other formats I think.

Please fix this if it is possible.

zeffii:
Currently we use the issuetracker to store images for the docs, this allows us to keep the download size of sverchok down. In order to embed the images, I think we'd need to set up a separate repository for the docs, and store the images on the repo.

(i'm not certain about this..)

nortikin:
we need separate them, it's time

zohozer:
Yes please, because without images it is more difficult to understand certain parts from documentation.

zeffii:
https://github.com/Sverchok/sverchok_docs

zeffii:
No ETA @zohozer

zeffii:
there is a problem with this separation, 

- we store the docs locally in the github repo, this allows us to include the docs to be read offline (without images. Images are handy, but not always essential)

A reasonable solution at the moment is to regularly sync the separated docs, with the most uptodate on the main repository..

zeffii:
this is slightly offtopic for this Pullrequest. More soon.

zeffii:
my mistake, compiling the docs doesn't automatically download the images: further discussion about this here: https://github.com/nortikin/sverchok/issues/1456

kalwalt:
@zeffii i added a minimal documentation for the scalar math node, i renamed to the new version so i think i can delete the scalar.rst. Do you confirm this? (well, i have already deleted... but i must resolve the conflict)

zeffii:
it should probably be
```
.. toctree::
   :maxdepth: 2

   exponential
   fibonacci
   float
   float_to_int
   formula2
   integer
   list_input
   mix_numbers
   random
   range_float
   range_int
   range_map
   scalar_mk2
   scalar
   numbers

```

both nodes for the time being..

kalwalt:
ok, i will fix this. In scalar.rst i will point to the new node doc, maybe only a note that are two different versions.

zeffii:
let me know when you're done and i will resolve the conflict .

kalwalt:
i will recheck, for typos or other improvements, i think in the next days i will be fine.

zeffii:
for corrections simply make a new PR ok? :)

zeffii:
i like to suck in improvements to docs asap :)

kalwalt:
ok! there so many nodes to improve the docs!

zeffii:
yep. so many nodes :)

half the nodes in Beta are there because i can't be bothered supporting them or writing docs..or both.

kalwalt:
> half the nodes in Beta are there because i can't be bothered supporting them or writing docs..or both.

Writing docs is also time consuming, i think you prefer improve the code or develop new nodes or impoving globally Sverchok, than spent your time in writing.

zeffii:
at one point there where no docs, then we did a massive crawl through the nodeset and made sure all nodes had a doc.

After a while we added a Beta and Alpha category because we wanted to be able push nodes, but didn't want to keep rewriting the docs during a 'fluid' state. We know docs are important, but we really do try to make nodes work in the most obvious way, making the documentation only really necessary for the moderately complicated nodes. 

Writing docs is hard, it's a skill not unlike programming... programming the reader's mind to follow a train of thought and parking information in the brain. I prefer, indeed, to program

zeffii:
of-course the code is the best form of documentation, it's unbiased ..brutally honest. The people that will get most out of sverchok eventually dig through the code of the nodes and get a better understanding of why something works (or doesn't ) .

kalwalt:
If i can and if i will able i would help writing the docs, I want now to finish the noise/random tutorial, it's something that i absolutely should finish.

zeffii:
no rush

kalwalt:
yes of course!

PR #1440, opened 2017-03-30
Restructure the MixNumber node into Mix Inputs nodes

DolphinDream:
Now the node interpolates vectors too.

Note: Be warned though that changing types will swap out input/output sockets with new type sockets which will disconnect any previously connected inputs/outputs. Perhaps a better version would be able to keep the input/output connection on with proper conversion to the input/ouput types ?

Update the documentation to include vector as well.

### DIFF
--- a/docs/nodes/number/mix_inputs.rst
+++ b/docs/nodes/number/mix_inputs.rst
@@ -1,10 +1,10 @@
-Mix Numbers
-===========
+Mix Inputs
+==========
 
 Functionality
 -------------
 
-This node mixes two values using a given factor and a selected interpolation and easing functions.
+This node mixes two values using a given factor for the selected interpolation & easing functions.
 
 For a factor of 0.0 it outputs the first value while the factor of 1.0 it outputs the last value. For every factor value between 0-1 it will output a value between the first and second input value. (*)
 
@@ -14,18 +14,22 @@ Note:
 Inputs & Parameters
 -------------------
 
-All parameters except for **Type**, **Interpolation** and **Easing** can be given by the node or an external input.
+All parameters except for **Mode**, **Interpolation** and **Easing** can be given by the node or an external input.
+
+All inputs are vectorized and they will accept single or multiple values.
 
 This node has the following parameters:
 
 +-------------------+---------------+-------------+-----------------------------------------------------+
 | Parameter         | Type          | Default     | Description                                         |
 +===================+===============+=============+=====================================================+
-| **Type**          | Enum:         | Float       | Type of inputs values to interpolate.               |
-|                   |  Int          |             | When Float is selected the input value1 and value2  |
-|                   |  Float        |             | expect float values                                 |
-|                   |               |             | When Int is selected the input value1 and value2    |
-|                   |               |             | expect int values.                                  |
+| **Mode**          | Enum:         | Float       | Type of input values to interpolate.                |
+|                   |  Int          |             | The input and output socket types will change       |
+|                   |  Float        |             | according to the selected mode to match the type.   |
+|                   |  Vector       |             |                                                     |
+|                   |  Color        |             | Note: changing the mode will not disconnect the     |
+|                   |  Quaternion   |             | connected input and output sockets.                 |
+|                   |  Matrix       |             |                                                     |
 +-------------------+---------------+-------------+-----------------------------------------------------+
 | **Interpolation** | Enum:         | Linear      | Type of interpolation.                              |
 |                   |  Linear       |             |   f(x) ~ x                                          |
@@ -45,9 +49,19 @@ This node has the following parameters:
 |                   |  Ease Out     |             |  Ease Out = slowly approaches the ending value      |
 |                   |  Ease In-Out  |             |  Ease In-Out = slowly departs and approaches values |
 +-------------------+---------------+-------------+-----------------------------------------------------+
-| **Value1**        | Int or Float  | 0 or 0.0    | Starting value                                      |
+| **A**             | Int           | 0           | Starting value                                      |
+|                   | Float         | 0.0         |                                                     |
+|                   | Vector        | (0,0,0)     |                                                     |
+|                   | Color         | (0,0,0,1)   |                                                     |
+|                   | Quaternion    | (1,0,0,0)   |                                                     |
+|                   | Matrix        | Identity    |                                                     |
 +-------------------+---------------+-------------+-----------------------------------------------------+
-| **Value2**        | Int or Float  | 1 or 1.0    | Ending value                                        |
+| **B**             | Int           | 0           | Ending value                                        |
+|                   | Float         | 0.0         |                                                     |
+|                   | Vector        | (0,0,0)     |                                                     |
+|                   | Color         | (1,1,1,1)   |                                                     |
+|                   | Quaternion    | (1,1,1,1)   |                                                     |
+|                   | Matrix        | Identity    |                                                     |
 +-------------------+---------------+-------------+-----------------------------------------------------+
 | **Factor**        | Float         | 0.5         | Mixing factor (between 0.0 and 1.0)                 |
 +-------------------+---------------+-------------+-----------------------------------------------------+
@@ -66,15 +80,13 @@ For certain interpolation types the node provides extra parameters on the proper
   Extra parameters to adjust the attenuation of the bounce and the number of bounces. The defaults are 0.5 and 4.
 
 * Elastic
-  Extra parameters to adjust the base and the exponent of the damping oscillation as well as the number of bounces (oscillations).
- The defaults are 1.6, 6.0 and 6.
+  Extra parameters to adjust the base and the exponent of the damping oscillation as well as the number of bounces (oscillations). The defaults are 1.6, 6.0 and 6.
 
 Outputs
 -------
 
-This node has one output: **Value**.
+This node has one output corresponding to the selected mode listing the interpolated values between A and B inputs.
 
-Inputs and outputs are vectorized, so if series of values is passed to one of inputs, then this node will produce several sequences.
 
 Example of usage
 ----------------
--- a/index.md
+++ b/index.md
@@ -137,6 +137,7 @@
     SvDatetimeStrings
     SvPolygonSortNode
 
+
 ## List Main
     ListJoinNode
     ZipNode
@@ -180,7 +181,9 @@
     SvRndNumGen
     RandomNode
     SvEasingNode
+    ---
     SvMixNumbersNode
+    SvMixInputsNode
 
 ## Vector
     GenVectorsNode
--- a/nodes/number/mix_inputs.py
+++ b/nodes/number/mix_inputs.py
@@ -1,0 +1,375 @@
+# ##### BEGIN GPL LICENSE BLOCK #####
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software Foundation,
+#  Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# ##### END GPL LICENSE BLOCK #####
+
+
+import bpy
+from bpy.props import IntProperty, FloatProperty, BoolProperty, EnumProperty, FloatVectorProperty
+from sverchok.node_tree import SverchCustomTreeNode
+from sverchok.data_structure import updateNode, match_long_repeat
+from sverchok.data_structure import Matrix_generate
+from sverchok.utils.sv_easing_functions import *
+from mathutils import Matrix, Quaternion
+import re
+
+idMat = [[tuple(v) for v in Matrix()]]  # identity matrix
+
+input_info = {
+    'INT': {
+        "Index": 0,
+        "SocketType": "StringsSocket",
+        "PropType": ('IntProperty', dict()),
+        "PropDefault": {"A": 0, "B": 1},
+        "Mixer": lambda t, v1, v2: (int(v1 * (1 - t) + v2 * t)),
+        "Inputs": lambda ia, ib: [ia.sv_get()[0], ib.sv_get()[0]],
+        "Outputs": lambda o, v: o.sv_set([v])
+    },
+
+    'FLOAT': {
+        "Index": 1,
+        "SocketType": "StringsSocket",
+        "PropType": ('FloatProperty', dict()),
+        "PropDefault": {"A": 0.0, "B": 1.0},
+        "Mixer": lambda t, v1, v2: v1 * (1 - t) + v2 * t,
+        "Inputs": lambda ia, ib: [ia.sv_get()[0], ib.sv_get()[0]],
+        "Outputs": lambda o, v: o.sv_set([v])
+    },
+
+    'VECTOR': {
+        "Index": 2,
+        "SocketType": "VerticesSocket",
+        "PropType": ('FloatVectorProperty', dict(size=3, subtype='XYZ')),
+        "PropDefault": {"A": (0.0, 0.0, 0.0), "B": (1.0, 1.0, 1.0)},
+        "Mixer": lambda t, v1, v2: tuple([v1[i] * (1 - t) + v2[i] * t for i in range(3)]),
+        "Inputs": lambda ia, ib: [ia.sv_get()[0], ib.sv_get()[0]],
+        "Outputs": lambda o, v: o.sv_set([v])
+    },
+
+    'COLOR': {
+        "Index": 3,
+        "SocketType": "SvColorSocket",
+        "PropType": ('FloatVectorProperty', dict(size=4, subtype='COLOR', min=0, max=1)),
+        "PropDefault": {"A": (0.0, 0.0, 0.0, 1.0), "B": (1.0, 1.0, 1.0, 1.0)},
+        "Mixer": lambda t, v1, v2: tuple([v1[i] * (1 - t) + v2[i] * t for i in range(4)]),
+        "Inputs": lambda ia, ib: [ia.sv_get()[0], ib.sv_get()[0]],
+        "Outputs": lambda o, v: o.sv_set([v])
+    },
+
+    'QUATERNION': {
+        "Index": 4,
+        "SocketType": "SvQuaternionSocket",
+        "PropType": ('FloatVectorProperty', dict(size=4, subtype='QUATERNION')),
+        "PropDefault": {"A": (0.0, 0.0, 0.0, 0.0), "B": (1.0, 1.0, 1.0, 1.0)},
+        "Mixer": lambda t, v1, v2: quaternionMix(t, v1, v2),
+        "Inputs": lambda ia, ib: [ia.sv_get()[0], ib.sv_get()[0]],
+        "Outputs": lambda o, v: o.sv_set([v])
+    },
+
+    'MATRIX': {
+        "Index": 5,
+        "SocketType": "MatrixSocket",
+        "Mixer": lambda t, v1, v2: matrixMix(t, v1, v2),
+        "Inputs": lambda ia, ib: [ia.sv_get(default=idMat), ib.sv_get(default=idMat)],
+        "Outputs": lambda o, v: o.sv_set(v)
+    }
+}
+
+interplationItems = [
+    ("LINEAR", "Linear", "", "IPO_LINEAR", 0),
+    ("SINUSOIDAL", "Sinusoidal", "", "IPO_SINE", 1),
+    ("QUADRATIC", "Quadratic", "", "IPO_QUAD", 2),
+    ("CUBIC", "Cubic", "", "IPO_CUBIC", 3),
+    ("QUARTIC", "Quartic", "", "IPO_QUART", 4),
+    ("QUINTIC", "Quintic", "", "IPO_QUINT", 5),
+    ("EXPONENTIAL", "Exponential", "", "IPO_EXPO", 6),
+    ("CIRCULAR", "Circular", "", "IPO_CIRC", 7),
+    # DYNAMIC effects
+    ("BACK", "Back", "", "IPO_BACK", 8),
+    ("BOUNCE", "Bounce", "", "IPO_BOUNCE", 9),
+    ("ELASTIC", "Elastic", "", "IPO_ELASTIC", 10)]
+
+easingItems = [
+    ("EASE_IN", "Ease In", "", "IPO_EASE_IN", 0),
+    ("EASE_OUT", "Ease Out", "", "IPO_EASE_OUT", 1),
+    ("EASE_IN_OUT", "Ease In-Out", "", "IPO_EASE_IN_OUT", 2)]
+
+
+def matrixMix(t, v1, v2):
+    m1 = Matrix_generate([v1])[0]
+    m2 = Matrix_generate([v2])[0]
+    m = m1.lerp(m2, t)
+    return m
+
+
+def quaternionMix(t, v1, v2):
+    q1 = Quaternion(v1)
+    q2 = Quaternion(v2)
+    q = q1.slerp(q2, max(0, min(t, 1)))
+    return q
+
+
+def make_prop(mode, label):
+    ''' Property Factory '''
+    description = "Mix " + mode + " value " + label
+    default = input_info[mode]["PropDefault"][label]
+    # general parameters
+    params = dict(name=label, description=description, default=default, update=updateNode)
+    # add type specific parameters
+    params.update(input_info[mode]["PropType"][1])
+    propType = input_info[mode]["PropType"][0]
+    return getattr(bpy.props, propType)(**params)
+
+
+class SvMixInputsNode(bpy.types.Node, SverchCustomTreeNode):
+    """
+    Triggers: Interpolate, Slerp
+    Tooltip: Interpolate between values of different types
+    """
+    bl_idname = 'SvMixInputsNode'
+    bl_label = 'Mix Inputs'
+    bl_icon = 'IPO'
+
+    # SV easing based interpolator
+    def get_interpolator(self):
+        # get the interpolator function based on selected interpolation and easing
+        if self.interpolation == "LINEAR":
+            return LinearInterpolation
+        else:
+            ''' This maps the Strings used in the Enumerator properties to the associated function'''
+            interpolatorName = self.interpolation + "_" + self.easing
+            interpolatorName = re.sub('SINUSOIDAL', 'sine', interpolatorName)  # for the exception
+            interpolate = globals()[re.sub(r'[_]', '', interpolatorName.lower().title())]
+
+            # setup the interpolator with prepared parameters
+            if self.interpolation == "EXPONENTIAL":
+                b = self.exponentialBase
+                e = self.exponentialExponent
+                settings = prepareExponentialSettings(b, e)
+                return lambda v: interpolate(v, settings)
+
+            elif self.interpolation == "BACK":
+                s = self.backScale
+                return lambda v: interpolate(v, s)
+
+            elif self.interpolation == "ELASTIC":
+                n = self.elasticBounces
+                b = self.elasticBase
+                e = self.elasticExponent
+                settings = prepareElasticSettings(n, b, e)
+                return lambda v: interpolate(v, settings)
+
+            elif self.interpolation == "BOUNCE":
+                n = self.bounceBounces
+                a = self.bounceAttenuation
+                settings = prepareBounceSettings(n, a)
+                return lambda v: interpolate(v, settings)
+
+            else:
+                return interpolate
+
+    def update_mode(self, context):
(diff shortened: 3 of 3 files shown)


zeffii:
have a look at the Vector Lerp node.. i think this does what the Mixnumber vector additions do.

DolphinDream:
Maybe. Though lerp doesn't do easing.

zeffii:
Alright, should be ok.

zeffii:
then maybe a better, more descriptive, name for the node would be ' Mix Inputs '.

zeffii:
(and category List Mutator )

DolphinDream:
Or maybe create separate nodes Mix Floats, Mix Ints, Mix Vectors, Mix Colors (e.g. which is what Animation Nodes have btw). And be all derived from a Mix Input class (for code reuse). This way they will be placed in the appropriate category (Numbers, Vectors, Colors? etc). The “list mutator” category seems to be a bit strange. Personally I wouldn’t think to look for it there. Also having separate node will solve the issue of input disconnection when switching types.

zeffii:
We have a simple `socket.replace_socket` function, that can change the current socket in place, without disconnecting.

```python
      A = self.inputs['some_int']
      A.replace_socket('VerticesSocket, 'some_float')
```

DolphinDream:
Will this do a conversion from a float to vector internally btw? 
And what about the idea of having separate nodes one for each type?

zeffii:
refresh the page i made a typo in that last comment

zeffii:
nope... there's no auto `float->vector` conversion implemented.

zeffii:
> And what about the idea of having separate nodes one for each type?
yep am on board with that... you stick the superclass in `sverchok/utils/sv_mixnode_base.py` or something.

DolphinDream:
Is there an example of a derived node somewhere? 

I was thinkgin of something like this:

```python
import bpy
from bpy.props import IntProperty, FloatProperty, BoolProperty, EnumProperty

from sverchok.nodes.number.mix_numbers import SvMixNumbersNode


class SvMixFloatsNode(SvMixNumbersNode):

    """docstring for SvMixFloatsNode"""

    bl_idname = 'SvMixFloatsNode'
    bl_label = 'Mix Floats'
    bl_icon = 'IPO'

    numType = "FLOAT"

def register():
    bpy.utils.register_class(SvMixFloatsNode)


def unregister():
    bpy.utils.unregister_class(SvMixFloatsNode)
```
But that doesn't work apparently.  I get a bunch of errors like these:

```
rna_uiItemR: property not found: SvMixFloatsNode.value_float1
/home/marius/.config/blender/2.78/scripts/addons/sverchok/node_tree.py:286
rna_uiItemR: property not found: SvMixFloatsNode.value_float2
```

zeffii:
I think your better off using a 'mixin', 

- define a bare class (not CustomNode class ) with everything that all the nodes share

```python
class SvMixInputsMixin():
    ....
```

then in `FloatMixing` (or whatever) you do

```python

class SvFloatMixing(bpy.types.Node, SverchCustomTreeNode, SvMixInputsMixin):
    ... all unique stuff
```

I think.

zeffii:
similar to how one might use the generic operator mixins: https://github.com/nortikin/sverchok/blob/master/utils/sv_operator_mixins.py

zeffii:
similar to how one might use the generic operator mixins: https://github.com/nortikin/sverchok/blob/master/utils/sv_operator_mixins.py

used here
https://github.com/nortikin/sverchok/blob/master/nodes/viz/texture_viewer.py#L39-L49

zeffii:
in this case ,  i think socket switching  is a bit more convenient for the user.

zeffii:
i'll give a prime example of another node that really needs socket switching,  the Int / Float node..

zeffii:
additionally, the power user should be acquainted with the Search feature, and type the nodename.. 

The more a node does the harder it becomes to nicely categorize it without inventing a new category or bending the definition of an existing category.

DolphinDream:
Should I take from your previous comment you do favor multiple mix node than one node with multiple types ?

zeffii:
not really, i'm justifying why i suggested adding it into the List Mutator cat. But i understand why that seems a little out of place. (but it does take lists , and would produce mutated output...).

We are OK with multi-purpose nodes, doing simple socket switching (that's what `socket.replace_socket` is for). If you were going to offer Int/Float switch, then adding a Int/Float/Vector option isn't that much of a stretch.

from the zen of Python " Practicality beats purity"

zeffii:
excuse the simple representation
```

  _________________________________
  |____Mix Input _________________|
  |                         int   o
  |  settings..  mode=int         |
  |                               | 
  o  factor                       |
  o  int                          | 
  o  int                          |
  --------------------------------- 


   _________________________________
  |____Mix Input _________________|
  |                       float   o
  |  settings..  mode=float       |
  |                               | 
  o  factor                       |
  o  float                        | 
  o  float                        |
  --------------------------------- 


   _________________________________
  |____Mix Input _________________|
  |                      vector   o
  |  settings..  mode=vector      |
  |                               | 
  o  factor                       |
  o  vector                       | 
  o  vector                       |
  --------------------------------- 

  
```

zeffii:
it be welcome to add an Interpolation   category.. we have tonnes of nodes that do it.

DolphinDream:
What do you mean by "adding an interpolation category" ?

zeffii:
a NodeCategory called `Interpolation` .


.. i'm starting to question my ability to communicate with you.

DolphinDream:
Ok. cool. Thx @zeffii 

Btw, you'll have to stop and consider the fact I've been dealing with pyton/nodes developing for about a month now (more or less).. on and off (whenever I find the time) and therefore I lack both the language and the concepts around this kind of coding. Plus, I'm not an native English speaker, if that wasn't clear enough. Your patience is highly appreciated. :+1:

DolphinDream:
> not really, i'm justifying why i suggested adding it into the List Mutator cat. But i understand why that seems a little out of place. (but it does take lists , and would produce mutated output...).

Aren't all nodes (more or less) mutating lists (values) ? :) Besides, this node has node values and also accepts inputs as single or vectorized (lists) values which then are processed and spit out a single value or a vectorized (list) value. (well, most outputs are list regardless of the input). I don't look at this node as necessarily as "mutating lists". I guess I'm starting to question my ability to connect the dots in your reasoning :)  :v:

DolphinDream:
Btw, does interpolating INTs make any sense? Would it be useful at all?

zeffii:
@DolphinDream  sometimes we need quantized/discrete values and not float..

DolphinDream:
@zeffii I see your discrete vs continuous point.. however, what does it  **mean** to interpolate between 0 and 1, for instance ? :) Such node would essentially give in a steppy output similar to using a float mix node followed by a rounding node. btw, I'm not arguing against an integer mixing node; just wondering what its practical uses would be. While designing the node I found it that I almost never used the INT type, but I'm sure it must have some uses that I can't think of.

zeffii:
isn't that the point of your node?... giving a higher range than 0....1 ?

zeffii:
if someone picks the Int mode, assume they are trying to interpolate between much larger values... 0 - 255 or something.

DolphinDream:
I agree.. I just pick an extreme situation.

DolphinDream:
I'm having some difficulties with the derived/base separation.
I want to keep the UI elements that are common (e.g. interpolation type, easing type) in the base class. But that means I have to make the base class derived from the SverchCustomTreeNode since these props have the updateNode as the callback.   Also the draw_buttons and draw_buttons_ext need to be kept in the base class since they draw the UI elements that are shared. The process method is also generic and should be kept in the base class. I'm assuming these methods are part of the SverchCustomTreeNode class/scheme, which is perhaps another reason why the base class should be derived from SverchCustomTreeNode. Any ideas how to handle this ?

DolphinDream:
Actually.. i think i got it :) I’ll push a update soon.

DolphinDream:
Ta-Da! Here come the Mix Input band.. MixFloats, MixInts, MixVectors and MixQuaternions.
I know.. quaternions are not quite at home in SV (why not i wonder), but I added them anyhow. Perhaps it would be a good to add a new category and nodes, sockets to deal with quaternions? 

sv-mixinputnodes

sv-mixinputs-demo1

DolphinDream:
Add MixLists node into the mix.

sv-mixlists-demo1

DolphinDream:
@zeffii to go back to your suggestion: I like the idea of a new Interpolation node category, though to be honest I also like the Animation Node organization of nodes as it keeps everything belonging to main categories: number, vectors, matrices, colors inside these categories.. there is a mix/interpolation node in every of these main categories. Intuitively I would go to the vector section if i’m looking for a vector mixing node rather than going to a mixing section to look for all sort of input mixing nodes. But that may be just the way my brain works. Although adding a new interpolation category could also allow us to give them a separate theme color (if needed).

zeffii:
> I also like the Animation Nodes organization of nodes in the search-menu 

maybe you should be trying to make AN more like  SV then, instead of the other way around?

DolphinDream:
@zeffii  1. I don’t know AN as much as I know SV. 2. And adopting things from other addons is not a bad thing. 3. And why would I make AN more like SV since I just stated how I think it’s more intuitive to me the way AN has it organized ?

zeffii:
fine, but you're forgetting that we have made SV the way it is because it is intuitive to us..

DolphinDream:
I see. yeah i keep forgetting that.

zeffii:
there's nothing stopping you from making your own menu addon, with exactly the arrangement you want.

DolphinDream:
Add Mix Color node into the mix

DolphinDream:
Add Mix Matrix node into the mix .. here’s a little preview:

sv-mixmatrices-demo1

DolphinDream:
another mix matrix demo

DolphinDream:
another mix matrix demo

sv-mixmatrices-demo3

zeffii:
I keep getting that nagging feeling that the mix node shouldn't be split up, and that instead it should have adaptive sockets.

DolphinDream:
@zeffii Yeah, I know what you mean. I had that feeling too. I'll update the MixNumber node to have the adaptive sockets and selective types and see how that feels. :) 

One thing about the one input mix node though is that (I anticipate) the workflow is not going to be necessarily easier even with adaptive sockets; for instance say you have a mixInputFloat hooked up with other nodes, then you switch the mixInputMatrix. It's likely that the float inputs even if they stay connected (with adaptive sockets) they are not quite what you would want and you may end up disconnecting the float nodes and reconnect some other matrix nodes instead.  At this point it's almost as if you're starting from scratch with your node setup. I can't think of a situation where changing the mix input type and having the sockets adapt would work out of the box without you having to actually rewire other nodes in/out. Perhaps int/float could be adaptive.. and even vector/matrix, though I have the feeling that in general a manual rewiring will be unavoidable.

zeffii:
I often use the nodeview as a "palette" while building up trees.  If I need a node, and I see an existing node in the view, i'll duplicate it and move it to where i want it. Being able to wire anything into the node then becomes useful. It's a split second decision between 

  1. Do I search for it, or..
  2. Do I can I duplicate an existing one..

The "automatic adaptive socket" is one way, or the 'select mode' is another.

portnov:
You actually have third option: to have separate parameter to switch modes.

DolphinDream:
@portnov I think that is what i had in mind.. a drop down to select from (int,float,vector,list,color,quaternion,matrix).

zeffii:
but rather than the names of the sockets you can use a node socket 'nipple' :)

DolphinDream:
A mix input node that adaptively selects the mode may not be kosher :) What if you have a float and a matrix inputs connected to a mix input node ? Or if you have two vector inputs and output connected to a node expecting a vector (vertice), then swap the inputs to be matrices ? how would the node adapt to that?  convert internally form matrix to a vector and keep the node in Vector mode? or have the node change to Matrix Mode and let the nodes connected to the mix node (which expect a vector) result in error ?

DolphinDream:
Do we have a NippleSocket class ? :)

zeffii:
no, but you can draw the dot in a layout..
like I did in the Monad interface

image

zeffii:
```python
r = _column.row(align=True)
r.template_node_socket(color=the_color)
```

zeffii:
ah.. that won't work inside an enum...   ahwell..

zeffii:
The option to pick a mode is preferable (Select / Switch ..whatever you want to call it). keep it simple.

zeffii:
and rather than calling the sockets `V1 V2` ,.. got for `A, B`  -- this at least is consistent with the Vector Math node..  ..or `"input 1"  "input 2"`  the socketnames are just strings..and can have spaces

zeffii:
using `replace_socket` however you can just rename the socket to whatever you want..

DolphinDream:
Alrighty then.. here’s the updated MixNumber node (the old node) augmented to have more types (colors, quaternions, matrices) and using replace_socket to keep the node connections when changing input types. Check it out and see how it feels :)

Note: the code is not cleaned up yet.. that’ll be in the next step if we want the unified mix input node rather than the separate mix input nodes. Also the node name will change to MixInputs. 

https://gist.github.com/76f8137d577377e47e3f4814d182858e

sv-mixinputs-demo1

DolphinDream:
With a single MixInput node it would be nice though if we could search for MixColors node, let’s say, and have it instantiate a MixInput node with a preselected color type instead of searching for MixInput and then changing to Color. Not a big deal, but I thought it would be slightly more convenient. Maybe that ought to be the tradeoff between having one vs many nodes and the usage convenience.

zeffii:
since we switched over to dynamic menu for shift+a,  we can add 'faux nodes' to the index.md where perhaps something like this _could_ make sense.

```text
## Vector
    GenVectorsNode
    VectorsOutNode
    SvAxisInputNodeMK2
    SvVectorMathNodeMK2
    VertsDelDoublesNode
    SvVectorRewire
    ---
    SvVertSortNode
    VectorDropNode
    macro SvMixInput {'selected_mode': "Vector"}

```

zeffii:
(even tho i'm not certain about the exact syntax)

zeffii:
see https://github.com/nortikin/sverchok/issues/1481 for further discussion on macros

DolphinDream:
Alrighty.. here’s the updated and cleaned up and compact the code. I changed the node name to MixInputs too.

The branch includes now both the MixInput node (with changeable type, socket replacing) and also the subclassed individual nodes (MixInt, MixColor etc). Check it out and see how both feel and decide which ones to keep.

and some gists for testing:
(scaling/rotating cube)
https://gist.github.com/e98d86a342358bc21becc889a091fd48

(various nodes to plug and play)
https://gist.github.com/bd7e860b5f285bc95855e497d98d228a

zeffii:
i'm not going to be too critical of this.. but the mixinputs node's code is a little obfuscated for my liking. I'd like you to learn from mistakes that i've made already.  Take a leap of faith and accept that `float_a` in the following example contains a value, as all variables do.. calling it `value_xxxx_x` in this case just means more complexity.
```diff
-        self.inputs.new('StringsSocket', "Float A").prop_name = 'value_float_a'
-        self.inputs.new('StringsSocket', "Float B").prop_name = 'value_float_b'
+        self.inputs.new('StringsSocket', "A").prop_name = 'float_a'
+        self.inputs.new('StringsSocket', "B").prop_name = 'float_b'
```

`replace socket`, can replace the socket kind without needing to change the name. [(see implementation)](https://github.com/nortikin/sverchok/blob/master/data_structure.py#L485).

You can be certain we as users know what when we select a mode (Float, Vector.. whatever) that we understand that the A and B expect floats. also the socket color does give some degree of hint too.

DolphinDream:
I agree. I’ll get rid of the silly “value” prefix. As for the A/B vs “Float A/Float B” thing.. i would agree with that too, except that when I make a connection to a socket I typically look at the socket name to figure out where to plug my inputs, and looking somewhere else to see the type first may be a bit inconvenient.  Also for inputs like the Vector vs Color vs Quaternions the color of the socket does not give it away what input is expected unless you look at the type. Perhaps for float/int is easy since you can see the current value (0, vs 0.0). Either way.. it’s an easy fix to switch everything to A/B.

DolphinDream:
refresh my last comment.

DolphinDream:
Perhaps if Colors and Quaternions (if we decide we need these) get their own sockets they could be colored differently.

zeffii:
really, keep it simple and use A/B   , use the `daw_label()` function to indicate on the header of the node what mixmode is being used.

DolphinDream:
Super. I kept it simple (A/B). I also added the draw label and updated documentation.

DolphinDream:
@zeffii let me know if the one mix input node is preferred over the individual mix nodes (currently in this branch). If so I can go ahead and remove the derived/base individual mix nodes. Also, as mentioned earlier we should probably have some faux nodes called MixColors, MixVectors etc (as you suggested perhaps via macros) so that they appear in the search as such.

zeffii:
also.. currently we use the StringsSocket for color information, see the Color In / Out node to show how to set the 'nodule_color' from green to the suggested yellow.

zeffii:
```diff
-    def update_sockets(self):
-        mode = self.mode
-        socketType = input_info[mode]["SocketType"]
-        # replace input sockets and link them to corresponding props
-        for i, l in enumerate(['A', 'B']):
-            prop = mode.lower() + "_" + l.lower()
-            self.inputs[i+1].replace_socket(socketType, l).prop_name = prop
-        # replace output sockets
-        self.outputs[0].replace_socket(socketType, mode.title())    def update_sockets(self):

+    def prop_name_from_mode(self, socket):
+        return self.mode.lower() + "_" + socket.name.lower()
+
+
+    def update_sockets(self):
+        # replace input and output sockets, linking input socket to corresponding props
+        new_socket_type = input_info[self.mode]["SocketType"]
+        for socket in self.inputs[1:]:
+            socket.replace_socket(new_socket_type).prop_name = self.prop_name_from_mode(socket)
+
+        self.outputs[0].replace_socket(new_socket_type, self.mode.title())  
```

zeffii:
you mention macros, i'm implementing [a secondary search tool](https://github.com/nortikin/sverchok/pull/1521), using standard blender ui methods. see here for `search+ and macros`

DolphinDream:
I switch colors to StringsSocket. I thought maybe Quaternions could do the same but unfortunately it represents the values in a row rather than dropwdown list (which is not kewl), so i think they should either be left as VerticesSocket for (eventually) get their own socket type. I tried to color the Quaternion sockets a different color, btw, but that only works if they are StringsSocket; if they are VerticesSocket it appears that the default color for VerticesSocket takes priority.

sv-mixinputs-quaternioncolors
sv-mixinputs-colors

zeffii:
For StringsSocket I implemented the `nodule_color`. specifically for the "fake" color socket.

have a look at the implementation of StringsSocket..  check out `custom_draw`  ..you can draw anything as long as you define the function ahead of time. see the implementation of `A Number` node.

zeffii:
```python

    def sv_init(self, context):
        self['float_'] = 0.0
        self['int_'] = 0
        self.inputs.new('StringsSocket', "Float").prop_name = 'float_'
        self.outputs.new('StringsSocket', "Float").custom_draw = 'mode_custom_draw'


    def mode_custom_draw(self, context, layout):

        if not self.show_limits:
            r = layout.row(align=True)
            r.prop(self, 'selected_mode', expand=True)
            r.prop(self, 'show_limits', icon='SETTINGS', text='')
        else:
            c = layout.column(align=True)
            kind = self.selected_mode
            c.prop(self, kind + '_min', text='min')
            c.prop(self, kind + '_max', text='max')
            c = layout.column()
            c.prop(self, 'show_limits', icon='SETTINGS', text='')
            c.prop(self, 'to3d', icon='UI', text='')

```

zeffii:
so if you wanted you could have a flip up menu..or a compact menu on the same socket. row()/multirow

zeffii:
i'll preface this by saying that I don't propose you do it this way, but it should hint at what can be done
```python
# ##### BEGIN GPL LICENSE BLOCK #####
#
#  This program is free software; you can redistribute it and/or
#  modify it under the terms of the GNU General Public License
#  as published by the Free Software Foundation; either version 2
#  of the License, or (at your option) any later version.
#
#  This program is distributed in the hope that it will be useful,
#  but WITHOUT ANY WARRANTY; without even the implied warranty of
#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
#  GNU General Public License for more details.
#
#  You should have received a copy of the GNU General Public License
#  along with this program; if not, write to the Free Software Foundation,
#  Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
#
# ##### END GPL LICENSE BLOCK #####

import bpy
from bpy.props import FloatVectorProperty
# import mathutils
# from mathutils import Vector
# from bpy.props import FloatProperty, BoolProperty
from sverchok.data_structure import updateNode

socket_colors = {
    "COLOR": (0.899, 0.8052, 0.0, 1.0),
    "QUATERNION": (0.9, 0.4, 0.7, 1.0)
 }


class SvFKB(bpy.types.Node, SverchCustomTreeNode):
    ''' a SvFKB f '''
    bl_idname = 'SvFKB'
    bl_label = 'SvFKB _ :)'
    bl_icon = 'GREASEPENCIL'

    quaternion_a = FloatVectorProperty(subtype='QUATERNION', min=0, max=0, size=4, default=(0,0,0,1))

    def sv_init(self, context):
        A = self.inputs.new('StringsSocket', "Float")
        A.prop_name = 'quaternion_a'
        A.custom_draw = 'draw_one_socket'
        A.nodule_color = socket_colors['QUATERNION']
        OUT = self.outputs.new('StringsSocket', "Float")
        OUT.nodule_color = socket_colors['QUATERNION']

        layout.label('helios')

    def draw_one_socket(self, context, layout):
        if not self.inputs[0].is_linked:
            c = layout.column(align=True)
            c.prop(self, 'quaternion_a', text='')

    def process(self):
        _in = self.inputs[0].sv_get()
        self.outputs[0].sv_set(_in)


def register():
    bpy.utils.register_class(SvFKB)


def unregister():
    bpy.utils.unregister_class(SvFKB)


if __name__ == '__main__':
    register()
```
then
```python
>>> nodes.new('SvFKB')
```

what I should do is modify the custom_draw feature to pass the socket too so..
```
    def draw_one_socket(self, context, layout):
        if not self.inputs[0].is_linked:
            c = layout.column(align=True)
            c.prop(self, 'quaternion_a', text='')
```
would become
```python
    def draw_one_socket(self, socket, context, layout):
        if socket.is_linked:
            c = layout.column(align=True)
            c.prop(self, socket.prop_name, text='')
```
image

but i'd prefer to have the option to include a flipup/down to show this ..or a collapsed version.

zeffii:
so hold off with that.. i will sort it out :)

zeffii:
https://github.com/nortikin/sverchok/pull/1532/files

zeffii:
i've commited that
```python
 def draw_one_socket(self, socket, context, layout):
        if not socket.is_linked:
            c = layout.column(align=True)
            c.prop(self, socket.prop_name, text='')
```
this will now work..

zeffii:
on second thoughts.... dont use this yet.

DolphinDream:
I'll wait. I'd be nice to have an expander button to toggle between and expanded list and a drop down list. I wonder if this could be implemented at the lower level for all sockets so they all nodes can expand/collapse the multi value inputs (vertices, colors, quaternions)

zeffii:
One hand I like giving full draw control to the node, on the other hand it's paramount to have logical convenient default drawing for a socket type - no need to make it complicated by default.

we're constantly on a tightrope between stable features and experimental, that's why I like the custom_draw stuff.. and it's not a problem when we all understand how it works..

DolphinDream:
Are you going to add the socket as the argument to the custom_draw method ? I think we need that to distinguish between the sockets within the same method, otherwise I’d have to define another method one for each socket.

DolphinDream:
@zeffii btw.. how do i draw the multi row property as a drop down (the same way the vertex socket is drawn now). I though this would work but it doesn’t:

```python
def mode_custom_draw(self, context, layout):
        c = layout.column(align=True)
        split = c.split(percentage=.14, align=True)
        c1 = split.column(align=True)
        c2 = split.column(align=True)
        if self.expand_a: # expand
            c1.prop(self, "expand_a", icon='TRIA_UP', text='')
            c1.label(text="A")
            c2.prop(self, "quaternion_a", text="", expand=True)
        else:
            c1.prop(self, "expand_a", icon='TRIA_DOWN', text='')
            c2.prop(self, “quaternion_a", expand=False)
```

sv-mixinputs-demo3

DolphinDream:
Also, your code example for custom_draw which only draws whenever the socket is not linked, it will not show anything when the socket IS linked, while it should actually show the socket name and the number of input items (just like other nodes).

DolphinDream:
something like this... (except that I still need to solve the drawing when socket is linked to show input name and socket input info.

sv-mixinputs-demo4

zeffii:
I did already commit the socket version of custom draw, am tempted to add 'text' as a last argument too (see StringsSocket implementation of draw)

I'd prefer to see the dropdown button to the right of the values, but other than that the UI as you've made it is essentially the same as I had hoped it would be. I think (and you probably agree) that it does make sense to let the node decide how to draw a socket -- but for most cases we should still use the standard drawing function for consistency.

DolphinDream:
The reason why I added the dropdown button to the left is to save some space, since I added the label (a/b) also to the left. If i add the label to the left and button to the right, it leaves less space for the value fields. Adding the label to the top and button to the right is also an option but it increase the node high a bit.

Here are other versions. The one on the left has the issue that the expand button moves down once you click it. The one on the right doesn’t have that. both of them take additional space, one on the top the other on the left. 

sv-mixinputs-demo6 sv-mixinputs-demo7

zeffii:
> Perhaps a slight improvement would be to make the label A/B be centered with the socket nodule but I don’t know how to do that :)

that's not possible :/

zeffii:
well... " not possible " is not true... but you'd need to generate empty rows... and that's painful code to read.

anyway..
```python
    def draw_one_socket(self, socket, context, layout):
        node = self
        if not socket.is_linked:
            c = layout.column(align=True)
            c.prop(node, socket.prop_name, text='')
        else:
            if socket.prop_name:
                prop = node.rna_type.properties.get(socket.prop_name, None)
                t = prop.name.split('_')[1].upper()
            else:
                t=''
            layout.label(t + '. ' + SvGetSocketInfo(socket))

```

just parking the code here https://gist.github.com/zeffii/c1429b16c591c1469527e1636c8f5da5

zeffii:
stick with this one : https://cloud.githubusercontent.com/assets/2083719/25109893/28730248-23ae-11e7-9b64-a589d6e2e740.gif   i'm starting to like it .

zeffii:
image

```python
    def draw_expander_template(self, context, prop_origin, prop_name="prop"):
    
        if self.use_expander:
            split = layout.split(percentage=.2, align=True)
            c1 = split.column(align=True)
            c2 = split.column(align=True)
            if self.expanded:
                c1.prop(self, "expanded", icon='TRIA_UP', text='')
                c1.label(text=self.name[0])
                c2.prop(prop_origin, prop_name, text="", expand=True)
            else: 
                c1.prop(self, "expanded", icon='TRIA_DOWN', text="")
                row = c2.row(align=True)
                row.template_component_menu(prop_origin, prop_name, name=self.name)
        else:
            layout.template_component_menu(prop_origin, prop_name, name=self.name)



   def draw(self, context, layout, node, text):

            if self.prop_name:
                self.draw_expander_template(context, prop_origin=node, prop_name=self.name)
            elif self.use_prop:
                self.draw_expander_template(context, prop_origin=node)
            else:
                layout.label(text)

        elif self.is_linked:
            layout.label(text + '. ' + SvGetSocketInfo(self))

        else:
            layout.label(text)
```

zeffii:
updated ( github doesn't warn you if i make changes )

image

```python
    def draw_expander_template(self, context, layout, prop_origin, prop_name="prop"):
    
        if self.use_expander:
            split = layout.split(percentage=.2, align=True)
            c1 = split.column(align=True)
            c2 = split.column(align=True)
            if self.expanded:
                c1.prop(self, "expanded", icon='TRIA_UP', text='')
                c1.label(text=self.name[0])
                c2.prop(prop_origin, prop_name, text="", expand=True)
            else: 
                c1.prop(self, "expanded", icon='TRIA_DOWN', text="")
                row = c2.row(align=True)
                row.template_component_menu(prop_origin, prop_name, name=self.name)
        else:
            layout.template_component_menu(prop_origin, prop_name, name=self.name)



   def draw(self, context, layout, node, text):

            if self.prop_name:
                self.draw_expander_template(context, layout, prop_origin=node, prop_name=self.name)
            elif self.use_prop:
                self.draw_expander_template(context, layout, prop_origin=node)
            else:
                layout.label(text)

        elif self.is_linked:
            layout.label(text + '. ' + SvGetSocketInfo(self))

        else:
            layout.label(text)
```

DolphinDream:
I removed the socket dev code from this branch. The socket dev to be continued on its own branch (#1551)

DolphinDream:
@zeffii I’m working on finishing this PR (now that we have Quaternion nodes).. though, given that the R is kind of old I was wondering if it’s better to just close it and reopening another one to do the MixNumber -> MixInput restructuring. This PR has helped develop the MixInput node, but it also kind of diverged into other topics like the "socket restructuring" as well :)

Let me know what you prefer. I can either continue to add the latest development to this PR or start another one.

zeffii:
start a new PR by all means.

DolphinDream:
Closing this PR and opening a new PR #2347 to continue the restructuring development/discussion.

DolphinDream:
Closing this PR and opening a new PR #2347 to continue the restructuring development/discussion.

PR #1441, opened 2017-03-31
GP Node Push 2

zeffii:
the less talking happens on this pull request, the faster it will be completed.

### DIFF
--- a/nodes/viz/viewer_gp.py
+++ b/nodes/viz/viewer_gp.py
@@ -40,6 +40,12 @@ def set_correct_stroke_count(strokes, coords):
         # remove excess strokes
         for _ in range(diff):
             strokes.remove(strokes[-1])
+    strokes.update()
+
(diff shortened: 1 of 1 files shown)


zeffii:
image

zeffii:
my experience shows that GP is unstable when controller dynamically, expect many occurrences of Blender closing to the desktop without warning.

nortikin:
i think it is from creating new layers for GP

zeffii:
currently it doesn't create new layers, only strokes.  (add / removes  / update strokes )

zeffii:
seen as it's not possible to get the Layername, from the Stroke..in `strokes`, (not even using strokes.id_data - which just returns the GP name (like id_data returns the NodeTree, not the node )

zeffii:
a slight redesign is needed. It will look more like the other viewers i think..

zeffii:
this will not be the PR..

PR #1449, opened 2017-04-01
adds ui reshuffle

zeffii:


zeffii:
image

zeffii:
```md
[12:39] <zeffii> ideasman42, is there any way to figure out what we need to disable or set?  
        https://github.com/nortikin/sverchok/blob/master/nodes/viz/texture_viewer.py  
        it only happens on osx..not on windows/linux
[12:40] <ideasman42> zeffii, just set texture state back to how it was originally
        <ideasman42> zeffii, there are many gl-debugging tools
```
@kalwalt

kalwalt:
@zeffii  i like the new layout,

i will look into it again if i have a bit of time.

enzyme69:
I would probably never mix float scalar modulo and degree that way, but thanks for showing.

PR #1454, opened 2017-04-01
Vector In: get 3D cursor location

portnov:
This adds a button to Vector In node, which assigns current location of 3D Cursor to X,Y, Z parameters of node. This button is available only when no inputs are connected.

### DIFF
--- a/docs/nodes/vector/vector_in.rst
+++ b/docs/nodes/vector/vector_in.rst
@@ -18,6 +18,11 @@ Outputs
 
 **Vector** - Vertex or series of vertices
 
+Operators
+---------
+
+This node has one button: **3D Cursor**. This button is available only when no of inputs are connected. When pressed, this button assigns current location of Blender's 3D Cursor to X, Y, Z parameters.
+
 Examples
 --------
 
--- a/nodes/vector/vector_in.py
+++ b/nodes/vector/vector_in.py
@@ -17,11 +17,28 @@
 # ##### END GPL LICENSE BLOCK #####
 
 import bpy
-from bpy.props import FloatProperty, BoolProperty
+from bpy.props import FloatProperty, BoolProperty, StringProperty
 from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode, fullList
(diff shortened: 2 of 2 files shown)


zeffii:
why not hijack the XYZ node instead?

zeffii:
@portnov 

how about a dedicated 3d cursor node, that can also store a number of locations?

enzyme69:
@zeffii @portnov "A number of locations" you mean multiple 3D Cursor Node? Like Empty? I remember having Sverchok reading 3D cursor and we can sort of have interactive setup like objects moving into 3D cursor position. Or maybe I was dreaming.

enzyme69:
@portnov Can this "button" have an option to continuously updating and tracking when Frame Updates so that objects can fly into this 3D cursor position?

zeffii:
A separate Node, just for 3d cursor. makes more sense than to add these extra features to the Vector In node.

enzyme69:
@zeffii @portnov Ah, yeah, I thought about that too hijacking "XYZ node" might keep "Vector In" cleaner. 

When a user plugging Range Float to Vector In, the might be surprise on why the button disappear, but it's all optional, I guess. I leave it to you guys~

zeffii:
but the reason i retracted the offer to use XYZ instead, is because the XYZ node is really just for _axis_.

nortikin:
agree, idea is good deserve separate node with additional functions for cursor.

portnov:
And what should it do?

nortikin:
personally i agree zeffii, separate to new cursor node. leave vector node as it was before

nortikin:
new cursor node can have input position and output position also

zeffii:
we can even add a nice bgl drawing into 3d view of stored 3d locations.

zeffii:
with custom colours per 3d reference..

enzyme69:
And custom option where it can be triggered when:
- frame updated
- cursor position changed
- other
-- 


- Jimmy Gunawan

enzyme69:
screen shot 2017-04-04 at 15 25 42

Gotta admit that sometimes this button to get 3D Cursor position can be handy, in this case for me getting the pivot for the spider legs.

PR #1455, opened 2017-04-01
Support multiple "centers" in mesh_select node

portnov:
In "by center and radius" mode, if many centers are passed, then "distance" means "distance from vertex to nearest of the centers".


portnov:
by-center-many
by-center-many-move

nortikin:
ahaha!!! good, will reblog

enzyme69:
@portnov That's totally wish came true! Thanks for this one. I didn't think of if this way, I was thinking using Displace Modifier trick or maybe some kind of "influence" or Transfer Attribute node, but this masking will work too!

Now if only Grease actually stick to the mesh surface, but that's another thing.

Thanks for this one.

@nortikin You have some ideas how to use this? 👯‍♂️

nortikin:
@enzyme69 proportional edit not so clear, Try this one:
[gp+selector+proportional_2017_04_03_16_29.zip](https://github.com/nortikin/sverchok/files/890852/gp.selector.proportional_2017_04_03_16_29.zip)
gp selector proportional

nortikin:
or like this
[gp+selector+proportional_2017_04_03_16_32.zip](https://github.com/nortikin/sverchok/files/890856/gp.selector.proportional_2017_04_03_16_32.zip)

import as blend zip directly

enzyme69:
@nortikin Studying it now, thanks for examples~

enzyme69:
@portnov @nortikin Testing it again, but only works on frame one and not multiple strokes. Rather strange...

[proportional_edit_002_2017_04_04_01_33.zip](https://github.com/nortikin/sverchok/files/891867/proportional_edit_002_2017_04_04_01_33.zip)

screen shot 2017-04-04 at 11 32 35

PR #1462, opened 2017-04-03
Nodeview Console

zeffii:
## do not merge

Maybe i should call this `SV Console` , and make search just a feature..just like the other functionality that you mention @nortikin . I was going to make a Console at some point, but this provides all the practical infrastructure I needed anyway.. 
```lisp
[-] >>> somename       # searches the registered Node list (bl_idname, filename, bl_label)
             |
             \.... this is a basic search implementation, i invite different engines at some point.

[ ] >>> !macroname     # get from a list of macros.
[x] >>> term?bpy       # open bpy docs searching the string
[x] >>> term?py        # open bpy docs searching the string
[x] >>> term?sv        # search sv docs for the string (read the docs)
[x] >>> term?gh        # search github sv issues
[x] >>> term?se        # search BlenderStackExchange
[x] >>> term?ghc       # search github sv code
[ ] >>> #templatename  # search templates for SNLite
[ ] >>> #ls            # list templates for SNLite
[ ] >>> .templatename  # search sverchok demo templates
[ ] >>> .ls            # list sverchok demo templates
[ ] >>> man            # throw the console manual to a webbrowser
[ ] >>> ls gist        # LS , list
[ ] >>> ?gist <term>   # parse all current csv into a mega tree once , and allow search of term in node tree/blendfile name

```

blisssgif


DolphinDream:
Any shapshot previews ? :)
Btw, in the previous screehshot... why is the "Exec Node Mod" listed for a "ve" search string ? It doesn't seem to have a "v" followed by an "e" in it.

zeffii:
because the bl_idname is SvExecNodeMod - The algorithm isn't something i've put any time into yet.. just being able to draw the names and pick a node will be step1. Then i hand it over to anyone wanting to make a better search algorithm.

zeffii:
but rather than it just being a search tool... it could also have 'canned' responses to certain input. So we generate a map between 'short input vs expected output'

```python
{   'vin' : Vector Input
    vo : Vector Output
    ci ..Color In
    co :ColorOut
   etc...
}

**Comment by DolphinDream** [comment]
Looking gewd. :) I know this is WIP.. but wouldn’t it make more sense to have the node name first.. then the class name ? 

Also, you know what I was thinking i would be super awesome down the road.. if you can have a preview of the node once you select one from the list. Often time I search for node and I’m not sure which node i’m looking for and basically I have to create them to see which one they are and delete if they are not the ones i’m looking for. A preview could help with showing the node before you create it. Just an idea.

fuzzysearch-nodepreview

**Comment by DolphinDream** [comment]
also.. some sort of predefined priority could be given to the results that would help the most frequent uses. For instance if i type **int** i pretty much want an **integer** node not an “intersect edges”,  but the regular search does not have int the first item in the list.

**Comment by zeffii** [comment]
>  I know this is WIP.. but wouldn’t it make more sense to have the node name first.. then the class name ?

of course.

- Regarding the shorthand stuff, that's really down to adding a `mapping dict` i think
- The algorithm for fuzz search is not implemented yet (it must be designed first....)
- I like the Node UI display notion, it could be done 3 ways..
    - (easy) include a way to add a selected node at the mouse cursor, and upon changing selection replace it with the other node. upside is the node exists already if you accept it..downside the view might have to be automatically zoomed to the extent where the node becomes clearly readable
    - (unlikely) include raster images in a folder ...forget about this.
    - (more unlikely) include an inspector function to turn a node reference into a bgl drawing.

but first..there's a small issue with kb handling to solve first.

**Comment by nortikin** [comment]
this is another reason for adding predicate for category i.e. 
ListItem
or even short 
L_Item
M_in
V_in
G_PlaneMK2
A_Volume

etc

**Comment by zeffii** [comment]
image

**Comment by nortikin** [comment]
maybe search can handle docs and also templates or SN1 SN light
example 
you write `script` and it propose you scripts from folder
you write `docs` and it propose you some 

or better 
`#` symbol or easier `.`
popup SN1template and SNLtemplate and Docs points
choose oтe of them, appears sublevels of docs folder or templates to choose from

**Comment by zeffii** [comment]
maybe i should call this `SV Console` , and make search just a feature..just like the other functionality that you mention @nortikin . I was going to make a Console
````
>>> ?somename    # searches for this 
>>> somename?bpy   # open bpy docs searching the string
>>> ls tmpl snl    # list templates for SNLite
>>> ls tmpl sv   # list sverchok demo templates
>>> man   # throw the console manual to a webbrowser
```
etc...  not yet.

zeffii:
i'll update the first message in this thread to contain my proposals for the various modes of the `console`.

zeffii:
gagagif

zeffii:
macros could be defined this way

```python
user_macros = {
    # "macro_string":  [add("bl_idname"), set("property_name", value)]
    "mixcol": [add("SvMixInputs"), set("selected_mode", "Colors")]

}
```

DolphinDream:
>(zeffii) I like the Node UI display notion, it could be done 3 ways..
(easy) include a way to add a selected node at the mouse cursor, and upon changing selection replace it with the other node. upside is the node exists already if you accept it..downside the view might have to be automatically zoomed to the extent where the node becomes clearly readable

A note on the node UI regarding the zooming.. In order to see the node at a readable size I can think of a few solutions:
- Automatically zoom the node editor to the normal zoom while searching .. and zoom back to the user defined zoom after the node is dropped into the node editor canvas. this may be visually disturbing though to have the node editor zoom in /out every time you search for a node.. but  it may be actually fine. We’d need to play with it to see how it feels.
- (not sure if this is possible) but have the node rendered to offscreen buffer into a texture.. then rendered into the search panel (not actual node). On selection the image of the node zooms in or out to the match the actual node placed in the node editor.
- have SV automatically generate screenshots of all the nodes and save them to disk on load and display as images in the search panel .. and also have them regenerate if sverchok is updated to new version. The images could also be generated (cached on disk) on the fly as they are searched for the first time.

DolphinDream:
I know this is more of a styling for later thing.. but can the console be semitransparent :) for slickness ?

DolphinDream:
Note on the “obj vd” macro (multiple node instantiation) that is super cool. I was thinking that a similar functionality could be added for nodes already in the node tree. I like the CTRL+SHIFT + LMB on a node in material editor / compositor that links the current node to a viewer node. I use that a lot. The question is what would a “viewer” be of a generic node that doesn’t have things to be displayed by a viewer draw node.. perhaps in most cases the viewer would be the stethoscope.. for some nodes it could be either the viewer draw node or a stethoscope. perhaps with different key combination it could cycle the output nodes into a stethoscope node, while other key combo would cycle link vertex/edge or vertex/poly to a viewer draw node. 

Alternatively (@zeffii) though I know you hate me mentioning AN .. we could have some sort of pie chart we can invoke for a node via a key.. and let the user select an operation to be done on a node, whether is a connection to a stethoscope, linking to viewer draw or other conceivable operations.

enzyme69:
I have some kind of idea: maybe we could also load GITS using keyword. Or
maybe loading "presets"
using typed way. which is awesome. oke no emoji but till awesome!!
-- 


- Jimmy Gunawan

zeffii:
Try harder to limit the scope of the discussion to what makes sense as features of the console vs what should maybe be a keyboardshortcut triggered dynamic menu like nodewrangler.

zeffii:
Contextual menus for rightclik or whatever should be discussed separately, there are things I'd to see implemented in that arena too. Like rclik node, insert before/after..   etc different topic.


Semi transparant bg.. could be a user setting.
Regarding the node.visualization, ill opt for simplest route first.

zeffii:
@enzyme69  i'm on board about the gist stuff. You might not be aware of this... but Sverchok stores 
 - the gist url and 
 - blendfile name
 - date and time 

of each gist you upload anonymously, in a folder called `datafiles/sverchok/`

image

i imagine a console command like 
```python
>>> ls gist            # LS , list
>>> ?gist <term>   # parse all current csv into a mega tree once , and allow search of term in blendfile name
```

zeffii:
@DolphinDream 

> may be visually disturbing though to have the node editor zoom in /out every time you search for a node.. but it may be actually fine

i'm not too concerned about showing the node, I think the described method will work (auto zoom), i'll get to that in due course.

> node rendered to offscreen buffer into a texture.

I asked devs a long time ago for that feature to help make screenshots for the docs too, rendering the node ui to a buffer would be great. Someone else will have to ask / lobby. With the growth of new node systems I think it's something that should be generalized and not implemented by us.

> auto screenshots for docs / menu

You are welcome to write such a process, but make it not Sverchok specific / separate ..so other systems can use it.

enzyme69:
gist thing I like a lot!!
-- 


- Jimmy Gunawan

zeffii:
actually, the csv stores the node tree name, not the blend name... perhaps it should store the blend name too.. but that's a separate issue.

zeffii:
please note:  I will periodically sanitise the thread here. i'm allergic to scrolling.

zeffii:
@nortikin 

> more lines maybe to choose from

yeah, maybe a user setting.. but i'd like to first enable a scroll if more items than 10 are returned.. and display the number of (num_total_results minus num_showing) so user gets an idea of how many results were found

zeffii:
a few search tools are added now , github issues is one of them

image

zeffii:
the routing table for search online
```
    if input_string.endswith(r'?\bpy'):
        type_of_search = 'bpy docs'
    elif input_string.endswith(r'?py'):
        type_of_search = 'py docs'
    elif input_string.endswith(r'?sv'):
        type_of_search = 'sv docs'
    elif input_string.endswith(r'?ghc'):
        type_of_search = 'github code'
    elif input_string.endswith(r'?gh'):
        type_of_search = 'github issues'
```

portnov:
This looks like super-mega-tool...
* Maybe it should be separate sverchok-indenepdent plugin?
* Isn't it possible to use blender's standard widgets system, with theming and so on?

zeffii:
> Isn't it possible to use blender's standard widgets system, with theming and so on?

nope, you can't get the content of the currently typed input string using `wm.invoke_search_popup`

> Maybe it should be separate sverchok-indenepdent plugin?

well, i already have a console addon for non sverchok stuff... so i'm just adding in the things I'd like regular sverchok users to have. When i get around to macros ..i think it should be very interesting.

zeffii:
ok. i won't commit this to Sverchok :)   it will be a separate addon.

zeffii:
i think this should be a separate addon too, but i'll probably not make it sverchok specific.

nortikin:
cool. not like pie menu. this line approach better as for me

enzyme69:
@zeffii I will be very happy to test this as separate add-on. Maybe faster that way too right? Is this like a command line way to:
- generate node
- connect node
etc?

zeffii:
Completely unrelated to speed

PR #1466, opened 2017-04-04
Math MK2 (v2)

zeffii:


kalwalt:
i added a minimal doc with this commit: https://github.com/nortikin/sverchok/pull/1438/commits/b1d527e02909c809f35cadda3ae284d505a3d246 
if you think it is missed some function or other let me know.

zeffii:
that's good thank you. appreciated! 

i'm still not ready to commit the node, but we can certainly commit the docs ahead of time.

zeffii:
image

zeffii:
no.. you should export the monad while viewing a regular node tree, not from inside the monad.

zeffii:
almost ready to push.

kalwalt:
Could be increased the float precision of the node? In the `random monad` for the noise nodes tut.  i should multiply for 43758.5453123  instead i can only do with 43758.55, this make a big difference. I have a similar problem for the Vector in node the input should be 12.9898 for the x component by a i can't, i must round to 12.991 with a slight difference in the result...

zeffii:
Ill see what can be done about adjusting precision.

zeffii:
i'm afraid 32bit float offers only limited precision.. http://stackoverflow.com/q/872544/1243487

zeffii:
there's a argument we can set when defining the property, called precision. Which by default is set to 2 (if we don't set it explicitely) . That value however just represents the visual precision of the number on the slider, you can for example still type a more elaborate number.. but only expect 5 or 6 significant figures for 32bit float (single precision).

so `43758.55`, or `43.75855`, but not  ` 43758.5453123`

kalwalt:
Thank you for the answer. but not understood, if i type in the slider:
`43758.54531` the number in the slider became `43758.55`. you mean that internally is stored as 5 values after the float but i can see only 2?

zeffii:
i'm saying that the precision you see on the UI doesn't relflect the precision that FloatProperty is capable of, however.. the precision is not very high. only about 5 or 6 (or 7..but that's pushing it) significant digits. `43758.54531`  is..going to lose precision around `43758.55`

zeffii:
so.. something like 
```
437585.5
43758.55
4375.855
437.5855
43.75855
4.375855
```

kalwalt:
Wait, you are saying about the new node math_mk2? at the moment i'm using the old one. Anyway plugging a `debug print node` after a math node with `4.375854969024658` in the slider but show me `4.38` in the console i have: `[[4.375854969024658]]`. So i need to use a debug print node to inspect the real value?

zeffii:
yeah i'm talking about the current scalar math node. but also in general about the UI effects of the FloatProperty slider.  Also yes, but if you keep the blender python console open you can press ctrl+C on the slider, and go into the console and ctrl+v it will paste the value.

image

zeffii:
image

zeffii:
here i typed ` 2.56746 ` into the slider value

image

kalwalt:
Thank you. this help me. i will try on my nodes now.
especially this: 
```python
>>> nodes["Math"].y
2.567460060119629
```
i don't need a debug print node anymore, better than a console ( in linux)

PR #1480, opened 2017-04-08
IcoSphere node.

portnov:
I think this node was waited enough :)

Some time ago @zeffii said that it would be better to have one node representing all bmesh primitives. But, for now, Sverchok already have more extensive set of primitives then bmesh does. Only primitives left (implemented in bmesh ,but not in sverchok) are Icosphere and Suzanne.

### DIFF
--- a/docs/nodes/generators_extended/generators_extended_index.rst
+++ b/docs/nodes/generators_extended/generators_extended_index.rst
@@ -6,6 +6,7 @@ Generators Extended
    :maxdepth: 2
 
    box_rounded
+   icosphere
    generative_art
    hilbert
    hilbert3d
--- a/docs/nodes/generators_extended/icosphere.rst
+++ b/docs/nodes/generators_extended/icosphere.rst
@@ -1,0 +1,40 @@
+IcoSphere Node
+==============
+
+Functionality
+-------------
+
+This node creates an IcoSphere primitive. In case of zero subdivisions, it simply produces right icosahedron.
+
+Inputs
+------
+
+This node has the following inputs:
+
+- **Subdivisions**
+- **Radius**
+
+Parameters
+----------
+
+This node has the following parameters:
+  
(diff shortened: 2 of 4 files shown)


portnov:
icosphere

zeffii:
oooooookaay :)   i think though you might want to implement a `limit` not only a soft_limit on the IntProperty for subdiv... but also a "hard limit" property set to a sane 5 by default in `draw_buttons_ext` ,..because 6 or 7 is going to be a very unpleasant experience..especially vectorized.

zeffii:
and the hard limit would also apply to values incoming from sockets..

zeffii:
icosphere can be a prime generator,  seems a lot more useful than having it in `generators extended` .

zeffii:
@portnov   please :)

kalwalt:
This is a generator that i wait a lot, really appreciated that is in the family now.
I agree with you @zeffii , maybe better to limits the num. of subdivisions on the node ( but maybe let this freedom in the properties?)

nortikin:
basic generators also ring planemk2 linemk2

enzyme69:
I imagine: Procedural Suzanne one can tweak the eyes, mouth. Maybe one day...

portnov:
@zeffii 
Is it possible to set hard_max for property programmatically, not in IntProperty() constructor? If so, how? :)

portnov:
@enzyme69 
Blender's Suzanne primitive creation procedure does not have any parameters to tweak at all. But, if you want, you may use "mesh expression" for that. Though writing down required formulas for such a model can get a lot of time...

zeffii:
Check out the new " A.number " node for example on how to use .get .set, unfortunately hard and soft limits can not be overwritten _ get/set is the nearest solution.

portnov:
@zeffii 
See #1489.

@nortikin , @zeffii 
I'd suppose not only move Plane and Line Mk2 to Generators, but also deprecate old Line and Plane and remove them from menu. Why would we need that old nodes, new nodes have more features...

nortikin:
do it, @portnov not even ask.

PR #1520, opened 2017-04-15
adds macro loading from index

zeffii:
image

```text
## Number

    # this means, register the node (bl_idname)  but hide it from custom menu
    <<< SvNumberNode

    # this means add a macro for this entry, 
    #   | bl_idname | menu label | (parameters,....)
    >>> SvNumberNode Int (selected_mode="int")
    >>> SvNumberNode Float (selected_mode="float")
    Float2IntNode
    ScalarMathNode
    SvScalarMathNodeMK2

```
image

##### not implemented now

I think for now i'll force single string Labels... like `Int` or `Float` above.. space separated labels should be done using `some_space_separated_name` , it's easy to convert underscores to space and change it to title case.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -92,7 +92,7 @@ utils_modules = [
     "text_editor_submenu", "text_editor_plugins",
     # UI operators and tools
     "sv_panels_tools", "sv_gist_tools", "sv_IO_panel_tools", "sv_load_zipped_blend",
-    "monad", "sv_help",
+    "monad", "sv_help", "sv_macro_utils",
     #"loadscript",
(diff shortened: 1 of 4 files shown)


zeffii:
the "search by string" feature ofcourse won't find the macro. for this the search feature would need to be entirely replaced. I've already written this code.. it all seems a bit messy.

zeffii:
image

image

zeffii:
I'm not happy with the extra complexity of this.. OK. not going to happen. 

a custom simpler searchbox will be needed i think..

PR #1523, opened 2017-04-16
Bmesh ops mk2

Kosvor2:


Kosvor2:
still broken (needs to update entire node-tree to apply changes of parameters to bmesh without error), but now looks better. All this venture with creating a Bmesh in one nodes and processing it in others was not very smart.

PR #1521, opened 2017-04-15
shift + ]  menu

zeffii:
image

### Usage

- `shift+]` will trigger a new search bar.  
- The search database is loaded with 
    - All nodes (shows Node Name (bl_label) and Description
    - Several default macros ( shown by `>` )
    - optionally User Driven Macros ( shown by `<` )
        These need to be created by the user in `/datafiles/sverchok/user_macros/macros.py` . See an example of such a file here

### Implementation details

- [x] implement macro searching
     - ( i can not control the width of the window.. unless there's a trick somewhere...)
- [x] select / place node
- [x] implement default macros
    - [x] `obj vd`   (single object to vdmk2)
    - [x] `objs vd`   (multi object to vdmk2)
    - [x] `sn petal`  (auto loads petalsine in snlite)
    - [x] `zen` (the zen of sverchok, shows [my tumblr post about the zen of it all](http://blenderpython.tumblr.com/post/91951323209/zen-of-sverchok))
    /- [ ] swap viewer  (hot swap viewer nodes  vdmk3 <-->  bmeshview )
- [x] Auto Line splitting on `///` for description
- [x] `>` will indicate the result is a macro
- [x] `<` will indicate the result is a user macro
- [x] Auto Line wrapping for description
- [x] implement user driven macro definitions.
- [x] pass mouse location.

PR #1529, opened 2017-04-16
"Select similar" node

portnov:
This is the implementation of Blender's "select similar" (Shift+G) feature.

### DIFF
--- a/docs/nodes/analyzers/analyzers_index.rst
+++ b/docs/nodes/analyzers/analyzers_index.rst
@@ -13,6 +13,7 @@ Analyzers
    kd_tree_edges_mk2
    mesh_filter
    mesh_select
+   select_similar
    proportional
    normal
    normals
--- a/docs/nodes/analyzers/select_similar.rst
+++ b/docs/nodes/analyzers/select_similar.rst
@@ -1,0 +1,74 @@
+Select similar
+==============
+
+Functionality
+-------------
+
+This node allows to select mesh elements (vertices, edges and faces), which are similar to elements already selected in some aspect.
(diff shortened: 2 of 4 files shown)


zeffii:
cool..

can you make the docstring `""" <short descrption>  ///   <longer description, ...rest > """`

zeffii:
for instance:
```python
class SvProfileNode(bpy.types.Node, SverchCustomTreeNode):
    '''
    svg-like 2d profiles ///

    SvProfileNode generates one or more profiles / elevation segments using;
    assignments, variables, and a string descriptor similar to SVG.
```
image

portnov:
@zeffii 
Done, please check.

zeffii:
sorry to be a pain :)

zeffii:
it will show `bl_label    <    >    description  `   best make the description give information not present in the bl_label

zeffii:
maybe.. "Like blender's ctrl+G ///    """ i think that fits

kalwalt:
should be added in all the nodes this?

portnov:
@zeffii 
ok :)

zeffii:
well... if we're going to be using the extended search menu more... it would be nice yes :)

zeffii:
and the docstring is used currently _nowhere_

kalwalt:
ok...so plans for the near future!

kalwalt:
maybe when it will be in the master, we can add to all nodes. (or at list one part of them)

kalwalt:
of course i was refering to the search feature

zeffii:
https://github.com/nortikin/sverchok/pull/1521 discuss it further here

PR #1533, opened 2017-04-17
Subdivide node

portnov:
This is actually an interface for bmesh.ops.subdivide_edges.

Interesting, that for some reason bmesh's operator has different options set from blender editing mode one. Some features are only present in bmesh interface, for example "falloff type".

Please review.

### DIFF
--- a/docs/nodes/modifier_change/modifier_change_index.rst
+++ b/docs/nodes/modifier_change/modifier_change_index.rst
@@ -6,6 +6,7 @@ Modifier Change
    :maxdepth: 2
 
    bevel
+   subdivide
    delete_loose
    edges_intersect
    extrude_edges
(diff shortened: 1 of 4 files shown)


portnov:
box
smooth_fractal
smooth_one_face
smooth_torus
suzanne

portnov:
Bmesh's operator has a number of options and several types of output, so when I first placed them all to the node it was really huge :) So I added toggle buttons to hide some options and outputs by default.

kalwalt:
Subdivide node is a must i think, thanks for this!

portnov:
Node with all options shown:
node

portnov:
A bit late, but still, an easter cake:
cake
:)

zeffii:
brilliant!

zeffii:
haha.. the full blown node volcano 18 sockets total is a bit of an attack on the senses.   : )

PR #1551, opened 2017-04-20
Socket Expander - WIP

DolphinDream:
Start developing a socket expander infrastructure to allow multi-row input sockets to have an expand/collapse button to expand/collapse the property. 

Currently this is added to the VerticesSocket.

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -210,36 +210,36 @@ class VerticesSocket(NodeSocket, SvSocketCommon):
         else:
             return default
 
-    def draw_prop_with_expander(self, layout, obj, prop_name):
-        split = layout.split(percentage=.2, align=True)
-        c1 = split.column(align=True)
-        c2 = split.column(align=True)
-        if self.expanded:  # expand ?
-            c1.prop(self, "expanded", icon='TRIA_UP', text='')
-            c1.label(text=self.name[0])
-            c2.prop(obj, prop_name, text="", expand=True)
-        else:  # collapse
-            c1.prop(self, "expanded", icon='TRIA_DOWN', text="")
-            row = c2.row(align=True)
-            row.template_component_menu(obj, prop_name, name=self.name)
+    def draw_expander_template(self, context, layout, prop_origin, prop_name="prop"):
+
+        if self.use_expander:
+            split = layout.split(percentage=.2, align=True)
+            c1 = split.column(align=True)
+            c2 = split.column(align=True)
+            if self.expanded:
+                c1.prop(self, "expanded", icon='TRIA_UP', text='')
+                c1.label(text=self.name[0])
+                c2.prop(prop_origin, prop_name, text="", expand=True)
+            else: # collapsed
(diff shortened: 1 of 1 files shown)


zeffii:
https://github.com/nortikin/sverchok/pull/1440#issuecomment-295628064

I prefer you use the suggested function name and signature here.

DolphinDream:
yeah.. i had implemented the same thing independently. I wasn’t quite happy with my naming and I was going to revise that. I’ll use your suggested naming.

DolphinDream:
Updated the expander code. 

I’m thinking of coming up with a generalized expander template code that other socket could use. i.e. Colors would have a different kind of collapsed display than the vector for instance.

zeffii:
that's the logical conclusion.

DolphinDream:
Factorized the draw expander template code and added new Color and Quaternion sockets (WIP).

zeffii:
please prefix all new classes with `Sv` , ..`SvColorSocket`

zeffii:
(and correspondingly ..bl_idnames )

DolphinDream:
@zeffii sorry. i accidentally merged this. can you undo the merge ?

DolphinDream:
Can this pull be reopened  now ?

zeffii:
i think you need to make another pull request..

DolphinDream:
Hmm darn. Btw. isn't it better to reset hard the master branch back one commit ? instead of reverting. (i know the "don't rewrite public history" rule, but in this case it won't be a big deal.

zeffii:
you asked me to revert, i just pressed the button. and commited the revert. no big deal in this instance.

DolphinDream:
you're right. I asked to "undo" the merge, I just didn't expect that it will be done via github's revert (which does not do a reset back to previous commit, but simply adds another commit to revert the last changes). It's not a big deal, i just thought the history would look cleaner if reset was done instead. I could do it but I didn't know if such move is forbidden.

zeffii:
yes, resetting head would have been a neater way.  let me suggest using the _Squash and merge_ from the web based merging/commit feature in future. (for larger commits - easy to revert when needed )

DolphinDream:
I’m using this CodeHub iPhone app and I accidentally clicked on the “unmerged” button and automatically merged the branch without any warning/confirmation, which is kind of stoopeed. Any good mobile apps you guys use for GitHub ?

nortikin:
i use only reader for issues
all coding in notebook or pc on my job.

DolphinDream:
I’m taking another pass at this “socket expander” feature. @zeffii when you get a chance could you take a look and see what else needs to be done? I know that the draw method was supposed to be consolidated into the base class (partially done). Are there any other considerations that I should be aware of?

I removed the flags from the settings that would allow the use of preferences to decide whether the nodes socket expander would be used and whether they would be expanded by default when a node is created (I agree with you it was too much to add for a single PR). The fix behavior would be to create the nodes with the socket collapsed.

PR #1555, opened 2017-04-21
"Mask converter" node

portnov:
This node allows to convert masks that are to be applied to different types of mesh elements. For example, it can convert mask on vertices to mask for faces, or mask for edges to mask for vertices, and so on.


portnov:
from_vertices

PR #1561, opened 2017-04-22
Socket expander (take2) - WIP

DolphinDream:
Reopen the pull request (after accidental merge/close) (#1551)

So far  this includes: 
- SvColorSocket
- SvQuaternionSocket
- Add socket expander in base class
- add link in spawners for Matrix and Vertices sockets
- return default values in case of no links (instead of raising exception)
- organize socket colors in dictionary

### DIFF
--- a/core/socket_conversions.py
+++ b/core/socket_conversions.py
@@ -85,18 +85,16 @@ def get_quaternions_from_matrices(data):
     quaternions = []
     collect_quaternion = quaternions.append
 
-    def get_all(m, container):
-
-        for sublist in m:
+    def get_all(data):
+        for sublist in data:
             if is_matrix(sublist):
                 mat = Matrix(sublist)
                 q = tuple(mat.to_quaternion())
                 collect_quaternion(q)
             else:
-                new_sublist = []
-                get_all(sublist, new_sublist)
+                get_all(sublist)
 
-    get_all(data, quaternions)
+    get_all(data)
     return [quaternions]
 
 
--- a/core/socket_conversions.py
+++ b/core/socket_conversions.py
@@ -18,13 +18,16 @@
 
 from sverchok.data_structure import get_other_socket
 
+from mathutils import Matrix, Quaternion
+from sverchok.data_structure import Matrix_listing, Matrix_generate
 
-## conversion tests, to be used in sv_get! 
+
+## conversion tests, to be used in sv_get!
 
 def cross_test_socket(self, A, B):
     """ A is origin type, B is destination type """
     other = get_other_socket(self)
-    get_type = {'v': 'VerticesSocket', 'm': 'MatrixSocket'}
+    get_type = {'v': 'VerticesSocket', 'm': 'MatrixSocket', 'q': "SvQuaternionSocket"}
     return other.bl_idname == get_type[A] and self.bl_idname == get_type[B]
 
 def is_vector_to_matrix(self):
@@ -32,6 +35,12 @@ def is_vector_to_matrix(self):
 
 def is_matrix_to_vector(self):
     return cross_test_socket(self, 'm', 'v')
+
+def is_matrix_to_quaternion(self):
+    return cross_test_socket(self, 'm', 'q')
+
+def is_quaternion_to_matrix(self):
+    return cross_test_socket(self, 'q', 'm')
 
 # ---
 
@@ -50,6 +59,22 @@ def get_matrices_from_locs(data):
 
     get_all(data)
     return location_matrices
+
+
+def get_matrices_from_quaternions(data):
+    matrices = []
+    collect_matrix = matrices.append
+
+    def get_all(data):
+        for item in data:
+            if isinstance(item, (tuple, list)) and len(item) == 4 and isinstance(item[0], (float, int)):
+                mat = Quaternion(item).to_matrix().to_4x4()
+                collect_matrix(Matrix_listing([mat])[0])
+            else:
+                get_all(item)
+
+    get_all(data)
+    return matrices
 
 
 def is_matrix(mat):
--- a/core/socket_conversions.py
+++ b/core/socket_conversions.py
@@ -22,7 +22,7 @@ from mathutils import Matrix, Quaternion
 from sverchok.data_structure import Matrix_listing, Matrix_generate
 
 
-## conversion tests, to be used in sv_get!
+# conversion tests, to be used in sv_get!
 
 def cross_test_socket(self, A, B):
     """ A is origin type, B is destination type """
@@ -33,16 +33,20 @@ def cross_test_socket(self, A, B):
 def is_vector_to_matrix(self):
     return cross_test_socket(self, 'v', 'm')
 
+
 def is_matrix_to_vector(self):
     return cross_test_socket(self, 'm', 'v')
 
+
 def is_matrix_to_quaternion(self):
     return cross_test_socket(self, 'm', 'q')
+
 
 def is_quaternion_to_matrix(self):
     return cross_test_socket(self, 'q', 'm')
 
 # ---
+
 
 def get_matrices_from_locs(data):
     location_matrices = []
@@ -75,6 +79,25 @@ def get_matrices_from_quaternions(data):
 
     get_all(data)
     return matrices
+
+
+def get_quaternions_from_matrices(data):
+    quaternions = []
+    collect_quaternion = quaternions.append
+
+    def get_all(m, container):
+
+        for sublist in m:
+            if is_matrix(sublist):
+                mat = Matrix(sublist)
+                q = tuple(mat.to_quaternion())
+                collect_quaternion(q)
+            else:
+                new_sublist = []
+                get_all(sublist, new_sublist)
+
+    get_all(data, quaternions)
+    return [quaternions]
 
 
 def is_matrix(mat):
--- a/node_tree.py
+++ b/node_tree.py
@@ -109,7 +109,7 @@ class SvSocketCommon:
             else:  # collapsed
                 c1.prop(self, "expanded", icon='TRIA_DOWN', text="")
                 row = c2.row(align=True)
-                if self.bl_idname == "ColorSocket":
+                if self.bl_idname == "SvColorSocket":
                     row.prop(prop_origin, prop_name)
                 else:
                     row.template_component_menu(prop_origin, prop_name, name=self.name)
--- a/node_tree.py
+++ b/node_tree.py
@@ -82,31 +82,9 @@ emptyQuaternion = [[(1, 0, 0, 0)]]
 def process_from_socket(self, context):
     """Update function of exposed properties in Sockets"""
     self.node.process_node(context)
-
-def get_use_expander():
-    print("Use expander")
-    with sv_preferences() as prefs:
-        if prefs:
-            print("I have prefs")
-            return prefs.use_socket_expander
-        else:
-            print("i dont have prefs")
-            return True
-    return True
-
-def get_use_quicklink():
-    print("Use quicklink")
-    with sv_preferences() as prefs:
-        return prefs.use_socket_quicklink
-    return True
-
-def get_expand_socket():
-    print("Expand socket")
-    with sv_preferences() as prefs:
-        return prefs.expand_input_sockets
-    return True
-
 # this property group is only used by the old viewer draw
+
+
 class SvColors(bpy.types.PropertyGroup):
     """ Class for colors CollectionProperty """
     color = FloatVectorProperty(
@@ -117,20 +95,36 @@ class SvColors(bpy.types.PropertyGroup):
 
 
 class SvSocketCommon:
+    use_prop = BoolProperty(default=False)
 
     # use_expander = BoolProperty(default=get_use_expander())
     use_expander = BoolProperty(default=False)
     use_quicklink = BoolProperty(default=False)
     expanded = BoolProperty(default=False)
 
-    def init(self):
-        print("initializing socket")
-        self.use_expander = get_use_expander()
-        self.use_quicklink = get_use_quicklink()
-        self.expanded = get_expand_socket()
+    @property
+    def use_socket_expander(self):
+        with sv_preferences() as prefs:
+            # print("use expander called")
+            return prefs.use_socket_expander
+        # return sv_preferences.use_socket_expander
+
+    @property
+    def use_socket_quicklink(self):
+        with sv_preferences() as prefs:
+            # print("use expander called")
+            return prefs.use_socket_quicklink
+        # return sv_preferences.use_socket_expander
+
+    @property
+    def use_expand_sockets(self):
+        with sv_preferences() as prefs:
+            # print("use expander called")
+            return prefs.expand_input_sockets
 
     def draw_expander_template(self, context, layout, prop_origin, prop_name="prop"):
-        if self.use_expander:
+        if self.use_socket_expander:
+            # if self.use_expander:
             split = layout.split(percentage=.2, align=True)
             c1 = split.column(align=True)
             c2 = split.column(align=True)
@@ -148,15 +142,46 @@ class SvSocketCommon:
         else:
             layout.template_component_menu(prop_origin, prop_name, name=self.name)
 
-    def draw_link_new_node(self, settings, idname, x_offset, y_offset):
-        if self.use_quicklink:
-            context, layout, node, index = settings
+    def draw_quick_link(self, context, layout, node):
+        if self.use_socket_quicklink:
+            if self.bl_idname == "MatrixSocket":
+                new_node_idname = "SvMatrixGenNodeMK2"
+            elif self.bl_idname == "VerticesSocket":
+                new_node_idname = "GenVectorsNode"
+            else:
+                return
+
             op = layout.operator('node.sv_quicklink_new_node_input', text="", icon="PLUGIN")
-            op.socket_index = index
+            op.socket_index = self.index
             op.origin = node.name
-            op.new_node_idname = idname
-            op.new_node_offsetx = x_offset
-            op.new_node_offsety = y_offset
+            op.new_node_idname = new_node_idname
+            op.new_node_offsetx = -200
+            op.new_node_offsety = 0
+
+    @property
+    def extra_info(self):
+        print("getting base extra info")
+        return ""
+
+    def draw(self, context, layout, node, text):
+        if self.is_linked:  # linked INPUT or OUTPUT
+            info_text = text + '. ' + SvGetSocketInfo(self)
+            info_text += self.extra_info
+            layout.label(info_text)
+
+        elif self.is_output:  # unlinked OUPUT
+            layout.label(text)
+
+        else:  # unlinked INPUT
+            if self.prop_name:  # has property
+                self.draw_expander_template(context, layout, prop_origin=node, prop_name=self.prop_name)
+
+            elif self.use_prop:  # no property but use default prop
+                self.draw_expander_template(context, layout, prop_origin=self)
+
+            else:  # no property and not use default prop
+                self.draw_quick_link(context, layout, node)
+                layout.label(text)
 
     @property
     def other(self):
@@ -210,6 +235,15 @@ class MatrixSocket(NodeSocket, SvSocketCommon):
     prop_name = StringProperty(default='')
     num_matrices = IntProperty(default=0)
 
+    @property
+    def extra_info(self):
+        print("getting matrix extra info")
+        info = ""
+        if is_vector_to_matrix(self):
+            info = (" (" + str(self.num_matrices) + ")")
+
+        return info
+
     def get_prop_data(self):
         return {}
 
@@ -235,7 +269,7 @@ class MatrixSocket(NodeSocket, SvSocketCommon):
         else:
             return default
 
-    def draw(self, context, layout, node, text):
+    def draw2(self, context, layout, node, text):
         if self.is_linked:
             draw_string = text + '. ' + SvGetSocketInfo(self)
             if is_vector_to_matrix(self):
@@ -290,7 +324,7 @@ class VerticesSocket(NodeSocket, SvSocketCommon):
         else:
             return default
 
-    def draw(self, context, layout, node, text):
+    def draw2(self, context, layout, node, text):
         if not self.is_output and not self.is_linked:
 
             if self.prop_name:
@@ -354,7 +388,7 @@ class SvQuaternionSocket(NodeSocket, SvSocketCommon):
         else:
             return default
 
-    def draw(self, context, layout, node, text):
+    def draw2(self, context, layout, node, text):
         if not self.is_output and not self.is_linked:
 
             if self.prop_name:
@@ -405,7 +439,7 @@ class SvColorSocket(NodeSocket, SvSocketCommon):
         else:
             return default
 
-    def draw(self, context, layout, node, text):
+    def draw2(self, context, layout, node, text):
         if not self.is_output and not self.is_linked:
 
             if self.prop_name:
@@ -444,7 +478,7 @@ class SvDummySocket(NodeSocket, SvSocketCommon):
     def sv_type_conversion(self, new_self):
         self = new_self
 
-    def draw(self, context, layout, node, text):
+    def draw2(self, context, layout, node, text):
         layout.label(text)
 
     def draw_color(self, context, node):
@@ -490,7 +524,7 @@ class StringsSocket(NodeSocket, SvSocketCommon):
         else:
             raise SvNoDataError
 
-    def draw(self, context, layout, node, text):
+    def draw2(self, context, layout, node, text):
 
         # just handle custom draw..be it input or output.
         if hasattr(self, 'custom_draw'):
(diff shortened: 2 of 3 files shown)


DolphinDream:
sv-expandsockets-demo2

DolphinDream:
I’m not sure if this is a socket related bug or a MixInputs related bug but when I reload a blend file with MixInput nodes (and the new Sockets with socket conversion) the defaults for the node seem to be messed up. Any clues why this happens?

mixinputreloadbug

zeffii:
if it's red, it should be printing some exception to the console.

zeffii:
seems to me.. without looking at the code..  that you are enumerating the Int Enum using a dict.. or some hashed collection ( therefor it wont be iterated in the same sequence each time )

```
for idx, (k, v) in enumerate(some_dict.items()):
    print(idx, k v)
``` 
if you have say a dict of 10 key/value pairs.. and keep executing that.. changes are you'll get a different output most of the time.  Dicts are not stable by default... not in Python 3.5 anyway.  in 3.6 that does seem to be implemented.

zeffii:
(Edited) 

seems to me.. without looking at the code..  that you are enumerating the Mode Enum using a dict.. or some hashed collection ( therefor it wont be iterated in the same sequence each time )

```
for idx, (k, v) in enumerate(some_dict.items()):
    print(idx, k v)
``` 
if you have say a dict of 10 key/value pairs.. and keep executing that.. chances are you'll get a different output most of the time.  Dicts are not stable by default... not in Python 3.5 anyway.  in 3.6 that does seem to be implemented.

zeffii:
because of the aforementioned instability in dict iteration, the enum 'selected' will only match the correct value.. if the iteration is the same.. hence we do stuff like in the Vector and Scalar  math nodes to populate the selected mode enums.

zeffii:
we enumerate selected modes by hardcoding the index, that way you can easily add new items in the enum list, and not break existing blends because the enumeration indices differ. Just keep the enum simple.

```python

from collections import OrderedDict

type_items_dict = OrderedDict(INT=0, FLOAT=5, VECTOR=10, COLOR=15, QUATERNION=20, MATRIX=25)
type_items = [(k, k.title(), '', v) for k, v in sorted(type_items_dict.items())]

print(type_items)
```

zeffii:
actually, go with this and you'll be able to control the menu items order (regardless of the index)
```python
type_items_list = [('INT', 0), ('FLOAT', 5), ('VECTOR', 10), ('COLOR', 15), ('QUATERNION', 20), ('MATRIX',25)]
type_items = [(k, k.title(), '', v) for k, v in type_items_list]

```

DolphinDream:
thx. So, this may actually not work then? I thought if i sorted the list it would be the same every time:
```python
typeItems = [(n, n.title(), "", "", i) for i, n in enumerate(input_info.keys())]

    mode = EnumProperty(
        name="Mode",
        default="FLOAT", items=sorted(typeItems),
        update=update_mode)
```

DolphinDream:
This seems to have fixed the problem (putting the sorting before creating the list of tuples:

```python
typeItems = [(n, n.title(), "", "", i) for i, n in enumerate(sorted(input_info.keys()))]

    mode = EnumProperty(
        name="Mode",
        default="FLOAT", items=typeItems,
        update=update_mode)
```

the only shortcoming is the list is ordered alphabetically rather than the dictionary definition.
I assumed that the int, float, vector at the beginning of the list would be preferable, but maybe nobody cares. :)

zeffii:
@DolphinDream   you need to hardcode the last element of the enum items, else you will not be able to add new items to the enum options without breaking all .blends that use that node.

zeffii:
to reiterate
```python
[(n, n.title(), "", "", hard_code_me) ........ ]
```

zeffii:
" This seems to have fixed the problem "   for now.. yes, but as soon as you add an item, (which may or may not happen...) it will break... and you will regret it. and you'll have to go MK2 instead.

DolphinDream:
Btw.. working with the socket conversion I think I stumbled over a matrix utils lerp bug :) (https://developer.blender.org/T51287)

DolphinDream:
Alrighty.. I did it this way:
I added an extra “Index” field to the dictionary to hardcode their order:

```python
input_info = {
    'INT': {
        “Index”: 0, ...
    'FLOAT': {
        “Index”: 1, ...
}
    # then build the sorted tuples
    typeDict = {input_info[n]["Index"]: n for n in input_info.keys()}
    typeItems = [(v, v.title(), "", "", k) for (k,v) in typeDict.items()]
```
I wanted to keep the input_info as a dictionary for accessing its values the same way as I already implemented.

DolphinDream:
Btw I don’t think the hardcoded index at the end of the property list items sorts the list. i think the order of items in the list itself establishes the order in the drop down.

i.e
```python
typeItems =[ (“A”, “a”, “”, “”, 3), (“B”, “b”, “”, “”, 2), (“C”, “c”, “”, “”, 1) ]
```

drowdown-indexorder

zeffii:
> Btw I don’t think the hardcoded index at the end of the property list items sorts the list.

i didn't say that.

zeffii:
that last value is to give a consistent/permanent index to an element. so if you add new items , then indexing is invariant. except of course that you must give the new element a unique index.. and you should keep existing item indices unchanged.

DolphinDream:
@zeffii So if you want the items to appear in a certain order it’s important to have both the hardcoded values and the items listed in the desired order then ? 

Btw what does “stored in file data” mean?

Do you think the latest implementation will work fine with blend file reloads and future insertion of new types ?

zeffii:
> @zeffii So if you want the items to appear in a certain order it’s important to have both the hardcoded values and the items listed in the desired order then ?

don't ask me . test it.

node_tree.py
```diff
@@ -307,6 +478,46 @@ def draw_color(self, context, node):
         return self.nodule_color
 
 
+class SvLinkMatrix(bpy.types.Operator):
```
zeffii: seriously. no. don't do this. 

make a SvLinkNewNode, and make it accept a bl_idname and an offset. so you can reuse the same operator for any number of new "New Node"s .  Here the _action_ of adding a new node is more important than the node kind..

zeffii: `bl_idname = 'node.sv_quicklink_new_node_input'`

zeffii: most nodes are the same width, so a X offset could have a default of 200, which it uses unless the operator is explicitely passed a different X offset.

DolphinDream: Super. I factorized the spawn/link new node code.

DolphinDream: It would be cool if the spawned nodes can be offset in a smart way to not overlap with each other. Perhaps add slightly different offset depending on the index of the input socket ?

zeffii: well.. i think that's a separate issue (spawning non-overlap) 

don't get stuck on small usability details in a large PullRequest.

zeffii: if.. over time we get annoyed at it.. sure.. that's going to be solved.

DolphinDream:
How do i find out the parent node of a socket? I think I asked this again and we came up with the “socket_index” solution .. however, if i want to spawn a node so that it doesn’t overlap other connected nodes I would have to be able to inspect the nodes connected to the caller node. I can traverse the inputs of the caller nodes, and get the “other” end of the socket in the connected nodes, but how do I get a hold of these nodes at the end of the sockets link?

node_tree.py
DolphinDream: Fair enough. We’ll deal with overlap later. Btw. the “small usability details” help me understand more about blender/sverchok node coding... i don’t just do it for the detail sake.

zeffii:
Grep.  .from_node  .from_socket

zeffii:
One could fill a kdtree with nodelocations and check if any are within radius. And only then nudge to avoid overlap.

core/socket_conversions.py
```diff
@@ -52,6 +65,41 @@ def get_all(data):
...
+
+    get_all(data)
+    return matrices
+
+
+def get_quaternions_from_matrices(data):
+    quaternions = []
+    collect_quaternion = quaternions.append
+
+    def get_all(m, container):
```
zeffii: container is unused?

DolphinDream:
Is there a socket method that is called once during instantiation ? I want to set the use expander/quicklink flags based on setting values and I want to initialize them in the socket during instantiation but I don’t know how to do that. Any suggestions ?

If there would be a similar method like the sv_init() that would let socket do some one time initialization that would probably work.

DolphinDream:
Is there a socket method that is called once during instantiation ? I want to set the use expander/quicklink flags based on preference setting values and I want to initialize them in the socket during instantiation but I don’t know how to do that. Any suggestions ?

If there would be a similar method like the sv_init() that would let socket do some one time initialization that would probably work.

zeffii:
The default should be sensible for the vast majority, ill make some macros that show how to apply props after creation.  Resist doing too much in a single pr.

zeffii:
>  I want to set the use expander/quicklink flags based on preference setting values and I want to initialize them in the socket during instantiation 

because these properties only affect how the layout is drawn for the socket, it is literally not gaining you any speed or convenience to try to set them up with some kind of config based on preferences. sockets and ui is redrawn a lot .. like ..sometimes 30 times a second.  

So applying the socket properties after node creation is both easier and less intrusive than trying to bolt on stuff onto the socket class that it does not need.

zeffii:
now..  while investing the node.add_node / search feature I stumbled into an unused part of the nodeitems_utils  module, and it is a `settings apply` routine that gets perform directly after a node is added. I think that's where we would add such settings/reading from preferences.

all in all, a task for a separate PR.

zeffii:
the shift+A menu can use our own implementation of 'bpy.ops.node.add_node'

DolphinDream:
I thought it would be better to have the settings off by default and let the user decide if they want to turn them on in the preferences.

zeffii:
then you add another property to the sockets, which can default to an sv_preferences setting (which might be off by default).  There's no `init` to speak of..

```python
class SvSocketCommon

    @property
    def use_expanded_sockets(self):
         return sv_preferences.use_expanded_sockets

    # this should never be called in a loop
    @use_expanded_sockets.setter
    def use_expanded_sockets(self, value):
        sv_preferences.use_expanded_sockets = value
```


I was under the impression that the expander code would be standard for all Vector/Matrix/Color/Quat sockets, I understand why one might want to suppress it for certain nodes... but you could do that in the sv_init of a node, or write a small function that takes a node and inspects its input sockets and switches off

zeffii:
i adjusted that code snippet. refresh if you havent

DolphinDream:
By the way what's your take on this PR so far? Should it be more refined? Are you're OK with integrating it in the current form? I know Still have to clean up the quaternion conversion but other than that I don't know what to do about it.

zeffii:
you'll have to give us some time to meditate about it, it will affect all node trees.

i'm OK with it, except for the massive code duplicating in the draw functions of VerticesSocket, SvQuaternion, and SvColorsSocket

```python

VerticesSocket

    def draw(self, context, layout, node, text):

            if self.prop_name:
                self.draw_expander_template(context, layout, prop_origin=node, prop_name=self.prop_name)
            elif self.use_prop:
                self.draw_expander_template(context, layout, prop_origin=self)
            else:
                if not self.is_output:
                    settings = [context, layout, node, self.index]
                    self.draw_link_new_node(settings, "GenVectorsNode", -200, 0)

                layout.label(text)

        elif self.is_linked:
            layout.label(text + '. ' + SvGetSocketInfo(self))

        else:
            layout.label(text)


SvQuaternionSocket

    def draw(self, context, layout, node, text):

            if self.prop_name:
                self.draw_expander_template(context, layout, prop_origin=node, prop_name=self.prop_name)
            elif self.use_prop:
                self.draw_expander_template(context, layout, prop_origin=self)
            else:
                layout.label(text)

        elif self.is_linked:
            layout.label(text + '. ' + SvGetSocketInfo(self))

        else:
            layout.label(text)

SvColorSocket


    def draw(self, context, layout, node, text):

            if self.prop_name:
                self.draw_expander_template(context, layout, prop_origin=node, prop_name=self.prop_name)
            elif self.use_prop:
                self.draw_expander_template(context, layout, prop_origin=self)
            else:
                layout.label(text)

        elif self.is_linked:
            layout.label(text + '. ' + SvGetSocketInfo(self))

        else:
            layout.label(text)
```

those could be all replaced by this .. i think  (untested)
```python

    def draw(self, context, layout, node, text):

            if self.prop_name:
                self.draw_expander_template(context, layout, prop_origin=node, prop_name=self.prop_name)
            elif self.use_prop:
                self.draw_expander_template(context, layout, prop_origin=self)
            else:
                if self.bl_idname == "MatrixSocket" and not self.is_output:
                    settings = [context, layout, node, self.index]
                    self.draw_link_new_node(settings, "GenVectorsNode", -200, 0)

                layout.label(text)

        elif self.is_linked:
            layout.label(text + '. ' + SvGetSocketInfo(self))

        else:
            layout.label(text)



```
and drop the explicit draw function VerticesSocket, SvQuaternion, and SvColorsSocket

zeffii:
the explicit draw function would ofcourse go in the SvSocketCommon mixin

zeffii:
Sharing that function in SvSocketCommon seems sane to me.

DolphinDream:
I guess this is what I was after (with the pref settings idea):

1) First, I wanted to keep the socket behavior/features the same for now not to affect the user and let them turn it ON in the settings if they want them. Later down the road when this feature matures enough we can make the default in the settings to be ON, but still keep it as a settings option so that the user can chose whatever behavior they like most. Your code above does help with that.

2) Once the expander and the quicklink buttons are shown/used (based on settings), I also wanted to have the “expand” flag be initilaized based on the settings so that the **new** nodes are either created with the inputs expanded or collapsed. The user can toggle this flag at will to expand or collapse each input, so its value should be per socket based (not per settings based) which is why I was trying to figure out how to make a property in the socket class be initialized once based on the settings.

Basically the settings would have:
[1] Use Socket Expander
[2] Use Socket QuickLink
[3] Expand Socket on Creation

zeffii:
1. OK.

2. There's no reason to think in terms of extra initialization as part of the node.. or the socket. 

##### what I don't want to see, not for the PR.. not now. is this:
a `socket.new_with_settings()` function that would wrap socket.new() .. just like we wrap  self.hide with self.hide_safe ).  Why? because then you have to edit all the nodes... The settings / preferences should be separated entirely from the node and from the socket.  The settings should be applied just after the node is created .. that should happen so fast you won't even see the ui being modified.

##### what i'm more amicable towards is a custom add_node operator that can read a settings dict 

```python
     node = nodes.new(bl_idname)
     apply_settings(node, bl_idname)
```

DolphinDream:
So.. picking up where we left off. What’s the preferred way to have the socket be initialized (similar to sv_init) once per creation? 

SocketCommon class could have something like:

def sv_socket_init:
       self.expanded = prefs.socket_expanded

So that the initial “expanded” value of the socket is whatever is set in the preferences. E.g. if the prefs have the expanded OFF all new nodes are created with the sockets collapsed (the user can expand them after that if they want to).

I assume such init method could be called once after creation of the socket (perhaps with the method that creates the socket).

DolphinDream:
I tried to do something like overloading the __init__ to do this kind of one time initialization (based on prefs values) but that seemed to be a terrible idea because of blender register shenanigans I get the AttributeError: Writing to ID classes in this context is not allowed:  type of error.

zeffii:
Socket inits are not needed. Ive said this like 4 times already

zeffii:
what would make sense is to have two methods

used for defaults (in sv_init), custom defaults (defaults.json), and iojson packing/unpacking
```
socket.get_state()          
socket.set_state(........)
```

zeffii:
- just make the default (without having to use `socket.set_state` ) as close to the current UI design as possible. 
- tweaking which node has different socket defaults is another pull request (those would be the sverchok defaults)
- custom defaults will override all of that, and are another pull request.

zeffii:
the more you try to do in one PR the more i will resist, out of caution.  like a Non-Newtonian fluid.

DolphinDream:
I’m taking another pass at this “socket expander” feature. @zeffii when you get a chance could you take a look and see what else needs to be done?

Summary of things done so far:
- draw calls were consolidated into base class
- draw label calls were consolidated in the base class
- socket color array was added for all socket types based on their bl_idname
- add expander option (button) to input sockets
- add quick link option (button) to input sockets
- draw expander template method was added to aid the draw methods draw the expanded/collapsed props
- draw quick link method was added to aid the draw methods draw the quick link to spawn new nodes
- add Quaternion and Color socket classes
- add quaternion/matrix conversion methods to socket_conversion

Are there any other considerations that I should be aware of?

Btw, I removed the flags from the settings that I originally put in this PR which were to let the user enable the expander/quicklink functionality via the preference settings. I agree with you it was too much to add for a single PR. For now the default behavior is let the nodes be created with the socket collapsed.

zeffii:
Too much time has passed by since then and now, it will take some time to asses, appreciate, confirm the level of impact. if it's all working nicely, it will be a cool addition and nice cleanup. But we'd need more certainty. 

I'm not in a position to spend much time on this at the moment, but will have a look shortly.

zeffii:
This PR will need to confirmed working with the GIST IO system before I'd be happy with it. Properties stored inside the sockets class instances are not yet stored/restored in the json. I'll be happy to help you navigate that swamp.

A lot of people use the gist IO and might not be aware that it will fail for nodes that rely on values stored in socket properties (as is the case for `.use_prop` )

zeffii:
### storing in json

Seen as this just happens in the input sockets..

```python

input_socket_storage = { }
for socket in node.inputs:
    if socket.use_prop:
        # what kind of prop...at the moment it's just named .prop
        input_socket_storage[socket.index] = serialize_prop(socket.prop)

```

serielize_prop could store a dict with any number of properties..

### import from json

in the part that sets the properties

```python

for socket_index, value in input_socket_storage.items():
    socket = node.inputs[socket_index]
    socket.use_prop = True
    apply_props_to_socket(socket, value)

```


I think things like `use_expander, use_quicklink, expanded` will need to be stored only if the current value is not the default value.

zeffii:
storing to json could look something like this ( " sparse storage ")

```python

input_socket_storage = { }
for socket in node.inputs:

    storable = {}
    tracked_props = 'use_expander', 'use_quicklink', 'expanded', 'use_prop'

    for tracked_prop in tracked_props:
        if hasattr(socket, tracked_prop) and getattr(socket, tracked_prop):

            storable[tracked_prop] = True

            if tracked_prop == 'use_prop':
                storable['socket_prop_value'] = serialize_prop(socket.prop)

    if storable:
        input_socket_storage[socket.index] = storable
      

```

zeffii:
The current commented out code ( we never settled this...)

https://github.com/nortikin/sverchok/blob/1ef52eec82ff5907af3312866cc67006f1293b58/utils/sv_IO_panel_tools.py#L258-L266

zeffii:
would result in something like

```python
        input_socket_storage = { }
        for socket in node.inputs:

            # if not tracked_socket(socket): continue

            storable = {}
            tracked_props = 'use_expander', 'use_quicklink', 'expanded', 'use_prop'

            for tracked_prop in tracked_props:
                if hasattr(socket, tracked_prop) and getattr(socket, tracked_prop):

                    storable[tracked_prop] = True

                    if tracked_prop == 'use_prop':
                        storable['socket_prop_value'] = serialize_prop(socket.prop)

            if storable:
                input_socket_storage[socket.index] = storable

        if input_socket_storage:
            node_dict['custom_socket_props'] = input_socket_storage

```

zeffii:
maybe the import schema would happen in the ` apply_post_processing()` function, towards the end, that's one of the last things done prior to connecting the nodes.

zeffii:
what about if I want a node to `use_quicklink` and `use_expander`,

i'm tempted to add a rightclick menu for nodeview, with a smart menu populated by useful nodes for the socket being rightclicked. But that will be a different topic.

DolphinDream:
About the GIST IO.. I guess you’re referring to the export/import of node trees via gist, correct? If so, I think the import/export still works except that you’ll lose the status of the new flags (the expanded flag really, since the other ones are set to default value True for now). In the current PR implementation when you export a node tree (via gist) with some node sockets expanded, upon re-import those nodes (via gist) will have their sockets collapsed as the nodes/sockets are recreated based on the **default** expanded value (which is currently set to False / collapsed in the base class). Having these flags saved as part of the gist would be a nice feature indeed, but it’s not necessary to preserve the nodes (other than their socket expanded/collapsed state). Is there something else I’m missing?

As far as the use_prop goes .. this was already defined in the specific socket classes. I just added one to the base class, thinking that it will eventually get removed from the specific classes. The only new flags added are : use_expander, use_quicklink and expanded (all having default/fix values in the base class for now). Btw, if you want certain nodes/sockets to not use expander/quicklink feature you should be able to set that within that node when creating the sockets, right ?

zeffii:
> Is there something else I’m missing?

We have this thing for a while, it's not something you introduced - but we never made much use of it.

For example in `SvQuaternionSocket`  you have a prop called `prop`, the value of which is not stored in the gist. So it will be fine if the user doesn't modify the initial value, but any changes from the default will be unstored.

- if the socket uses 'use_prop', then the property (when unconnected) comes from a property stored in the socket instance, a property called 'prop'.
- if the socket uses 'prop_name', then the value (when unconnected) comes from a property defined in the node class, obtainable via `getattr(node, socket.prop_name)`.

zeffii:
> Btw, if you want certain nodes/sockets to not use expander/quicklink feature you should be able to set that within that node when creating the sockets, right ?

yes, or using the node presets system i made a while back.

DolphinDream:
Btw, after factorizing the draw methods, the StringsSocket (also factorized) looses the “custom draw” feature, which I think you introduced mainly to handle the missing “color” socket type/class. Now that there is a dedicated ColorSocket class I don’t see the need for the custom draw feature in the StringsSocket class, unless you had other things in mind.

DolphinDream:
@zeffii can you give me an node tree example where the props are not handled properly via export/import to/from gist ? I tried a few cases with some nodes and alter the prop values in the node but they seem to export import correctly.

importexport-gist

DolphinDream:
Actually I think I got the “Vector Attraction” property “Center” to fail the import/export test, though this is not related to changes made in this PR. I used the **master** branch and the import/export of this node’s prop still fails.

zeffii:
nodes that use `some_socket.use_prop = True`, and then rely on the socket to store the parameters value have never been supported by IO/Gist. If you do a search on the repo for `.use_prop` any node that employs it will not work correctly if that socket is unconnected and the standard value is modified. They are ofcourse valid node trees, but data is lost when exporting to json with GIST/IO.

So if you make your new sockets use this same system, and I agree you probably should, it's right to fix the IO code for these new sockets at the same time.

zeffii:
also the value of Vector (unconnected) in the Matrix In node, is stored on the node , because it uses socket.prop_name = 'l_'   (some FloatVectorProperty).

zeffii:
Vector attractor uses `use_prop`

https://github.com/nortikin/sverchok/blob/1ef52eec82ff5907af3312866cc67006f1293b58/nodes/vector/attractor.py#L97-L101

works fine in regular node tree / .blends,  but information is lost during gist IO

DolphinDream:
So, to clarify, the issue with the sockets that rely on use_prop not being imported/exported correctly via gist is **not** an issue introduced by this PR, but existed even before. And what you are suggesting is that we should fix this within this PR ? If I get this correctly, your earlier comment 

> This PR will need to confirmed working with the GIST IO system before I’d be happy with it.

was not about this PR introducing some issues with gist io, but that it would be nice if the gist issue (related to the use_prop, which this PR touches) would also be fixed. Yep/Nope/Kinda ?

zeffii:
the PR does implement new flags that will be ignored by IO/JSON, even if you ignore use of `.use_prop` this will still produce slightly awkward json representations of the node tree / certain nodes - unless adressed in the IO code.

A sane solution might be to branch master , call it `development_io`, then merge socketExpander into it, and we'll fix the IO code in `development_io`, then when it works we merge `development_io` into `master`.

DolphinDream:
Btw, i noticed the gist export/import doesn’t preserve the custom node/frame colors (the ones you can select via property panel). is this an oversight or by design ?

DolphinDream:
@Zeffii Can you please elaborate on the serialize_prop? Can this just be = socket.prop[:] ?

zeffii:
`socket.prop[:]` will work for most props, except `Int/Float/BoolProperty` , because they aren't subscriptable..
```python
>>> nodes['A Number'].int_max[:]    # an IntProperty()
TypeError: 'int' object is not subscriptable

>>> nodes['Vector Math'].v3_input_0  # a FloatVectorProperty()
bpy.data...nodes["Vector Math"].v3_input_0

>>> nodes['Vector Math'].v3_input_0[:]
(0.0, 0.0, 0.0)

```
because the `props` used in sockets are not in that Set, `socket.prop[:]` will suffice for now.

DolphinDream:
Close this PR since development moved to #1847 , and was just merged into master.

PR #1569, opened 2017-04-23
Gtext node ++

zeffii:
this forces users to update to a very recent blender (not official 2.78c ) if they want to use GP's new features.


zeffii:
am tempted to add the original GText font (font 0) as optional... it's just caps tho..

image

PR #1575, opened 2017-04-24
Csg boolean mk2

Kosvor2:
Now accepts a nested lists of meshes. In addition, there is a new mode in which the nested objects are connected to each other one by one.

csg_booleanmk2_1


kalwalt:
Did you see this my post in blender-stackexchange? :smile: 
https://blender.stackexchange.com/questions/78104/what-is-the-best-approach-to-do-boolean-operations-with-multi-meshes-in-sverchok

Kosvor2:
СSG creates unconnected islands of geometry which become noticeable if you subdivide the resulting mesh and smooth the vertices after that. I just want to say that this is not caused by changes in this pull request, but this problem was originally there.

Kosvor2:
Added option to output only result of last iteration instead of each one like on first pic.

kalwalt:
It works perfectly! Thank a lot for this!

enzyme69:
@Kosvor2 This matrix thing does CSG Boolean to first Object?

For example:
- Start with Monkey
- Next sequences = bunch of boxes

Monkey + Cube + Cube + Cube?
Or Monkey - Cube - Cube - Cube?

Kosvor2:
@enzyme69
When nested mode is used, it does apply next object to result of applying of previous object to previous result and so on. In normal mode, it applies object "A" to object "B" until they all ends (A shorter list of objects it extends to be of the same size as longer list).

zeffii:
looks like this now
image

PR #1579, opened 2017-04-25
add small bw comp flow control

zeffii:


zeffii:
adds backward compatibility for old text in json storage,

Issue #1526, opened 2017-04-16
templates and deprecated nodes


kalwalt:
Recently i looked at some sverchok templates and i found that some has some deprecated nodes, so i start to look to other and i think many of them has some deprecated node. Example line node instead line mk2. Should we fix them? maybe we can change them with the newest version?

zeffii:
good point. definitely doesn't make sense to have deprecated nodes in templates. 

If you're lucky all you have to do is change the bl_idname in the .json. That'll take someone an hour or more to go through..

kalwalt:
i think i can do, so i will try as you suggested. I think open all the templates will be a crazy thing...

zeffii:
i suggest replacing the node in situ, and uploading the gist here including the name of the template that it replaces.

or

make a branch, and save/overwrite gists as you get through them ;) then do a pullrequest :)

kalwalt:
i think better to make a new branch and overwrite the json file in the branch. maybe some simpler json can be changed in the text editor but remain the fact you have to know which node is deprecated or not in the json.

kalwalt:
I would to fix some small things more, (some broken connections) but when i try to export (for saving the json) i get this error:
```python
failed total 3
[['Text Input.001', 'Objects input.002:Matrixes', 'Matrix out.001', 'Matrix'], ['Text Input', 'Objects input:Vertices', 'Bisect', 'vertices'], ['Text Input', 'Objects input:Polygons', 'Bisect', 'edg_pol']]
Processed  Matrix out.001 in: 0.0000
cache miss: Vectors out -> Vectors from: Matrix out.001 -> Location
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/core/update_system.py", line 322, in do_update_general
    node.process()
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/nodes/vector/vector_out.py", line 40, in process
    xyz = self.inputs['Vectors'].sv_get(deepcopy=False)
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/node_tree.py", line 185, in sv_get
    return SvGetSocket(self, deepcopy)
  File "/home/walter/blender-2.78a-linux-glibc211-x86_64/2.78/scripts/addons/sverchok-master/core/socket_data.py", line 106, in SvGetSocket
    raise SvNoDataError
[...]

PR #1580, opened 2017-04-25
Template upgrade 02

kalwalt:
There are some small fixes in the layout of some nodes. Though there is a little issue:
https://github.com/nortikin/sverchok/issues/1526#issuecomment-297046952
wait before merging.


kalwalt:
Regarding the issue mentioned, in some cases (look at waffel minmal template) it save the nodes with a very long distance in between.... at least this happens to me.

Issue #1586, opened 2017-04-26
Custom PNG saving


kalwalt:
Discussion from https://github.com/nortikin/sverchok/issues/1566
Before to ask a pull request, i'm testing the code that @zeffii pointed out in this link http://stackoverflow.com/a/19174800/1243487
original code from @ideasman42 and @Evgeni_Sergeev

**PNG_test.py**

```python
'''
code taken from:
http://stackoverflow.com/questions/902761/saving-a-numpy-array-as-an-image/19174800#19174800
thanks to @ideasman42 and @Evgeni Sergeev
'''
"""
in buf s d=[[]] n=1
out some s
"""

def write_png(buf, width, height):
    """ buf: must be bytes or a bytearray in Python3.x,
        a regular string in Python2.x.
    """
    import zlib, struct

    # reverse the vertical line order and add null bytes at the start
    width_byte_4 = width * 4
    raw_data = b''.join(b'\x00' + buf[span:span + width_byte_4]
                        for span in range((height - 1) * width_byte_4, -1, - width_byte_4))

    def png_pack(png_tag, data):
        chunk_head = png_tag + data
        return (struct.pack("!I", len(data)) +
                chunk_head +
                struct.pack("!I", 0xFFFFFFFF & zlib.crc32(chunk_head)))

    return b''.join([
        b'\x89PNG\r\n\x1a\n',
        png_pack(b'IHDR', struct.pack("!2I5B", width, height, 8, 6, 0, 0, 0)),
        png_pack(b'IDAT', zlib.compress(raw_data, 9)),
        png_pack(b'IEND', b'')])
        

def saveAsPNG(array, filename):
    import struct
    if any([len(row) != len(array[0]) for row in array]):
        raise ValueError("Array should have elements of equal size")

                                #First row becomes top row of image.
    flat = []; map(flat.extend, reversed(array))
                                 #Big-endian, unsigned 32-byte integer.
    buf = b''.join([struct.pack('>I', ((0xffFFff & i32)<<8)|(i32>>24) )
                    for i32 in flat])   #Rotate from ARGB to RGBA.

    data = write_png(buf, len(array[0]), len(array))
    f = open(filename, 'wb')
    f.write(data)
    f.close()
    
saveAsPNG(buf, 'test_grid.png')
#saveAsPNG([[0xffFF0000, 0xffFFFF00],
           #[0xff00aa77, 0xff333333]], 'test_EU_grid.png')

```

my nodeTree:

[...]

PR #1588, opened 2017-04-26
save png feature

kalwalt:
reference Custom png saving: https://github.com/nortikin/sverchok/issues/1586

to save a png:

```python
"""
in data s d=[[]] n=1
in size s d=64 n=2
out some s
"""

import numpy as np
from sverchok.utils import sv_export_png as png

png.save_png('test01.png', data, 'RGBA', size)
```

### DIFF
--- a/utils/sv_export_png.py
+++ b/utils/sv_export_png.py
@@ -1,0 +1,57 @@
+"""
+code taken from:
+http://stackoverflow.com/questions/902761/saving-a-numpy-array-as-an-image/19174800#19174800
+thanks to @ideasman42 and @Evgeni Sergeev
+"""
+import numpy as np
+
+
+color_t = {'BW': 1, 'RGB': 3, 'RGBA': 4}
+
+
+def convert(buf, type, size):
+    array = np.array(buf)
+    print('array', array)
+    d = np.interp(array, [0, 1], [0, 255])
+    print('from converter', d)
+    data_uint = np.array(d, dtype=np.uint8)
+    res = np.reshape(data_uint, (size, size, color_t[type]))
+    return res.flatten().tolist()
+
+
+def write_png(buf, width, height):
+    """ buf: must be bytes or a bytearray in Python3.x,
+        a regular string in Python2.x.
+    """
+
+    import zlib, struct
+
+    # reverse the vertical line order and add null bytes at the start
+    width_byte_4 = width * 4
+    raw_data = b''.join(b'\x00' + buf[span:span + width_byte_4]
+                        for span in range((height - 1) * width_byte_4, -1, - width_byte_4))
+
+    def png_pack(png_tag, data):
+        chunk_head = png_tag + data
+        return (struct.pack("!I", len(data)) +
+                chunk_head +
+                struct.pack("!I", 0xFFFFFFFF & zlib.crc32(chunk_head)))
+
+    return b''.join([
+        b'\x89PNG\r\n\x1a\n',
+        png_pack(b'IHDR', struct.pack("!2I5B", width, height, 8, 6, 0, 0, 0)),
+        png_pack(b'IDAT', zlib.compress(raw_data, 9)),
+        png_pack(b'IEND', b'')])
+
+
+def save_png(filename, buf, type, size):
+    if buf[0]:
+        data = convert(buf, type, size)
+        # data = bytearray([int(p * 255) for p in buf[0]])
+        print(data)
+        d = bytearray([int(p)for p in data])
+        print(d)
+        final_data = write_png(d, size, size)
+        with open(filename, 'wb') as fd:
+            fd.write(final_data)
+            print('png saved')


zeffii:
this reminds me that I should add a way to trigger a callback for snlite (SN1 had this feature). What i mean is, when doing a custom draw, you could do something like
```python
"""
in data s d=[[]] n=1
in size s d=64 n=2
out some s
"""

import numpy as np
from sverchok.utils import sv_export_png as png


def draw_extra(self, context, layout):
    layout.prop("node.generic_callback_....", text="save png").fn_name = 'sn_operator_savepng'

def sn_operator_savepng(operator, node, context):
    # self is available locally as node.
    png.save_png('test01.png ............ )

...rest of node

```

kalwalt:
@zeffii an operator for a ui button?

zeffii:
yeah, snlite isn't fully implemented yet :)  but this was regression since SN1.

utils/sv_export_png.py
```diff
@@ -15,11 +16,11 @@ def convert(buf, type, width, height):
     d = np.interp(array, [0, 1], [0, 255])
     print('from converter', d)
     data_uint = np.array(d, dtype=np.uint8)
-    res = np.reshape(data_uint, (width, height, color_t[type]))
+    res = np.reshape(data_uint, (width, height, color_size[type]))
```
kalwalt: Probably this reshaping stuff is not needed i have to check.

kalwalt:
Now it save png by matrix or flatten list of data.

kalwalt:
@zeffii I think we could add a compression factor. Btw it could be implemented a lot of other features, but maybe for now it's ok.

kalwalt:
not ready yet to be merged, i'm checking if it save correctly.

kalwalt:
saving a png, comparison from `sv_export_png` and saving from `texture viz` node:

from sv_export_png:
test

from texture viz:
test_texture_viewer

i think that it is something not correct in my code....

kalwalt:
a rgba image:
from sv_export_png:
test_rgba

from texture viz:
test_rgba_texture_viewer
much better this.

zeffii:
that's cool!  (i'm not sure i'll have any immediate use for it... but whatever!  )  also interesting as educational tool

kalwalt:
i think i have to read carefully this: https://en.wikipedia.org/wiki/Portable_Network_Graphics
not sure if we have to do something similar as in the `texture_viz` for the Grayscale images. there we have to replicate a single layer for three times, I believe that was a blender procedure but maybe i was wrong.

> also interesting as educational tool

i think this is the most interesting part!

kalwalt:
from wikipedia:
> The PLTE chunk is essential for color type 3......and it must not appear for color types 0 and 4 (grayscale and grayscale with alpha).

Maybe this is the reason?

kalwalt:
no i haven't added this...

kalwalt:
i get it!!
```python
# BW
width_byte_4 = width
# RGB
width_byte_4 = width * 3
# RGBA
width_byte_4 = width  * 4
 
```

zeffii:
lol! ! !   and when you look at that...    how could it have been any other way?! :)

kalwalt:
yeah, also the reason is that i have no yet a complete understanding of the python language....

kalwalt:
with the latest:
**grayscale image**
test_02bw
**rgb image**
test_02rgb
**rgba image**
test_02rgba

not bad....

kalwalt:
i will add also a function to save a png directly from a numpy matrix of array.

kalwalt:
I switched to only one fuction for exporting, has much sense for me.

utils/sv_export_png.py
```diff
@@ -55,9 +55,7 @@ def save_png(filename, buf, type, width, height, compression=5):
     elif buf:
         data = convert(buf)
 
-    _type = color_type[type]
-
-    if _type not in [0, 2, 6]:
+    if type not in ['BW', 'RGB', 'RGBA']:
```
zeffii: avoid using keyword "type" , use _kind_ instead
```diff
   if not kind in {'BW', 'RGB', 'RGBA'}:
```

zeffii: or _mode_ :)

kalwalt: i will change in mode

utils/sv_export_png.py
```diff
@@ -39,7 +39,7 @@ def png_pack(png_tag, data):
                 chunk_head +
                 struct.pack("!I", 0xFFFFFFFF & zlib.crc32(chunk_head)))
 
-    t = color_type[type]
+    t = color_type[mode]
```
kalwalt: to improve code readibility i think i will turn this line and others to something:

```python
def png_header():
     return b'\x89PNG\r\n\x1a\n'
```

```diff
+ return b''.join([
- b'\x89PNG\r\n\x1a\n',
+ png_header(),
```

kalwalt:
Unfortunately, again, i have some issue with my pc (probably hardware..) i can't boot and so i can't work on this and other stuff at the moment, plus tomorrow i'm leaving for Paris for 3 months, i will work [here](https://www.studio-orta.com/en/les-moulins), hoping to have time to continue my Sverchok adventure and to fix my pc troubles...

kalwalt:
I'm back to this. Added the changes to the save function. @zeffii we can add something else or improve the code in some way?

kalwalt:
I add here some pictures for documentations:
**save_png** with normal input:
savepng_custom_normal


**save_png** from a numpy array input:
savepng_custom_w_numpy

code to use with script node lite:
```python
"""
in data s d=[[]] n=1
in width s d=64 n=2
in height s d=64 n=2
out some s
"""

from sverchok.utils import sv_export_png as png

png.save_png('test/test_03nump', data, 'BW', width, height, 6)

```
with animation:
savepng_custom_anim

code for animation (script node lite):

```python
"""
in frame_current s d=0 n=2
in frame_start s d=0 n=2
in frame_end s d=4 n=2
in data s d=[[]] n=1
in width s d=64 n=2
in height s d=64 n=2
out some s
"""
import bpy
from sverchok.utils import sv_export_png as png


image_name = 'test_anim'

if data:
    if frame_current <= frame_end:

        path = 'test/' + image_name  + str(frame_current)
        png.save_png(path, data, 'BW', width, height, 6)
        
    if frame_current == frame_end:
        
        bpy.ops.screen.animation_play()
        print('stop animation')



```

kalwalt:
@zeffi it sounds good to me, It would be nice have some sort of save animation feature... waiting for the 2.8 release.

PR #1591, opened 2017-04-27
3dview center align tool

zeffii:
this is both 
- a nodeview operator, that works on an active node.. 
- a node operator that is added directly to the viewer interface (where it makes sense )

image


enzyme69:
So this thing will exist as a button click under Viewer draw? Or more like Macro as well?

zeffii:
re-read the first post on the top.

zeffii:
it's not currently a default macro but i might make it one...  i was thinking of `kw: zm` if I do.  but a button on the Viewer UI seems like the most convenient.

zeffii:
current implementation only jumps to first item in matrices, or vectors*matrices

PR #1627, opened 2017-05-05
add annotations and svpep

zeffii:


zeffii:
i think this solves some pesky problems, mentioned in the commit log

portnov:
@zeffii do you by any chance remember: why did we decide to identify node sockets in JSON by index, not by name?

zeffii:
yes, because names can be duplicate on a node, where index can not.

portnov:
Ah, I see. Theoretically there is "identifier", but mostly we do not provide these identifiers for our sockets...

PR #1632, opened 2017-05-08
adding np.resize to the buffer to allow partial filling

kalwalt:
I added the resize feature to the texture viz node as the lite version does. As @Kosvor2  did in the lite version i changed the input buffer, not real a big effort to do this but hadn't the time before to code and to test it.
@zeffii Please test and review it,  for me it's ok but check if it works in other Os.  Merge it when you want.

### DIFF
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -369,7 +369,7 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
             pixels = np.array(self.inputs['Float'].sv_get(deepcopy=False)).flatten()
             width, height = self.texture_width_height
 
-            transfer_to_image(pixels, self.texture_name, width, height, mode)
+            transfer_to_image(np.resize(pixels, self.total_size), self.texture_name, width, height, mode)
 
         if self.activate:
             texture = self.get_buffer()
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -369,7 +369,8 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
             pixels = np.array(self.inputs['Float'].sv_get(deepcopy=False)).flatten()
             width, height = self.texture_width_height
 
-            transfer_to_image(np.resize(pixels, self.calculate_total_size()), self.texture_name, width, height, mode)
+            resized_np_array = np.resize(pixels, self.calculate_total_size())
+            transfer_to_image(resized_np_array, self.texture_name, width, height, mode)
 
         if self.activate:
             texture = self.get_buffer()
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -369,7 +369,7 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
             pixels = np.array(self.inputs['Float'].sv_get(deepcopy=False)).flatten()
             width, height = self.texture_width_height
 
-            transfer_to_image(np.resize(pixels, self.total_size), self.texture_name, width, height, mode)
+            transfer_to_image(np.resize(pixels, self.calculate_total_size()), self.texture_name, width, height, mode)
 
         if self.activate:
             texture = self.get_buffer()
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -286,7 +286,7 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
         data = np.array(self.inputs['Float'].sv_get(deepcopy=False)).flatten()
         self.total_size = self.calculate_total_size()
         self.make_data_correct_length(data)
-        texture = bgl.Buffer(bgl.GL_FLOAT, self.total_size, data)
+        texture = bgl.Buffer(bgl.GL_FLOAT, self.total_size, np.resize(data, self.total_size))
         return texture
 
     def draw_buttons(self, context, layout):
@@ -294,7 +294,7 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
         c.label(text="Set texture display:")
         row = c.row()
         row.prop(self, "selected_mode", expand=True)
-        
+
         nrow = c.row()
         nrow.prop(self, 'activate')
         nrow.prop(self, 'to_image_viewer')
@@ -354,7 +354,7 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
     def process(self):
         if not self.inputs['Float'].is_linked:
             return
-        
+
         n_id = node_id(self)
         self.delete_texture()
         nvBGL2.callback_disable(n_id)
--- a/nodes/viz/texture_viewer.py
+++ b/nodes/viz/texture_viewer.py
@@ -208,11 +208,11 @@ class SvTextureViewerNode(bpy.types.Node, SverchCustomTreeNode):
         default="S", update=wrapped_update)
 
     width_custom_tex = IntProperty(
-        min=0, max=1024, default=206, name='Width Tex',
+        min=1, max=1024, default=206, name='Width Tex',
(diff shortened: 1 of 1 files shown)


kalwalt:
This works for displaying in the node tree but will not works if we want to pass the data to the internal blender image editor. Do you want also this feature?

kalwalt:
with the latest commit allow partial filling also for **BW** mode, **RGB** and **RGBA** is not allowed, need to figure out how to do this.

zeffii:
i can't push this right now

kalwalt:
no worries @zeffii  i will do this in my spare time, i will see for the **RGB RGBA** part...

zeffii:
cool! take your time

nodes/viz/texture_viewer.py
```diff
@@ -369,7 +369,7 @@ def process(self):
             pixels = np.array(self.inputs['Float'].sv_get(deepcopy=False)).flatten()
             width, height = self.texture_width_height
 
-            transfer_to_image(np.resize(pixels, self.total_size), self.texture_name, width, height, mode)
+            transfer_to_image(np.resize(pixels, self.calculate_total_size()), self.texture_name, width, height, mode)
```
kalwalt: this allow **rgb** and **rgba** mode but i get some 'red' alert ( i mean the texture viz node became red)

kalwalt:
stopping for now because too late in the night, i will see in the next days this issue...

kalwalt:
this from the console:
```
  File "/home/walter/build_linux/bin/2.78/scripts/addons/sverchok/data_structure.py", line 447, in updateNode
    self.process_node(context)
  File "/home/walter/build_linux/bin/2.78/scripts/addons/sverchok/node_tree.py", line 485, in process_node
    process_from_node(self)
  File "/home/walter/build_linux/bin/2.78/scripts/addons/sverchok/core/update_system.py", line 423, in process_from_node
    do_update(update_list, nodes)
  File "/home/walter/build_linux/bin/2.78/scripts/addons/sverchok/core/update_system.py", line 349, in do_update
    do_update_general(node_list, nodes)
  File "/home/walter/build_linux/bin/2.78/scripts/addons/sverchok/core/update_system.py", line 313, in do_update_general
    done_nodes = set(procesed_nodes)
RecursionError: maximum recursion depth exceeded while calling a Python object
File "/home/walter/build_linux/bin/2.78/scripts/addons/sverchok/data_structure.py", line 442, in updateNode

```

zeffii:
are you sure you pushed all commits? I don't see any reason for recursion in the current code..

zeffii:
it might be that the update into the Image is causing a scene update, which in turn causes a sverchok update..

kalwalt:
>  are you sure you pushed all commits? I don't see any reason for recursion in the current code..

yes, i'm sure.

> it might be that the update into the Image is causing a scene update, which in turn causes a sverchok 
update.

i will look into it when i have a bit of time.

kalwalt:
The recursion error appear when in you set the the dimension of the image to user(custom) and if width or heigth go to zero and re-assign a value greater than zero. The data are passed to the image viewer but the node stays in the `red` state.

kalwalt:
How can avoid this?

kalwalt:
min for width and heigth are set to 0 i will set to 1, i will see if this make a difference...

kalwalt:
The simplest and obvious solutions are always the best ones!

kalwalt:
...and also min = 0 does not make any sense to me...!

nodes/viz/texture_viewer.py
```diff
@@ -369,7 +369,7 @@ def process(self):
             pixels = np.array(self.inputs['Float'].sv_get(deepcopy=False)).flatten()
             width, height = self.texture_width_height
 
-            transfer_to_image(pixels, self.texture_name, width, height, mode)
+            transfer_to_image(np.resize(pixels, self.calculate_total_size()), self.texture_name, width, height, mode)
```
kalwalt: @zeffii maybe better to put in two lines this?
```python
            transfer_to_image(np.resize(pixels, self.calculate_total_size()),
                              self.texture_name, width, height, mode)
```

zeffii: or
```
        resized_np_array = np.resize(pixels, self.calculate_total_size())
        transfer_to_image(resized_np_array, self.texture_name, width, height, mode)
```

kalwalt:
i get again that error while increasing the size of the tex in the custom mode:
```
  File "/home/walter/build_linux/bin/2.78/scripts/addons/sverchok/data_structure.py", line 447, in updateNode
    self.process_node(context)
  File "/home/walter/build_linux/bin/2.78/scripts/addons/sverchok/node_tree.py", line 485, in process_node
    process_from_node(self)
  File "/home/walter/build_linux/bin/2.78/scripts/addons/sverchok/core/update_system.py", line 423, in process_from_node
    do_update(update_list, nodes)
  File "/home/walter/build_linux/bin/2.78/scripts/addons/sverchok/core/update_system.py", line 349, in do_update
    do_update_general(node_list, nodes)
  File "/home/walter/build_linux/bin/2.78/scripts/addons/sverchok/core/update_system.py", line 313, in do_update_general
    done_nodes = set(procesed_nodes)
RecursionError: maximum recursion depth exceeded while calling a Python object
File "/home/walter/build_linux/bin/2.78/scripts/addons/sverchok/data_structure.py", line 442, in updateNode
```
but at least the node now not became red.

kalwalt:
maybe we should do something like this:
```python
# inside Texture viever class

def update_to_image():
     mode = self.color_mode
     self.activate = False
     pixels = np.array(self.inputs['Float'].sv_get(deepcopy=False)).flatten()
     width, height = self.texture_width_height
     resized_np_array = np.resize(pixels, self.calculate_total_size())
     transfer_to_image(resized_np_array, self.texture_name, width, height, mode)
     updateNode(self, context)

if self.to_image_viewer:
    self.update_to_image()
```

kalwalt:
`updateNode` is intented only for the enum properties?

kalwalt:
the `update_to_image` solution doesn't works of course, but i found that this line of code is causing the 'recursion' issue : 
            `self.activate = False`
 https://github.com/kalwalt/sverchok/blob/b69a429774bdbeca2f651ae372897342d4553025/nodes/viz/texture_viewer.py#L368

we can simply delete it or find another solution but i  have any idea....

zeffii:
then wrap updateNode for that scenario
```
    def wrapped_updateNode_(self, context):
        self.activate = False
        self.process()

    to_image_viewer = BoolProperty(
        name='Pass', description='Transfer pixels to image viewer',
        default=False, update=wrapped_updateNode_)
```

zeffii:
meaning it doesn't need to be set from the process function .. (which is whats causing the retrig as you point out)

zeffii:
or just
```
    def wrapped_updateNode_(self, context):
        self.activate = False
        # self.process()
```

zeffii:
changing self.activate will trigger `self.process` anyway, so that is safe to drop from the wrapper..

zeffii:
and ofcourse remove that line to `self.activate = False` from the process function itself.

kalwalt:
i think this will solve the issue... i'm trying right now...

kalwalt:
yes! but for safety i will check sunday, since i will have any time tomorrow.... thanks @zeffii !

kalwalt:
but shoul i add also updateNode(self, context) in the wrapped update function?

zeffii:
nope.. because  if you modify `self.activate` from inside the `wrapped_updateNode_` , that action will still call updateNode.. because that's the update function of `self.activate`

zeffii:
the point of removing the explicit call to `self.activate=False` from inside the process function is to prevent the loop 

```
    def process
    
        self.activate = X   # this calls self.process again
```

zeffii:
good catch btw.

kalwalt:
understood. 

> good catch btw.

I went for exclusion, with a bit of lucky...!

kalwalt:
i should do also another refinement in the code and after i think this will be fine.

kalwalt:
I'm fine with it. What do you think @zeffii?

PR #1640, opened 2017-05-09
add second mode, multi

zeffii:
image


zeffii:
i'm intentionally splitting the row used for the single/multi toggle, because otherwise you will accidentally press the toggle if you increase the count.

zeffii:
added specific path for objects switching... the end user might have to unwrap or pick index from final output, depending on use case.

PR #1647, opened 2017-05-11
Simplify repeat_last()

cclauss:


cclauss:
Yes.  This version is slightly faster and removes 3 lines of Python code, 30 Python bytecodes, and is easier to read and reason about.

import dis
print(dis.dis(repeat_last_old))  # 71 Python bytecodes
print(dis.dis(repeat_last_new))  # 41 Python bytecodes

import timeit
test = 'x = [repeat_last_old([x for x in "abcde"]) for i in range(10)]'   # on my iPad...
print(timeit.timeit(test, setup="from __main__ import repeat_last_old"))  # 25.039370791986585
test = 'x = [repeat_last_new([x for x in "abcde"]) for i in range(10)]'
print(timeit.timeit(test, setup="from __main__ import repeat_last_new"))  # 24.785957959014922

zeffii:
there's quite a bit of cobwebbs in `data_structure.py`, i welcome the patch.

i'll do some timing tests with representative data before accepting the PR (we've learnt in the past that less code is not necessarily faster code.. hence my caution) - it looks much cleaner indeed.

zeffii:
or
```python
def repeat_last_new2(lst):
    """
    creates an infinite iterator the first each element in lst
    and then keep repeating the last element,
    use with terminating input
    """
    yield from lst
    while lst:
        yield lst[-1]
```

cclauss:
That limits you to [Python 3.3](https://docs.python.org/3/whatsnew/3.3.html) or later but it works.

```python
for i, x in enumerate(repeat_last([x for x in "abcdef"])):
    if i > 9:
        break
    print(i, x)
```

zeffii:
we've put python 3.5 as the prerequisite.

zeffii:
I hope you understand we might be a bit slow in making changes to that function, it gets used in a couple of the common nodes.

zeffii:
```python
def repeat_last_new2(lst):
    """
    creates an infinite iterator the first each element in lst
    and then keep repeating the last element,
    use with terminating input
    """
    yield from lst
    while lst:
        yield lst[-1]



def repeat_last_new(lst):
    """
    creates an infinite iterator the first each element in lst
    and then keep repeating the last element,
    use with terminating input
    """
    for x in lst:
        yield x
    while lst:
        yield lst[-1]


def repeat_last_old(lst):
    """
    creates an infinite iterator the first each element in lst
    and then keep repeating the last element,
    use with terminating input
    """
    i = -1
    while lst:
        i += 1
        if len(lst) > i:
            yield lst[i]
        else:
            yield lst[-1]




import dis
print(dis.dis(repeat_last_old))  # 71 Python bytecodes
print(dis.dis(repeat_last_new))  # 47 Python bytecodes
print(dis.dis(repeat_last_new2))  # 34 Python bytecodes

print('------')

import timeit
import random
rnd = random.random

data = str([[rnd(), rnd(), rnd()] for i in range(3)])
test = 'x = [repeat_last_old({0}) for i in range(20)]'.format(data)
print(timeit.timeit(test, setup="from __main__ import repeat_last_old"))
test = 'x = [repeat_last_new({0}) for i in range(20)]'.format(data)
print(timeit.timeit(test, setup="from __main__ import repeat_last_new"))
test = 'x = [repeat_last_new2({0}) for i in range(20)]'.format(data)
print(timeit.timeit(test, setup="from __main__ import repeat_last_new2"))

print('-------')
```

i think paints a different picture.

PR #1650, opened 2017-05-13
add boilerplate for custom defaults

zeffii:
this PR allows us to write a `node_defaults.json`, with two forms of directive

1. `props`
2. `function`

it will be extended to add

3. `sockets_states` , when _`socket.set_state()`_ is implemented.

##### datafiles/sverchok/node_defaults/node_defaults.json
```
{
    "SvGenFloatRange": {"props": [["stop_", 1.0], ["count_", 20]]},
    "ViewerNode2": {"props": [], "function": "view3d.init_vdmk2"}
}

```

- it's possible to push props in sequence with a list of `prop_name`, `prop_value` pairs.
- it's possible to push a function, in the example above the function is referenced as `view3d.init_vdmk2`. read that as `filename.function`. and filename shall be located in `datafiles/sverchok/node_defaults/`.

##### datafiles/sverchok/node_defaults/view3d.py
```python
import random
import colorsys

import bpy



def init_vdmk2(node):

    def fold_color(color, shift):
        h, s, v = colorsys.rgb_to_hsv(*color)
        return colorsys.hsv_to_rgb(shift, s, v)

    A = [0.938000, 0.948000, 0.900000, 1.000000]
    B = [0.500000, 0.752000, 0.899000, 1.000000]
    C = [0.030100, 0.488000, 0.899000, 1.000000]

    shift = random.random()
    node.vertex_colors = fold_color(A[:3], shift)
    node.edge_colors = fold_color(B[:3], shift)
    node.face_colors = fold_color(C[:3], shift)

    node.edge_width = 1.0
    node.vertex_size = 2.0

```

------------------

Each time you add a node, the `node_default_dict` is checked if the `bl_idname` has an entry. 

- this has no effect on nodes that you duplicate, they clone their properties according to `self.copy()`
- this doesn't offer a visual way to set the preferences yet, that would be the next step.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -79,7 +79,7 @@ root_modules = [
 
 core_modules = [
     "monad_properties",
-    "handlers", "update_system", "upgrade_nodes", "upgrade_group", "monad",
+    "handlers", "update_system", "upgrade_nodes", "upgrade_group", "monad", "node_defaults"
 ]
 
 utils_modules = [
@@ -157,6 +157,7 @@ if reload_event:
 
 import bpy
 from sverchok.utils import ascii_print
+from sverchok.core import node_defaults
 
 
 def register():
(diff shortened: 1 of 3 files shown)


zeffii:
image

PR #1658, opened 2017-05-18
add compact / depth  and remove print statement for defaults

zeffii:


zeffii:
this may increase usability of stethoscope.

PR #1660, opened 2017-05-19
small restruct

zeffii:


zeffii:
image

zeffii:
active frame will behave differently with the new frame socket, if you connect to it. if you do not connect, it will behave as it currently does.

forcing active frame from index / Frame Info only works when playing the scene in forwards direction.

PR #1661, opened 2017-05-20
add monad loop mode

zeffii:


zeffii:
you will experience many freezes, until you use it correctly.

kalwalt:
@zeffii thank you for this, i will try as i will have a bit of time!

PR #1678, opened 2017-05-25
Default macros cp+

zeffii:


zeffii:
@enzyme69 , `git checkout master` , then `git pull`.

load blender, draw a few GP strokes,  then `ctrl+space`. and type ` gp` (for a grease pencil macro)

image

means all we need to do is pick the GP layer :) and `pickframe` or `active_frame`.

enzyme69:
OMG, @zeffii so many good things you pump into SV.

PR #1703, opened 2017-06-04
pcloud and isinside (nodes)

zeffii:


enzyme69:
What is Hom Ran Vec, Zeff?

zeffii:
##### pcloud
formerly "hom ran vec" ( read the green stuff in "Files Changed"  for the full words ).  
pcloud is a generator that produces a vector grid, with 2 main options
  - randomization factor (with seed)
  - rm doubles distance :_( this is what lets us make point clouds with relatively homogenous distances between neighbouring points )_

##### point_inside_mesh
takes a list of points and a mesh (must be verts + faces) based, and returns the mask list for List Mask Out.

##### Why?
this is handy for all kinds of things, and i wanted to have a single node for each of these features.

portnov:
@zeffii 
Consider an option of adding "point_inside_mesh" functionality into "select mesh elements by location" node. Not sure if it will be much better, but may be. For example, "mesh select" can also select edges and faces connected to selected vertices.

zeffii:
@portnov  i understand and sympathise with that comment, but i really want a simple node for this.

zeffii:
@portnov  we could make it possible to load the `point_inside_mesh` into the node tree upon picking _points inside mesh_ from the enum on "Select mesh elements by location" node. (place it beside the node), but the shorthand for spacebar search will be `pin`

zeffii:
@portnov  or a more advanced mode of points_inside_mesh specifically for the "select mesh elements by location"  node.

zeffii:
The reason this algorithm fails for simpler geometry, is because the normals of the vertices are being considered. 

image

- black dot is the point being probed. 
- red lines/points is the low poly mesh from sideview, 
- blue line is the normal of the closest vert to the black dot.

this scenario makes clear why that point would be considered outside of the mesh; namely the normal is pointing away from the black point.

therefore subdivision locally is needed, when the mesh itself is too lowpoly to offer intended results

image

##### multisampling mode

the extra subdivision step should be automatic for meshes with low vertex counts, and adaptive to only subdivide faces that are in contact with the originally found  'closest' point in the `bm`.

zeffii:
```python
from mathutils.bvhtree import BVHTree

bvh = BVHTree.FromBMesh(bm, epsilon=0.0)
for point in points:
    # should return points on polygons.. not just real vertices
    co = find_nearest(pt)
    ray_direction = co - point
    co, normal, distance = ray_cast(co, ray_direction

```

zeffii:
def are_inside(points, bm):
    rpoints = []
    addp = rpoints.append
    bvh = BVHTree.FromBMesh(bm, epsilon=0.0001)
 
    # return points on polygons
    for point in points:
        fco, normal, _, _ = bvh.find_nearest(point)
        p2 = fco - Vector(point)
        v = p2.dot(normal)
        addp(not v < 0.0)  # addp(v >= 0.0) ?
    
    return rpoints

zeffii:
`find_nearest` finds the closest face, (co, normal, index,, distance) ,  but it's not very robust either on low poly input..

zeffii:
i'm commiting these nodes, with minimal documentation. -- both nodes are not vectorized implicitely.

PR #1706, opened 2017-06-08
Convex hull2

zeffii:
just code so far, haven't ran it.

### DIFF
--- a/index.md
+++ b/index.md
@@ -100,6 +100,7 @@
     LineConnectNodeMK2
     SvLatheNode
     SvConvexHullNode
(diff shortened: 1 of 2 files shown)


zeffii:
seems to work OK in brief testing. auto sort is quite convenient :)

zeffii:
it should probably hide the Topo sort option if "inside" is return values

zeffii:
docs to come shortly.. or someone else can write them.. it's not a complicated node.

zeffii:
the code is a little verbose, the alternative was almost unreadable obfuscation. This at least can be debugged.

PR #1709, opened 2017-06-10
profile understands now multyobject multyspline selection

nortikin:


nortikin:
the subject to you is x working not only for L line but closing after C curve

zeffii:
> x .... closing after C curve?

maybe that doesn't work. it's possible

zeffii:
if so I will fix it.

zeffii:
yeah, it doesn't work.. if C/c is the last segment type. ( i will need to think about this)
```
M 0,0
L 0,1
C 1,2 3,2 3,0 20 1
x

M 5,0
l 0,1
c 3,b a,2 c,d 20 1
x
```

zeffii:
it _might_ be easier for me to rewrite the parser in a much simpler way... than to fix this in the current code. But I don't think I can get my brain around it right now. maybe later today.

nortikin:
don't worry, it works and it is ok. avoided with zero-length line before 'x'
Will merge now.

nortikin:
oh, no, i want one feature with selected knot!

nortikin:
ok, i'm ready

nortikin:
From selection makes not only organized XML/JSON text with comments, but also making nodes for user convenience to work with (not go to manual)
User experience for both nodes increased.
Also profile node has information in indices viewer, user can orient in curve, witch knot to correct.
Also When you select knot it appears in inputs socket as 'a' veriable to manipulate, automatically (if not plugged allready)

mesh expression-fixing2
mesh expression-fixing1
mesh expression-fixing

PR #1713, opened 2017-06-13
Uilist in ob3

zeffii:
- [x] auto self update after operator removes items (Doesn't work for the last object.. not sure why yet)
- [x] fix "Matrixes" typo in socket name
- [ ] Sort is not used for ui list display order
- [ ] maybe (want) : objs which are not connected from their verts,edges,polygons sockets, shall only do work to obtain their matrix. 
   - this might mean adding an explicit `def update` to detect a change in socket connections.. -- remember, when you disconnect outgoing sockets on node X, node X's process function is not called.

~maybe : display VEFM (Vectors, Edges, Faces, Matrix) icons~

### DIFF
--- a/nodes/scene/objects_mk3.py
+++ b/nodes/scene/objects_mk3.py
@@ -27,23 +27,44 @@ from sverchok.utils.context_managers import hard_freeze
 from sverchok.utils.sv_bmesh_utils import pydata_from_bmesh
 
 
+class SvObj3DataCollection(bpy.types.PropertyGroup):
+    name = bpy.props.StringProperty()
+    icon = bpy.props.StringProperty(default="BLANK1")
+
 
 class SvOB3NamesList(bpy.types.UIList):
 
     def draw_item(self, context, layout, data, item, icon, active_data, active_propname):
-        # groups = datas
+
+        layout.label(item.name, icon=item.icon)
         
-        # if self.layout_type in {'DEFAULT', 'COMPACT'}:
-            
-        #     layout.prop(groups, "name", text="", emboss=False, icon_value=icon)
-        #     layout.label(text="Rawr", translate=False)
+        # This is needed to help disambiguate the origin of this click. The receiver needs
+        # to know from which node tree and node it originated.
+        action = layout.operator("node.sv_ob3_collection_operator", icon='X', text='')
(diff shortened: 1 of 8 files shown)


zeffii:
image

zeffii:
was tempted to use this on a single row for the UI.. but won't.
```python
        # OR !!

        # col = layout.column(align=True)
        # row = col.row(align=True)
        # row.prop(self, 'sort', text='', toggle=True, icon='SORTALPHA')  # sort alpha
        # row.prop(self, "modifiers", text='', toggle=True, icon='MODIFIER')  # post modifier
        # row.prop(self, "vergroups", text='', toggle=True, icon='GROUP_VERTEX') # vertex groups
        # row.operator(callback, text="Select").fn_name = 'select_objs'
```

nortikin:
but we always can reduce height a little

zeffii:
are we ready for this? i'd like to push soon. but will test a bit with iojson first.

nortikin:
when i try to remove objects from list, it rempoves wrong items. it seems we need rebuild list dictionary again each time

zeffii:
Ok. i'll wait till there is time to test / fix

nortikin:
it seems uilist count every time wrong index of item

nortikin:
i found how to fix

nodes/scene/objects_mk3.py
```diff
@@ -34,7 +34,7 @@ class SvOB3BDataCollection(bpy.types.PropertyGroup):
 
 class SvOB3BNamesList(bpy.types.UIList):
 
-    def draw_item(self, context, layout, data, item, icon, active_data, active_propname):
+    def draw_item(self, context, layout, data, item, icon, active_data, active_propname, index):
```
zeffii: cool. looks correct.

zeffii: maybe variable `self.active_object_index` is no longer needed then..

zeffii: oh.. i think that variable _is_needed, but wasn't being used correctly without the `index` parameter

zeffii:
could add an operator to each list item 'hide', 'select', 'edit' .

PR #1723, opened 2017-06-21
add color where it should be added

zeffii:


zeffii:
from what I understand, color needs to be added at the point of the creation of the new stroke. and only reassigned a new colour if the color of a current stroke doesn't match a new suggested color

zeffii:
this is about as good as it gets.

zeffii:
@nortikin 
we could make it this way..
- if input layer is not connected, then offer the GP selection (with option to + new one) directly on the drawer.

yes.. i think i'll do that :)

PR #1733, opened 2017-07-04
removes all bgl and updates

zeffii:


zeffii:
lacking a way to update, and that means adding sockets :)

PR #1738, opened 2017-07-08
docs: Improve lesson 02 text. Add missing transition to lesson 03.

faerietree:


zeffii:
Although I think `inputted` is technically correct.

image

faerietree:
You are right. My bad. I see language like programming. Consistency is a beautiful queen. So if one wanted to become king "exceptions" should have no say in your rule(s). But I make many mistakes myself and it is picky. Sorry for that one.
With the increasing extent of the versioning spider webs, becoming king becomes less attractive. Thus knotting more nodes into such webs than really necessary is unwise. Definitely I made a mistake here.

(I did not want to put you to the pillory. Just wanted to strike some of your TODOs. Because your work is great. Thanks for your effort.)

zeffii:
I'm 100% sure that I only understand 10% of that last comment.

people can strike off my todo's all day / every day. Thanks :)

PR #1745, opened 2017-07-20
Pinside mod

zeffii:


zeffii:
Warning to users: May cause existing layouts that use the "old" node to require that you replace that node from the layout ( add a p_inside_mesh node from the menu, hook it up, delete old p_inside_mesh node from layout... bam! )

PR #1754, opened 2017-08-17
Add  HyperCube node to extended generators (WIP)

DolphinDream:
This node generates the 3D projection of a hypercube rotated in 4D space (via 6 rotation angles).

**

This is work in progress. Not a major attraction kind of node I presume, but I wanted to get some feedback on it while I’m developing it :) Any suggestions appreciated.

Also, there seems to be some sort of memory leak with this node that I can’t figure out where it comes from. Here’s a node setup: (https://gist.github.com/6b03981e3673c05a1aec2add0c7a8b09)

sv-hypercube-setup1

sv-hypercube-demo1


nodes/generators_extended/hypercube.py
```diff
@@ -0,0 +1,253 @@
...
+_hypercube = {}
+
+
+def project(vert4D, d):
+    '''
+        Project a 4D vector onto 3D space.
+    '''
+    cx, cy, cz = [0.0, 0.0, 0.0]  # center
+    x, y, z, t = vert4D
+    return [x + (cx - x)*t/d, y + (cy - y)*t/d, z + (cz - z)*t/d]
```
zeffii: can d ever be `0`  ?

zeffii: i guess it can... `[project(verts4D[i], d) for i in range(len(verts4D))]`

DolphinDream: correct. d could be zero and I should add a check for that.  This doesn’t explain of course the memory leak i observed (as I don’t set the d to zero in the node input)

zeffii: well, `d` seems to me like it will always be `0` at least once (at the start of iterating through `range(len(verts4D))`.  if len(verts4D) is larger than 0.

how do you know the bug is the result of a _memory leak_? does it give an error.. or crash blender?  i'm too occupied to try.

DolphinDream: actually **d** comes from the function parameter.. the **i** is the one iterating from 0 to len(verts4D).

There was no error or crash. The reason why I suspect a memory leak is that I left the above nodetree/animation running all night and blender was really, really slow in the morning. Checking the activity monitor (memory) (mac os x) it showed that blender has eaten up 2GB of memory. I restarted blender and monitored how much memory it eats up as the same node tree / animation is running and it seems to grow at a visible rate (about 2MB/sec or so).  Other node setups don’t seem to consume that much memory.

Thanks for the feedback, zeffi. don’t worry about it if you’re busy. I thought it may have been obvious by looking at the code if my implementation was wrong.

zeffii: of course -- my bad, `d` isn't the iterated value there.

you might want to cache `_hypercube` inside a `self.node_dict` using `node_id(self)` as a key, it might not really matter in a single-thread process.  If you had 2 hypercube nodes they would both use the same dict, same with writing multiple hypercubes within the same node.

another way to narrow down the culprit is to try the same experiment with a BMeshViewer instead, it could be that the bgl drawlist is not cleared.

zeffii: i tried and can confirm that using bmeshviewer with fixed vert count (n panel option) keeps the memory usage on a flatline..

zeffii:
additionally I think the edges socket needs corrected wrapping of the output.

image

DolphinDream:
Point noted about the edge list format. Unfortunately the edge list has no output as of now (on my todo list).. only the verts and polys are defined for now.

zeffii:
At least it appears the memory "congestion" isn't happening in your node, but rather the vdmk2. When I have enough time i'll take a closer look at the display list. I can completely imagine that we never properly flush it.

DolphinDream:
Yep. You’re right. I checked the bmesh viewer node instead of vdmk2 node and the memory consumption is flatlined (growing at a much slower rate). It’s curious though why other node setups using vdmk2 don’t seem to have the same anomalous memory consumption behavior I observed with the hypercube node.

zeffii:
> It’s curious

i concur. best share more gists of these scenarios then. The problem may be two fold: for the vdmk2 to be at fault (or at least partially implicit) the same would need to happen when hypercube node isn't present..)

DolphinDream:
I see a similar behavior with just the box node (no hypercube node) so it may be vdmk2.

https://gist.github.com/f561d16b0b670bd676f001d1c94f8f07

memoryleakhunt

zeffii:
This might be an effect of bgl upgrades in Blender over the past months. I'll make a dedicated issue for this .

nortikin:
hypercubenode

portnov:
What is the status of this?

DolphinDream:
Still working on it. I’m considering splitting the TRSP manipulation functionality (translation, rotation, scale projection) out of this node and put it into its own 4D math node, and let the Hypercube node be just a generator. A 4D math node would also allow for other things to be manipulated in 4D not just the hypercube verts. My longer term goal is to also have nodes to do 4D intersections etc. 

@portnov are you asking because you are interested in this node or you’re just wondering if it’s a stale PR?

enzyme69:
I am awaiting in anticipation. The amount of details you put into a node is
fascinating. One day, make useful node.

DolphinDream:
Here’s a preview of a 3D plane intersection node I’m developing to hopefully extend to 4D.
A rotating 3D round box passing through an array of slicing planes.. predator camouflage style :) I’m curious to know what a 4D counterpart would look like.

sv-planeintersection-demo1

portnov:
I was asking just because I see a PR that seems to be almost finished, but for some reason not merged yet. Maybe anyone forgot about it or what...

zeffii:
> 3D plane intersection node

what does it do that the `cross-section` and `Bisect` nodes don't do ?

DolphinDream:
@zeffii yeah..  i did not know about those nodes :) I tried the boolean node btw and it did not do what I expected so I created a separate node to explore the concepts of intersection, which would help me to further extend to 4D. I’ll take a look at these nodes to see if I can borrow some ideas.

DolphinDream:
Btw, both cross-section and bisect seem to have problems intersecting suzzane ..

https://gist.github.com/77d0dcdeafa12bab28424a8d353912e1

sv-bisect-crosssection-bug

enzyme69:
Suzanne can slice ok I think...
screen shot 2017-12-02 at 12 42 08

DolphinDream:
@enzyme69 yeah. for oblique planes seems to work fine. try using the horizontal planes like in my gist. Also, try to only show the cross section (not the volume part of the suzanne).

DolphinDream:
in interest of keeping this feed about hypercube .. i removed some comments.  I opened an issue for the cross-section/bisect bug.

DolphinDream:
I broke the hypercube node into its 4D math components and put each one into its own node. The initial hypercube node did everything: generated the 4D mesh, generated a 4D matrix, transformed the verts it in 4D and projected them into a 3D space. With the upcoming update these disticnt operations are hosted by individual nodes. This will allow any 4D objects (not just the hypercube) to be transformed and projected into 3D, but also allows for more complex 4D manipulation of any 4D mesh (e.g using the 4D Matrix Math node which does matrix composition)

The proposed (implemented) nodes are:
- Hypercube (generator only node)
- 4D Matrix In (equivalent to the Matrix In node)
- 4D Transform (equivalent to Matrix Apply node)
- 4D Matrix Math (equivalent to Matrix Math node)
- 4D Project

sv-hypercube-4dnodes

sv-hypercube-demo11

portnov:
Huh. That may require a separate nodes category :)
Nodes for other 4-polytopes are on their way I suspect? :)

DolphinDream:
@portnov of course. A 4D Polytope node would be a natural extension. I thought that perhaps such node could replace the hypercube node and provide a set of different polytopes (including the Hypercube).  As for a separate node category, you may be right :)

I also was thinking whether to have these nodes be stuck with 4D or should generalize them to N-D. It would probably be simpler to start with 4D and see how to extend them to N-D later.

portnov:
I think this PR should be rebased to 2.8 branch and merged someday.

DolphinDream:
Absolutely. Btw, What is the status of the 28 branch? Are the structural changes mostly done? halfway done? Honestly one major thing that discourages me to convert nodes to 28 are the sort of structural bugs similar to the one that I encountered when I converted the switch node which kind of drained me. Personally I don’t have a whole lot of time, but when I do and I dedicate it to developing nodes I would rather not waste it on figuring out ancillary bugs that have nothing to do with the node itself I’m developing. I’ve been waiting for the 28 branch to become a little bit more stable first before I convert a bunch of nodes that I started or developed for 2.79.

zeffii:
the safest node is a node that doesn't try to go 9000-level transformer - AA Abrahams - style.  Most nodes only have minor UI manipulation. So switchnode is really an exception..

i'll commit my version of the switchnode, it gets to the core of the functionality needed - but at the expense of some (what i consider) superficial features.

zeffii:
I recommend not adding / removing sockets - to avoid headaches.  use `hide_safe`

Durman:
This can be inconvenient for users who get used to use Ctrl + h.

zeffii:
ugh : (

zeffii:
ok.. the change to b28 hasn't made using `node.inputs.new()` / `.remove` less correct. But avoid doing clever things in a node's `update` function.

switchnode is an extreme example.  and i've had no problems worth discussing with any other nodes in the transition.

portnov:
@DolphinDream any update on status of this PR? :)

PR #1782, opened 2017-09-12
Add spin twist option to the torus node

DolphinDream:
The "spin twist" feature allows to rotate the spin sections of the torus such that the edges/polys appear as a continuous twist around the torus tube. The smallest spin twist is the "spin section delta angle" (= 2*pi / # spin sections). For this angle, for a complete revolution the last spin section is rotated relative to the first spin section by an angle as to shift the vertices in the spin section list by one vertex (essentially, one polygon shift in the spin direction). The total spin is typically an integer number of this smallest spin twist angle, which means in order to have the torus make a 2*pi (full) twist for a complete revolution, the "spin twist" value has to be equal to the number of spin sections T = Ns. For double full twists N = 2Ns etc.

sv-twistedtorus-demo1

### DIFF
--- a/nodes/generator/torus.py
+++ b/nodes/generator/torus.py
@@ -27,7 +27,7 @@ from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode, match_long_repeat
 
 
-def torus_verts(R, r, N1, N2, rPhase, sPhase, Separate):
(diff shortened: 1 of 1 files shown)


stansu:
Great!

enzyme69:
Do I click on Merge Pull Request?

DolphinDream:
I was hoping for a quick review first.

DolphinDream:
I will add this new feature to the documentation soon.

PR #1791, opened 2017-09-26
Add new MatrixBasisChangeNode (to beta nodes) - WIP

DolphinDream:
This is a generalization of the "matrix normal" node, and it takes a set of 3 vectors T,B,N (orthonormalized or not) and generates a homogeneous matrix that transforms the orthonormal XYZ axes such that they align with the given 3 vectors, based on a given XYZ to TBN one to one mapping provided by the node. If the 3 given vectors TBN are orthonormal (and right-handed), the generated matrix is in essence a "basis change matrix" between XYZ and TBN. If the 3 vectors TBN are not orthonormal, the node will align the XYZ axes to the 3 given vectors TBN based on an "orthogonalization order" setting provided by the node.

Example:

If the orthogonalizing order is ZXY and Z is mapped to T and X is mapped to N, the Z axis (having the highest priority) will be aligned precisely with T, while X will be perpendicular to Z and lie within the plane defined by T and N (so N is loosly determining the direction of X). The last axis, Y is derived automatically to be orthonormal to the ZX plane.

sv-matrixbasischange-x
sv-matrixbasischange-1
sv-matrixbasischange-2
sv-matrixbasischange-3


DolphinDream:
And a live demo :) The node setup is a bit complicated (though not that much). It basically takes the location of 3 empty objects in the scene to define the T B and N vectors.. and it renders them along with the aligned XYZ vectors .. plus the cube and the cross section plane for illustration. 

and the node gist: https://gist.github.com/8961425418abd90db35a14e41d52f4ac
(you’ll have to add the 3 empty objects though to make it work with the node)

sv-matrixbasischange-demo1

portnov:
A time ago I thought about "matrix math" node, which could add, multiply matrices, calculate SVD and so on... I think I even got some code somewhere...

DolphinDream:
@portnov A matrix math node would be nice to have.

nortikin:
that is must have

DolphinDream:
Btw, is there a simpler way to get the location of an empty in the scene other than using the object select node + matrix-out nodes ?

DolphinDream:
I can work on a matrix math node i you guys can give me some feedback on must have operations..
Some I can think of:

- Add
- Subtract
- Multiply
- Scale
- Determinant
- Orthogonalize
- Normalize
- Homogenize
- Inversion
- SVD ?

DolphinDream:
@nortikin oh yeah. i forgot about the automatic socket conversion :)

nortikin:
i donno much of matrices, as for me it is enough

DolphinDream:
btw.. any kewl redux tricks how I can condense this into fewer lines ?
can I use the orthogonlizer_order string (minus the spaces) to create a lambda that has the name of one of the class functions suffixed with that string ?

```python
def orthogonalizer(self):
        if self.orthogonalizing_order == "X Y Z":
            orthogonalizer = lambda X, Y, Z: self.orthogonalizeXYZ(X, Y, Z)
        elif self.orthogonalizing_order == "X Z Y":
            orthogonalizer = lambda X, Y, Z: self.orthogonalizeXZY(X, Y, Z)
        elif self.orthogonalizing_order == "Y X Z":
            orthogonalizer = lambda X, Y, Z: self.orthogonalizeYXZ(X, Y, Z)
        elif self.orthogonalizing_order == "Y Z X":
            orthogonalizer = lambda X, Y, Z: self.orthogonalizeYZX(X, Y, Z)
        elif self.orthogonalizing_order == "Z X Y":
            orthogonalizer = lambda X, Y, Z: self.orthogonalizeZXY(X, Y, Z)
        elif self.orthogonalizing_order == "Z Y X":
            orthogonalizer = lambda X, Y, Z: self.orthogonalizeZYX(X, Y, Z)
        else:  # default is XYZ (fall through case)
            orthogonalizer = lambda X, Y, Z: self.orthogonalizeXYZ(X, Y, Z)
```

DolphinDream:
Another question on the socket labels:

T: Tangent
B: Binormal
N: Normal

Since I refer to these three vectors as T, B and N I thought it would be easier if i prefixed the socket labels with these letters, but maybe that is not necessary. Any thoughts on that ? Would it be obvious that TBN stand for Tangent, Binormal and Normal ? :)

DolphinDream:
Btw I selected the T, B, N labels more or less arbitrarily. These vectors don’t necessarily need to be tangent, binormal and normal to a surface, but can be any vectors. A simpler notation wold have been X’, Y’, Z’.. or even V1, V2, V3 .. but I tried to stay away from X’,Y’,Z’ since it is very confusing to look at a mapping from XYZ -> X’Y’Z’. As for the V# notation I thought this has one too many letters ;) and it works better to have one letter for the horizontal (compact) layout of the UI.  Plus V1V2V3 doesn’t really inspire the idea of an orthonormal basis (like the TBN or XYZ do).

portnov:
@DolphinDream See https://github.com/portnov/sverchok/tree/matrix_math and https://github.com/portnov/sverchok/issues/17 .

DolphinDream:
This is more of a python question.. when I normalize the vectors I do a check: if n != 0 then x  = x/n .. Would it be better/faster to do a try catch instead and always do x/n without a if check, and in case of a  division by zero the catch clause will then skip the normalization  ?

portnov:
"They say" that exceptions in python are almost as fast as if/else. But, it is generally a bad idea to use exceptions for control flow. So, I'd vote for "if".

DolphinDream:
@portnov When you say try/catch is almost as fast as if/else do you mean this is true anytime or specifically when an exception is thrown ? My thinking was if the division by zero happens very rarely, it may be faster to do try x/n .. and when the n is zero then the try/catch of that division by zero is probably as fast as if/else. I wonder if the “preparation” of the try clause underneath takes a bunch of other calls.. in which case it makes sense to just do a if/else check.

portnov:
Well, https://stackoverflow.com/questions/2522005/cost-of-exception-handlers-in-python :)

nortikin:
https://github.com/nortikin/sverchok/pull/1791#issuecomment-333274008
what if enumerate combinations of XYZ ZYX XZY... and one function will evaluate order and choose indexed letters for cross operations.
It will help avoid XXY or ZZZ inputs

DolphinDream:
Perhaps this is what you had in mind @nortikin with the more compactness idea? 

The only problem with this layout is that the node wiould have to be made wider so that the letters in the X,Y,Z dropdowns don’t get trimmed (and disappear even, especially when selecting the negative ones).

sw-matrixbasischange-04

DolphinDream:
> #1791 (comment)
what if enumerate combinations of XYZ ZYX XZY... and one function will evaluate order and choose indexed letters for cross operations.
It will help avoid XXY or ZZZ inputs
>
>Or even input not XYZ, but 012 or leave label as XYZ, but enumerate data will be 012 120 210 021 etc

@nortikin can you please clarify what you mean by this comment? 

I know that user can currently select X->T and Y->T which would create a malformed matrix (since the vectors are linearly dependent) but the user can easily change that. I could impose a restriction on one of the mapping to only list the allowed choices to avoid creating a singularity, but then it will be somewhat difficult to make a XYZ mapping selection. e.g. if X->T and Y options are limited to N -N B -B it makes it hard to make Y->T or -T as it will first require to change X to be not T (or -T) then make Y T or -T. Besides, even with such restrictions once you select T->B you can change X->B creating a singularity. I don’t think there is an automatic way to set the selections for XYZ based on just one of them. Also, even if the mapping is always well formed (no XXY, ZZZ etc) the result matrix can still be malformed since the input vectors themselves (TBN) can be linearly dependent (T = (0,0,1), B = (0,0,2) . and X->T, Y->B, Z->N will result in a malformed matrix since T and B are linearly dependent).

DolphinDream:
Alternatively I can call TBN simply ABC.. since there is no obligation for the user to plugin a Tangent vector into the Tangent socket, nor there is an obligation for the TBN to be orthogonal to each other. 

sv-matrixbasischange-05

DolphinDream:
What does MK stand for btw?

nortikin:
https://github.com/nortikin/sverchok/pull/1791#issuecomment-335038359
zeffii added this detail not remember

DolphinDream:
The more I think about it..  it makes sense to drop the third vector altogether since only two vectors are used at one time. The “Orthogonalizing Order” (XYZ, ZYX...) determines which is the “Track” vector (the first axis in the list) and the “Up” vector (the second in the list).. while the third is defined via cross product. 

sv-matrixbasischange-06

DolphinDream:
Here’s another iteration. I simplified the input to only take two vectors: A and B to be used as arbitrary Track and Up vectors to create the orthonormal XYZ axis/matrix.  I also added the scale input to allow to create full matrix composition Translation + Rotation + Scale. 

sv-matrixbasischange-demo2

DolphinDream:
another live demo ... 

sv-matrixbasischange-demo3

DolphinDream:
and some examples.. 

gist: https://gist.github.com/45f1a492d1986837ef45c0c498e035b7

sv-matrixbasischange-example1

sv-matrixbasischange-example2

DolphinDream:
Node name poll. Any preference how to call this node ? 

- Matrix Basis Change
- Matrix Track To
- Matrix Align

DolphinDream:
I updated the UI to have the Track/Up selected at the same time from the list of all possible combination AB, BA, -AB, B-A etc.. 

The logic of the node should be pretty straightforward. The first letter in each dropdown represents the **Track** and the second represents the **Up**. The third letter in the first drop down (the orthogonalizing order) is derived via cross product of the first two axis. The Track/Up dropdown basically allows to map the Track/Up axis defined in the first drop down to any of the (A,B,-A,-B) permutations. 

e.g.

**Z**X Y  / **-A** B 

means **Z** points precisely in the direction of **-A** (tracks -A) and **X** points in approximately direction of **B** (up is B). Of course, X can only point exactly in the direction of B if A and B are orthogonal to each other, otherwise X (perpendicular to Z=-A) will only point in a direction close to B and within the plane defined by A & B.

sv-matrixtrackto-demo1

DolphinDream:
If there are no objectsions I was gonna merge this into master. Any additional feedback is welcomed of course.

DolphinDream:
And a few more refinements... 

I decided to drop the third letter in the XYZ (“orthogonalizing order”) and only keep the first two, which are the axes mapped to **Track** and **Up** (the third is derived via cross product and thus the user has no influence over it). More over even though the “orthogonalizing  order” make sense from the implementation (math) point of view the name is a bit obscure, so I decided to change this to be “**Track/Up Axes**” instead and the second drop down to be “**Track/Up mapping**”.  With this renaming/cleanup the node’s logic is: the “Track/Up axes” lets you select which two axes (from X,Y and Z) are Track and Up axes.. and the “Track/Up mapping” lets you select to which input (A, B or its negatives) to map the Track and Up vectors. 

I also moved the “normalize vectors” option to the property panel (button extended) since this is an option that is probably not going to be used very often. It basically normalizes the output XYZ vectors (ON by default) and it exists simply to allow the user to turn OFF normalization to optimize computation if the normalization is not needed.

sv-matrixtrackto-9

DolphinDream:
Another example:

gist: https://gist.github.com/5ec2d2b68cda7840b9adf87a46096639

sv-matrixtrackto-10

sv-matrixtrackto-11

sv-matrixtrackto-demo2

DolphinDream:
And same as above but with “track axis” pointing towards the center of the torus knot.

https://gist.github.com/1e1c3f567a3b5cd41e168284890dd166

sv-matrixtrackto-12

sv-matrixtrackto-demo3

portnov:
@DolphinDream see #1884 :)

DolphinDream:
@portnov 👌🏼you wish you had this node, eh? 🤓 or had you already created this functionality for your PR?

portnov:
@DolphinDream 
"Objects along edge" historically had "autorotate" function (algorithm of how to rotate object along vector) based on Householder reflection matrix. In the referred PR, I introduced two more algorithms, based on blender's .track_to() and .rotation_difference() methods. Maybe you will be interested.

nortikin:
that is reason to start thread n issue before start coding.

DolphinDream:
I think the current node (Matrix Track To) has unique features to worth its own node. Sorry, but I fail to see how the “Duplicate Object Along Edge” node can orient an object (get the transformation matrix) given two arbitrary vectors A and B. I’m sure it does it internally, but it’s not obvious I can (or should) use this node for that purpose. Perhaps it gets these vectors from the edges entries (?). Anyway, just to compute a matrix (which then could be composed with another matrix stack) having to use the DOAE node is a bit non-intuitive. The algorithm to compute the matrix in MTT was not a big deal I’d say; the tricky part was coming up with an intuitive, simple enough UI that would let user get a track-to matrix given two arbitrary vectors.  (MTT is an extension of the “matrix normal” node to any two arbitrary vectors, plus it adds the ortho-normalizing order feature).

PR #1793, opened 2017-09-28
Torus knot node updates - add tangent output + optimization

DolphinDream:
Add tangent output to the node and conditionally generate normals and tangent only when outputs are connected. Also made some minor optimizations throughout the node.

### DIFF
--- a/nodes/generator/torusKnot.py
+++ b/nodes/generator/torusKnot.py
@@ -21,23 +21,27 @@
 import bpy
 from bpy.props import IntProperty, FloatProperty, BoolProperty, EnumProperty
 
-from math import sin, cos, pi, sqrt, radians, gcd
-from random import random
+from math import sin, cos, pi, sqrt, gcd
 import time
 
 from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode, match_long_repeat
 
-DEBUG=False
-
-options_plus=True
+DEBUG = False
 
 
 ########################################################################
 ####################### Knot Definitions ###############################
 ########################################################################
-def Torus_Knot(R, r, p, q, u, v, h, s, rPhase, sPhase, flipP, flipQ, N, linkIndex=0):
+def Torus_Knot(flags, settings, linkIndex=0):
     '''
+        flags =
+        compute_normals    : compute normal vectors
+        compute_tangents   : compute tangent vectors
+        normalize_normals  : normalize normal vectors
+        normalize_tangents : normalize tangent vectors
+
+        settings =
         R         : major radius of the torus
         r         : minor radius of the torus
         p         : number of REVOLUTIONS around the torus hole
@@ -51,17 +55,14 @@ def Torus_Knot(R, r, p, q, u, v, h, s, rPhase, sPhase, flipP, flipQ, N, linkInde
         flipP     : flip REVOLUTION direction (P)
         flipQ     : flip SPIN direction (Q)
         N         : number of vertices in the curve (per link)
-        linkIndex : link index in a multiple link knot (when q & p are not co-primes)
+
+        linkIndex : link index in a multiple-link knot (when q & p are not co-primes)
     '''
     if DEBUG:
         start = time.time()
 
-    # use plus options only when they are enabled
-    if not options_plus:
-        u = 1
-        v = 1
-        h = 1
-        s = 1
+    compute_normals, compute_tangents, normalize_normals, normalize_tangents = flags
+    R, r, p, q, u, v, h, s, rPhase, sPhase, flipP, flipQ, N = settings
 
     # scale the radii
     R = R * s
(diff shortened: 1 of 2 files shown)


nodes/generator/torusKnot.py
```diff
@@ -320,9 +343,9 @@ def sv_init(self, context):
 
         self.outputs.new('VerticesSocket', "Vertices")
         self.outputs.new('StringsSocket',  "Edges")
+        self.outputs.new('VerticesSocket', "Tangents")
```
zeffii: adding sockets inbetween existing sockets will cause .blends that use the existing sockets to output unexpected data when loaded with the new version of the Node. However, if you add the new socket as a last socket things will be OK.

also json IO works on index and not socketname.

things to consider.

DolphinDream: good point. I thought that regardless of what you change in a node you'll have to update those node-trees using an older version of the node anyway.. so if the tanget output i added at the end of within other sockets it wouldn't have mattered (it made more sense to me for the tangent to come before normal). 

And btw..
1. It would be nice if you could reorder sockets and have the node still work correctly (not sure if that is possible). 
2. Sometimes I get this nodes with a red box around them hinting that the node is obsolete / outdated. But I don't get the same for the torus knot node once i updated it to a new version. Any ideas why? 
3. The new() for sockets take a 3rd argument which i see in many nodes being populated with the same as the second argument. Any reasons for that? I thought that one is supposed to be the name of the socket, and the other the socket label displayed in the UI. If that would be the case you could easily rename the socket LABEL in one place and not have to update all the code referencing the NAME of that socket.

DolphinDream:
General question.. am I free to merge some of these changes in the master or should I let the other more senior contributors do it ? Regardless, I would still like to get my PR changes reviewed of course to catch any possible errors and also prompt any reasonable optimizations before it's integrated.

zeffii:
If you aren't making an MK2 - the general approach we take is:

1. have a demo blend that works with the original node
2. does the new branch load that .blend without incident

- if yes, push if you want. (if it's your own node.. you'll probably know it best..and have good test blends to cross check).
- if no, then 
    - you can nudge to mk2. ( and stick old node with the list of old nodes )
    - or reason about why the new branch doesn't load old .blends.
         - main reasons:  different socket order, new sockets inbetween existing sockets.

Regarding "can we make the node system agnostic to socket location" ?  Anything is possible :) but does the work / new complexity required to make that work.. justify the feature ( we've stayed away from it in favour of MK2...MK3).

lots more to be said about the topic

zeffii:
(i updated the last response, hit f5)

zeffii:
let me know if I missed anything in your question.

DolphinDream:
And my third question about the socket.new("VerticesSocket", "name", "name"). What's the use of the last parameter? And why not use the last parameter as the UI label of the socket, and the second parameter as the name of the socket.

zeffii:
a red box will appear if a loaded .blend contains a Node (bl_idname) that isn't in `index.md`, the postload handler will seek in the currently loaded nodes that index.md produces, then in `/oldnodes` and load from there... and if it finds it there it will do the extra red+frame

zeffii:
`socket.new("VerticesSocket", "name", "name")`  this question is maybe more suitably directed at lukastoenne who did the pynodes API.

often we find it handier to omit the third param because : effort++  and it looks goofy to repeat it.

zeffii:
i may be wrong.. but I think we also like to keep `socket names` and `identifier` the same to aid introspection. A consistency thing.

zeffii:
otherwise you can run into problem like if you try to script the nodelinks of certain nodes. Forexample the Color mix node. (Cycles) that node has two sockets called "Color", and the only way to correctly identify the desired socket is by index.

DolphinDream:
@zeffii though I understand your point.. would this rationale be pertinent to this node: given that this node is somewhat of a less popular one, and given that it was introduced fairly recent and perhaps not many people had used it for many projects, it would be fine to reorder the sockets as it would only break a few blend files / node trees ? :) In the end I don’t think the order of the outputs is paramount for this node.. it was just a personal preference given that I made this extension in parallel with the MatrixBasisChange node which has the sockets listed in the order: tangent, binormal, normal. I somehow think of normal more like Z direction, which is why I listed as the last input.

zeffii:
With nodes that are out of beta and alpha we have the policy that we avoid breaking existing .blends. Normally we only do an MK2 for substantial updates. It's up to you to decide whether the new socket's position is worth breaking existing .blends.  I don't know how many users' files that will impact, and really wish I could tell you.

from experience, take the easiest solution that breaks least. Only when you get pissed off enough should you do an MK2.

DolphinDream:
Fair enough. I reversed the order to keep the old blend files happy (tangents come after normals). I wish  there was an easy way though you could reorder sockets without breaking old blends :) (without going the MK# route).

PR #1811, opened 2017-10-10
adds alternative route for Text objects

zeffii:


zeffii:
this modification will only affect "FONT" objects, but other object types that don't offer `x.ray_cast` can be added in the check too.

zeffii:
not sure if this needs a test to remove the `NOB`, so there's no dangling objects..
```python
if OB.type == 'FONT':
    del NOB
```

Issue #1755, opened 2017-08-19
vdmk2 not being flushed during animation updates


zeffii:
mentioned here : 
https://github.com/nortikin/sverchok/pull/1754

The memory usage of vdmk2 goes up and up during uninterrupted animation. I suspect a more rigorous use of .GLflush may be needed.

-----

it should perhaps be mentioned that the vdmk2 is designed to optimally display static mesh data, by caching drawing calls once as a function, and calling that function repeatedly when a view update is called (for example: rotating a view around an unchanging object). This works great as long as the mesh isn't continuously updating/recaching. There's a small computation price to pay for caching the initial draw call (slower by `1.2 * non cached speed`), but redrawing from the cached function is about 10 times faster than non-cached.

In the scenario where animation comes in, and you intentionally cause mesh updates then you'd expect constant slower drawing (because it never uses the display list / caching feature )

DolphinDream:
Any ideas if this can be easily fixed ? For some node-settings I get huge memory consumption making the SV animations with viewer draw a bit prohibitive (I need to restart blender often to clear the memory)

zeffii:
at the moment the bmesh viewer might be a good fit? at least it's output will render anti-aliased if you an openGL render animation. (definitely not the case with vdmk2)

i've not thought much about this, but i'll poke around in it today.  it's a fundamental node and requires more cautious update policy :)

zeffii:
I think this is the section that needs modification (it's been a long time since we really manipulated the opengl drawlist code...)

https://github.com/nortikin/sverchok/blob/a0fe08cb430fe403f0267231d066dc0dfce6d797/ui/viewer_draw_mk2.py#L420-L449

zeffii:
it may have to be an option that you would enable if you are doing an animation, there's a chance that the option won't be the most efficient drawing solution for regular ui interaction. (tho ...we'll have to see .. but that's a suspicion of mine)

DolphinDream:
I like to do animation with vdmk2 first before creating any mesh since i believe it is faster. Once i’m satisfied with the way the animation looks of course i have to use the bmesh node to make it renderable. Which btw, I wish there was a way I could render the SV output the vdmk2 creates without having to create a mesh/materials first :)

zeffii:
> since i believe it is faster

[...]

PR #1814, opened 2017-10-15
finally reuse existing lists, reuse works in-place-of deletelists too

zeffii:


DolphinDream:
Sweet ! :) no more leaks.. blender has 108MB constant usage for 10 min now :) 

At least using the same node tree I posted in #1755 

gist: https://gist.github.com/8c48cf8455888e021a27eacef59e9deb

zeffii:
thanks @DolphinDream for noticing and reporting it

Issue #1295, opened 2017-03-03
Is there a node to create both inset & outset ?


DolphinDream:
The inset node can create inset or outset but not both. I’m after turning a circle into a circle band (centered ON the circle line). Basically, for the “flower of life" patterns (generated with the hexa grid) I would like to create thick circle lines that I can render. I don’t want to skinny the mesh which makes a tube around the edges since I don’t care about the 3D lines (and i want to keep the triangle count to a minimum); I only need to render the 2D circle patterns. So, I thought that I could use a node like inset to do the job, but the inset moves the circle radius either in or out, which is why i need the inset node to do both In  & Out around the edge (like in the image below).  I could augment the Inset node to do this if that’s ok. And if the node is updated I also think that the “mask inner” should be made an input socket in its full rights so I don’t have to attach an int node just to set it. Any thoughts ?

The node tree in the screenshot does not actually solve my problem; just illustrating the outcome i’m after.
![sv-insetoutset-example](https://cloud.githubusercontent.com/assets/2083719/23535515/bd493108-ff8c-11e6-9c90-53adf1368a08.png)

zeffii:
1. I agree inset specials should have those properties on the node, i'll add them.
2. Augmenting the node? Make it a separate node first. 

Also see 
![image](https://cloud.githubusercontent.com/assets/619340/23543805/2dc055d0-fff4-11e6-9324-a7523a0a7ab0.png)

- consider adding a `2d_donut` to `utils.geom` ( closely related function is _arc_slices_ ) , 
- consider a 2d mode to the donut node :) so we pass it only radius and thickness and it outputs VEP.

A while back i sank some time into a generalized 2d solution to go from edgenet to edgenet with thickness, it was never completed - i might have another stab at it :)  https://github.com/nortikin/sverchok/issues/623

![img](https://cloud.githubusercontent.com/assets/619340/7418615/61d52cca-ef6f-11e4-8a2c-864d76b22a1c.png)

zeffii:
![image](https://cloud.githubusercontent.com/assets/619340/23544336/e8bcc11e-fff6-11e6-8038-91d9299404ef.png)
sadly this only outputs one set of VEP, a dedicated 2d donut is definitely on the cards (if you won't, I will )

zeffii:
... or a circle MK2 . with thickness value, thickness 0 just outputs as normal, any positive thickness value would generate the 2d donut.

zeffii:
[...]

PR #1815, opened 2017-10-16
Add new spiral node (WIP)

DolphinDream:
Starting a new Spiral node. (anyone missed such a node ? :)

Currently this includes a few spiral types: Archimedean, Logarithmic, Spherical, Cornu and one of my own making called Exo (double exponential spiral between two circles). The spherical spiral is the only 3D spiral so far but I’ll be extending these to 3D later on to create helixes and such. Any suggestions are welcome. 

sv-spiralnode-1

### DIFF
--- a/docs/nodes/beta_nodes/spiralNode.rst
+++ b/docs/nodes/beta_nodes/spiralNode.rst
@@ -1,0 +1,198 @@
+Spiral
+========
+
+Functionality
+-------------
+
+Generates various types of spirals: Archimedean, Logarithmic, Spherical, Cornu, Ovoidal, Exo (to coin the term) and Spirangle.
+
+All spirals are defined as smooth curves (given enough curve resolution per turn) except the Spirangle spiral, which is a polygonal type of spiral.
+
+Most of the spirals are 2D and some are 3D, but any of the 2D spirals can be turned into 3D by setting the height parameter to a non-zero value.
+
+Inputs
+------
+
+All inputs are vectorized and they will accept single or multiple values.
+
+- **Exterior Radius** [1]
+- **Interior Radius** [1]
+- **Exponent** [2]
+- **Turns**
+- **Turn Resolution**
+- **Scale** [3]
+- **Height** [4]
+- **Phase**
+- **Arms**
+
+Notes:
+[1] : For some spirals the Exterior/Interior radii are repurposed to represent other spiral parameters like spherical (3D) radius, spiral length (Cornu) etc.
+[2] : For some spirals the exponent setting is repurposed to alter spiral shape in some way.
+[3] : For some spirals the scale settings affect xy radii, while for others it scales xyz radii.
+[4] : For some spirals the height setting spreads the spiral along z, for other it is ignored (e.g. spherical spiral has a predefined height to define a sphere)
+
+
+Parameters
+----------
+
+All settings except **Type**, **Flip** and **Preset** can be given by the node or an external input.
+
++----------------------+--------------+-------------+----------------------------------------------+
+| Param                | Type         | Default [1] | Description                                  |
++======================+==============+=============+==============================================+
+| **Spiral Type**      | Enum:        | Archimedean | Type of the spiral. Each type is defined by  |
+|                      |  Archimedean |             | a unique formula (See "Spiral Type" section  |
+|                      |  Logarithmic |             | for details).                                |
+|                      |  Spherical   |             |                                              |
+|                      |  Ovoidal     |             |                                              |
+|                      |  Cornu       |             |                                              |
+|                      |  Exo         |             |                                              |
+|                      |  Spirangle   |             |                                              |
++----------------------+--------------+-------------+----------------------------------------------+
+| **Exterior Radius**  |  Float       |   0.1       | Exterior radius of the spiral.               |
++----------------------+--------------+-------------+----------------------------------------------+
+| **Interior Radius**  |  Float       |   0.0       | Interior radius of the spiral.               |
++----------------------+--------------+-------------+----------------------------------------------+
+| **Exponent**         |  Float       |   1.0       | Exponent, attenuator, slope.                 |
++----------------------+--------------+-------------+----------------------------------------------+
+| **Turns**            |  Int         |   7         | Number of turns in the spiral (>= 1).        |
++----------------------+--------------+-------------+----------------------------------------------+
+| **Turn Resolution**  |  Int         |   100       | Number of points in one turn (>= 3).         |
++----------------------+--------------+-------------+----------------------------------------------+
+| **Scale**            |  Float       |   1.00      | Scale the radii of the spiral.               |
++----------------------+--------------+-------------+----------------------------------------------+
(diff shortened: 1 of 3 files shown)


DolphinDream:
Any suggestions for what’s the best way to render lines ? I tried skin mesh or curve viewer but this makes the animation way less interactive. I wish there was a node to create an inset/outset around a line :) https://github.com/nortikin/sverchok/issues/1295 to create a very simple mesh (2D would do for now).

zeffii:
regarding "best way" to render lines.

The options for live / interactive rendering aren't really fantastic at the moment. I'll stress that this isn't an area we've considered imperative to fix either, with 2.8 around the corner and new bgl/gawain paradigms this may change.

- outputting verts+edges to a BMeshViewer gives reasonable speed, and can be rendered using the Animation OpenGL option (A Blender feature)
- _Curve Viewer_ or _PolyLine Viewer_ will work as they just push data into objects of type Curve, there may be some slowdown but if you are rendering this won't be significant compared to the rendertime of a frame.

an inset/outside node might help you, but will involve drawing more than double the amount of geometry (and that's a tick in the slowdown category)

DolphinDream:
A little preview .. into the Ovoidal spirals. I call this “The Spinning Duality”  :)

sv-spiralnode-demo1

DolphinDream:
First round of development is coming to an end (to be merged in shortly). Here is the latest update of the UI and the types of spirals currently implemented.

sv-spiralnode-ui3

sv-spiralnode-collage2

DolphinDream:
And an interactive demo of the types/presets. 
(the node has a list of presets with various settings for each type of spiral).

sv-spiralnode-demo5

zeffii:
I hope to one day find a use for it :)

DolphinDream:
@zeffii I doubt you will.. unless you go camping 😉 
(btw, are you hinting that I shouldn’t have merged this node in?)

zeffii:
this node is fine, just not something for me / can't promise i'll test it much. looks like good fun

enzyme69:
I want this Spiral node :D will use it for Live Noding.

DolphinDream:
@enzyme69 It’s merged in.. Please keep in mind though that I kept this as a beta node for now since I’m considering making further adjustments to it. Feedback is always appreciated 👍

DolphinDream:
Settings demo - Archimedean spiral

sv-spiralnode-demo6-archimedean

DolphinDream:
Settings demo : Exo spiral

sv-spiralnode-demo6-exo

DolphinDream:
Settings demo - Ovoidal spiral:

Note: the eR/iR are not quite intuitive for the moment (and the height is ignored). Next update will make the iR change the horizontal cross section radii without changing the height of the spiral and the eR/H will both affect the height and vertical cross section radius.  

sv-spiralnode-demo6-ovoidal

DolphinDream:
Some fun examples.. :) two seven arm, counter rotating fibonacci vortices.. opening a portal between Sun and Earth 

sv-spiralnode-demo7-a
sv-spiralnode-demo7-b

enzyme69:
Very cool! By the way, I am still not getting push update. Wonder why...

zeffii:
@enzyme69  assuming you are using git...
```git
git checkout master
git pull
```

enzyme69:
Ooh... "Show Latest Commit" does the job!

Cheers for the Git tips again, Zeff!

enzyme69:
screen shot 2017-10-31 at 22 02 31

DolphinDream:
@zeffii btw.. if i have incremental changes to this node do I need to create a PR or should i just merged them in ?

DolphinDream:
I meant to ask this a while back when i was working on the Torus Knot node.. for some values of the parameters the result output is a set of disjoint curves and if I use vectorization that creates different versions of the spiral each having multiple arms (thus disjoint curves) when i try to pozition them via Viewer Draw + Line, each individual spiral (arm) is being positioned on the line grid instead of each spiral (with all of its arms). Any ideas how to go about this issue? Do i need to create a tripple nested lists or something ? Does the Bmesh viewer or Viewer Draw node know how to handle tripple nested lists this way ?

DolphinDream:
Some more ovoidal spiral fun demos: “The Flux Capacitor” :)

sv-spiralnode-demo8

Durman:
I have not well understood your problem but if we want to create some structure inside object we can use mask list. For example:

list of vectors - [[(V1),(V2),(V3)],...]
mask - [[0,0,1],...]

Where zero will select vectors V1 and V2, one will select vector V3.
In this case we can receive more difficult structure without triple nested list of vectors.

zeffii:
> @zeffii btw.. if i have incremental changes to this node do I need to create a PR or should i just merged them in ?

because it's a beta node you can just push, it's up to the users to avoid these nodes in critical works.

> . for some values of the parameters the result output is a set of disjoint curves and if I use vectorization that creates different versions of the spiral each having multiple arms (thus disjoint curves) when i try to position them via Viewer Draw + Line, each individual spiral (arm) instead of each spiral (with all of its arms) is being positioned on the line grid. Any ideas how to go about this issue?

I'm not sure I follow, show images.

DolphinDream:
@zeffii Here’s a setup. I have a spiral setup with 2 arms (arms are separate lists in the spiral) and I create (vectorize) versions with 1,2,3,4 turns. The purple draw shows all of them overlapping. the blue draw shows the spread/moved at different locations (via line). As you can see the location of each (separate) arm is distributed at the given line locations instead of each spiral as a whole (with both arms). What I want of course is to put 2-arm spiral with 1 turn at L1, 2-arm spiral with 2turns at L2 etc. I know i can do all sorts of grouping and merging with other nodes but I was wondering if there is something simpler I can do within the node so that this setup just works (e.g. Viewer Draw to just know to put the first to lists at L1, next two lists at L2 etc).

gist: https://gist.github.com/6b4db73d06abc30c0501f33b5fabe5e3

sv-multiarmspiralreplication

zeffii:
I see. Let's reduce the example to 4... instead of 8

When a node outputs `verts. 4` in my opinion it should really represent 4 sets of verts, each part of an individual "object" `[verts, verts, verts, verts]`. At present we don't really have a way for the list of verts to carry metadata about itself to say "this `verts` is made up out of this number of disjoint parts, starting by index 0, 12, 23, 45".. Mostly this is because it's an extra layer of complication.

You came up with a solution, an output mast lists. it would output something like
```
output_meta_data = [[0,0,0,0,0,0,0,0],[0,0,0,0,0,0,1,1,1,1,1,1,1],[0,0,0,0,0,0,1,1,1,1,1,1,2,2,2,2,2,2,2]......]
```
where each vert inside a list of verts is indexed by a mask. That takes the complexity of dealing with the sublists out of the node, and might require a matrix apply node with understanding of the masking. And if that's workable, new nodes that need this sub-list-indexing can use the same scheme. (like Generator/Cylinder' s "separate" functionality )

nortikin:
just WOWOWOW!!!
Amazing work.

0000888888880000:
GNP

Issue #1817, opened 2017-10-16
Curve In node


zeffii:
`Object In` can already load Curve data, but strips all data except points.

As @Durman [alludes to in this thread](https://github.com/nortikin/sverchok/issues/1816) there's a weakness in what we offer regarding Curve Object data. I'd like to offer extra introspection into the spline data, there's a lot we can add. Some of this data can only be [conveniently accessed](https://gist.github.com/zeffii/90fa1a83ce821d6ae03a71ae91e7cdd5) via scripts at this point.
```
Bezier Curve
    - points
    - num points per subcurve (set this property)
    - radius at points ( interpolate according to `LINEAR`..etc )
Spline Curve
    - points
    - num points per subcurve (set this property)
    - radius at points ( interpolate according to `LINEAR`..etc )
optional 
    - edges 
         - inferred if no thickness :   connect between points, 
         - if thickneses then use obj.to_mesh
    - polygons output  (when thickness present this can use obj.to_mesh)
```
No ETA.

Durman:
I already have tried to receive radiuses of curve with script node.

```Python
if objs_in[0].data.splines[0].type == 'BEZIER':

    for i in objs_in[0].data.splines[0].bezier_points:
        bas_r.append(i.radius)
        bas_c.append(i.co)
else:
    for i in objs_in[0].data.splines[0].points:
        bas_r.append(i.radius)
        bas_c.append(i.co)
```
But i had problem with interpolation value of radius between intermediate points. I did not manage to receive exact distance between points by curve.

zeffii:
yeah, i think the radii must be calculated by hand for interpolated points - i'm not aware of any convenience methods for this.. that's OK though.

Durman:
Interpolation of radiuses i did thus: 
![2017-10-16_18-57-47](https://user-images.githubusercontent.com/28003269/31619396-751df662-b2a5-11e7-9163-ac63ac524324.png)
Where D axis is distance between points, R axis is radiuses of points, C1,2 is points of curve, Pn is interpolation points.
Pny = dyC2 * dyPn / dyC2 + C1y
But i do not know how to find distance between points on curve.

zeffii:
@Durman we can offer multiple radius interpolation approaches. I'll implement linear and maybe cubic. If you're interested you can look at the Blender source for the exact algorithms. I'm not ready to do this yet

zeffii:
anyone want to write a function that interpolations radii?
```python

[...]

PR #1818, opened 2017-10-16
Curve in

zeffii:


zeffii:
https://www.mvps.org/directx/articles/catmull/

zeffii:
or...maybe one of the various already implemented versions in SV.. [cubic](https://github.com/nortikin/sverchok/search?p=1&q=cubic&type=&utf8=%E2%9C%93)

zeffii:
catmul.. maybe.. just convert from konts to sections
```
def interpolate_catmul(knots, cyclic, num_segments):
    """
    from https://www.mvps.org/directx/articles/catmull/

    q(t) = 0.5 *( (2 * P1) +
                (-P0 + P2) * t +
                (2*P0 - 5*P1 + 4*P2 - P3) * t2 +
                (-P0 + 3*P1- 3*P2 + P3) * t3)
    """

    if cyclic:
        sections = [] 
    else:
        sections = []

    theta = 1 / num_segments
    for P0, P1, P2, P3 in sections:
        for xt in range(num_segments+1):
            t = theta * xt
            t2 = t*t
            t3 = t2*t
            radt = 0.5*((2*P1)+(-P0+P2)*t+(2*P0-5*P1+4*P2-P3)*t2+(-P0+3*P1-3*P2+P3)*t3)
```
if not cyclic and knots are 1.0  2.0  3.0
```
P0a=1.0; P1a=1.0; P2a=2.0; P3a=3.0  # notice the first knot is weighted twice
P0b=1.0; P1b=2.0; P2b=3.0; P3b=3.0
```
this would allow all calculations between 1.0....2.0,... and 2.0...3.0

zeffii:
```python
KNOTS1 = ["A", "B"]
KNOTS = ["A", "B", "C"]
KNOTS2 = ["A", "B", "C", "D"]

cyclic = False
for knots in [KNOTS1, KNOTS, KNOTS2]:

    sections = []
    if not cyclic:
        if len(knots) == 2:
            sections.append((0, 0, 1, 1))
        else:
            for i in range(len(knots)-1):
                if i == 0:
                    indices = i, i, i+1, i+2
                elif i == len(knots)-2:
                    indices = i-1, i, i+1, i+1
                else:
                    indices = i-1, i, i+1, i+2
                sections.append(indices)
    else:
        ...


    print(sections)
```

zeffii:
or with much fewer checks..
```python
def get_sections(knots, cyclic):
    sections = []
    if not cyclic:
        if len(knots) == 2:
            sections.append((0, 0, 1, 1))
        else:
            sections.append((i, i, i+1, i+2))
            for i in range(1, len(knots)-2):
                sections.append((i-1, i, i+1, i+2))
            sections.append((i-1, i, i+1, i+1))
    else:
        if len(knots) == 2:
            sections.append((1, 0, 1, 0))
        else:
            last_idx = len(knots)-1
            sections.append((last_idx, 0, 1, 2))
            for i in range(len(knots)):
                sections.append((i-1, i, i+1, (i+2) % len(knots)))
            sections.append((last_idx-1, last_idx, 0, 1))

    return sections
```

zeffii:
overlapping sections detected.. vertices affected, tho i think both radii calculation types are correct.

zeffii:
image

zeffii:
if cyclic mode, it still produces one too many radii

zeffii:
all my goals are completed here, i guess tilt output could be easy enough..with sample interpolation options

zeffii:
these two are equivalent
```python
def get_sections(knots, cyclic):
    sections = []
    if not cyclic:
        if len(knots) == 2:
            sections.append((0, 0, 1, 1))
        else:
            for i in range(len(knots)-1):
                if i == 0:
                    indices = i, i, i+1, i+2
                elif i == len(knots)-2:
                    indices = i-1, i, i+1, i+1
                else:
                    indices = i-1, i, i+1, i+2
                sections.append(indices)
    else:
        if len(knots) == 2:
            sections.append((1, 0, 1, 0))
        else:
            last_idx = len(knots)-1
            for i in range(len(knots)):
                if i == 0:
                    indices = last_idx, 0, 1, 2
                elif i == last_idx:
                    indices = last_idx-1, last_idx, 0, 1
                else:
                    indices = i-1, i, i+1, (i+2) % len(knots)
                sections.append(indices)

    return sections


def get_sections(knots, cyclic):

    # end early
    if len(knots) == 2:
        return [(0, 0, 1, 1) if cyclic else (1, 0, 1, 0)]

    sections = []

    if not cyclic:
        j = len(knots)-2 # second_last_idx 
        sections.append((i, i, i+1, i+2))
        for i in range(1, len(knots)-2):
            sections.append((i-1, i, i+1, i+2))
        sections.append((j-1, j, j+1, j+1))
    else:
        last_idx = len(knots)-1
        sections.append((last_idx, 0, 1, 2))
        for i in range(len(knots)):
            sections.append((i-1, i, i+1, (i+2) % len(knots)))
        sections.append((last_idx-1, last_idx, 0, 1))

    return sections
```
i think

portnov:
screenshot_20171203_163653

zeffii:
yep it should perhaps use the `radius` + `bevel width` as the final float output, right now I think it only outputs the calculated radius.

Durman:
It would be not bad to work with several curves in one object. Different curves should be separated to different objects or maybe better to have mask of vertices such as [0,0,0,1,1,1,2,2,2] where 0,1,2 means different objects.

several curves
#1817

Durman:
How it due be as i think:

work of node

PR #1822, opened 2017-10-20
Execnode mod copy

zeffii:


zeffii:
maybe the get text from Text datablock operation should print/report something if `self.text` hasn't been set by the user.

nodes/script/multi_exec.py
```diff
@@ -143,9 +146,25 @@ def insert_blank(self, context):
...
 
     def copy_from_text(self, context):
-        for i, i2 in zip(self.dynamic_strings, bpy.data.texts[self.text].lines):
+        """ make sure self.dynamic_strings has enough strings to do this """
+        slines = bpy.data.texts[self.text].lines
+        num_str_lines = len(slines)
+        if len(self.dynamic_strings) < num_str_lines:
+            num_to_add = num_str_lines - len(self.dynamic_strings)
+            for _ in range(num_to_add):
+                self.dynamic_strings.add()
```
zeffii:
```python
def copy_from_text(self, context):
    slines = bpy.data.texts[self.text].lines
    while (len(self.dynamic_strings) < len(slines)):
        self.dynamic_strings.add()

    for i, i2 in zip(self.dynamic_strings, slines):
          i.line = i2.body

```

PR #1825, opened 2017-10-22
Update proportional.py

Durman:
If all values is False in mask output list is zeros.

### DIFF
--- a/nodes/analyzer/proportional.py
+++ b/nodes/analyzer/proportional.py
@@ -105,19 +105,24 @@ class SvProportionalEditNode(bpy.types.Node, SverchCustomTreeNode):
 
             # build KDTree
             base = [v for v, mask in zip(vertices, masks) if mask]
-            tree = kdtree.KDTree(len(base))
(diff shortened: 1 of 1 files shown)


nodes/analyzer/proportional.py
```diff
@@ -105,19 +105,24 @@ def process(self):
...
+                coeffs = []
+                for vertex, mask in zip(vertices, masks):
+                    if mask:
+                        coef = 1.0
+                    else:
+                        _, _, rho = tree.find(vertex)
+                        coef = self.falloff(radius, rho)
+                    coeffs.append(coef)
+
+            else: coeffs = [0 for _ in masks]
```
zeffii: make it 
```python
    else:
        coeffs = [0.0 for _ in masks]
```
we keep the else and the code it executes on different lines (per convention)

for consistency i'd user `0.0` as the coefficient

zeffii:
should be OK now. i've committed the pull request (PR)

PR #1830, opened 2017-10-31
Matrix math node (WIP)

DolphinDream:
Started a Matrix Math node.. currently **Multiply**, **Invert** and **Filter** operations are implemented (WIP).

Filter operation allows you filter out the Translation, Rotation and/or Scale components of the 4x4 homogeneous matrix.

Note: UI is not cleaned up yet and other operations may be added if needed. Feedback is welcomed as usual.

[will post some demos soon]

sv-matrixmathnode-ui1

### DIFF
--- a/docs/nodes/matrix/matrix_math.rst
+++ b/docs/nodes/matrix/matrix_math.rst
@@ -1,0 +1,81 @@
+Matrix Math
+===========
+
+Functionality
+-------------
+
+Matrix Math node allows for various matrix operations to be performed on the input matrices, such as: Multiply, Invert and Filter.
+
+Inputs
+------
+
+All inputs are vectorized and they will accept single or multiple values.
+
+- **A**
+- **B**  [1]
+
+Notes:
+[1] : The second input is only available for the operations that require a second operand, like Multiply.
+
+Parameters
+----------
+
+The **Operation** parameter allows to select one of following operations: Multiply, Invert, Filter.
+
+All parameters except **Operation**, **PrePost** and **Filter T/R/S** can be given as an external input.
+
++---------------+------------+----------+--------------------------------------------------+
+| Param         | Type       | Default  | Description                                      |
++===============+============+==========+==================================================+
+| **Operation** | Enum:      | Multiply | Multiply: A,B => C = A*B  [1]                    |
+|               |  Multiply  |          |   Invert: A => C = A^-1                          |
+|               |  Invert    |          |   Filter: A => C = [AT]*[AR]*[AS]                |
+|               |  Filter    |          |                                                  |
++---------------+------------+----------+--------------------------------------------------+
+| **PrePost**   | Enum:      | Pre      | Determines the order the operands  [2]           |
+|               |  Pre       |          |                                                  |
+|               |  Post      |          |                                                  |
++---------------+------------+----------+--------------------------------------------------+
+| **Filter T**  | Bool       | False    | Filter out the Translation component  [3]        |
++---------------+------------+----------+--------------------------------------------------+
+| **Filter R**  | Bool       | False    | Filter out the Rotation component  [3]           |
++---------------+------------+----------+--------------------------------------------------+
+| **Filter S**  | Bool       | False    | Filter out the Scale component  [3]              |
++---------------+------------+----------+--------------------------------------------------+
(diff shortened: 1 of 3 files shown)


DolphinDream:
Filter / Multiply demo. 

The tree has 3 matrix in that are composed in various orders. The first matrix in randomizes the location and rotation, the other two have a frame driven rotation with some fix translation and scales. The second level of nodes are Matrix Filter nodes that allows to turn on/off certain components of the matrix in nodes.. Then Multiply M1,M2,M3 in different orders (pre/post multiply). It’s fun to see that pre multiply these matrices make the objects flow almost like a school of fish ;) 

sv-matrixmath-demo1

DolphinDream:
Anybody else tired of using Matrix Apply to do matrix composition  ? :) 
I think it’s even faster to compose the matrices first and then apply the result to a mesh.

DolphinDream:
I thought of adding (based on prior discussions) to add a Determinant, Add, Subtract, Transpose operations, though I can’t think of a use for any of these. I don’t know what any of these mean in the context of a homogeneous 4x4 matrix (that has translation, rotation and scale in one). Any clues?

zeffii:
i'm all for this, Matrix Composition is not super convenient in Sverchok.. i'd also like to see a node that can take an arbitrary sequence of operations like..
````
rotate -> scale -> translate -> rotate
```
to reduce node-spaghetti more. Exactly how the UI for such a node would be implemented is up for debate. but maybe not in this PR..

DolphinDream:
Should this node be renamed **Matrix Composition** then? And reserve the matrix Math node for all sorts of other math related operations relevant to matrix in general (necessarily homogeneous matrices)

zeffii:
Matrix composer.. I dunno it can be tricky to name them sometimes. Especially nodes that do a wiďe range of operations.

DolphinDream:
Updated the sockets based on selected operation (add/remove input B).

sv-matrixmath-demo1

DolphinDream:
> i’m all for this, Matrix Composition is not super convenient in Sverchok.. i’d also like to see a node that can take an arbitrary sequence of operations like.. rotate -> scale -> translate -> rotate

@zeffii Are you thinking of a node that takes an arbitrary number of matrix inputs similar to the **List Join** and performs matrix composition on them ? I don’t know if we could differentiate between what is a **translation**, **rotation** and **scale**, so I guess it would be up to you to plug the TRS in the order you want them to be multiplied.

DolphinDream:
An idea would be (if possible / advisable) is to have a matrix math node that takes an arbitrary number of matrices (matrix format) (similar to how the **List Join** node creates new sockets) but also allow each socket input to be reordered (e.g. an up/down arrow next to each socket to allow them to be reordered).

sv-matrixmath-socketreorder

another variation (with up/down buttons next to the sockets): 

sv-matrixmath-socketreorder2

I think this could be easily implemented in the lower level socket code (node_tree) the same way the socket expander is handled.  And it would basically swap the inputs to the sockets to avoid having to manually reconnect two inputs. I’m not sure if this would work in general or only for sockets of the same type though.

DolphinDream:
Latest updated of the node UI: (I dropped the “operation” label since it’s cropped and it’s pretty obvious what that setting really is; similar to the Math MK2 node UI). 

sv-matrixnode-ui2

DolphinDream:
Brainstorming on the “arbitrary number of input matrix sockets” topic.. while this would make sense for Matrix Multiplication operation it doesn’t make much sense for the other two (invert/filter), unless.. we want that invert/filter operations to act on ALL inputs the same way and output a join list of all inputs with the operation applied to them (e.g. if 4 inputs are connected, the invert/filter will generate a list of 4 inverted/filtered matrices).

DolphinDream:
If there are no objections i’ll merge this into master for now... and continue developing while in beta to explore the multiple input idea in a separate PR.

zeffii:
> so I guess it would be up to you to plug the TRS in the order you want them to be multiplied.

yeah, i'd settle for a vector input version of that ..

DolphinDream:
Do you suggest a vector input into the matrix math node? How would the node know that the input is a T, R or S ? And actually R cannot be a 3D vector unless an angle is also provided.  A 4D vector perhaps could be interpreted as a rotation axis + an angle.

PR #1847, opened 2017-11-07
development_io (quick_link, expander... socket props and support in IO/JSON)

zeffii:
Things that are still issues before hitting master

- [x] socket custom props are stored, but isn't sparse.
- [x] nodule_color  (but only used in `color-nodes` for a limited period from this point...and `gp node`
- [x] restore custom_draw
- [x] correctly store use_custom_color

condensed changelog.

* Return zero vector for no input in VerticesSocket
* Add Sv prefix to Color and Quaternion socket classes
* Add "link matrix" UI/operator for MatrixSocket
* Add "link vector" UI/operator for VerticesSocket
* Add quaternion to matrix socket conversion
* Update collapse expander for color to use SvColorSocket name
* Add matrix to quaternion socket conversion
*Add draw_link_new_node to socket base (common) to allow various inputs to add button to spawn and link different input nodes.
* Minor class/idname changes to the link new node
* Add "use quicklink" flag (To allow the quick link feature to be turned on/off if needed.)
* Socket specific draw calls were replaced by the draw call in the base class.
* adjust the quicklink new node offset to avoid overlapping new nodes (new nodes are now offset based on their index).
* Add code to handle export/import of socket custom props
* Optimization to `draw_expander_template`

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -505,7 +505,7 @@ class StringsSocket(NodeSocket, SvSocketCommon):
 
     prop_type = StringProperty(default='')
     prop_index = IntProperty()
-    nodule_color = FloatVectorProperty(default=(0.6, 1.0, 0.6, 1.0), size=4)
+    nodule_color = FloatVectorProperty(default=socket_colors["StringsSocket"], size=4)
 
     custom_draw = StringProperty()
 
@@ -534,6 +534,9 @@ class StringsSocket(NodeSocket, SvSocketCommon):
             return default
         else:
             raise SvNoDataError(self)
+
+    def draw_color(self, context, node):
+        return self.nodule_color
 
 
 class SvLinkNewNodeInput(bpy.types.Operator):
--- a/node_tree.py
+++ b/node_tree.py
@@ -85,7 +85,6 @@ emptyQuaternion = [[(1, 0, 0, 0)]]
 def process_from_socket(self, context):
     """Update function of exposed properties in Sockets"""
     self.node.process_node(context)
-# this property group is only used by the old viewer draw
 
 class SvDocstring(object):
     """
@@ -188,6 +187,7 @@ class SvDocstring(object):
         else:
             return self.docstring.strip()
 
+# this property group is only used by the old viewer draw
 class SvColors(bpy.types.PropertyGroup):
     """ Class for colors CollectionProperty """
     color = FloatVectorProperty(
--- a/node_tree.py
+++ b/node_tree.py
@@ -86,6 +86,7 @@ def process_from_socket(self, context):
     """Update function of exposed properties in Sockets"""
     self.node.process_node(context)
 
+
 class SvDocstring(object):
     """
     A class that incapsulates parsing of Sverchok's nodes docstrings.
@@ -157,7 +158,7 @@ class SvDocstring(object):
         """
         Get shorthand to be used in search menu.
         If fallback == True, then whole docstring
-        will be returned for case when we can't 
+        will be returned for case when we can't
         find valid shorthand specification.
         """
 
@@ -171,7 +172,7 @@ class SvDocstring(object):
             return self.docstring
         else:
             return None
-    
+
     def has_shorthand(self):
         return self.get_shorthand() is not None
 
@@ -188,6 +189,8 @@ class SvDocstring(object):
             return self.docstring.strip()
 
 # this property group is only used by the old viewer draw
+
+
 class SvColors(bpy.types.PropertyGroup):
     """ Class for colors CollectionProperty """
     color = FloatVectorProperty(
@@ -326,6 +329,7 @@ class SvSocketCommon:
 
     def draw_color(self, context, node):
         return socket_colors[self.bl_idname]
+
 
 class MatrixSocket(NodeSocket, SvSocketCommon):
     '''4x4 matrix Socket type'''
@@ -641,7 +645,6 @@ class SverchCustomTree(NodeTree, SvNodeTreeCommon):
     sv_process = BoolProperty(name="Process", default=True, description='Process layout')
     sv_user_colors = StringProperty(default="")
 
-
     def update(self):
         '''
         Tags tree for update for handle
@@ -763,7 +766,7 @@ class SverchCustomTreeNode:
         """
 
         return cls.get_docstring().get_tooltip()
-    
+
     @classmethod
     def get_shorthand(cls):
         """
@@ -801,7 +804,6 @@ class SverchCustomTreeNode:
             # print('applying default for', self.name)
             set_defaults_if_defined(self)
 
-
     def process_node(self, context):
         '''
         Doesn't work as intended, inherited functions can't be used for bpy.props
@@ -837,7 +839,6 @@ classes = [
 def register():
     for cls in classes:
         bpy.utils.register_class(cls)
-
 
 
 def unregister():
--- a/node_tree.py
+++ b/node_tree.py
@@ -75,12 +75,6 @@ socket_colors = {
     "TextSocket": (0.68, 0.85, 0.90, 1),
 }
 
-# default values returned when no input is connected to socket
-identityMatrix = [[tuple(v) for v in Matrix()]]
-emptyVertex = [[(0, 0, 0)]]
-emptyColor = [[(0, 0, 0, 1)]]
-emptyQuaternion = [[(1, 0, 0, 0)]]
-
 
 def process_from_socket(self, context):
     """Update function of exposed properties in Sockets"""
@@ -404,7 +398,7 @@ class VerticesSocket(NodeSocket, SvSocketCommon):
         elif self.use_prop:
             return [[self.prop[:]]]
         elif default is sentinel:
-            return emptyVertex
+            raise SvNoDataError(self)
         else:
             return default
 
@@ -445,7 +439,7 @@ class SvQuaternionSocket(NodeSocket, SvSocketCommon):
         elif self.use_prop:
             return [[self.prop[:]]]
         elif default is sentinel:
-            return emptyQuaternion
+            raise SvNoDataError(self)
         else:
             return default
 
@@ -477,7 +471,7 @@ class SvColorSocket(NodeSocket, SvSocketCommon):
         elif self.use_prop:
             return [[self.prop[:]]]
         elif default is sentinel:
-            return emptyColor
+            raise SvNoDataError(self)
         else:
             return default
 
--- a/node_tree.py
+++ b/node_tree.py
@@ -277,6 +277,7 @@ class SvSocketCommon:
                 layout.template_component_menu(prop_origin, prop_name, name=self.name)
 
     def draw_quick_link(self, context, layout, node):
+
         if self.use_quicklink:
             if self.bl_idname == "MatrixSocket":
                 new_node_idname = "SvMatrixGenNodeMK2"
@@ -293,6 +294,14 @@ class SvSocketCommon:
             op.new_node_offsety = -30 * self.index
 
     def draw(self, context, layout, node, text):
+
+        # just handle custom draw..be it input or output.
+        # hasattr may be excessive here
+        if self.bl_idname == 'StringsSocket' and hasattr(self, 'custom_draw'):
+            if self.custom_draw and hasattr(node, self.custom_draw):
+                getattr(node, self.custom_draw)(self, context, layout)
+                return
+
         if self.is_linked:  # linked INPUT or OUTPUT
             info_text = text + '. ' + SvGetSocketInfo(self)
             info_text += self.extra_info
--- a/node_tree.py
+++ b/node_tree.py
@@ -255,24 +255,27 @@ class SvSocketCommon:
         return ""
 
     def draw_expander_template(self, context, layout, prop_origin, prop_name="prop"):
-        if self.use_expander and self.bl_idname != "StringsSocket":
-            split = layout.split(percentage=.2, align=True)
-            c1 = split.column(align=True)
-            c2 = split.column(align=True)
-            if self.expanded:
-                c1.prop(self, "expanded", icon='TRIA_UP', text='')
-                c1.label(text=self.name[0])
-                c2.prop(prop_origin, prop_name, text="", expand=True)
-            else:  # collapsed
-                c1.prop(self, "expanded", icon='TRIA_DOWN', text="")
-                row = c2.row(align=True)
-                if self.bl_idname == "SvColorSocket":
-                    row.prop(prop_origin, prop_name)
-                else:
-                    row.template_component_menu(prop_origin, prop_name, name=self.name)
+
+        if self.bl_idname == "StringsSocket":
+            layout.prop(prop_origin, prop_name)
         else:
-            if self.bl_idname == "StringsSocket":
-                layout.prop(prop_origin, prop_name)
+            if self.use_expander:
+                split = layout.split(percentage=.2, align=True)
+                c1 = split.column(align=True)
+                c2 = split.column(align=True)
+
+                if self.expanded:
+                    c1.prop(self, "expanded", icon='TRIA_UP', text='')
+                    c1.label(text=self.name[0])
+                    c2.prop(prop_origin, prop_name, text="", expand=True)
+                else:
+                    c1.prop(self, "expanded", icon='TRIA_DOWN', text="")
+                    row = c2.row(align=True)
+                    if self.bl_idname == "SvColorSocket":
+                        row.prop(prop_origin, prop_name)
+                    else:
+                        row.template_component_menu(prop_origin, prop_name, name=self.name)
+
             else:
                 layout.template_component_menu(prop_origin, prop_name, name=self.name)
 
--- a/node_tree.py
+++ b/node_tree.py
@@ -298,16 +298,18 @@ class SvSocketCommon:
         # just handle custom draw..be it input or output.
         # hasattr may be excessive here
         if self.bl_idname == 'StringsSocket':
-            if hasattr(self, 'custom_draw'):
-                if self.custom_draw and hasattr(node, self.custom_draw):
+            if hasattr(self, 'custom_draw') and self.custom_draw:
+
+                # does the node have the draw function referred to by 
+                # the string stored in socket's custom_draw attribute
+                if hasattr(node, self.custom_draw):
                     getattr(node, self.custom_draw)(self, context, layout)
                     return
 
             if node.bl_idname in {'SvScriptNodeLite', 'SvScriptNode'}:
-                if not self.is_output and not self.is_linked:
+                if not self.is_output and not self.is_linked and self.prop_type:
                     layout.prop(node, self.prop_type, index=self.prop_index, text=self.name)
                     return
-
 
         if self.is_linked:  # linked INPUT or OUTPUT
             info_text = text + '. ' + SvGetSocketInfo(self)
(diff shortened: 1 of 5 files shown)


zeffii:
not read for merging yet. i'm going to modularize the changes to `io_panel_tools`

resist modifying anything in the branch for a while. please.

zeffii:
am enjoying the convenience of the `quick_link`,  tho i think maybe in a later iteration of `matrix_in` i would add quick_link and expander on by default to the vector inputs,

zeffii:
https://gist.github.com/b96be887d2332db75edb8915c761f9aa

zeffii:
now... i have no idea if Monads will be stored correctly using this...

zeffii:
preliminary testing is OK. but don't feel comfortable pushing it to master yet. need more eyes/testers
@nortikin @portnov @Durman @enzyme69 @Kosvor2

zeffii:
```json
"0": {
          "expanded": false,
          "use_expander": true,
          "use_prop": false,
          "use_quicklink": true
        }
```
am tempted to make socket prop storage more sparse, and only store values that are not default. This means we would need to agree on the defaults being unchanged going forward.

DolphinDream:
> am tempted to make socket prop storage more sparse, and only store values that are not default. This means we would need to agree on the defaults being unchanged going forward.

I was just about to ask you this.. I noticed that gist export does only save the values for the props that deviate from the default values.. while I understand this can save space, the export integrity may be compromised if the defaults change (as you just noted). The question is: do we want to export the values that the defaults had at the export time or do we want to capture that some props were not changed  (thus had some default values) at export time and let the import set them to (hopefully) the same default values. I personally would vote for the former (save everything) but that may make the gists unnecessarily large. Just my 2cents.

DolphinDream:
Just a reminder (I made a comment in the PR #1561 a while back) that the StringSocket now relies on draw method factorized in the common (base) socket class and it does not have the “custom draw” anymore. If this is something needed for some nodes we’ll have to do something about it. I think one place where it was used was for the color in/out nodes, as these relied on a StringsSocket as the socket type (?). Now that we have a SvColorSocket we can update the color nodes to use this socket type instead. Additionally we could probably add a SvTextSocket if needed.

node_tree.py
```diff
@@ -116,6 +146,71 @@ def replace_socket(self, new_type, new_name=None):
         return the new socket, the old reference might be invalid"""
         return replace_socket(self, new_type, new_name)
 
+    @property
+    def extra_info(self):
+        # print("getting base extra info")
+        return ""
+
+    def draw_expander_template(self, context, layout, prop_origin, prop_name="prop"):
+        if self.use_expander and self.bl_idname != "StringsSocket":
```
DolphinDream: This could be further optimized by doing a test for StringSocket once:
```python
        if self.bl_idname == "StringsSocket":
            layout.prop(prop_origin, prop_name)
        else: # all the other socket types
            if self.use_expander:
                split = layout.split(percentage=.2, align=True)
                c1 = split.column(align=True)
                c2 = split.column(align=True)
                if self.expanded:
                    c1.prop(self, "expanded", icon='TRIA_UP', text='')
                    c1.label(text=self.name[0])
                    c2.prop(prop_origin, prop_name, text="", expand=True)
                else:  # collapsed
                    c1.prop(self, "expanded", icon='TRIA_DOWN', text="")
                    row = c2.row(align=True)
                    if self.bl_idname == "SvColorSocket":
                        row.prop(prop_origin, prop_name)
                    else:
                        row.template_component_menu(prop_origin, prop_name, name=self.name)
            else: # not using expander
                layout.template_component_menu(prop_origin, prop_name, name=self.name)

```

zeffii:
- Im against optimizations in this pass.. simply because the pr alreaady touches too much for comfort.
- ill look at custom draw.
- socket defaults couldbe stored in nodedict header once, only deviations per node/socket Need to be recorded

zeffii:
well.. custom_draw will need to be reinstated :) because I use it in `A Number`.
https://github.com/nortikin/sverchok/search?utf8=%E2%9C%93&q=custom_draw&type=

zeffii:
it should take prime-precedence, nothing else matters is `socket.custom_draw = <some string>`

zeffii:
somehow development_io is still tracking changes from DolphinDream/DevIO ? ..args.. not your fault.. but I can't handle these surprises :) .. now there's a conflict with my new edits..

zeffii:
not sure why GIT is so picky sometimes..

zeffii:
https://gist.github.com/4c883ec762d06868ee914e668893b8f4

DolphinDream:
About nodule color.. i made a PR #1848 to update color in/out to use the new SvColorNodes, so I don’t know if nodule color is really needed, unless it’s used for other nodes.

DolphinDream:
@zeffii  Btw, there seem to be replicated history on this branch now. Did you do a git pull/rebase after my comments updates? To cleanup you’ll need to do a git rebase -i master and get rid of the replicated commits (in the terminal since i don’t think github can do much here). I can clean it up for you if you want.

zeffii:
it's a dirty process, but 6 dupe commit messages doesn't bug me.

zeffii:
exactly. if this PR is squashed ( that was the intention anyway ) none of this matters..really.

zeffii:
as you can probably tell, I have limited patience for this. If you want to fix it as offered - go ahead

DolphinDream:
@zeffii you’re right. Don’t worry about it. I understand not everyone has patience with this git shenanigans. I pushed the cleaned up version to upstream anyway since I already did it at my end. You just need to bring your local developer_io up to date with the upstream developer_io. 

This should do it:

```
git checkout developer_io
git fetch upstream 
git pull --rebase upstream developer_io
```

zeffii:
just a `git pull --rebase` here, i think.

nortikin:
crazy stuff!!!
later on have to make generators insertion occurred with viz draw node insertion also. Or to quick link in output with viz node? What do you think?

zeffii:
Yeah or a kb combo operating on active node. But for now pull request is big enough

portnov:
Just occasionally noted that 2pt spline node seems to be not ready for this: it checks if inputs are linked.

zeffii:
@portnov feel free to relax `2pt spline` 's process function.

portnov:
Should I push change to development_io branch or should I create a separate PR?

zeffii:
i think for now keep everything focused on development_io branch, including individual node updates..

portnov:
I looked deeper into code and saw that the problem does not have anything to deal with current PR. It is just broken. So I decided to put it into separate PR... #1880.

zeffii:
Well, the question arises now what is more useful. A default value for unconnected sockets.. or a sane error message upon Exception raised by an unconnected socket.

@DolphinDream @portnov @nortikin 

resolving the conflict isn't hard, but what do we prefer/ personally I'm leaning towards the default value for the socket types where it makes sense. ( all? )

portnov:
IMHO the default value is good exactly where it makes sense for specific node :) In general:
* If the specific input should contain data for node to process (e.g. vertices to be transformed) - there is no sense in defaults, except maybe for specific cases
* If the specific input should contain some parameter of processing - then in most cases there should be some sane default value.
So imho this should be decided at each input, by providing or not the `default=` parameter to sv_get().

DolphinDream:
My 2¢ ..
I personally favor the default values in all instances. 

I suspect there is little to no harm in having a default value for most of the nodes and the only useful outcome I can think of for not having an _implicit_ or _explicit_ default value is so that the node can report an error (changing color) to alert the user that no output is generated unless all the requested inputs are connected. This may be a bit too restrictive and I find it a bit annoying in most cases I came across. Also, some nodes, for instance, (I don’t remember which ones) provide the identity matrix as the _explicit_ default value (set during the reading of the socket input) even though you could bring the same argument against using both an explicit or implicit default since the node doesn’t really transform to the input in a meaningful way if the _identity matrix_ is provided. However, as a counter argument, when a matrix input is expected, the identity matrix is a valid input so why not provide that as the default value (implicit or explicit). I’d say this discussion could be extended to every type of input (not just vertices and matrices): ints, floats etc and the question is : what makes the selected (coded) default values acceptable for any of the inputs? 

In any case, if we really want to shy away from the implicit default values we can at least offer the nodes the chance to _opt in_  or _opt out_ but first we would need to figure out what is the most frequent case where we would not want or want to have an implicit value and make either the opt-in or opt-out of the default the default :) The nodes can then be given the option to **opt in** via new Socket(...).use_default = True (assuming the default is to not use defaults) OR to **opt out** via new Socket(...).use_default = False (assuming the default is to use defaults). This adds to the complexity (not needed in my view) but it would offer some flexibility.

zeffii:
i'll side with  if  `default=(some_value)` then return the value, else raise the Exception.

zeffii:
the todo for this pull request relies on two things now
- resolve the conflict according to the agreed default/exception convention
- resolve the io/json storage of socket properties (it's almost there but i've not had head-space to tackle it)

possibly tomorrow, because this really should make it into master post-haste -- i think a lot of people will love it.

DolphinDream:
@zeffii I could cleanup the conflicts if you want and add the default/exception convention.

zeffii:
ok make it so @DolphinDream 

i'm not happy with the verbose socket-properties storage per node. that should be resolved before a merge..

utils/sv_IO_panel_tools.py
```diff
@@ -145,7 +142,41 @@ def get_superficial_props(node_dict, node):
...
+def collect_custom_socket_properties(node, node_dict):
+    print("** PROCESSING custom properties for node: ", node.bl_idname)
+    input_socket_storage = { }
+    for socket in node.inputs:
+
+        # if not tracked_socket(socket): continue
+        print("Socket %d of %d" % (socket.index+1, len(node.inputs)))
+
+        storable = {}
+        tracked_props = 'use_expander', 'use_quicklink', 'expanded', 'use_prop'
```
zeffii: somewhere here you should be able to check against   the socket.<properties>.default , and store only if the current value is modfied.  like... `socket.bl_rna.properties[property_name].default == socket.property_name`  @DolphinDream

zeffii: socket.bl_rna.properties[property_name].default == getattr(socket, property_name)

DolphinDream:
Okidoki.. I rebased the branch on top of latest master and resolved conflicts (so please **reset** your local branch to its upstream instead of git pull otherwise with merge it will necessarily duplicate commits). On top I added more commits to fix the latest issues with this PR: put back nodule color, raise exception if no default and save property to io only if it deviates from the default.

Of course, this needs to be reviewed and tested (@nortikin, @zeffii, @portnov).

```
git fetch upstream 
git checkout development_io
git reset --hard upstream/development_io
```

or.. (if you dont’ have any local changes to this branch) just delete it and check it out again.

```
git checkout master
git branch -d development_io
git checkout developent_io
```

portnov:
@DolphinDream Am I getting it right that main thing that needs testing is json export/import?

portnov:
I tried to export and import simple setup which was using a monad (https://gist.github.com/5be7ffb6e2cb9ec4c594f5a320832072).

Import log:

      ############ 0.07
      ==== loading monad ====
      ############ 0.07
      2017-12-09 12:24:56,402 [INFO] root: Initializing Sverchok logging. Blender version 2.79 (sub 0), Sverchok version 0.5.9.6, branch development_io, commit 55a3c0e6
      applying node props for node:  SvExtrudeRegionNode
      sockets wiped, skipped update
      applying node props for node:  SvGroupInputsNodeExp
      sockets wiped, skipped update
      applying node props for node:  SvGroupOutputsNodeExp
      applying node props for node:  MatrixApplyNode
      applying node props for node:  SvNGonNode
      update lists:
      ['Group Inputs Exp', 0, 'NGon', 1]
      ['NGon', 0, 'Matrix Apply (verts)', 0]
      ['Group Inputs Exp', 2, 'Matrix Apply (verts)', 1]
      ['Matrix Apply (verts)', 0, 'Extrude Region', 0]
      ['NGon', 2, 'Extrude Region', 2]
      ['Group Inputs Exp', 1, 'Extrude Region', 5]
      ['Extrude Region', 0, 'Group Outputs Exp', 0]
      ['Extrude Region', 2, 'Group Outputs Exp', 1]
      no failed connections! awesome.
      applying node props for node:  SvNumberNode
      applying node props for node:  SvNumberNode
        File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/utils/sv_IO_panel_tools.py", line 570, in add_node_to_tree
          node = unpack_monad(nodes, node_ref)
        File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/utils/sv_IO_monad_helpers.py", line 89, in unpack_monad
          setattr(node, k, params[k])
      KeyError: 'height_'

      SvMonadGenericNode not currently registered, skipping
      applying node props for node:  SvNGonNode
      applying node props for node:  ViewerNode2
      update lists:
      ['A Number', 0, 'Monad', 0]
      ['A Number.001', 0, 'Monad', 1]
      ['NGon', 0, 'Monad', 2]
      ['Monad', 0, 'Viewer Draw', 0]
      ['Monad', 1, 'Viewer Draw', 1]
      no failed connections! awesome.

Everything imported correctly in the end, though.

zeffii:
"  SvMonadGenericNode not currently registered, skipping  " 

lol.  wtf :)

zeffii:
as for the _KeyError_ , there's still a number of unresolved issues regarding monad properties. But if it loads correctly, then perhaps the order of operation in the monad import can be switched around to update the `cls_dict` first. I thought we already do this..

DolphinDream:
If no one objects I was going to rebase this branch on top of master to test everything with the latest fixes.

DolphinDream:
@portnov I think everything needs testing, export/import in particular, but also the expand/quicklink functionality in general for different nodes to make sure they work as expected. So far playing with various nodes to expand/collapse/quicklink it seems to work just fine at my end, including the export/import (which saves sparse json as it should).

zeffii:
OK. There's a part of `unpack_monad` that can be skipped during a gist import. The reason we get

image


```python
    File "/sverchok-master/utils/sv_IO_monad_helpers.py", line 89, in unpack_monad
      setattr(node, k, params[k])
  KeyError: 'height_'
```
is because *height_* (`k`) is not in params, but stored in two other locations in params.. neither of which really need to be accessed in this scenario.


  - `params.get('float_props').get(k).get('default')` 
  - `params.get('input_template')` -> wormhole -> ...get('prop_name') --> 'height_'

The reason this doesn't matter is because eventually the cls_dict and the monad's own magic sets the socket data.


-- because this function is only called in io/json, i'll add a hold-off.

zeffii:
the example image above touches this part of Monad "socket acquisition" that we've not completed yet.

> - [ ]vector sockets that use prop_name, like Vector Math node..
    image
   this is especially tricky.

https://github.com/nortikin/sverchok/blob/28022409909413126ae61d3b8a02c4e381360624/core/monad.py#L118-L121

i'm inclined to get this PR merged, as it will benefit all. Monads will effect less users..and these features don't work at the moment anyway :)

zeffii:
will print this for now
```
floatvec prop ignored (normal behaviour since day one)
-- prop_func: <built-in function FloatVectorProperty>
-- prop_dict: {'size': 3, 'default': (0, 0, 0), 'name': 'input a', 'update': <function updateNode at 0x0000017136728598>, 'attr': 'v3_input_0'}
```
maybe when i stare at that for a while in the console it will piss me off enough to fix it.

DolphinDream:
Rebased on master and resolved conflicts. (set your local branch to upstream/development_io)

DolphinDream:
Rebased on master and resolved conflicts. (set your local branch to upstream/development_io)

DolphinDream:
@zeffii I not clear about what you mean by the “vector sockets that use prop_name, like Vector Math node”.  You want vector sockets that use_prop to have a quicklink? or both?

zeffii:
It's late so my answer will be brief. In Monad - Properties like IntProperty and FloatProperty are picked up using a pair of default collections called int_props and float_props respectively.

Adding desired handling of floatvector props by the monad is more invasive than I want to deal with in this PR. I will remove this from the checklist at the top.

DolphinDream:
Is there anything else we are waiting for  to integrate this into master?

zeffii:
other than extensive testing?   nope..

portnov:
As my last several PRs shown, "extensive testings" begins after merging to master :)

zeffii:
A small bug:
- SNLite does not display its sliders for Int / Float 

i'll have a look.

nortikin:
SNlight not handle interface template with color curves easing

zeffii:
SNlite had a special relationship with the socket classes, i want to fix it tonight.

zeffii:
StringsSocket may need a `force_origin` attribute to allow certain nodes to control where the socket will take the slider value from. at the moment if use_prop is set, then it is assumed that the socket will provide this property (and this will be true for 99% of nodes) but SNLite will need to get it's property from the node.  (for backward compatibility...)

zeffii:
closer..

old draw function of `StringsSocket`
```
    elif self.prop_type:
            layout.prop(node, self.prop_type, index=self.prop_index, text=self.name)
```
note there's no event that triggers `index` in this new draw sequence.. OK. think it's tracked now

image

so...
```python

        if self.bl_idname == 'StringsSocket' and node.bl_idname in {'SvScriptNodeLite'}:
                layout.prop(node, self.prop_type, index=self.prop_index, text=self.name)
                return
```

zeffii:
maybe..for the time being..
```python
        if self.bl_idname == 'StringsSocket':
            if hasattr(self, 'custom_draw'):
                if self.custom_draw and hasattr(node, self.custom_draw):
                    getattr(node, self.custom_draw)(self, context, layout)
                    return

            if node.bl_idname in {'SvScriptNodeLite', 'SvScriptNode'}:
                    layout.prop(node, self.prop_type, index=self.prop_index, text=self.name)
                    return
```

zeffii:
the `node_tree` in master only supports custom_draw on `StringsSocket`, i had not intended to modify that in the coming Pull Request.. and I thought a simple copy/paste would suffice.. but no.

hey. atleast we got error messages this time :)   -- i'm on a branch of development_io so the line numbers won't match up if you're reading along. ( `snlite_fix_for_devio` )
```
\sverchok\node_tree.py:308
rna_uiItemR: property not found: SvScriptNodeLite.
```

zeffii:
well.. it does work on this end..
image

however I don't think it ever displayed the slider beside the StringsSocket when custom drawing.. so this is not different. The template could be updated to draw something...

zeffii:
preliminary testing suggests this is covering things... i've tested :)

```python
        # just handle custom draw..be it input or output.
        # hasattr may be excessive here
        if self.bl_idname == 'StringsSocket':
            if hasattr(self, 'custom_draw'):
                if self.custom_draw and hasattr(node, self.custom_draw):
                    getattr(node, self.custom_draw)(self, context, layout)
                    return

            if node.bl_idname in {'SvScriptNodeLite', 'SvScriptNode'}:
                if not self.is_output and not self.is_linked and self.prop_type:
                    layout.prop(node, self.prop_type, index=self.prop_index, text=self.name)
                    return
```

zeffii:
I'm ok with this now.

portnov:
If nobody has objections, I'd propose to merge this, and then fix bugs that will arise :)

zeffii:
But merge squash

zeffii:
```
*runs for the hills*
```

DolphinDream:
yey! 😄 finally

PR #1849, opened 2017-11-08
Minor optimization in draw_expander_template

DolphinDream:
Reorder conditions to only test for StringSocket once


zeffii:
i've added it anyway

PR #1854, opened 2017-11-09
Auto oldnodes init2

zeffii:


zeffii:
```python
import os
import ast

path = r"C:\Users\zeffi\Desktop\GITHUB\sverchok\old_nodes"

def get_old_nodes_list(path):
    for fp in os.listdir(path):
        if fp.endswith(".py") and not fp.startswith('__'):
            yield fp

def get_registered_nodeclasses(path, old_node_file):
    collection = []
    file_path = os.path.join(path, old_node_file)
    with open(file_path, errors='replace') as file:
        
        node = ast.parse(file.read())
        classes = [n for n in node.body if isinstance(n, ast.ClassDef)]
        for c in classes:
            for k in c.bases:
                if hasattr(k, 'id') and k.id == 'SverchCustomTreeNode':
                    collection.append([c.name, old_node_file])

    return collection


for old_node_file in get_old_nodes_list(path):
    items = get_registered_nodeclasses(path, old_node_file)
    print(items)
```

nortikin:
thanks

zeffii:
i am happy with it so far :)

PR #1857, opened 2017-11-10
add slight code change

zeffii:


zeffii:
@Durman  i think  the matrix  code still isn't correct ... try it on a donut/

zeffii:
image

zeffii:
image

zeffii:
i think the matrix orientation in vector_in mode needs more consideration. It means that the Normal alone isn't enough to generate the a correct "space" matrix. because the rotation component will be arbitrary..

zeffii:
the matrix shown here describes the normal just fine... but can not convey the face's orientation..
image
exactly how to solve that is open for debate

zeffii:
pushing, as end result is the same -- but slightly less clunky code. still needs a good solution for non ordinal axis/normal

Durman:
@zeffii 
I did not expect this behaviour of polygons.)) I thought if Y axis always will have direction to up it will be normal working but i see now that it's wrong. May be it is necessary to calculate bound rectangle for polygon and send axis of matrix to centers of rectangle sides. how do you think?

scale

Durman:
How to locate matrix for such polygons or this case is not interesting for us?

strange polygon

zeffii:
@Durman this _is_ *the* question..  " by which criteria do we decide a polygons orientation ". This definitely needs some debate. 

peronally I was going to treat Triangles one way, Quads another way..and Ngons also different. But the downside is all these operations are not fast in Python for a heavy object. This is also already one of the slowest nodes in Sverchok. (I don't know if it needs to be this slow, or it's this slow because we use these bmesh operations...)

If you experiment with "Inset Specials" node, it is much faster and has similar features.

Durman:
I don't see point to create scaling by axis for triangles and Ngones at all.
I decided to choose upper middle point of rectangular for X axis direction but it is wrong too.

new scaling

```Python
    #calculating top middle point of polygon
    mid_verts = []
    for e in face.edges:
        mid_verts.append(e.verts[0].co.lerp(e.verts[1].co, 0.5))
    max_z = max([i[2] for i in mid_verts])
    max_vert = [i for i in mid_verts if i[2] == max_z][0]
```

zeffii:
> I don't see point to create scaling by axis for triangles and Ngones at all.

even in that case we could implement it so that if the polygon is Ngon or Triangle, that the first component of the scale vector is used. `vec=(x, x, x)`, and if Quad then maybe consider the longest side as Y. If the quad is not uniform/rectangle the rectangle to consider would be the bounding rect. (not a cheap thing to process for each polygon... )

Durman:
It is difficult task. I suppose that the only solution of orientation of matrix is analyze of topology of object. It can allow us to locate matrixes in right position and it will be independent from location of object.

Durman:
I have thought to do something clever for analysis of mesh, I mean to create object for polygons where we can keep all information about them.
```Python
"""
in verts_in v d=[] n=1
in pols_in s d=[] n=1
out something s
"""
class Face:
    def __init__(self,id,ngon=None):
        self.id = id
        self.ngon = ngon
        
    def get(self):
        return 'Id = %s, Ngon = %s' % (self.id,self.ngon)
        
smart_face = []
for c,pol in enumerate(pols_in):
    ngon = len(pol)
    smart_face.append(Face(id=c, ngon=ngon))
    
print([i.get() for i in smart_face])
```

Durman:
I did something like analyzing of topology but it's working only with torus yet and code is not fast :/

donate

zeffii:
@Durman that's cool! 

as you have learned, the cpu time needed for extensive analysis can be very high compared to the simplest case of `vec=(scale, scale, scale)`.

Durman:
I have slowly function of searching neighbours but I managed to improve it and now is faster.
```
Time test with standart Torus - [0.279484556413081] - first version of algorithm
Time test with standart Torus - [0.00020413738093338907] - second
Time test with standart Torus - [0.0067695655598072335] - total node
```
May be it is possible to analyze mesh only one time if mesh doesn't change?

zeffii:
generally when you modify the values of nodes downstream from the node, then the upstream nodes are not updated/recalculated. However, if you consider the idea that your node has input for a mesh, and some other input for scale.. because scale is changed, the whole node is recalculated.  

I understand you would want to cache the calculations for part of the input that is unchanged,. Sverchok doesn't offer this feature (yet).....   

but ! it is possible I think. Each object (python object... list integer..string..etc)  has a reference to a point in memory where the information is stored.
```python
   # new property
   self.reference_storage = bpy.props.StringProperty()

   ## def process
   verts = some_vert_socket.get()
   faces = some_faces_socket.get()

   reference_verts = hash(id(verts))
   reference_faces = hash(id(faces))
   reference = (reference_verts + reference_faces)
   if self.reference_storage == reference:
      # unchanged input , obtain cached calculations
      print('unchanged verts + faces input ') 
   else:
      print('changed     verts + faces input ') 
      # calculate fresh  , store for future updates
      self.reference_storage = reference
```
if that works, then you need to decide where to store the cached data.. in a dict in the .py file or a dict inside the node class. You can use a dict for testing..

```
# somewhere at the top of the file
mesh_calculations = {}
```
then this part changes
```python
   if self.reference_storage == reference:
      # unchanged input , obtain cached calculations
      processed_data = mesh_calculations[hash(self)].get('processed_mesh_data')
      print('unchanged verts + faces input ') 
   else:
      print('changed     verts + faces input ') 
      processed_data = calculate_data(verts, faces)
      mesh_calculations[hash(self)]['processed_mesh_data'] = processed_data
      self.reference_storage = reference
```
i'm not sure this works..

zeffii:
OK.. i think you can cache, but its very strict.

```python
        vertices_s = inputs['Vertices'].sv_get(deepcopy=False)
        edges_s = inputs['Edges'].sv_get(default=[[]])
        faces_s = inputs['Polygons'].sv_get(default=[[]], deepcopy=False)
        masks_s = inputs['Mask'].sv_get(default=[[1]])
        heights_s = inputs['Height'].sv_get()
        scales_s  = inputs['Scale'].sv_get()

        cache_id = str(hash(id(vertices_s)) + hash(id(faces_s)))
        if cache_id == self.cache_id:
            print('verts and faces are unchanged - the same id')
        else:
            print('changed objects... :(')
            self.cache_id = cache_id
```

Durman:
I completed my function of analysis of mesh. It works with all primitive objects now but there is topology with which result of the function will look as bug. I did not make out yet how to cache the data. It works weary slowly - Time test with standard torus - [0.23496227516125145] sec. 
@zeffii  Is the code suitable?
```Python
"""
in verts_in v d=[] n=1
in pols_in s d=[] n=1
enum = Work Test
out mat m
out time_test s
"""
from mathutils import Vector, Matrix
from mathutils.geometry import normal
from sverchok.data_structure import Matrix_listing

class Face:
    
    id_face = {}
    number_log = 0
    first_todo_faces = set()
    todo_faces = set()
    done_faces = set()
    
    def __init__(self,id=None, pol=None, verts=None):
        self.id = id
        self.pol = pol
        self.verts = verts
        self.neibs = []
        self.x_edge = None
        self.y_edge = None
        self.x_axis = None
        self.y_axis = None
        self.matrix = None
        
    def print_log(self,text, what):
        #pass
        Face.number_log += 1
        print(Face.number_log,'. ',text,what)
                
    def analyse(self):
        
        def get_center(verts, ngon):
            if ngon == 4:
                v1 = verts[0].lerp(verts[1], 0.5)
                v2 = verts[2].lerp(verts[3], 0.5)
                return v1.lerp(v2, 0.5)
            else:
                return 'Algaritm absent'
        
        self.z_normal = normal(self.verts)
        self.ngon = len(self.pol)
        self.center = get_center(self.verts, self.ngon)
        self.edges = [(self.pol[i:] + self.pol[:i])[:2] for i in range(4)]
        self.id_face[self.id] = self
        
    def search_neighbors(self):
        # Very slolwly algoritm!
        for key in self.id_face:
            if self.id_face[key].ngon == 4:
                for key2 in self.id_face:
                    if self.id_face[key2].ngon == 4:
                        if len(self.id_face[key].neibs) == 4: 
                            break
                        else:
                            if self.id_face[key2].id not in self.id_face[key].neibs:
                                if len(set(self.id_face[key].pol) - set(self.id_face[key2].pol)) == 2:
                                    self.id_face[key].neibs.append(self.id_face[key2].id)
                                    self.id_face[key2].neibs.append(self.id_face[key].id)

    def first_face(self):
        self.x_edge = self.edges[0]
        self.y_edge = self.edges[1]
        self.x_axis = self.verts[0].lerp(self.verts[1], 0.5) - self.center
        self.y_axis = self.verts[1].lerp(self.verts[2], 0.5) - self.center
        self.cal_mat()
        Face.first_todo_faces.add(self.id)
        
    def cal_xy_edges(self):
        
        def loop_x(doner_face, current_neib):
                
            x_edges = [i for i in current_neib.edges if len(set(i) - set(doner_face.pol)) == 0 
                                                    or len(set(i) - set(doner_face.pol)) == 2]

            x_bol = False
            for i in x_edges:
                if len(set(doner_face.x_edge) - set(i)) == 0:
                    x_bol = True

            if x_bol:
                current_neib.x_edge = [i for i in x_edges if len(set(i) - set(doner_face.x_edge)) == 2][0]
            else:
                current_neib.x_edge = [i for i in x_edges if len(set(i) - set(doner_face.y_edge)) == 1][0]

            y_edges = [i for i in current_neib.edges if len(set(i) - set(x_edges[0])) == 1]
            current_neib.y_edge = [i for i in y_edges if len(set(i) - set(doner_face.y_edge)) == 1][0]
                        
            self.cal_xy_axis(current_neib)
            self.cal_mat(current_neib)
                
        def loop_y (doner_face, current_neib):
            x_edges = [i for i in current_neib.edges if len(set(i) - set(doner_face.pol)) == 1]
            current_neib.x_edge = [i for i in x_edges if len(set(i) - set(doner_face.x_edge)) == 1][0]
                            
            y_edges = [i for i in current_neib.edges if len(set(i) - set(x_edges[0])) == 1]
                        
            y_bol = False
            for i in y_edges:
                if len(set(doner_face.y_edge) - set(i)) == 0:
                    y_bol = True
                    
            if y_bol:
                current_neib.y_edge = [i for i in y_edges if len(set(i) - set(doner_face.y_edge)) == 2][0]
            else:
                current_neib.y_edge = [i for i in y_edges if len(set(i) - set(doner_face.x_edge)) == 1][0]
                current_neib.y_edge = [i for i in y_edges if len(set(i) - set(doner_face.x_edge)) == 1][0]
    
            self.cal_xy_axis(current_neib)
            self.cal_mat(current_neib)
        
        loop = 0
        while Face.todo_faces or Face.first_todo_faces:            
                        
            flag = None

            if not Face.first_todo_faces:
                Face.first_todo_faces.add(Face.todo_faces.pop())
                loop = 0
                
            if len(Face.first_todo_faces) == 1:
                flag = list(Face.first_todo_faces)[0]
                doner_face = self.id_face[list(Face.first_todo_faces)[0]]
            else:
                doner_face = self.id_face[Face.first_todo_faces.pop()]
            
            Face.done_faces.add(doner_face.id)

            for neib in doner_face.neibs:
                current_neib = self.id_face[neib]
                if not current_neib.matrix:

                    if (len(set(current_neib.pol) - set(doner_face.x_edge)) == 2 
                    or len(set(current_neib.pol) - set(doner_face.x_edge)) == 4):
                        
                        Face.first_todo_faces.add(current_neib.id)                            
                        loop_x(doner_face, current_neib)
                        
                    else:
                        
                        if not loop:
                            
                            if current_neib.id not in Face.done_faces:
                                Face.todo_faces.add(current_neib.id)
                                
                            loop_y(doner_face, current_neib)
                                                
            if flag != None:
                Face.first_todo_faces.remove(flag)
            loop += 1
        
    def cal_xy_axis(self, current_face):
        px1 = current_face.verts[current_face.pol.index(current_face.x_edge[0])]
        px2 = current_face.verts[current_face.pol.index(current_face.x_edge[1])]
        py1 = current_face.verts[current_face.pol.index(current_face.y_edge[0])]
        py2 = current_face.verts[current_face.pol.index(current_face.y_edge[1])]
        current_face.x_axis = px1.lerp(px2, 0.5) - current_face.center
        current_face.y_axis = py1.lerp(py2, 0.5) - current_face.center
        
    def cal_mat(self, current_face=None):
        if not current_face:
            current_face = self
        else:
            current_face = current_face

        matrix_t = Matrix.Translation(current_face.center)
        matrix_r = Matrix(list([*zip(current_face.x_axis.normalized()[:], 
                                    current_face.y_axis.normalized()[:], current_face.z_normal[:])])).to_4x4()        
        current_face.matrix = matrix_t * matrix_r

smart_face = []
def process():               
    for c,pol in enumerate(pols_in):    
        smart_face.append(Face(id=c, pol=pol, verts=[Vector(verts_in[i]) for i in pol]))
        smart_face[c].analyse()

    smart_face[0].search_neighbors()
    smart_face[0].first_face()
    smart_face[0].cal_xy_edges()
        
    mat = Matrix_listing([i.matrix for i in smart_face if i.matrix])
    return(mat)

#----------------------Test-----------------------
import timeit
def test():
            
    time_t = timeit.Timer(process).repeat(1, 1)
    time_test = 'Time test with standart Torus - %s' % time_t
    return time_test
    
    layout.prop(self, 'custom_enum', expand=True)
    
if self.custom_enum == 'Work':
    mat = process()
else:
    time_test = test()
```

zeffii:
cool @Durman - looks complicated, I don't have time to closely analyse your code. If it works for your needs then it's good :) . Not all algorithms can be expected to run in realtime in python. This brings up an interesting problem. When we handcode routines (as you have done here) instead of calling bmesh ops, there comes a point where we should be strongly considering writing in a compiled language (Rust/Go..C) and calling modules from Python. The speed increase can be dramatic.

SNLite is suitable for testing your algorithm, but not suitable for using the caching mechanism i've described above.

Durman:
The thought about protection from extra recalculation still does not leave me.)) 
Time of running hash function:
x - number of points, 
y - time


<details> 
  <summary>code</summary>  

```Python
import random
import time
from matplotlib import pyplot as plt

random.seed(0)

def get_data(number):
    r = random.random()
    return([(r,r,r) for _ in range(number)])

def get_hash(data):
    return hash(tuple(data))

def timed(f, *args, n_iter=10):
    acc = float('inf')
    for i in range(n_iter):
        t0 = time.perf_counter()
        f(*args)
        t1 = time.perf_counter()
        acc = min(acc, t1 - t0)
    return acc

def compare(f, args):
    plt.plot([len(arg) for arg in args], [timed(f, arg) for arg in args], label=f.__name__)
    plt.legend()
    plt.grid(True)
    plt.show()

compare(get_hash, [get_data(i) for i in range(0,1000000,20000)])
```

</details>

It looks like that hash function is slower than I thought

PR #1859, opened 2017-11-11
add rclick stub for nodeview

zeffii:
image

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -100,7 +100,7 @@ utils_modules = [
 
 ui_modules = [
     "color_def", "sv_IO_panel", "sv_templates_menu",
-    "sv_panels", "nodeview_space_menu", "nodeview_keymaps",
+    "sv_panels", "nodeview_rclick_menu", "nodeview_space_menu", "nodeview_keymaps",
     "monad", "sv_icons",
     # bgl modules
     "viewer_draw", "viewer_draw_mk2", "nodeview_bgl_viewer_draw", "nodeview_bgl_viewer_draw_mk2",
--- a/ui/nodeview_keymaps.py
+++ b/ui/nodeview_keymaps.py
@@ -55,6 +55,10 @@ def add_keymap():
         kmi = km.keymap_items.new('node.sv_extra_search', 'SPACE', 'PRESS', ctrl=True)
         nodeview_keymaps.append((km, kmi))
 
+        # Right Click (Nodeview)  | show custom menu !
+        kmi = km.keymap_items.new('wm.call_menu', 'RIGHTMOUSE', 'CLICK')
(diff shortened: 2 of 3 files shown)


zeffii:
can add menu items 
```
Connect outputs (of Active Node)
    - to VDMK2
    - to BMVMK2
```

nortikin:
nodes not needed, i guess. nodes menu exists in shift+a allready :-|

zeffii:
i'm not sure i like the nodes menu in rclick either.
i might keep it if the current nodeview has no nodes in it.. or no active node..
when an active node is present there can be many types of menu showing depending on the type of active node.. the rclick menu can be contextual.

nortikin:
rclick can handle old nodes as well

zeffii:
ishi_ban

zeffii:
maybe keep track of ideas here

- [x] add vdmk2 to active node
- [ ] if active node is vdmk2:
    - [ ]  offer to add+connect indexviewer to the same incoming sockets
    - [ ] offer to randomize hue and vertex color

zeffii:
nodes could encode a set of menu items dedicated them, when right clicked. The code would be located in the node class instead of in the rightclick menu file. The upside to localising the code is that you can keep all code associated with a node in one place (the node's python file), the downside might be that there exists the possibility to have much duplicated code...

nortikin:
how is better? when creating generator auto connect vdmk2 initially or to leave your RMB menu?

nortikin:
macros?
for users who too lazy to write >_< symbols

portnov:
> offer to randomize hue and vertex color
I'd propose to add this to N pane of this node.

zeffii:
the menu entries and operators listed can be dynamic  - my point with adding the menu stub, is to 
- give a convenient way to add options to all nodes, or classes of nodes.
- offer contextual menu that reflects the selected node(s)

i have purposefully not added much contextual options yet, but encourage anyone who likes to customize their UI to contribute ideas to make sverchok more accessible.

zeffii:
@portnov  sure vdmk2 should maybe have that randomize colors in external buttons, and if the functionality exists in the node, it can appear as an option in that node's rclick menu. As long as we are chained to the mouse anyway, we can at least reduce the kilometers travelled by having the option near the node.. or ... indeed...on the main node ui.

Durman:
@zeffii 
I suggest to add stethoscope on rclick on a the socket.

zeffii:
Sounds good

zeffii:
i'm not aware of any convenience feature in the node's API to tell me where on the node the rightclick came from. It would be possible to test the location of the cursor vs the locations of all sockets on the node, and attach it to the nearest one. (warning.. you will always need to make sure the Node is selected/active first...-- else the algorithm would need to filter through all nodes in the tree and return the closest socket that way)

PR #1868, opened 2017-11-19
Metaball output node

portnov:
Inspired by this video: https://www.youtube.com/watch?v=PP14rVhgX9M

Unfortunately, Blender does not have API for metaballs similar for bmesh for meshes; Metaballs can only exist in scene. So we cannot "process" metaballs, but we can create them.

This PR introduces a new node, "Metaball out". It has almost all parameters that can be set for metaballs through python API. I put it directly to the Viz category, because I do not think it gonna change a lot.

Documentation is pending.

### DIFF
--- a/docs/nodes/viz/metaball_out.rst
+++ b/docs/nodes/viz/metaball_out.rst
@@ -40,7 +40,11 @@ Parameters
 This node has the following parameters:
 
 - **UPD**. The node will process data only if this button is enabled.
-- **Base name**. Base part of name for Meta object to create (or update). Default is "SvMetaBall".
+- **Hide View** (eye icon). Toggle visibility of generated object in viewport.
+- **Hide Select** (pointer icon). Toggle ability to select for generated object.
+- **Hide Render** (render icon). Toggle renderability for generated object.
+- **Base name**. Base part of name for Meta object to create (or update). Default is "Meta_Alpha".
+- **Select Toggle**. Select / deselect generated object.
 - **Material**. Material to be assigned to created object.
 - **Threshold**. Influence of meta elements. Default is 0.6.
 - **Resolution (viewport)**. Resolution of Meta object for viewport. Lesser value mean better resolution. Default is 0.2. This parameter can be set only in the N panel.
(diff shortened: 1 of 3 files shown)


portnov:
screenshot_20171119_182754

portnov:
screenshot_20171119_193312

portnov:
It is probably not easy at all to create chicken from that video with Sverchok+metaballs, but it is pretty easy to create something like this:

meta_2

portnov:
meta_ico

nortikin:
ддддд
кек
🤣 
change not lot - only as bmesh viewer - hidden/selectable/renderable three buttons

portnov:
added some functionality similar to bmesh viewer.

portnov:
meta-torus-2

portnov:
meta-torus-3

zeffii:
stick more light in those renderings @portnov  :)

The metaball algorithm was updated this year. this should be relatively snappy.

portnov:
I consider this ready for merge. Any cons?

zeffii:
just object reference output :)

enzyme69:
I am waiting to test this node ... merge?

enzyme69:
Ok I agressive-ly use the node:
https://youtu.be/nEyM4T20rRo

Feedbacks:
- Need **Objects Out** slot
- Somewhat does not show up when I use **Objects In**
screen shot 2017-11-23 at 21 23 18

So far I really enjoy it, this node is BIG for me. Top notch. My favourite next to Grease Pencil and what Zeffii have made gp+

zeffii:
@enzyme69  the nodetree will have an arbitrary update order when you have disjoint graphs in the tree. This means that the object data-to-svdata process can reflect the right data or the previous state of the object, depending on whether the metaball node updates first or second. Hence metaball should have an object out, and you should use BPY->OBJECT ID OUT node to go from metaball to mesh.

I imagine suzanne with modifiers on will produce far too many vertices for metaballs to be fast. remember each vertex will generate between 6 and 40 verts depending on where the vertex is in the volume and how high the subdivision is.

Durman:
I didn't quite understand in which menu could I already find this node?

portnov:
@Durman this merge was reverted, i hope #1871 will be merged soon. After that (or in my portnov:metaball_out branch) you can find this under Viz menu.

PR #1870, opened 2017-11-23
Revert "Metaball output node"

zeffii:
Reverts nortikin/sverchok#1868


zeffii:
@enzyme69  you don't test a branch by merging. you first checkout the branch using git.

i have reverted your merge, for now.

@nortikin  i think we should not allow people to merge code they haven't read / contributed to / understood / tested.  maybe @enzyme69 should be promoted to "Contributor" instead of "Collaborator" ?

zeffii:
what we could do is offer an abstraction of git in the update panel in nodeview. Assuming the user has git installed, and it can be started from their terminal using `$git` or `c:\git`  we could make an operator that lists the current branches, and does a `git checkout <branch_name>` under the hood. and a `change to master` would run `git checkout master; git pull`. etc..

portnov:
A dropdown list with branches near "update" button would be good.

@zeffii Should I now create a new PR with added object reference output?

zeffii:
Yes please portnov. Why don't you have coming access?

portnov:
#1871 

> Why don't you have commit access

Don't know. What should I do to have it? :)

zeffii:
@portnov if it was up to me you'd have collaborator status :) but I think only @nortikin can give that privilege.

nortikin:
lets see about contributor status

nortikin:
added @portnov to collaborators. There is no status of contributor, only by fact of contribution

PR #1871, opened 2017-11-23
Metaball out (with Objects out)

portnov:
This is a redo of  #1868, with Objects output included.

### DIFF
--- a/nodes/viz/metaball_out.py
+++ b/nodes/viz/metaball_out.py
@@ -1,0 +1,294 @@
+# ##### BEGIN GPL LICENSE BLOCK #####
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software Foundation,
+#  Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# ##### END GPL LICENSE BLOCK #####
+
+import bpy
+from mathutils import Matrix
+from bpy.props import StringProperty, BoolProperty, FloatProperty, EnumProperty
+
+from sverchok.node_tree import SverchCustomTreeNode
+from sverchok.data_structure import SvGetSocketAnyType, node_id, Matrix_generate, match_long_repeat, updateNode
+from sverchok.utils.sv_viewer_utils import greek_alphabet
+
+class SvMetaballOperator(bpy.types.Operator):
+
+    bl_idname = "node.sv_callback_metaball"
+    bl_label = "Sverchok metaball general callback"
+    bl_options = {'REGISTER', 'UNDO'}
+
+    fn_name = StringProperty(default='')
+
+    def hide_unhide(self, context, type_op):
+        node = context.node
+        metaball = node.find_metaball()
+
+        if type_op in {'hide', 'hide_render', 'hide_select'}:
+            op_value = getattr(node, type_op)
+            setattr(metaball, type_op, op_value)
+            setattr(node, type_op, not op_value)
+
+        elif type_op == 'mesh_select':
+            metaball.select = node.select_state
+            node.select_state = not node.select_state
+
+    def execute(self, context):
+        self.hide_unhide(context, self.fn_name)
+        return {'FINISHED'}
+
+
+class SvMetaballOutNode(bpy.types.Node, SverchCustomTreeNode):
+    '''Create Blender's metaball object'''
+    bl_idname = 'SvMetaballOutNode'
+    bl_label = 'Metaball'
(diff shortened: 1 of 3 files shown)


nodes/viz/metaball_out.py
```diff
@@ -0,0 +1,294 @@
...
+                setup_element(element, item)
+        else:
+            #print('Recreating metaball data')
+            metaball_object.data.elements.clear()
+
+            for item in items:
+                element = metaball_object.data.elements.new()
+                setup_element(element, item)
+
+        if 'Objects' in self.outputs:
```
Kosvor2: if self.outputs['Objects'].is_linked:

nodes/viz/metaball_out.py
```diff
@@ -0,0 +1,294 @@
...
+                    raise Exception("`Types' input expects an integer number from 1 to 5")
+                meta_type = self.meta_type_by_id[meta_type]
+            element.type = meta_type
+            element.radius = radius
+            element.stiffness = stiffness
+            element.rotation = rotation
+            element.size_x = scale[0]
+            element.size_y = scale[1]
+            element.size_z = scale[2]
+            element.use_negative = bool(negate)
```
Kosvor2: bool() here not needed.

Kosvor2:
How about using `match_long_cycle` instead of `match_long_repeat` ?
cycle_negate

nodes/viz/metaball_out.py
portnov: You mean Python will translate it automatically? Well, probably yes. For me personally, it seems better if I do explicit conversion; but if there is some convention in Sverchok or in Python in general, i'll change it.

nodes/viz/metaball_out.py
portnov: I added this line because I (and probably @enzyme69 :) ) already have several scenes with this node, which would break if I tried to set value for unexisting output. Afaik check for .is_linked for output is not necessary anymore in Sverchok, correct me if I'm wrong.

portnov:
Well, this imho is one of the most outstanding questions with most of Sverchok's nodes - matching type is hardcoded, but for many uses some other matching algorhithm would be better. I remember discussing the general solution for 0.6. I could add a dropdown with matching modes (to the N panel I think).

portnov:
Objects output maybe not working the way someone expected (i don't exactly know what is expected though). If I plug it to "Object ID out", it does not output vertices. The same if I select generated meta in "object ID out" manually. I think this is just because META objects do not actually have any vertices (it is not mesh). But, if I select it with "Objects in mk3", It outputs vertices. Maybe it converts objects to mesh internally somehow? Will look in source.

portnov:
Aha, Object In node does object.to_mesh(). Should I do so in Metaball node? My opinion is not, that should be part of some object-manipulating node.

nodes/viz/metaball_out.py
zeffii: `.is_linked` is optional, and if the data is calculated anyway you don't need to test if_linked, just push to the socket. Most of what we are doing by pushing to a socket using .sv_set is passing object ID's (like passing by reference)

zeffii:
>  Should I do so in Metaball node?
indeed . don't do that.   Instead in OBJECT ID OUT the type of object should be tested, if it's metaball then similar code to `objects in mk3` should be thrown at it.

zeffii:
i think there will be an unfair expectation of meshes generated from metaballs... it's going to be big meshes for the combination _high subdiv / vert per ball / verts total passed in._ heavy meshes means converted to sv_data means you can forget near realtime viewport processing.

portnov:
meta-penta-1

enzyme69:
Click "Merge pull request" button? Just kidding, I let Zeffii do it. 🙃

nodes/viz/metaball_out.py
nortikin: python is such a circus:
True == 1
False == 0

nortikin:
from https://github.com/nortikin/sverchok/pull/1871#issuecomment-346667542 this comment conclusion:
in sockets to convert automatically from objects to matrices - matrix from objects. from objects to vertices - vertices? Or it will slow down node tree?

portnov:
> in sockets to convert automatically from objects to matrices - matrix from objects. from objects to vertices - vertices? Or it will slow down node tree?

This conversion works automatically right now as far as I can see.

nortikin:
типа когда объект вводишь, соединяешь такой сокет объекта с сокетом вершин, они и конвертируются (сейчас так матрицы и вектора, туда-сюда конвертятся)

portnov:
Socket conversions are to be discussed in separate issue I believe. Anyway this is to be implemented in core somewhere. Also consider @zeffii 's comment above: with high-resolution metaballs you can accidentally get millions of vertices with two clicks, that most probably is not what you want.

portnov:
I hope nobody has objections against merging this?

Durman:
I tried to animate radius parameter but blender crashes, any ideas?
[metaball1.zip](https://github.com/nortikin/sverchok/files/1502180/metabol1.zip)

portnov:
Well, it looks like radius becomes too large near frame 0 :)

portnov:
meta-pipes1
Generative Art + metaball (+freestyle)

Durman:
Something. Curve + metaball

enzyme69:
Yes, I was actually playing around with Twist Torus, it was pretty cool. There is something about Metaballs and Freestyle.

Durman:
I noticed that resolution property must be in node.

portnov:
@Durman It is accessible in the N panel.

PR #1874, opened 2017-11-26
Menu refactor

portnov:
I'm not sure if this is ready to merge yet, but at least we need to discuss, I think.

## What this adds is:

* Icons to node creation buttons in T panel
* Proper tooltips (from node docstrings) to buttons in T panel
* Proper tooltips for items in Shift-A menu.
* Possibility to create nodes from standard Blender's Space menu (obviously only in node editor context).

screenshot_20171128_201320

## More technical description of changes:

* Special operator class is registered for each node class. This is done in menu.py.
* Icons for operators are taken from node's bl_icon property. All icon-handling methods are moved from menu.py to sverchok.ui.sv_icons. This is done to use exactly the same code both for menu and for T panel: these functions are called both from sverchok.menu and from sverchok.ui.nodeview_space_menu.
* A new, more regular syntax is introduced for node class docstrings. RFC-822 is to be used as a standard, at least for new nodes:

      class SvBoxNode(bpy.types.Node, SverchCustomTreeNode):
          """
          Triggers: Box
          Tooltip: Generate a Box primitive
  
          This is the long description in free form, it can include technical
          or historical remarks.
          """
  
          bl_idname = 'SvBoxNode'
          ...

* For now, only two headers are used: Triggers is a shorthand to be used in Ctrl-Space search menu, and Tooltip is used as a tooltip. Other headers probably will be added in the future. Unknown headers are just ignored. Standard Python's `email` module is used to parse docstrings, so we can be more or less sure that we correctly understand RFC-822 subtle moments (multiline headers, escaping and so on).
* Old `///` syntax for search shorthands is also supported for compatibility.
* To incapsulate docstring parsing logic, there is SvDocstring class in node_tree.py introduced.
* An instance of SvDocstring is associated with each node class and can be obtained with `node_class.get_docstring()`.
* New classmethods are introduced for SverchCustomTreeNode: `get_shorthand()` and `get_tooltip()`. In the default implementation, they parse node class docstring as described above. Author of specific node can override these methods if he needs to. These new methods are used by sv_extra_search.py, menu.py and nodeview_space_menu.py.
* Box node docstring is updated to be an example of what docstrings should look like.

What I'm not sure about is:

* [x] Should "enable icons" setting in preferences affect T panel also?
* [x] Will current code play nicely with reloads?
* [ ] Can we simplify current code, or make it more robust against upcoming blender API changes (if any)?

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -167,6 +167,7 @@ def register():
             if hasattr(m, "register"):
                 #print("Registering module: {}".format(m.__name__))
                 m.register()
+    # We have to register menu module after all nodes are registered
     menu.register()
     # this is used to access preferences, should/could be hidden
     # in an interface
--- a/__init__.py
+++ b/__init__.py
@@ -163,8 +163,11 @@ from sverchok.core import node_defaults
 
 def register():
     for m in imported_modules + node_list:
-        if hasattr(m, "register"):
-            m.register()
+        if m.__name__ != "sverchok.menu":
+            if hasattr(m, "register"):
+                #print("Registering module: {}".format(m.__name__))
+                m.register()
+    menu.register()
     # this is used to access preferences, should/could be hidden
     # in an interface
     data_structure.SVERCHOK_NAME = __name__
--- a/__init__.py
+++ b/__init__.py
@@ -153,7 +153,6 @@ if reload_event:
     for node in node_list:
         importlib.reload(node)
     old_nodes.reload_old()
-    menu.reload_menu()
 
 import bpy
 from sverchok.utils import ascii_print, auto_gather_node_classes, node_classes
@@ -179,6 +178,7 @@ def register():
     auto_gather_node_classes()
     if reload_event:
         data_structure.RELOAD_EVENT = True
+        menu.reload_menu()
         print("Sverchok is reloaded, press update")
 
 
@@ -187,4 +187,5 @@ def unregister():
     node_classes.clear()
     for m in reversed(imported_modules + node_list):
         if hasattr(m, "unregister"):
+            #print("Unregistering module: {}".format(m.__name__))
             m.unregister()
--- a/menu.py
+++ b/menu.py
@@ -126,7 +126,7 @@ class SverchNodeItem(object):
         return getattr(bpy.types, self.nodetype)
 
     def get_idname(self):
-        return self.nodetype.lower()
+        return get_node_idname_for_operator(self.nodetype)
 
     def make_add_operator(self):
 
@@ -161,16 +161,42 @@ class SverchNodeItem(object):
 
     @staticmethod
     def draw(self, layout, context):
-        default_context = bpy.app.translations.contexts.default
-
-        props = layout.operator("node.sv_add_" + self.get_idname(), icon=self.get_icon(), text=self.label, text_ctxt=default_context)
-        props.type = self.nodetype
-        props.use_transform = True
+        add = draw_add_node_operator(layout, self.nodetype, label=self._label, icon=self.get_icon())
 
         for setting in self.settings.items():
-            ops = props.settings.add()
+            ops = add.settings.add()
             ops.name = setting[0]
             ops.value = setting[1]
+
+def get_node_idname_for_operator(nodetype):
+    rna = getattr(bpy.types, nodetype).bl_rna
+    if hasattr(rna, 'bl_idname'):
+        return rna.bl_idname.lower()
+    else:
+        return rna.name.lower()
+
+def draw_add_node_operator(layout, nodetype, label=None, icon=None, params=None):
+    default_context = bpy.app.translations.contexts.default
+
+    if label is None:
+        rna = getattr(bpy.types, nodetype).bl_rna
+        if hasattr(rna, 'bl_label'):
+            label = rna.bl_label
+        else:
+            label = rna.name
+
+    if params is None:
+        params = dict(text=label)
+    params['text_ctxt'] = default_context
+    if icon is not None:
+        params['icon'] = icon
+
+    add = layout.operator("node.sv_add_" + get_node_idname_for_operator(nodetype), **params)
+                            
+    add.type = nodetype
+    add.use_transform = True
+
+    return add
 
 def sv_group_items(context):
     """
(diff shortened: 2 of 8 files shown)


zeffii:
very cool!

zeffii:
> Should "enable icons" setting in preferences affect T panel also?

if it's extra work to make those not appear by default, don't bother.

> Will current code play nicely with reloads?

this has a definitive and testable answer. but if you are asking because it does not currently behave nicely, then it will need to be investigated.

> Can we simplify current code, or make it more robust against upcoming blender API changes (if any)?

i suggest first to get it working nicely, and keep as little of the other (yet potentially related) code unchanged for this code injection.

portnov:
This looks more or less stable (ish? I like when Ton pronounces "is" as "ish") to me. @zeffii , @nortikin please review and test.

Also, if this is going to be in master, as next step I think we will decide something with node docstrings, because they are going to appear in tooltips. Some of nodes do not have docstrings at all, other have too laconic ones. Also, part of docstring after "///" is currently used for search menu. What to do with this? We can for example use only last line of docstring as tooltip. Or only first. Or whatever we find better.

zeffii:
> Also, part of docstring after "///" is currently used for search menu. What to do with this? We can for example use only last line of docstring as tooltip. Or only first. Or whatever we find better.

i want to keep this format. Using the first part of the docstring up to `///` as the content for the search term. Don't change what is not broken.

zeffii:
the tooltip for each node means you are registering a new operator for every node?

portnov:
@zeffii yes.

zeffii:
i thought a bit more about this. We can add a line to the `__doc__` of each node and look for the pattern, and use the content if its found and use the part after `///` if it isn't.
```py
class SomeNode():
    """
    shorthand triggers + brief descriptor ///
    
    this node blablabla (longer description maybe with problem descriptions..who worked on what..)
    stuff for those of us who read these parts of the code.

    (#tooltip: This node allows. ....................................... #)
    """
```
We might want the tooltip to say very specific things about the node. I'm not sure if tooltips are intended to be multi-line?

zeffii:
> the tooltip for each node means you are registering a new operator for every node?

>>  @zeffii yes.

yeah, it's one reason I didn't already do this. The idea of an operator per node just to set the tool-tip wasn't very appealing. But...at some point we must make a choice in favour of users and usability. This helps users.

maybe down the road bpy API will permit less convoluted solutions.

portnov:
@zeffii 
About docstring syntax: I'd propose not to invent special syntax for each thing we want to add to docstring (`///`, `(#`, who knows if we will want to add inputs or outputs description for example), but to introduce some more or less regular and potentially extendable syntax. For example, I think MIME format could work:

    class SvBoxNode(bpy.types.Node, SverchCustomTreeNode):
        """
        Brief: Box
        Tooltip: Generate a Box primitive

        This is the long description in free form, it can include technical
        or historical remarks.
        """

        bl_idname = 'SvBoxNode'
        ...

Here long description is separated from headers with an empty line. If there are no headers detected, then whole docstring will be used as tooltip.

portnov:
Another way that we can use is to add specific class methods (get_tooltip(), get_brief(), ...) to main node class and override them in specific nodes.

We can also combine both approaches: get_tooltip(), get_brief() methods in base node class would parse information from docstring, but node author can override them if he wants for some reason.

zeffii:
Well ok. But search still needs to work using shorthands. Propose a change to the code that doesnt impact how search works

portnov:
ok, will do.

Just btw,
> the tooltip for each node means you are registering a new operator for every node?

This approach gives an interesting possibility: it is now possible to add nodes via standard Blender's Space menu :)

zeffii:
Yep. Whatever is most convenient for users. Ctrl+space will remain a power search feature

portnov:
Added some technical details to description. I'm not going to add new features to this PR, but there can be bugfixes. @DolphinDream , @nortikin please help @zeffii in reviewing and testing this :) For me this already works.

What to test:
* Ctrl-Space search is working the same way it was working before
* Shift-A menu works as before, but now displays correct tooltips
* T panel works as before, but now displays correct tooltips
* Registering, unregistering, reloading (F8) works as before.
* Nothing of named crashes or displays errors or prints stuff to console.

Well, I actually found #1877, but that is present in master too.

zeffii:
instead of "brief:" or in addition to it... i'd like to see "triggers:" , so for example vdmk2
```python
class ViewerNode2(bpy.types.Node, SverchCustomTreeNode):
    ''' vd View Geometry /// ViewerNode2 '''
    bl_idname = 'ViewerNode2'
    bl_label = 'Viewer Draw'
```
goes to

```python
class ViewerNode2(bpy.types.Node, SverchCustomTreeNode):
    """
    triggers: vd View Geometry
    tooltip: Viewer Node (MK2) for viewing geometry and matrices
   """
    bl_idname = 'ViewerNode2'
    bl_label = 'Viewer Draw'
```
if we're going to be parsing it using conventional methods, this might be the way to go.. unfortunately it does involve updating a few docstrings... and the search menu code..

portnov:
@zeffii 
ok, I can easily rename Brief to Triggers. I'd prefer header names with first later capitalized, more in classic rfc-822 style, though.

The search code just calls get_shorthand() method now, so it does not bother with such things as exact docstring syntax. We can even recognize several aliases for the same thing, but I do not think that several variants of syntax is better than one in this case.

As for existing docstrings - yes, they should be updated. But for now `ack --python "///"` finds only  9  nodes (of which one is an old_node), so it is not very much work.

nortikin:
ivan prytov making docs now, maybe should i say him to add something to docs?

portnov:
@nortikin 
ok, if someone wants to update docstrings - you're welcome.

portnov:
About future compatibility: mostly I'm not sure about `bl_operators` module, if it will be available in 2.8 (the naming alludes to being some sort of internal module). I tried today's build of 2.8, and the module is available. I know that python api of 2.8 is not stable yet though...

portnov:
Maybe I should better copy-paste code from bl_operators.node.NodeAddOperator.create_node, instead of calling this method? not sure...

portnov:
Well then, I think I'll better merge this and will wait for complains :) In master, there will be a bit more testers.

PR #1875, opened 2017-11-26
Implement Suzanne/Monkey primitive.

portnov:
I can't say it is very useful, but it looked strange that Sverchok did not have this primitive.

### DIFF
--- a/docs/nodes/generator/generator_index.rst
+++ b/docs/nodes/generator/generator_index.rst
@@ -19,6 +19,7 @@ Generators
    script1_lite
    sphere
    icosphere
+   suzanne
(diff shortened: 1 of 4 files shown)


portnov:
screenshot_20171126_151436

zeffii:
beside it being a useful primitive, it's also a useful example of the mechanics of a simplest node.

nortikin:
maybe sverchok-beast node?

zeffii:
Mascot node with switch between monkey and cricket

nortikin:
i have an idea, will test week later.
Cricket with different parameters.

PR #1876, opened 2017-11-26
Proper icons for some nodes

portnov:
Some nodes do not have icons, while there are more or less good standard icons in blender.
It would be nice to draw custom icons for all nodes, but...

This has some relation to #1874 , but this can be applied independently I think.


portnov:
Don't see any reason why this should not be implemented, so merging.

PR #1878, opened 2017-11-28
allows untick/retick sverchok

zeffii:


zeffii:
it silences the sverchok code related things. untested regarding the registration of monad properties. 

don't commit.

zeffii:
as for the printouts about the missing operators.. i'm not certain there's a solution for that.  In this PR i first unregister the keymaps..  then unregister the operators. In theory that would be the correct order to avoid these printouts.

zeffii:
@portnov  what do you think of this.. this does resolve the enable/disable bug.. but doesn't fix the printouts about the missing Operators. 

At first I thought it was happening because we where probably disabling sverchok with Nodeview open..and Userpreferences open, both of which would internally reference those operators (but just as strings... so.. huh? ) 

but then I closed node, and preferences and ran the `bpy.ops.wm.addon_disable(module='sverchok')`, didn't try..
```
import addon_utils
addon_utils.disable("sverchok")
```
it that does disable without printing out then possible it is the windowing system..

zeffii:
hmmm.. also `addon_utils` version hits the wm. ( what i suspect is the problem )

zeffii:
so... if we really want to dig...

maybe `__init__.py`'s `unregister` in the root of sverchok is not the first function called by the unregistration system..

portnov:
It looks like we have to put prints at the beginning of each unregister() method and see in which order they are called.

portnov:
Also, it wouldn't surprise me if blender people were going to change order of calling unregister in 2.79/2.8.

zeffii:
yeah.. or would be neat if there was a handler event for addon enable / disable

zeffii:
yeah, if that's the case i'll start with adding prints to some of the last nodes that are registered. it may be unregistered in reverse order too...

zeffii:
contrary to what I thought was happening, the `untick event` does not call unregister functions... at least not in the same way that `f8` does.

zeffii:
not even the main `__init__.py`'s unregister is called when you disable the add-on..

zeffii:
for now I think the best we can do is make it untick/retick correctly and ignore the printouts,  that would be some progress.

zeffii:
or maybe we have to register the keymap to sverchok.. i've wondered about this
```
>>> bpy.context.window_manager.keyconfigs.new('sverchok')
bpy.data.window_managers['WinMan']...KeyConfig

>>> bpy.context.window_manager.keyconfigs['sverchok']
bpy.data.window_managers['WinMan']...KeyConfig
```

zeffii:
only the `monad_propertes` registration was needed to make untick/retick work.   the missing operator issue is a different matter solved another way.

PR #1880, opened 2017-11-28
Update 2ptspline node.

portnov:
The code seems to be pre-prop_name-times. So in fact setting inputs as
properties was not supported.

This does not change algorithm; so the node still does not try to match
inputs data, I only added an exception instead of silently-not-working
in case when number of items does not match.

refs #1847


zeffii:
i think we might add default values to the spline knots/anchors like Blender does when adding a bezier. but we'd need to commute this to a `mk2` to avoid old blends producing unexpected results..

push this if you want. we'll upgrade later.

nortikin:
https://gist.github.com/d019de6023675e870de920f504bd3c1f
curve

PR #1884, opened 2017-11-29
"Objects along edge" update

portnov:
* Simplify default algorithm a bit.
* Introduce 2 new algorithms of object rotation calculation. It is in general not a trivial task to rotate a 3D object along a vector, because there are always 2 other axes of object and it is not clear where should they be directed to. So, this node supports 3 different algorithms of object rotation calculation: Householder, Tracking and Rotation Difference. In many simple cases, all these algorithms will give exactly the same result. But in more complex setups, or in some corner cases, results can be very different. So, just try all algorithms and see which one fits you better.
* For Tracking algorithm, "Up axis" parameter is introduced.
* Clean up UI a bit.

Default algorithm is the Householder, which was the only algorithm previously available. So the changes are fully backwards compatible, the new code should work in old scenes exactly the same way.

I'm more or less sure in this code, but gonna wait a couple of days before merging, in case someone has some ideas about code or smth.

### DIFF
--- a/docs/nodes/modifier_change/objects_along_edge.rst
+++ b/docs/nodes/modifier_change/objects_along_edge.rst
@@ -1,16 +1,29 @@
 Duplicate along Edge
 ====================
 
-*destination after Beta: Modifier Change*
-
 Functionality
 -------------
 
-This node creates an array of copies of one (donor) mesh and aligns it along given recipient segment (edge). Count of objects in array can be specified by user or detected automatically, based on size of donor mesh and length of recipient edge. Donor mesh can be scaled automatically to fill all length of recipient edge.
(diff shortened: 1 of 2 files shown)


portnov:
For the illustration, the same thingy with different algorithms:

Householder:
householder

Tracking:
tracking

portnov:
suzanne-spiral

enzyme69:
How did you make this one so tidy:
https://github.com/nortikin/sverchok/pull/1884#issuecomment-347963376

portnov:
@enzyme69 
https://gist.github.com/9b3e6902f0476e194f5fa471b287fb25

PR #1889, opened 2017-11-30
Vcol rgba fix1

zeffii:
DO NOT MERGE YET

Recently Blender 2.79 Buildbot builds have added a 4th channel to the Vertex Color Map , it now expects RGBA. This pull request hopes to address this update. It's a pretty cool new feature. This does mean old Blends will necessarily be broken in the following situations:

   -  if you run the new node in an old binary of Blender
   - if you run the old node in a new binary of Blender

To save us all a headache we will MK3 this node.  Sorry for any inconvenience, sometimes it's the price of progress. Make some cool stuff :)

### DIFF
--- a/index.md
+++ b/index.md
@@ -256,6 +256,7 @@
 ## Objects
     SvVertexGroupNodeMK2
     SvVertexColorNodeMK2
+    SvVertexColorNodeMK3
 
 ## Layout
     WifiInNode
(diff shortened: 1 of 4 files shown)


zeffii:
@Kosvor2 does this work as expected? i'm not sure how to access the Alpha channel from VertexColormap / Cycles Attribute node ..doesn't appear to be passed into the factor socket.

PR #1893, opened 2017-12-02
A bit better handling for SvNoDataError

portnov:
At least print name of missed socket.

Was:

      File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/core/update_system.py", line 322, in do_update_general
        node.process()
      File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/nodes/transforms/apply.py", line 43, in process
        mats_ = self.inputs['Matrixes'].sv_get(deepcopy=False)
      File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/node_tree.py", line 143, in sv_get
        raise SvNoDataError
    Node Matrix Apply (verts) had exception

now:

      File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/core/update_system.py", line 322, in do_update_general
        node.process()
      File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/nodes/transforms/apply.py", line 43, in process
        mats_ = self.inputs['Matrixes'].sv_get(deepcopy=False)
      File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/node_tree.py", line 143, in sv_get
        raise SvNoDataError(self)
    Node Matrix Apply (verts) had exception: No data passed into socket `Matrixes'

### DIFF
--- a/core/socket_data.py
+++ b/core/socket_data.py
@@ -103,12 +103,36 @@ def SvGetSocket(socket, deepcopy=True):
         else:
             if data_structure.DEBUG_MODE:
                 print("cache miss:", socket.node.name, "->", socket.name, "from:", other.node.name, "->", other.name)
-            raise SvNoDataError
+            raise SvNoDataError(socket)
     # not linked
-    raise SvNoDataError
+    raise SvNoDataError(socket)
 
 class SvNoDataError(LookupError):
-    pass
+    def __init__(self, socket=None, node=None):
+        if node is None and socket is not None:
+            node = socket.node
+        self.node = node
+        self.socket = socket
+
+        super(LookupError, self).__init__(self.get_message())
+
+    def get_message(self):
+        if not self.node and not self.socket:
+            return "SvNoDataError"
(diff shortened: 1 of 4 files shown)


portnov:
Any thoughts about how to report errors properly? I mean, display them in Blender's UI, not in console.

zeffii:
I'm all for more exact Exception messages

Sverchok is a complex beast constantly in a state of development, unlike thousands of other add-ons, it can fail in many but unspectacular ways. We can expect our users to have their terminal open (or get used to having it open).

I am not sure there's a nice UI way to display exceptions per node, other than drawing with bgl onto the nodeview as was the new approach in Sverchok Redux. But.. if we go that way it should be optional.

portnov:
Well, this PR is relatively simple change. Things like BGL will require more investigations, that will be separate PR anyway.

zeffii:
cool

PR #1894, opened 2017-12-02
Spline interpolation refactoring

portnov:
A story: I wanted to write some code that uses cubic spline interpolation similar to "Vector interpolation mk3", and... found out that we have 5 (five!) definitions of `cubic_spline` method in master. Which are all more or less copy-paste of

    # spline function modifed from
    # from looptools 4.5.2 done by Bart Crouch

(all implementations have this comment).

There is also yet another implementation, in form of two classes (more OOP-style) in geom.py by @ly29 , which is not used anywhere.

So when I started to write some new code, I was faced with an idea of writing 7th implementation... no way. So I couldn't stay from moving the implementation from "interpolation_mk3.py" (which seems to be the last) to geom.py, merging with ly29's code. So in this PR we have LinearSpline and CubicSpline classes, which are used in interpolation_mk3. I did not touch other copies for now.

Theoretically, end-user is not able to see any differences before and after this PR, as this is just a movement of the same code from one module to another.

@zeffii could you please review?

### DIFF
--- a/nodes/vector/interpolation_mk3.py
+++ b/nodes/vector/interpolation_mk3.py
@@ -28,116 +28,172 @@ from sverchok.data_structure import updateNode, dataCorrect, repeat_last
 # spline function modifed from
 # from looptools 4.5.2 done by Bart Crouch
 
+class Spline(object):
+    @classmethod
+    def create_knots(cls, pts, metric="DISTANCE"):
+        if metric == "DISTANCE":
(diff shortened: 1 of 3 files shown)


zeffii:
```diff
+ from sverchok.utils.geom import LinearSpline, CubicSpline
```
yep. this is sane :)

nortikin:
some interface principles in mk2 i use all time if something.

Issue #1895, opened 2017-12-03
Exception after recent Merge (Affects nightly Blender 2.79 build users)


zeffii:
```python
Exception in module register(): '\\scripts\\addons_contrib\\sverchok\\__init__.py'
Traceback (most recent call last):
  File "\scripts\modules\addon_utils.py", line 350, in enable
    mod.register()
 
  File "\scripts\addons_contrib\sverchok\menu.py", line 194, in get_node_idname_for_operator
    rna = getattr(bpy.types, nodetype).bl_rna
AttributeError: 'RNA_Types' object has no attribute 'SvLineNodeMK2'
```

zeffii:
since the registration changed in Blender (if using newer Blender versions 2.79 nightlies...) you can't access a node's class reference by doing `getattr(bpy.types, nodetype)` .  you have to use our lookup table instead.

portnov:
Ah. I'm using 2.79 release version.

What is the "our lookup table"?

portnov:
You mean sverchok.utils.get_node_class_reference? will try.

zeffii:
`cls = get_node_class_reference(cls_bl_idname)`

it will work fine in release, but many people don't use release anymore - including us.

zeffii:
Blender developers try to speed up the bpy.types lookups by removing as much as possible from being registered in that namespace by default. So we have to track our nodes manually using a lookup.

portnov:
@zeffii 
Hmm. I changed this locally and now my code can't find registered node by bl_idname "NodeReroute". But I can't grep such node anywhere. What is going on? Is this some special node?

zeffii:
yeah, that node is registered to bpy.types :)  i think.  it's the only node where this applies.

zeffii:
it's not really a "node" as we would define it.

portnov:
Then I'll add this exception directly to get_node_class_reference(). ok?

zeffii:
yes please.

portnov:
@zeffii 
Please test #1896. It works for me.

zeffii:
i recon you should use a nightly build for testing too.

portnov:
Well, will try :) 
But since blender api changes in nightly builds, we actually have to test both on nightly builds and on stable ones.

portnov:
@zeffii 
I tried the nightly build (today's) and it started normally. When I tried to open a scene with ViewerText node, I got

location: <unknown location>:-1
[...]

PR #1896, opened 2017-12-03
Change the way of accessing registered node classes.

portnov:
* Move menu module registration even later
* Add exceptions for NodeReroute class

refs #1895, #1874.

### DIFF
--- a/menu.py
+++ b/menu.py
@@ -133,7 +133,7 @@ class SverchNodeItem(object):
             return self.get_node_class().bl_rna.name
 
     def get_node_class(self):
-        return getattr(bpy.types, self.nodetype)
+        return get_node_class_reference(self.nodetype)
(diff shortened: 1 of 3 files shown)


menu.py
```diff
@@ -204,11 +208,13 @@ def draw_add_node_operator(layout, nodetype, label=None, icon_name=None, params=
     """
 
     default_context = bpy.app.translations.contexts.default
-    node_rna = getattr(bpy.types, nodetype).bl_rna
+    node_rna = get_node_class_reference(nodetype)
```
zeffii: isn't this missing a `.bl_rna` at the end?

portnov: Added. But I do not see any difference. It seems node class and it's rna both have the same bl_label property.

PR #1898, opened 2017-12-03
bpy.types cleanup in ViewerText

portnov:
This removes direct references to bpy.types.*Socket from viewer_text_mk2 node, for compatibility with nightly 2.79 builds.


zeffii:
enough for now :)

PR #1905, opened 2017-12-05
Object Out  to Empty Out

vicdoval:
Added "Object output" to node


vicdoval:
Just added tree lines of code to have the output in the node itself to play nice with the new get- set nodes.
I think it does not worth to call it MK2..

zeffii:
yeah, when adding a socket at the end of the current socket lists (in or out..or both) it won't cause any problems with existing setups. indeed no need to mk2.

PR #1909, opened 2017-12-07
Add multi socket input to Matrix Math node

DolphinDream:
- add multi-socket input feature to the node for the multiple input operations
- add new operation (Basis) to extract basis vectors from matrix
- cleanup the UI and nodes shown per operation
- cleaned up the code
- update documentation

sv-4dmatrixmath-demo3

### DIFF
--- a/docs/nodes/matrix/matrix_math.rst
+++ b/docs/nodes/matrix/matrix_math.rst
@@ -4,7 +4,7 @@ Matrix Math
 Functionality
 -------------
 
-Matrix Math node allows for various matrix operations to be performed on the input matrices, such as: Multiply, Invert and Filter.
+Matrix Math node allows for various matrix operations to be performed on the input matrices, such as: Multiply, Invert, Filter, Basis.
 
 Inputs
 ------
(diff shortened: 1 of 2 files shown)


zeffii:
looks good, and useful :)

DolphinDream:
Btw, one question .. when removing sockets while iterating over sockets I thought that the proper way to do this is by using a separate list, but I’m not sure if this is necessary (as I see in many places sockets being erased from socket list while iterating over the list). Perhaps it’s ok to do this with while loops but may not be ok with for loops? 

```python
ss = [s for s in inputs]
for s in ss:
    if s != inputs["A"]:
        inputs.remove(s)

# could this just be:

for s in inputs:
    if s != inputs["A"]:
        inputs.remove(s)
```

zeffii:
if you find code that does it one way (removing the socket while iterating over the sockets) and it works.. then i think the iterator of `self.inputs/outputs` is OK with it. (else we wouldn't write that code)

We have no rule about this. If it breaks in the future we fix it.
```python
for socket_name in excess_socket_names:
    self.inputs.remove(self.inputs[socket_name])
```

PR #1911, opened 2017-12-07
Introduce a proper logging facility

portnov:
I hope, this movement will enhance usability a lot. But, for now only core support of logging was implemented; it is to be supported across the whole codebase yet, see details below.

## For users

Sverchok is going to write all it's messages not only to console window, but also:
* to text buffer within Blender's scene (optional)
* to external file somewhere on disk (optional)

It is also possible to specify minimum severity level of messages that you want to see in the log.

screenshot_20171207_230748

So, when/if this (and possibly following PRs) is merged, the recommendation for users to have console window opened while working with Sverchok is going to be replaced with recommendation to have blender's text editor frame opened, which is much friendlier I think.

## For developers

The facility is implemented by means of standard python's `logging` module. All glue code lies in `sverchok.utils.logging`. 

The implementation is more or less straightforward; exceptions are:

* Due to "interesting" order of importing and initializing different things (bpy objects, addon preferences, ...) in blender's python environment, it is not possible to do just `logger = logging.getLogger(__name__)` in the beginning of each module, as many practices say; this is going to lead to exceptions. We have to obtain a logger at each event. To simplify things, a "standard" method family `sverchok.utils.logging.debug`, `info`, `warning`, `error`, `exception` is introduced (these methods use caller module name as logger name by default):

      from sverchok.utils.logging import debug

      def something():
          ...
          debug("got addon")
          ...

* It is possible also to use  `logging.debug` and siblings if you do not want to bother with logger name, and just write to root logger.
* For base node class, a family of logging methods is provided. So within any node class, you can just replace `print(xxx)` with `self.debug(xxx)`.
* It appears to be not possible to initialize output to blender's text buffer during `register()`. So, it is registered at the first call of `sverchok.utils.logging.getLogger()`, i.e. basically at first logged event.
* Due to mentioned initialization order things, there going to be some problems if one wants to replace different prints that are done during initialization with logging calls. I think it is simpler to do not touch these initialization printouts for now.

## Example of log output

      2017-12-07 23:34:00,558 [INFO] root: Initializing Sverchok logging. Blender version 2.79 (sub 0), Sverchok version (0, 5, 9, 6)
      2017-12-07 23:34:00,558 [DEBUG] root: Current log level: DEBUG, log to text buffer: sverchok.log, log to file: sverchok.log
      2017-12-07 23:34:00,558 [DEBUG] SvBoxNode: hello from the box
      2017-12-07 23:34:09,571 [DEBUG] SvBoxNode: hello from the box
      2017-12-07 23:34:09,683 [ERROR] sverchok.core.update_system: Node Rotation had exception: No data passed into socket `Vertices'
        File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/core/update_system.py", line 323, in do_update_general
          node.process()
        File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/nodes/transforms/rotation.py", line 139, in process
          Vertices = self.inputs['Vertices'].sv_get()
        File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/node_tree.py", line 297, in sv_get
          raise SvNoDataError(self)
      sverchok.core.socket_data.SvNoDataError: No data passed into socket `Vertices'

## What next

As I said in the beginning, this adds only core support of logging. We have a lot of `print`-s across the whole codebase. I do not think that it worth the effort to go and replace all of them right now. Instead, I'd propose to update some most used / most verbose parts of code now (either within this PR or in separate, to be discussed). All other code can be updated gradually: just when you are fixing a regular bug in some node, replace all prints in it, it is not going to take much time.

@nortikin @zeffii @DolphinDream  please review and test.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -158,8 +158,17 @@ if reload_event:
 import bpy
 from sverchok.utils import ascii_print, auto_gather_node_classes, node_classes
 from sverchok.core import node_defaults
+from sverchok.ui.development import get_branch, get_hash
 
-
+def get_version_string():
+    version = ".".join(map(str, bl_info['version']))
+    branch = get_branch()
+    if branch:
+        version += ", branch " + branch
+        hash = get_hash()
+        if hash:
+            version += ", commit " + hash
+    return version
 
 def register():
     for m in imported_modules + node_list:
@@ -170,7 +179,7 @@ def register():
     # this is used to access preferences, should/could be hidden
     # in an interface
     data_structure.SVERCHOK_NAME = __name__
-    print("** version: ", bl_info['version']," **")
+    print("** version: ", get_version_string()," **")
     print("** Have a nice day with sverchok  **\n")
     ascii_print.logo()
     node_defaults.register_defaults()
--- a/__init__.py
+++ b/__init__.py
@@ -158,17 +158,7 @@ if reload_event:
 import bpy
 from sverchok.utils import ascii_print, auto_gather_node_classes, node_classes
 from sverchok.core import node_defaults
-from sverchok.ui.development import get_branch, get_hash
-
-def get_version_string():
-    version = ".".join(map(str, bl_info['version']))
-    branch = get_branch()
-    if branch:
-        version += ", branch " + branch
-        hash = get_hash()
-        if hash:
-            version += ", commit " + hash
-    return version
+from sverchok.ui.development import get_version_string
 
 def register():
     for m in imported_modules + node_list:
--- a/__init__.py
+++ b/__init__.py
@@ -88,6 +88,7 @@ utils_modules = [
     "voronoi", "sv_script", "sv_itertools", "script_importhelper", "sv_oldnodes_parser",
     "csg_core", "csg_geom", "geom", "sv_easing_functions",
     "snlite_utils", "snlite_importhelper", "context_managers",
+    "logging",
     # UI text editor ui
     "text_editor_submenu", "text_editor_plugins",
     # UI operators and tools
--- a/nodes/text/debug_print.py
+++ b/nodes/text/debug_print.py
@@ -61,7 +61,7 @@ class SvDebugPrintNode(bpy.types.Node, SverchCustomTreeNode):
             return
         for i, socket in enumerate(self.inputs):
             if socket.is_linked and self.print_socket[i]:
-                print(socket.sv_get(deepcopy=False))
+                self.debug(socket.sv_get(deepcopy=False))
 
 def register():
     bpy.utils.register_class(SvDebugPrintNode)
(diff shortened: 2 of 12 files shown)


zeffii:
finally.

"to external file somewhere on disk (optional)"  this can be in `datafiles/sverchok/debug.log`

__init__.py
```diff
@@ -158,8 +158,17 @@ def make_node_list():
...
+
+def get_version_string():
+    version = ".".join(map(str, bl_info['version']))
+    branch = get_branch()
+    if branch:
+        version += ", branch " + branch
+        hash = get_hash()
+        if hash:
+            version += ", commit " + hash
+    return version
```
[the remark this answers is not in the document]
portnov: It usually says nothing if someone is reporting an error "in sverchok 0.5.9.6" :)

nodes/text/debug_print.py
```diff
@@ -61,7 +61,7 @@ def process(self):
             return
         for i, socket in enumerate(self.inputs):
             if socket.is_linked and self.print_socket[i]:
-                print(socket.sv_get(deepcopy=False))
+                self.debug(socket.sv_get(deepcopy=False))
```
zeffii: i'm pretty sure this node should always print its data, regardless of logging level?

portnov: But, it is called "debug print"... :) I'm not sure. I can change this to self.info instead, for example, to make output visible by default. Or, I can revert this back to print, but then the output will not be visible in logs, only in console.

zeffii: the node is meant to output the content of the incoming socket into console, that's what I use it for sometimes. so it would be a bit of pain to have to enable an option in preferences for it to work..and disable it when your done..

portnov: It is also possible to set logging level to DEBUG for this one line and then set it back. Not sure if it is a good idea.

zeffii: `self.info()` makes a bit more sense in this specific instance.

portnov:
I do not see other modules (except for already updated) that i'd want to update right now. You?

zeffii:
nope. i'll take my time adding support in SNLite :)

zeffii:
thanks for doing this @portnov  . it will be neat, and crimps my todo by 1 item. :)

zeffii:
i'd like to add "loading default dict.." to only appear in debug mode.

portnov:
Yeah, the main restriction is that we can't use log to internal text buffer from `register()`. It is possible to log that to console and file, though; i'm going to remove message about "Can't initialize logging to internal buffer...", it does not really tell anything useful. And I have some ideas about how to make initialization a tiny bit clearer.

PR #1913, opened 2017-12-08
rclick menu improvements

zeffii:
- [x] add `ve` as match of string start, for verts/vecs/etc
~- [ ] add location dependant stethoscope connection~  <-- not possible


zeffii:
i recall (maybe incorrectly) that it was possible to get the location of sockets, but can not seem to find in the API any longer. This means it's harder (and uglier) to determin the socket desired for the stethoscope connection. For now i might just add stethoscope and connect it to `outputs[0]` of the active_node

Durman:
> For now i might just add stethoscope and connect it to outputs[0] of the active_node

I think that it is better than nothing. Great

PR #1920, opened 2017-12-09
Profiling facility

portnov:
This adds some more advanced profiling facilities that we already have. I started to write this when wrote "bend along surface" node, to dig into details of it's performance.

In addon preferences, there is new option introduced, "Profiling mode", with following values:
* Disable - profiling is disabled (default)
* Only marked methods - only methods manually marked by developer with `@profile` (`sverchok.utils.profile.profile`) decorator are profiled. By default no methods are decorated in such way, so this option profiles nothing. These marks are supposed to be added either locally or in some debugging branches.
* Whole update process - the whole `do_update_general` method is profiled.

screenshot_20171211_223651

If profiling is not disabled in settings, there are two new buttons in nodetree N panel (near "Update all"):
* Start/stop profiling - start/stop gathering statistics. Statistics are not gathered by default.
* Dump stats - dump collected statistics to the log. Just writes a message if we do not have statistics to dump yet. In the popup window you can specify how to sort the data.

screenshot_20171211_223841

So the usual workflow if you want to profile something is:
* If you want to profile the whole nodetree update process, set profiling mode = "Node tree update" in settings.
* Or if you want to profile the specific node or some set of methods:
  * Edit the source and mark interesting methods with `@profile` decorator
  * Set profiling mode = "Marked methods only" in settings.
* Click "Start profiling"
* Do something that calls your methods, for example press "Update All", one or several times
* Click "Stop profiling"
* Repeat if necessary
* Click "Dump statistics"
* Observe the log.

Under the hood, standard python's `cProfile` module is used.

To generalize a bit, there is concept of "profiling section" used. Profiling section is a named set of methods to be profiled. The `@profile` decorator can be provided with name of profiling section, by default section is "MANUAL". The decorator only actually profiles the method if provided name of section equals to `profile_mode` option in settings. For now, we have only two sections: MANUAL for manually marked methods and UPDATE for `do_update_general` method. Later we may want to add other sections.

Example output: https://gist.github.com/portnov/a394ac0b4793f639c58c8e0f18cf5d47

With my examples, i do not see any performance overhead of profiling, it is less than measurement error. But it is possible that there is some overhead in some cases.

@zeffii @DolphinDream @nortikin please review and test.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -88,7 +88,7 @@ utils_modules = [
     "voronoi", "sv_script", "sv_itertools", "script_importhelper", "sv_oldnodes_parser",
     "csg_core", "csg_geom", "geom", "sv_easing_functions",
     "snlite_utils", "snlite_importhelper", "context_managers",
-    "logging",
+    "profile", "logging", 
     # UI text editor ui
     "text_editor_submenu", "text_editor_plugins",
     # UI operators and tools
--- a/core/update_system.py
+++ b/core/update_system.py
@@ -25,6 +25,7 @@ from mathutils import Vector
 from sverchok import data_structure
 from sverchok.core.socket_data import SvNoDataError, reset_socket_cache
 from sverchok.utils.logging import debug, info, warning, error, exception
+from sverchok.utils.profile import profile
 import sverchok
 
 import traceback
@@ -303,6 +304,7 @@ def reset_error_nodes(ng):
         del ng["error nodes"]
 
 
+@profile(section="UPDATE")
 def do_update_general(node_list, nodes, procesed_nodes=set()):
     """
     General update function for node set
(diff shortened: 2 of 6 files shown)


portnov:
Merging. Please write if you will see any performance impact with profiling disabled.

nortikin:
cool. never thought of this workflow before. will use in new nodes.

nortikin:
replace print() with what?

portnov:
@nortikin 
In node classes — with self.debug() (or self.info, self.warning, self.error, self.exception). Outside nodes — sverchok.utils.logging.debug ... exception.

portnov:
Also please note: these functions support python printf-style formatting (via %s etc). So you can do

```
self.debug("X is %s", x)
```

This is (at least theoretically) more preferrable, comparing to concatenating or formatting messages by hand, because actual formatting will only be done in case when debug output is enabled in settings, you can save some CPU cycles.

nortikin:
how did i leave without it before?

PR #1922, opened 2017-12-10
Gz zip archive handling

zeffii:
- handle gz files (loads from online source)


zeffii:
not sure if i'm ready to replace zip creation with .gz 
- the code is not significantly different in length / complexity 
- the zip code would need to be supported anyway for backwards support.

zeffii:
some reuse could work between zip/gz archiving..
 export code..

```python
import os
import gzip
import shutil
from time import gmtime, strftime

import bpy


def propose_archive_filepath(blendpath, extension='zip'):
    """ disect existing filepath, add timestamp """
    blendname = os.path.basename(blendpath)
    blenddir = os.path.dirname(blendpath)
    blendbasename = blendname.split('.')[0]
    raw_time_stamp = strftime("%Y_%m_%d_%H_%M", gmtime())
    archivename = blendbasename + '_' + raw_time_stamp + '.' + extension

    return os.path.join(blenddir, archivename)


def compress_blend_as_gzip(self):

    blendpath = bpy.data.filepath

    if not blendpath:
        msg = 'you must save the .blend first before we can compress it'
        # self.report({'INFO'}, msg)
        print(msg)
        return {'CANCELLED'}

    with open(blendpath, 'rb') as f_in:
        gzip_filepath = propose_archive_filepath(blendpath, 'gz')
        with gzip.open(gzip_filepath, 'wb') as f_out:
            shutil.copyfileobj(f_in, f_out)
        print('saved to', gzip_filepath)

    return {'FINISHED'}


compress_blend_as_gzip('dummy')
```
import test: 
[line to face2_2017_12_10_10_17.gz](https://github.com/nortikin/sverchok/files/1545513/line.to.face2_2017_12_10_10_17.gz)

portnov:
Cool, this works for me.

zeffii:
[line to face2_2017_12_10_20_06.gz](https://github.com/nortikin/sverchok/files/1546030/line.to.face2_2017_12_10_20_06.gz)

gzipped by the `blend_to_archive` Operator with `.archive_ext='gz'`

PR #1923, opened 2017-12-10
Archive options

zeffii:
add option to archive as .gz


zeffii:
@portnov 
i'm too lazy to add row.split() ..  so hopefully just a small icon will do.
image

PR #1933, opened 2017-12-15
Updates in profiling subsystem.

portnov:
* Add a button to dump statistics to binary file in cProfile format. Such file can be parsed, for example, by gprof2dot utility.
* UI improvements:
  * Separate button to reset statistics
  * Icons
  * Do not show buttons which can not be used at the moment.

### DIFF
--- a/ui/sv_panels.py
+++ b/ui/sv_panels.py
@@ -277,12 +277,16 @@ class SverchokToolsMenu(bpy.types.Panel):
 
         addon = context.user_preferences.addons.get(sverchok.__name__)
         if addon.preferences.profile_mode != "NONE":
-            profile_row = layout.row(align=True)
+            profile_col = layout.column(align=True)
             if profile.is_currently_enabled:
-                profile_row.operator("node.sverchok_profile_toggle", text="Stop profiling")
+                profile_col.operator("node.sverchok_profile_toggle", text="Stop profiling", icon="CANCEL")
             else:
-                profile_row.operator("node.sverchok_profile_toggle", text="Start profiling")
-            profile_row.operator("node.sverchok_profile_dump", text="Dump data")
+                profile_col.operator("node.sverchok_profile_toggle", text="Start profiling", icon="TIME")
+            if profile.have_gathered_stats():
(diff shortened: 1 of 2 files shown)


portnov:
gprof2dot, in turn, can generate more interesting pictures than text output of cProfile:

test

Durman:
Can you explain what happens on this picture? : /

portnov:
@Durman 
The top right red node is the main method which updates the whole tree. Arrows mean method calls. So this is basically a call graph. Nodes are colorized from red (each execution takes near 100% of time) to blue (takes almost 0 time).  Numbers with x mean how many times method was called. So from this picture I can see that 1) almost all time was spent in "eval" method at geom.py, line 722; 2) eval method at line 575 was called very very many times.
 More exactly numbers are documented in readme at https://github.com/jrfonseca/gprof2dot (scroll down to "Output" section).

DolphinDream:
@portnov nice profiling tool. I wonder if the number of digits in the statistics can be set to other than the default 3 digits. The percall values are usually zero within the default 3 digits after the decimal so these values are sometimes not useful.

portnov:
@DolphinDream i'm not exactly sure. Please refer to documentation: https://docs.python.org/2/library/profile.html#pstats.Stats
Theoretically, you should be able to extract required information from Stats objects and format it the way you want.

DolphinDream:
Also it would be cool if the SV profiler options would allow you to set a number of call counts to run the profiling for and automatically stop after it reaches that count. This way I can run the same setup for different versions of the node for the exact number of counts and compare the results. Currently I have to keep dumping the data and see if the total count is the same as in previous runs so I can compare apples to apples. Per call statistics is an average and would help not having to run for specific number of calls, but as I mentioned earlier these values are so small they appear as zero.

portnov:
@DolphinDream instead you can trigger methods that you need to profile programmatically. For example, in you debug branch, you can add a button to some panel, which does "for i in range(100000): call_my_method()". You will have to mark that method with `@profile` decorator for profiler to "see" it.

PR #1939, opened 2017-12-16
Fix the name for generator colors

DolphinDream:
The generator category name was changed recently from Generators to Generator, which fix the indexing of the generator colors in the themes. This update fixes the indexing string to match the new "Generator" name.

### DIFF
--- a/ui/color_def.py
+++ b/ui/color_def.py
@@ -35,7 +35,7 @@ default_theme = {
     "Text": (0.5, 0.5, 1),
     "Scene": (0, 0.5, 0.2),
     "Layout": (0.674, 0.242, 0.363),
-    "Generators": (0, 0.5, 0.5),
+    "Generator": (0, 0.5, 0.5),
 }
 
 nipon_blossom = {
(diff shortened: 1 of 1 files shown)


DolphinDream:
Btw, I’m not sure why “Generators”  was changed to “Generator” in index.md, but if we plan to use singular nouns for categories, shouldn’t we do that for every category name ?

```
Transforms -> Transform
Generators Extended -> Generator Extended
Analyzers -> Analyzer
Modifiers -> Modifier
List Masks -> List Mask
List Mutators -> List Mutator
Objects -> Object
```

zeffii:
you will need to also adjust the documentation categories.

zeffii:
but.. this stuff isn't really high on my radar

nortikin:
it was syncronized to docs....

DolphinDream:
@nortikin so should this fix be merged into master as is?

zeffii:
pushed since this only affects themes.

while I also have issues with inconsistent naming of the categories.. it's more important that they match the category names found in the docs and subsequent doc indices. So ... this should be good now.

PR #1942, opened 2017-12-17
Bend surface ungroup

portnov:
Added new option "Grouped". If checked, then the node expects list of     lists of vertices (level 4) instead of list of vertices (level 3).    Output always has corresponding nesting level.
    
Default value Grouped = true corresponds to old behaivour.

So, basically, instead of passing output via "List Del Levels", you can now just uncheck the "Grouped" option.


nortikin:
join flag better

PR #1943, opened 2017-12-17
Data Shape node.

portnov:
This is a very simple node, that displays human-readable description of shape of data that are passed to input. IMHO it can be used instead of stethoscope in many situations.

Example:

screenshot_20171217_174304


nortikin:
instead of text viewer node more than stheto...
because viewer text shows structure first of all,
cool!

PR #1945, opened 2017-12-17
but this works..

zeffii:


zeffii:
@portnov

portnov:
Agreed. Unnecessary copy-paste from Note node.

zeffii:
cool

PR #1947, opened 2017-12-18
add changes to small things doc

zeffii:


zeffii:
yeah makes sense.

zeffii:
renaming to come later.

PR #1948, opened 2017-12-18
Use logging functions in import code.

portnov:
This is basically a s/print/debug/. I tried to choose severity level according to context.

@zeffii  please review.

### DIFF
--- a/utils/sv_IO_panel_tools.py
+++ b/utils/sv_IO_panel_tools.py
@@ -674,7 +674,7 @@ def import_tree(ng, fullpath='', nodes_json=None, create_texts=True):
         if isinstance(nodes_json, str):
             nodes_json = json.loads(nodes_json)
             debug('==== loading monad ====')
-        print(('#' * 12) + nodes_json['export_version'])
+        info(('#' * 12) + nodes_json['export_version'])
 
         ''' create all nodes and groups '''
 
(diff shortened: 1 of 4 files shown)


utils/sv_IO_panel_tools.py
```diff
@@ -218,17 +219,17 @@ def create_dict_of_tree(ng, skip_set={}, selected=False):
         for k, v in node.items():
 
             if not isinstance(v, (float, int, str)):
-                print('//')
-                print(node.name, ' -> property:', k, type(v))
+                debug('//')
```
zeffii: i'm a little torn on this..i suppose if I really need to see this message i'd switch on debug mode..

zeffii: (updated my comment)

utils/sv_IO_panel_tools.py
```diff
@@ -674,8 +673,8 @@ def generate_layout(fullpath, nodes_json):
         # this will handle both scenarios
         if isinstance(nodes_json, str):
             nodes_json = json.loads(nodes_json)
-            print('==== loading monad ====')
-        print('#' * 12, nodes_json['export_version'])
+            debug('==== loading monad ====')
+        print(('#' * 12) + nodes_json['export_version'])
```
zeffii: info ? instead of print?

zeffii:
nothing jumps out at me that I don't agree with.

utils/sv_IO_panel_tools.py
portnov: ok, just overlooked

Issue #1936, opened 2017-12-16
Common "bpy.objects" class for the Viewers that generate objects.


zeffii:
### This is for after expander sockets are merged.

Right now we have quite a collection of nodes that output and update bpy.objects of type 'CURVE/MESH/METABALL`. all of these nodes share a considerable amount of 'procedural' code with respect to 
   - `getting children` of the node  (`objs = `)
   - generating object names
   - setting materials
   - generating materials
   - cleanup orphan objects
   - hide / hide_select / hide_render / select operations.
   - easy auto update the node switches https://github.com/nortikin/sverchok/issues/1930
   - include layer handling
   - include per object material passing.

When i started out with BMViewer..and later CurveViewer it was still manageable but with the 6 or 7 nodes that generate bpy.objects now duplicating these things i'd like to gradually streamline these nodes and allow an easier way to add features to all nodes of this "class" at the same time.

then consider pulling them out of Beta. unibrows and all.

portnov:
+1. Also I have beginnings of "Lamp out" node, which has all the same problems.

zeffii:
the PR is here. it's not in a working state yet ( and has no users to depend on it yet.. )

https://github.com/nortikin/sverchok/pull/1951

this may take a while.

zeffii:
am closing this Issue to focus the talk in one place. https://github.com/nortikin/sverchok/pull/1951

PR #1949, opened 2017-12-18
Lamp out node (WIP)

portnov:
TODO:
* [ ] #1936
* [x] Use SvColorSocket
* [ ] documentation

### DIFF
--- a/index.md
+++ b/index.md
@@ -252,6 +252,7 @@
     SvObjInLite
     SvObjEdit
     SvFrameInfoNodeMK2
+    SvLampOutNode
     SvEmptyOutNode
     SvInstancerNode
     SvDupliInstancesMK4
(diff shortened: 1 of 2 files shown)


portnov:
screenshot_20171218_201451

nortikin:
just ne rabotaet
colors defined by one object of colors and every color in separate object both approaches not work.


http://pasteall.org/pic/fe2e517b4e30f4acdd019701c5630213

Kosvor2:
@nortikin are you tested in current master or in this branch? I tried in master and
it works for both metods:
lamp_colors234

nortikin:
https://github.com/nortikin/sverchok/blob/master/nodes/viz/lamp_out.py#L210
and
https://github.com/nortikin/sverchok/blob/master/nodes/viz/lamp_out.py#L290
something wrong with this line and all next lines.
and why color without alpha come error? https://github.com/nortikin/sverchok/blob/master/nodes/viz/lamp_out.py#L248 here is no alpha needed

PR #1952, opened 2017-12-19
HexaGrid node MK1 - Replace join (OFF default) with separate (OFF default) 

DolphinDream:
- replace join with Separate option
The Hexa tiles in a grid are joined by default (Separate OFF). This provides the typical SV outputs: [[v11, v12,..., v16, v21, v22,..,v26,...,vn1, vn2,.. vn6]]. The Separate ON will separate the tiles into individual lists resulting in a [[[v11,v12,..v16], [v21, v22, ..., v26],...[vn1, vn2,...,vn6]]] output.

- Update documentation
- Adjust the UI
- minor code cleanup
- move old Hexa Grid node to the old nodes


DolphinDream:
sv-hexagridnodemk1

zeffii:
Minus the middle hex and you got JWST :)  can't wait to see the images that returns..

DolphinDream:
Yeah :) I can’t wait either. Btw, trying to figure out how to unfold the hexa grid from one tile to the fully open JWST grid.. not an easy task.. if I could only find an easy way to parent the tiles to one another and define axis of rotation (hinges).

DolphinDream:
I separate the MK1 transition into 3 commits to allow for an easy comparison between the old and the new (MK1) node. In general I think this is a good steps to follow when creating new nodes if review of the new node is desirable (useful if the new node is not a complete rewrite of the old node but an incremental update):

1. Clone the old node into a MK# prefixed node (class, filename, doc)  and add to index.md into an initial commit.
2. Do developing on the MK1 node (class/doc) into a separate commit (this will allow reviewers to see the changes to the new node versus the initial node) 
3. Move the old node to the old_nodes section and remove old node documentation and its reference from index.md

DolphinDream:
If there are no objections I can merge this into master.

DolphinDream:
Btw, @zeffii what’s the recommendation on the old nodes documentation ? the nodes are kept around in the old_nodes folder but I can’t see their documentation anywhere. Was the old documentation intentionally removed from history once a node migrates to old_nodes? Should the documentation for old nodes be kept around ? (i wonder if there is a case when it would make sense).

zeffii:
You're doing fine @DolphinDream.  That is also how I tend to approach more invasive changes.  
The old node doc is a valid question, I don't have any doctrine about old node docs. There's always the git history..

zeffii:
i could add logic into the "view docs" part of the Node properties panel, to deal specifically with bl_idnames loaded from the old_nodes directory. But, and this is just my opinion - maybe not what @nortikin would prefer - I'd prefer to drop docs of old_nodes and maybe include the hash of the last known working state of the old doc, and send the reader to that instead.

But it's not terrible if we can't be perfect in these things.

nortikin:
no problem for drop old docs

PR #1957, opened 2017-12-19
Color input node

portnov:
Maybe better call it Color Picker? I named it "Color Input" for better     ressemblance with Matrix Input, List Input etc.
    
It is also automatically added to 3D panel, if enabled.

### DIFF
--- a/docs/nodes/vector/color_input.rst
+++ b/docs/nodes/vector/color_input.rst
@@ -1,0 +1,32 @@
+Color Input
+===========
+
+Functionality
+-------------
+
+This node allows you to specify color by using a colorpicker widget. In a sense, this is an analog of "Matrix Input" or "List Input" nodes, but for colors.
+
+Inputs
+------
+
+This node does not have any inputs.
+
+Parameters
+----------
+
+This node has the following parameters:
+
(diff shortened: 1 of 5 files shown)


DolphinDream:
Should we have the current Color In node be a combination of the color picker and the Color In node ? The color in node seems to have the parameters of the builtin color picker already.. (RGB,HSV tabs _ values). A color picker node with the color in input sockets would look something like this:

sv-colorinnode
 
When sockets are connected the picker is less relevant of course. Perhaps if the inputs are connecter the picker sliders could change into a tile of all the colors represented by the input values like a gradient maybe or a color swatch (as a way to visualize the output?

sv-colorinnode2

DolphinDream:
Come to think about it, a color swatch node could be interesting.. with the ability to select a color, add remove colors etc.

portnov:
We need a decision here: whether we a) merge this, or b) add similar functionality to "Color In" node (smth like @DolphinDream  proposed). We obviously do not need both at the same time.

nortikin:
not need both

portnov:
screenshot_20180224_231605

portnov:
No objections for long time, so I take freedom to merge this :)

Issue #1914, opened 2017-12-08
Line into face in 2D space.


Durman:
I want to create algorithm that will add thickness for lines in 2D. It absents in Sverchek as I know. I imagine like this:
![2017-12-08_12-57-01](https://user-images.githubusercontent.com/28003269/33758468-c7cf63b4-dc17-11e7-8985-59bf9a4d6cc6.png)
I think Initially it is necessary to start from an analyze and understand how many neighbors have each point also it is necessary to counting angles between edges. After that it is possible to start a bypass around the mesh by edges. 
For first variant the pass will be 01,10 ; second 01,12,21,13,31,10, third 02,21,13,30 and 03,31,12,20.
![2017-12-08_13-29-57](https://user-images.githubusercontent.com/28003269/33759473-e67d11d6-dc1b-11e7-9eb4-bf9209ea2870.png)
For measurement angles it is better to use[ angle_signed](https://docs.blender.org/api/blender_python_api_2_78c_release/mathutils.html#mathutils.Vector.angle_signed). It have unique values for all 360 degrees.

zeffii:
it's difficult :)  especially for more interesting edge-nets  

https://github.com/nortikin/sverchok/issues/623 i never got to the end of it.

Durman:
@zeffii 
Either way I'll take a try and I'll write my experience here.

enzyme69:
A silhouette of 3D shapes is also hard I think.
-- 


------------------------------------------------------------
- Jimmy Gunawan
http://blendersushi.blogspot.com.au
http://mayaspiral.blogspot.com.au
http://houdoodles.blogspot.com.au
http://puppetar.blogspot.com.au

Durman:
I already have some result. )

![lit to plan](https://user-images.githubusercontent.com/28003269/33798360-53229bb6-dd30-11e7-8383-d7fa3e3b4cc5.gif)

Durman:
I have finished. It is ready for testing. I have done it for SLnode.
[line to face2.zip](https://github.com/nortikin/sverchok/files/1545327/line.to.face2.zip)
```Python
"""
in verts_in v d=[] n=1
in edges_in s d=[] n=1
in shift s d=1.0 n=2
out verts_out v
out edges_out s
"""
from mathutils import Vector, Matrix
from math import radians, pi, sin

verts_in = [Vector(v).to_2d() for v in verts_in]

#Searching neighbours for each point
neighbours = [[] for _ in range(len(verts_in))]
[...]

PR #1958, opened 2017-12-20
New offset line node.

Durman:
My first node)) not sure that I did all right :/
#1914


nodes/modifier_change/offset_line.py
```diff
@@ -0,0 +1,210 @@
...
+        # avoid zero offset
+        if not shift:
+            return
+        
+        verts_out = []
+        faces_out = []
+
+        from mathutils import Vector, Matrix
+        from math import radians, pi, sin
+        from mathutils.geometry import normal
```
zeffii: stick imports at the top of the file, around the bpy import. See how other nodes do this.  

- 1.  standard python imports, (math, collections... etc)
- 2. bpy/blender imports, (mathutils...etc)
- 3. sverchok imports (utils..etc)

nodes/modifier_change/offset_line.py
```diff
@@ -0,0 +1,210 @@
...
+    Only X and Y dimensions of input points will be taken for work.
+    """
+    bl_idname = 'SvOffsetLineNode'
+    bl_label = 'Offset line'
+    bl_icon = 'OUTLINER_OB_EMPTY'
+
+    offset = FloatProperty(
+        name='offset', description='distance of offset',
+        default=0.1,
+        options={'ANIMATABLE'}, update=updateNode)
```
zeffii: `option={'ANIMATABLE'}`  is not needed, properties have this attribute set to animatable by default.

nodes/modifier_change/offset_line.py
```diff
@@ -0,0 +1,210 @@
...
+        self.outputs.new('StringsSocket', "Faces")
+        self.outputs.new('StringsSocket', "OuterEdges")
+        self.outputs.new('StringsSocket', "VersMask")
+
+    def process(self):
+
+        if not ((self.outputs['Vers'].is_linked or self.outputs['Faces'].is_linked
+              or self.outputs['OuterEdges'].is_linked or self.outputs['VersMask'].is_linked) 
+              and (self.inputs['Vers'].is_linked or self.inputs['Edgs'].is_linked)):
+            return
```
zeffii: I would first test if the `input` vers and edges is linked. _Then_ test if there's also useful output links.

zeffii:
i've left a few comments, and have one bigger issue ( don't worry it will be cool ) with the current form of  the process function.

we need to handle multiple input meshes, users expect this, this is the common paradigm for most of our nodes. (except really painfully complicated ones)

what you have now is 
```text

def process(self):
   verts_in = socket*.get()[0]
   edges_in = socket*.get()[0]

   big_inline_code operating on `verts_in, edges_in`

   outputsockets*.sv_set( single verts )
   outputsockets*.sv_set( single faces)
   outputsockets*.sv_set( single masks)
```
what you really want is 
```text

def big_mesh_function(verts, edges):
    massive code
    return geometry

def process(self):
   verts_in = socket*.get()
   edges_in = socket*.get()
   ....
   geometry_out = []
   for verts, edges in zip(verts_in, edges_in):
        geometry = big_mesh_function(verts, edges):
        geometry_out.append(geometry)
   ...

   outputsockets*.sv_set( all verts )
   outputsockets*.sv_set( all faces)
   outputsockets*.sv_set( all masks)
```

zeffii:
something like this
```python
    from mathutils import Vector, Matrix
    from math import radians, pi, sin
    from mathutils.geometry import normal

    TWO_PI = 2 * pi

    def calc_angle(p1,p2):
        # Convert to from 0 to 2pi angle
        angle = p1.angle_signed(p2)
        return angle + TWO_PI if angle < 0 else angle


    def offset_mesh(verts_in, edges_in, shift, generate_outer_edges, generate_verts_mask):
        """
        take an input mesh (verts + edges ) and an offset property and generate the resulting geometry
        """

        verts_out = []
        faces_out = []

        verts_in = [Vector(v).to_2d() for v in verts_in]

        #Searching neighbours for each point
        neighbours = [[] for _ in verts_in]
        for edg in edges_in:
            neighbours[edg[0]].append(edg)
            neighbours[edg[1]].append(edg)

        #Sorting neighbours by hour hand
        for i, p_neighbs in enumerate(neighbours):
            if len(p_neighbs) != 1:
                angles = [0]
                first_item = list(set(p_neighbs[0]) - set([i]))[0]
                for another_neighb in p_neighbs[1:]:            
                    second_item = list(set(another_neighb) - set([i]))[0]
                    vector1 = verts_in[first_item] - verts_in[i]
                    vector2 = verts_in[second_item] - verts_in[i]
                    angles.append(calc_angle(vector1, vector2))
                sorted_ = list(zip(angles, p_neighbs))
                sorted_.sort()
                neighbours[i] = [e[1] for e in sorted_]

        def get_end_points(item_point):
            second_item = list(set(neighbours[item_point][0]) - set([item_point]))[0]
            vert_edg = verts_in[item_point] - verts_in[second_item]
            mat_r1 = Matrix.Rotation(radians(-45),2,'X')
            mat_r2 = Matrix.Rotation(radians(45),2,'X')
            vert_new1 = (vert_edg * mat_r1).normalized() * shift/sin(radians(45)) + verts_in[item_point]
            vert_new2 = (vert_edg * mat_r2).normalized() * shift/sin(radians(45)) + verts_in[item_point]
            return [vert_new1,vert_new2]

        def get_middle_points(item_point):
            points = []
            for i in range(len(neighbours[item_point])):
                current_edges = (neighbours[item_point][i:] + neighbours[item_point][:i])[:2]
                second_item1 = list(set(current_edges[0]) - set([item_point]))[0]
                second_item2 = list(set(current_edges[1]) - set([item_point]))[0]
                vert_edg1 = verts_in[second_item1] - verts_in[item_point]
                vert_edg2 = verts_in[second_item2] - verts_in[item_point]
                angle = calc_angle(vert_edg1, vert_edg2) / 2
                mat_r = Matrix.Rotation(angle, 2, 'X')        
                points.append((vert_edg1 * mat_r).normalized() * shift/sin(angle) + verts_in[item_point])
            return points

        #Seting points
        findex_new_points = [0]
        for i in range(len(verts_in)):
            if len(neighbours[i]) == 1:
                verts_out.extend(get_end_points(i))
                findex_new_points.append(findex_new_points[-1] + 2)
            else:
                p = get_middle_points(i)
                verts_out.extend(p)
                findex_new_points.append(findex_new_points[-1] + len(p))
        
        findex_new_points = findex_new_points[:-1]
        
        #Creating faces and mark outer edges and central points
        outer_edges = []
        current_index = len(verts_out) - 1
        vers_mask = [0 for _ in verts_out]
        nomber_outer_points = current_index
        position_old_points = [0 for _ in verts_out]

        for edg in edges_in:
            need_points = []
            for i in edg:
                if len(neighbours[i]) <= 2:
                    need_points.extend([findex_new_points[i], findex_new_points[i] + 1])
                else:
                    position = neighbours[i].index(edg)
                    nomber_points = len(neighbours[i])
                    variants_positions = list(range(findex_new_points[i], findex_new_points[i] + nomber_points))
                    need_points.extend((variants_positions[position - 1:] + variants_positions[:position - 1])[:2])
    
            vec_edg = verts_in[edg[0]] - verts_in[edg[1]]
            vec_1 = verts_out[need_points[0]] - verts_in[edg[0]]
            vec_2 = verts_out[need_points[1]] - verts_in[edg[0]]
            vec_3 = verts_out[need_points[2]] - verts_in[edg[1]]
            vec_4 = verts_out[need_points[3]] - verts_in[edg[1]]
            new_vecs = [vec_1, vec_2, vec_3, vec_4]

            angles = [vec_edg.angle_signed(vec) for vec in new_vecs]

            if position_old_points[edg[0]] == 0:
                verts_out.append(verts_in[edg[0]])
                vers_mask.append(1)
                current_index += 1
                position_old_points[edg[0]] = current_index
        
            if position_old_points[edg[1]] == 0:
                verts_out.append(verts_in[edg[1]])
                vers_mask.append(1)
                current_index += 1
                position_old_points[edg[1]] = current_index
            
            n_p = need_points
            if angles[0] < 0 and angles[2] < 0 or angles[0] >= 0 and angles[2] >= 0:
                new_edges = [[n_p[1], position_old_points[edg[0]], position_old_points[edg[1]], n_p[3]],
                             [n_p[0], position_old_points[edg[0]], position_old_points[edg[1]], n_p[2]]]
                             
                outer_edges.extend([[n_p[1],n_p[3]],[n_p[0],n_p[2]]])
            else:
                new_edges = [[n_p[0], position_old_points[edg[0]], position_old_points[edg[1]], n_p[3]],
                             [n_p[1], position_old_points[edg[0]], position_old_points[edg[1]], n_p[2]]]
                             
                outer_edges.extend([[n_p[0],n_p[3]],[n_p[1],n_p[2]]])
    
            if len(neighbours[edg[0]]) == 1:
                new_edges.append([need_points[1], position_old_points[edg[0]], need_points[0]])
                outer_edges.append([need_points[1],need_points[0]])
            if len(neighbours[edg[1]]) == 1:
                new_edges.append([need_points[2], position_old_points[edg[1]], need_points[3]])
                outer_edges.append([need_points[2],need_points[3]])
    
            for c,face in enumerate(new_edges):
                if normal(*[verts_out[i].to_3d() for i in face])[2] < 0:
                    new_edges[c] = new_edges[c][::-1]
        
            faces_out.extend(new_edges)

        verts_out = [i.to_3d()[:] for i in verts_out]
        return verts_out, faces_out, outer_edges, vers_mask


    def process(self):

        if not ((self.outputs['Vers'].is_linked or self.outputs['Faces'].is_linked
              or self.outputs['OuterEdges'].is_linked or self.outputs['VersMask'].is_linked) 
              and (self.inputs['Vers'].is_linked or self.inputs['Edgs'].is_linked)):
            return

        verts_in = self.inputs['Vers'].sv_get()
        edges_in = self.inputs['Edgs'].sv_get()
        shifter = self.inputs['Offset'].sv_get()[0]
        
        # avoid zero offset
        if not shifter:
            return

        # verts_out, faces_out, outer_edges, vers_mask
        out_geometry = [[], [], [], []]  

        for verts, edges, shift in zip(verts_in, edges_in, shifter)   # <--- not correct yet.
            geometry = offset_mesh(verts, edges, shift, generate_outer_edges=True, generate_verts_mask=True)
            _ = [out_geometry[idx].append(data) for idx, data in enumerate(geometry)]
       
        self.outputs['Vers'].sv_set(out_geometry[0])
        self.outputs['Faces'].sv_set(out_geometry[1])
        self.outputs['OuterEdges'].sv_set(out_geometry[2])
        self.outputs['VersMask'].sv_set(out_geometry[3])
```

zeffii:
the goal is to keep the `process` function abstract, as a "task master". 

   process is best for

1. get data
2. sanitize data into best form
3. send (sanitized) data to worker functions
4. get product from worker functions
5. fill outputs sockets with product.

The worker functions are cohesive blocks of code that are often specialized enough to stick into their own function.

Durman:
@zeffii Okay, I understood. I will fix.

zeffii:
@Durman 

you don't have to do this immediately
```python
offset_mesh(verts, edges, shift, generate_outer_edges=True, generate_verts_mask=True)
```
i was only thinking that maybe it's possible / efficient to only produce the outer_edge and vert_masks if those sockets are connected. But if it's not possible to efficiently make those optional, then always generate :)

zeffii:
also, maybe i'm wrong but this seems like it could be calculated once and used twice.
```python
        vert_new1 = (vert_edg * mat_r1).normalized() * shift/sin(radians(45)) + verts_in[item_point]
        vert_new2 = (vert_edg * mat_r2).normalized() * shift/sin(radians(45)) + verts_in[item_point]
```
or
```python
        amp = shift / sin(radians(45))
        vert_new1 = (vert_edg * mat_r1).normalized() * amp + verts_in[item_point]
        vert_new2 = (vert_edg * mat_r2).normalized() * amp + verts_in[item_point]
```

at a later stage we can discuss making it possible to set a unique shift value for every original vertex -- but also this is up to you. It may require two versions of the function (for speed)  one for constant shift per mesh, one for variable shift per mesh.

Durman:
@zeffii 
>i was only thinking that maybe it's possible / efficient to only produce the outer_edge and vert_masks if those sockets are connected.

I think it will not be efficient. I don't use any preparation for these sockets, just assign values for variables in different places of the code. I think that checking boolean values only can to increase time of runtime.

>at a later stage we can discuss making it possible to set a unique shift value for every original vertex

I tried to imagine situation when it can be useful. I suppose that in most cases `edges_in` will be unordered so it will be difficult to prepare data for offset socket and receive interesting result. May with `curve in` node it could be interesting to have such feature but this node haven't done yet as i know. At this moment I prepare offset socket only for different values per object.

```Python
TWO_PI = 2 * pi
```
I am not quite understand benefit of creating the variable. Why not just to use `2 * pi` into my function?

zeffii:
because `calc_angle` is a function you call many many times, and instead of recalculating 2*pi on each call, you can simply reference the answer (TWO_PI). It's a little bit more efficient. It's also standard graphics programming practice.

zeffii:
> I think that checking boolean values only can to increase time of runtime.

that depends how often the boolean is checked. 
--------------
anyway. If you want to always calculate both versMask and outerEdges in the beta version of the node that's OK.

Durman:
I added this node in Alpha. Must it be in beta?

zeffii:
No, alpha is fine )

Durman:
It is seems that I fixed all. ))

zeffii:
Why 2 calc_angle functions, keep the outermost that uses TWO_PI

enzyme69:
I am an Alpha Rebel.
-- 


- Jimmy Gunawan

Durman:
@zeffii 
> Why 2 calc_angle functions, keep the outermost that uses TWO_PI

Sorry, don't understand.

Durman:
@zeffii Yes, didn't notice :/

zeffii:
i have a question about the `verts mask` output.

- if you never rearrange the input verts and don't remove any
- then the final verts_mask is a single step

```python

def make_verts_new_mask(verts_in, verts_out):
    num_original_verts = len(verts_in)
    num_verts_out = len(verts_out)
    num_new_verts = num_verts_out - num_original_verts
    f = ([0] * num_original_verts) + ([1] * num_new_verts)
    return f
    

f = make_verts_new_mask(verts_in=[5,2,3,4,5], verts_out=[5,2,3,4,5,5,5,5,2,2,2,3,3,3,4])
print(f)  # [0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
```

zeffii:
and, this is equivalent
```diff
-        if not (self.outputs['Vers'].is_linked or self.outputs['Faces'].is_linked
-              or self.outputs['OuterEdges'].is_linked or self.outputs['VersMask'].is_linked):
-            return

+        if not any(socket.is_linked for socket in self.outputs):
+            return
```

zeffii:
This concludes my review :) looking forward to merging.

Durman:
@zeffii 
> i have a question about the `verts mask` output.

It's not quite exactly. In my script at first I add new points and after that add old points thus the order is changed. 
I have only three strings that is responsible for this.
```Python
vers_mask = [0 for _ in verts_out]
...
      vers_mask.append(1)
...
      vers_mask.append(1)
```
Is there point to change this? I think it must work faster than if writ seperate function for this.

zeffii:
OK  .

> It's not quite exactly. In my script at first I add new points and after that add old points thus the order is changed.

this is not a problem. I will merge if you think you're ready to have more people test it :)

Durman:
Ok. I am ready. ))

zeffii:
thanks for the cool new node @Durman .

hopefully the process was not too painful. First time is the most difficult (eye - opening).

zeffii:
i think i can implement variable offset, if you're interested.

zeffii:
imagine for a moment that we can generate a `list of offset radii` with the same length as `verts_in`, (we have a function for this already in utils/core "match_long_repeat") Then rather than shift being static per mesh, it can be variable.

```python
   #Setting points
   findex_new_points = [0]
   for i in range(len(verts_in)):
        if len(neighbours[i]) == 1:
            verts_out.extend(get_end_points(i))
            findex_new_points.append(findex_new_points[-1] + 2)
        else:
            p = get_middle_points(i)
            verts_out.extend(p)
            findex_new_points.append(findex_new_points[-1] + len(p))
```
and..
```python

   # (not displayed here) we always ensure radii is as long as verts_in
   # meaning we can enumerate radii and use the idx it produces

   findex_new_points = [0]
   for idx, radius in enumerate(radii):
        if len(neighbours[idx]) == 1:
            verts_out.extend(get_end_points(idx, radius))
            findex_new_points.append(findex_new_points[-1] + 2)
        else:
            p = get_middle_points(idx, radius)
            verts_out.extend(p)
            findex_new_points.append(findex_new_points[-1] + len(p))
```
for speed I would probably make variable offset optional. and modify the `offset_edge` code to pick which of the proposed loops to use depending on what's needed. Maybe it will make more sense if I just do it.

Durman:
@zeffii Okay, I agree with it.
I also can do this, and also to do experiments with Z coordinate. Not sure that the node have to have constraints with Z values and to exist only in XY surface. It is easy to add Z coordinate but difficult to calculate normals correctly.

zeffii:
Z stuff is not important / not interesting use of this node. @enzyme69 often tries to cut grass using a nail clipper. (wrong tool for job)

enzyme69:
I like cutting tree using nail clipper.

PR #1961, opened 2017-12-21
Add first iteration of mk2 - untested

zeffii:
- fix parenthesized output / input read
- create MK2 of TextIn and TextOut
- UI backend streamline
- reorder imports
- introduce common util for TextIO nodes (including operator: `utils/sv_text_io_common`)
- place Text Nodes within Text Category.
- move over to text=Stringproperty and obtain textblock using prop_search on `bpy.data.texts`.
- some logic shortcutting.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -86,7 +86,7 @@ utils_modules = [
     # non UI tools
     "cad_module", "cad_module_class", "sv_bmesh_utils", "sv_viewer_utils", "sv_curve_utils",
     "voronoi", "sv_script", "sv_itertools", "script_importhelper", "sv_oldnodes_parser",
-    "csg_core", "csg_geom", "geom", "sv_easing_functions",
+    "csg_core", "csg_geom", "geom", "sv_easing_functions", "sv_text_io_common",
     "snlite_utils", "snlite_importhelper", "context_managers",
     "profile", "logging", 
     # UI text editor ui
--- a/index.md
+++ b/index.md
@@ -225,8 +225,8 @@
 
 ## Text
     ViewerNodeTextMK2
-    SvTextInNode
-    SvTextOutNode
+    SvTextInNodeMK2
+    SvTextOutNodeMK2
     NoteNode
     SvDataShapeNode
     GTextNode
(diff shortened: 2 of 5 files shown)


zeffii:
small note to self.
```python
    def reload(self):
        if self.textmode == 'CSV':
            self.reload_csv()
        elif self.textmode == 'SV':
            self.reload_sv()
        elif self.textmode == 'JSON':
            self.reload_json()

        updateNode(self, None)

    def process(self):
        try:
            l = bpy.data.node_groups[self.id_data.name]
        except Exception as e:
            print(self.name, "cannot run during startup, press update.")
            return

        if not self.current_text: return

        if self.textmode == 'CSV':
            self.update_csv()
        elif self.textmode == 'SV':
            self.update_sv()
        elif self.textmode == 'JSON':
            self.update_json()

    def load(self):
        if self.textmode == 'CSV':
            self.load_csv()
        elif self.textmode == 'SV':
            self.load_sv()
        elif self.textmode == 'JSON':
            self.load_json()


=====================

    def perform_action(self, named_function):
        getattr(self, named_function + '_' + self.textmode.lower())()

    def reload(self):
        self.perform_action("reload")

    def process(self):

        # this try/except might be redundant now.. 
        try:
            l = bpy.data.node_groups[self.id_data.name]
        except Exception as e:
            print(self.name, "cannot run during startup, press update.")
            return

        if not self.current_text:
            return

        self.perform_action("update")

    def load(self):
        self.perform_action("load")
```

zeffii:
When sverchok tries to load a node tree that contains nodes that aren't in the current node list, it will look through the old_nodes list and find which file/module has that `bl_idname` and call that file/module's 'register' function. This works well for all nodes, but some nodes will notice a side effect if multiple nodes are defined in the same file/module.

I was testing what happens to old TextIn / Out nodes, because these nodes share a file, both nodes are registered using `register()`, this function is called by the old_nodes registration. If TextIn is registered, also TextOut is registered - when it comes time to add TextOut it will not be missing from the current node list, and you get this: 

in the following screenshot you'll see two "old nodes" but only one will be put inside a "Deprecated Node" frame:
image

zeffii:
fixing that might be best left to a new PR.

zeffii:
anyway.. this is almost working except for the CSV loading , which i seem to have broken.

soon.

zeffii:
i think it's not useful to try to output mesh data (verts, edges, faces) using the CSV format, which outputs columns. hence this is going to happen
image

but, i could add a mode to output unquoted data.

PR #1963, opened 2017-12-22
add new text io route

zeffii:
- [x] add support for IO json
     - [x] (json mode) store the output file used by TextInMk2
     - [x] (json mode) load works
     - [x] (json mode) restore sockets with the same order as existed during export (currently not the case, because json reading as dataform picks sub structures in an arbitrary order) 
- [ ] explore handling TextIn and TextOut bl_idname for old nodes [the 2 nodes per file issue described here](https://github.com/nortikin/sverchok/pull/1961#issuecomment-353583357)

### DIFF
--- a/nodes/text/text_in_mk2.py
+++ b/nodes/text/text_in_mk2.py
@@ -437,13 +437,32 @@ class SvTextInNodeMK2(bpy.types.Node, SverchCustomTreeNode):
             self.current_text = ''
             return
 
-        for item, data in json_data.items():
-            if len(data) == 2 and data[0] in {'v', 's', 'm'}:
-                new_output_socket(self, item, data[0])
-            else:
-                self.use_custom_color = True
-                self.color = FAIL_COLOR
-                return
+        # if you can still read this I have been out of time to refactor.
+
+        socket_order = json_data.get('socket_order')
+        if socket_order:
+            # this is necessary to avoid arbitrary socket assignment upon import
+            # not all gists/iojsons will have this key, hence the test. but it is
+            # important, and will let you define the order manually by editing the file
+            # if you have to.
+            for named_socket in socket_order:
+                data = json_data.get(named_socket)
+                if len(data) == 2 and data[0] in {'v', 's', 'm'}:
+                    new_output_socket(self, named_socket, data[0])
+                else:
+                    self.use_custom_color = True
+                    self.color = FAIL_COLOR
+                    return
+        else:
+            for named_socket, data in json_data.items():
+                if len(data) == 2 and data[0] in {'v', 's', 'm'}:
+                    new_output_socket(self, named_socket, data[0])
+                else:
+                    self.use_custom_color = True
+                    self.color = FAIL_COLOR
+                    return
+
+
 
     def reload_json(self):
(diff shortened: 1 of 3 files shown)


zeffii:
current code..
```python
    def load_json(self):
        n_id = node_id(self)
        self.load_json_data()
        json_data = self.json_data.get(n_id, [])
        if not json_data:
            self.current_text = ''
            return

        for item, data in json_data.items():
            if len(data) == 2 and data[0] in {'v', 's', 'm'}:
                new_output_socket(self, item, data[0])
            else:
                self.use_custom_color = True
                self.color = FAIL_COLOR
                return
```

zeffii:
there;s a dict structure called "text_lines" used in the storage of the json in IOJSON only, i think i'll add a key, value like  `text_lines['socket_order'] = [name0, name1, name2...]`

zeffii:
hmmm.. instead the json itself should just store a key called "socket_order", and enumerate as list the names of the keys. This would help restoring from a json.

zeffii:
something like this, but cleaner.
```python
    def load_json(self):
        n_id = node_id(self)
        self.load_json_data()
        json_data = self.json_data.get(n_id, [])
        if not json_data:
            self.current_text = ''
            return

        # if you can still read this I have been out of time to refactor.

        socket_order = json_data.get('socket_order')
        if socket_order:
            # this is necessary to avoid arbitrary socket assignment upon import
            # not all gists/iojsons will have this key, hence the test. but it is
            # important, and will let you define the order manually by editing the file
            # if you have to.
            for named_socket in socket_order:
                data = json_data.get(named_socket)
                if len(data) == 2 and data[0] in {'v', 's', 'm'}:
                    new_output_socket(self, named_socket, data[0])
                else:
                    self.use_custom_color = True
                    self.color = FAIL_COLOR
                    return
        else:
            for named_socket, data in json_data.items():
                if len(data) == 2 and data[0] in {'v', 's', 'm'}:
                    new_output_socket(self, named_socket, data[0])
                else:
                    self.use_custom_color = True
                    self.color = FAIL_COLOR
                    return

```

zeffii:
even..
```python
    def load_json(self):
        n_id = node_id(self)
        self.load_json_data()
        json_data = self.json_data.get(n_id, [])
        if not json_data:
            self.current_text = ''
            return

        # if you can still read this I have been out of time to refactor.

        socket_order = json_data.get('socket_order')

        if socket_order:
            # this is necessary to avoid arbitrary socket assignment upon import
            # not all gists/iojsons will have this key, hence the test. but it is
            # important, and will let you define the order manually by editing the file
            # if you have to.
            def iterate_socket_order():
                for named_socket in socket_order:
                    data = json_data.get(named_socket)
                    yield named_socket, data

            socket_iterator = iterate_socket_order()
        else:
            socket_iterator = json_data.items()


        for named_socket, data in socket_iterator:
            if len(data) == 2 and data[0] in {'v', 's', 'm'}:
                new_output_socket(self, named_socket, data[0])
            else:
                self.use_custom_color = True
                self.color = FAIL_COLOR
                return

```

zeffii:
the remaining issue... i have no immediate tidy solution for.

PR #1976, opened 2017-12-24
Mesh Expression: sample nodes are optional

portnov:
Creation of example nodes at "from selection" is now optional, and    disabled by default. One can enable it in N panel.

Documentation is updated accordingly.


portnov:
@nortikin I hope you do not mind.

It is possible to enable this option by default locally, via node_defaults.json mechanism.

nortikin:
ok, just leave profile node

Issue #1973, opened 2017-12-24
Offset Line node (Offset radius per point possible?)


enzyme69:
<img width="350" alt="screen shot 2017-12-24 at 8 41 55 pm" src="https://user-images.githubusercontent.com/3952687/34325624-e613d38e-e8ea-11e7-8fb8-0f2f22dfdd92.png">

[line_to_leaf_01_2017_12_24_09_41.zip](https://github.com/nortikin/sverchok/files/1584728/line_to_leaf_01_2017_12_24_09_41.zip)

I am kind of curious of the possibilities to modify Offset Line node so that we can "scale offset" per points. To make something like a Leaft from a few points+edges, and passing offset radius per point (similar to PolylineViewerMK1).

Currently it is very cool already that it is vectorizing and works for multiple Lines input.

enzyme69:
<img width="165" alt="screen shot 2017-12-24 at 8 41 18 pm" src="https://user-images.githubusercontent.com/3952687/34325634-55ea7366-e8eb-11e7-9be0-34e315cffc5e.png">

From the index, it looks like this might not be possible. 

But I suspect it is possible to do it another way...

zeffii:
i have no idea what you're talking about. please re-edit your posts.

zeffii:
if the essence of your question is this:

> Is it possible to modify it to allow "offset" per individual (original) vertex.

zeffii:
I spoke with durman about adding support for per vertex radius, last week (so it would act more like SkinModifier / PolylineViewer. and be generally more versatile). You're question is not unique and certainly is the logical next step.

We don't need an open issue for this.

nortikin:
please, vectorize someone )))

Durman:
@nortikin 
I am not sure what does vectorize mean if it is about multiple values of offset socket it works now.

nortikin:
yes, i see. it is vectorization itself

Issue #1971, opened 2017-12-24
Line Offset and Grease Pencil, but without that Center Line ...


enzyme69:
<img width="662" alt="screen shot 2017-12-24 at 2 23 49 pm" src="https://user-images.githubusercontent.com/3952687/34324048-4c41a512-e8b6-11e7-9b0a-8116624801f1.png">
<img width="798" alt="screen shot 2017-12-24 at 2 22 18 pm" src="https://user-images.githubusercontent.com/3952687/34324049-4c7294ce-e8b6-11e7-935b-3bd809a46d30.png">
[gp_to_lineXY_001_2017_12_24_03_22.zip](https://github.com/nortikin/sverchok/files/1584589/gp_to_lineXY_001_2017_12_24_03_22.zip)

I am testing the Line Offset with Grease Pencil. It works well and I can get a very nice mesh out of it as a flat XY projection.

Let say supposedly I wanted to ensure each points **not overlapping**, normally I would use something like **Vector Rewire**, offsetting by one.

But in this case, I am getting the center lines. That I wish I could just dissolve.

Just using the Outer Edges currently is the only solution, but it does not generate face along the lines, and if I just fill hole it does not always give a nice polygon face, because it is not "flat". Ideally there should be connecting lines for every points.

enzyme69:
<img width="859" alt="screen shot 2017-12-24 at 2 34 47 pm" src="https://user-images.githubusercontent.com/3952687/34324074-9d5f7856-e8b7-11e7-83a7-f8877e65146d.png">
<img width="611" alt="screen shot 2017-12-24 at 2 34 44 pm" src="https://user-images.githubusercontent.com/3952687/34324075-9d915e7a-e8b7-11e7-8d75-dd9f35ae91b8.png">
[train_track_001_2017_12_24_03_34.zip](https://github.com/nortikin/sverchok/files/1584593/train_track_001_2017_12_24_03_34.zip)

Grease Pencil does not really have a proper "angle direction orientation" because it is always flat, if I am not wrong we have "tangent" somewhere.

I don't know @Durman maybe using Bending like above is the best for now?

enzyme69:
Ok... after tinkering, probably this is the most ideal setup for now.

[train_track_002_2017_12_24_03_41.zip](https://github.com/nortikin/sverchok/files/1584594/train_track_002_2017_12_24_03_41.zip)

[...]

PR #1978, opened 2017-12-26
Offset line update value per vector and z cor

Durman:
Now it is available to set offset value for each point. Also Z coordinate is using. I added some explanation in file of documentation and several new examples.
#1971 #1973 #1914


zeffii:
nice work @Durman

nodes/modifier_change/offset_line.py
```diff
@@ -160,7 +176,7 @@ def get_middle_points(item_point):
         
         faces_out.extend(new_edges)
 
-    verts_out = [i.to_3d()[:] for i in verts_out]
+    verts_out = [(i.to_3d() + Vector((0,0,z_co[c])))[:] for c,i in enumerate(verts_out)]
```
zeffii:
```diff
- verts_out = [(i.to_3d() + Vector((0,0,z_co[c])))[:] for c,i in enumerate(verts_out)]
+ verts_out = [(v.x, v.y, z_co[idx]) for idx, v in enumerate(verts_out)]
```

zeffii: the convention is to use i or idx for the index of the enumerator. else this promotes confusion to the reader (even you... in 2 months time). and we tend to use `v` to indicate that an element is a vector (or similar triplet)

i suggest using `v.x, v,y, ...` here, but you can use `v[0], v[1], ....`  . bracket lookup might be faster, but v.x and v.y are a little more meaningful.

We like to avoid the use of Vector entirely, because casting from [x,y,z] to Vector([x,y,z]) introduces a slowdown -- I don't think this is a problem now, but to increase speed if it's ever a serious problem it's try implementing this using the verts/edges converted to a bmesh representation.

Durman: Ok, I got it.

zeffii:
consider:

```diff
-    verts_out = [(v.x, v.y, z_co[idx]) for idx,v in enumerate(verts_out)]
-    return(verts_out, faces_out, outer_edges, vers_mask)
```
or even..
```diff
+    verts_out = [(v.x, v.y, z) for v, z in zip(verts_out, z_co)]        # 1
+    return verts_out, faces_out, outer_edges, vers_mask                 # 2
```
`#1` because when both iterables (z_co, and verts_out) share the same length, and you know they directly map to eachother, the `zip` function is a good way to make this clear.
`#2` because any sequence of comma separated variables is a tuple, the surrounding parens (` (...)  `) are superfluous (but not incorrect).   but.. `return(.... )` is not nice style. (without space between return and parenthesis). .. it looks too much like a function call.

nodes/modifier_change/offset_line.py
```diff
@@ -189,27 +205,22 @@ def sv_init(self, context):
 
     def process(self):
 
-        if not (self.inputs['Vers'].is_linked or self.inputs['Edgs'].is_linked):
+        if not any(socket.is_linked for socket in self.inputs):
```
zeffii:
```diff
- if not any(socket.is_linked for socket in self.inputs):
+ if not all(socket.is_linked for socket in self.inputs[:2]):
```
the first two sockets must be connected, last socket is optional.

zeffii: there's differing opinions about whether we need these input checks, in my opininion they are fine. but we also have a mechanism that responds to sockets that are inconnected and don't have a "default" parameter in the `socket.sv_get(   )` call. -- in that scenario the node will say " socket x not connected " [(or something similar)](https://github.com/nortikin/sverchok/pull/1893)

zeffii: @portnov  in the case of two sockets been mandatory, we could maybe raise a specific error too ? like
(a proposal, this won't work at the moment)
```python
   raise SvNotFullyConnected(self, msg="node x needs input a and input b to perform work")
```

portnov: @zeffii good idea. I'm all for more specific exception classes for specific cases (for now we raise plain Exception in many cases).

Durman: @zeffii 
> but we also have a mechanism that responds to sockets that are unconnected and don't have a "default" parameter in the `socket.sv_get( )`

So will it be better to delete checking at all?

zeffii: for now use 
```python
if not all(socket.is_linked for socket in self.inputs[:2]):
    return
```
we will add the correct Exception class soon!

nortikin:
thanks

PR #1998, opened 2018-01-02
User-defined node presets

portnov:
I really think this is kind of feature which Sverchok lacked very much from user perspective.

The main ideas are:
* The main approach to solving complex problems (in this case - designing complex parametric setups) is to divide and conquer.
* As we know very well from our programming experience, the right way to divide and conquer is to have a library of building blocks and be able to compose them in different ways.
* As we know from using github, Maven central, pypi.org and others, a very good way to produce a large number of such building blocks is to allow users to produce them and share them.

My requirements as a user are:
* [x] F1. Be able to store settings of either one node (for nodes that have more than 2-3 properties), or some set of linked nodes, as a named preset for later usage in different setups.
* [x] F1.1. Special case 1 is to be able to save and use Monad node as a preset.
* [x] F1.2. Special case 2 is to be able to save and use a Scripted node with script in it. The same applies for Profile node, Mesh Expression and so on.
* [x] F2. Be able to manage my presets locally: rename them, remove unused and so on.
* [ ] F2.1. Extension of F2 (not need it right now but very probably will need in future): be able to organize my presets in some sort of categories or tags locally.
* [x] F3. Be able to share my presets with people. It can be done via simple files or via some web service.
* [ ] F3.1. Extension of F3 (want to see it implemented, but not sure whether it is possible to implement with our resources at all): be able to manage shared resources online. By that I mean some sort of asset management and sharing web-service, like opendesktop.org.
* [x] F4. Be able to use presets shared by someone else. It can be done via simple files or via some web service.

In current implementation, presets are stored in JSON format (the same as we use for gists), under `.../datafiles/sverchok/presets/*.json`. Rename/delete operators just do corresponding operations on json files.

I supported minimal set of metainformation attributes:
* Preset description (displayed as tooltip)
* Preset author
* Preset license

All of them are optional. Metainformation is saved under `"metainfo"` key in json.

Nice thing is that all already existing gists can be now used as presets (user can import gists to local machine and use them as presets).

screenshot_20180102_234335

screenshot_20180102_234406

screenshot_20180103_180341
screenshot_20180103_180445


portnov:
@zeffii  @DolphinDream @nortikin could you please review and test?
The first thing to review is requirements: maybe we need to clarify them or add something.

zeffii:
i'm glad you reused the gist storing/loading code :) this will be another way to test that code continuously. 

..still recovering from some kind of flu i'm not so energised to get actively involved, but I did read through the code changes+additions. Nothing worried me during the "mental compile".

utils/sv_IO_panel_tools.py
```diff
@@ -851,54 +854,65 @@ def invoke(self, context, event):
...
+    gist_id: gist ID. Passing full URL is also supported.
+    operator: optional instance of bpy.types.Operator. Used for errors reporting.
 
-    bl_idname = "node.tree_import_from_gist"
-    bl_label = "sv NodeTree Gist Import Operator"
-
-    id_tree = StringProperty()
-    new_nodetree_name = StringProperty()
-    gist_id = StringProperty()
+    Rerturns JSON dictionary.
```
zeffii: small typo 'Rerturns`

portnov:
@zeffii I wish you will be filling better soon :)

portnov:
Considerations about F3.1 (not for right now, but for future plans):
* We could think about separate service. But not now, maybe somewhere in distant shiny future. For now it seems we just do not have so much resources.
* We could ask people from opendesktop.org to create a category for sverchok. It requires some investigations: how much resources will we need to manage such a category (will we need to moderate it, for example). opendesktop would give us a storage, a sort-of-website, it provides API for both uploading and downloading things, it manages authentication question... Very good thing this opendesktop is, and imho the first we should consider. But again, I think not right now.
* Theoretically we can integrate with Blender 2.8's asset management subsystem. Main points about it for us are:
  * To actually import things, it uses standard blender's link-or-append mechanism; this means, that assets (presets) have to be stored in `.blend` files. This looks like overkill for what we currently have (just `.json` files), but it can be promising for future (for example we could share input objects for "Object In" nodes, maybe some textures or materials together with node tree).
  * "Asset management subsystem" by itself does not provide any place where to store assets. It is just an API. Actual storage is done by "asset engine", which is sort of addon. There will be simple standard asset engine, which stores assets as plain .blend files on local filesystem. To store assets somewhere in network, we will have to either write or use some existing "asset engine", which fits our needs. And we will have to think about storage separately anyway.

At one side, these considerations may look like being too early, as for now we probably do not have so much presets or gists to store, and not so much users willing to share their presets. But on the other hand, I consider this whole topic as point from where sverchok's user community may grow. This leads me to the next interesting questions:
* Do we have any idea, how many users of sverchok there are? Not very many obviously, but what that means? Tens, or hundreds, or?..
* How many shared gists (by all users) there are?

portnov:
@zeffii totally separate question about existing "gist storing/loading code". It looks like it started as generic code, but then special cases were added one by one, and now it is just a collection of special cases. I have a temptation to refactor this code (move all special cases handling to specific node classes). But current code looks fragile, and since we do not have automated / unit tests for it, I'm not ready to start refactoring. But some day in shiny future, one will have to sit and cover all this with tests, and refactor...

portnov:
I was going to write a user guide document for this feature, but when I looked in documentation, I found it partly obsolete, partly incomplete. So I better will do documentation update in another PR.

zeffii:
> But current code looks fragile,

yes. and no. mostly yes. Special cases make up about 30 (guestimate) of the 260+ nodes. The low revision number is not a coincidence. I consider this implementation utterly beta, but it does document the hairy details as I understood at the time. It started out as an experiment to see how far I could serialise the tree. As it progressed we encountered nodes that did need extra coercion. 

(i might solve these problems differently nowadays)

portnov:
An interesting link: https://blender.stackexchange.com/questions/67274/how-to-do-continuous-integration-with-gitlab-when-developing-blender-addons

portnov:
Going to merge this, I consider main features as complete. Having related plans for future: portnov#28, portnov#29, portnov#30.

nortikin:
impressed
will use dayly

Durman:
@portnov If preset has a monad and current file has already monad with same name, which behavior is expected? Monad from preset is replaced by monad from file at this moment, it can confuse an user.

portnov:
@Durman have no idea actually. What the presets mechanism is doing is just importing something from JSON files (same format as we use for gists).

zeffii:
the IO/json system is not optimal for importing into an existing node tree, it does no extensive checking for node conflicts with respect to Monads (it is a special class of Node )

Issue #1990, opened 2017-12-30
Performance considerations (discussion)


portnov:
While trying to improve performance of "bend along surface" node, I did some research about about whether it worth it to move some part of code to C++ (you can see the code at https://github.com/portnov/python_cpp_excersises). Here are the results:

* It is not so hard to integrate C++ and python nowadays. Keyword is "boost::python".
* It is possible to eliminate data copying almost completely in most situations.
* The most subtle thing is memory management. Python has it's own memory management, C++ has it's own "kind-of memory management". You must know at each moment who owns the particular pointer, who is gonna release it and when exactly.
* For non-numeric code (such as "get data and put it to the map of some sort", "traverse that map" and so on), C++ code is about 3-4 times faster than very similar python code.
* For vectorized numeric computations in C++, the keywords are "boost::ublas" and "eigen" (last one uses SIMD instructions when they are available).
* For numeric code, C++ can be about 10 times faster than python code. 
* But, in many places, for numeric computations we already use numpy. Numpy is actually a C++ library with support of SIMD instructions. So, the performance of python-numpy code is about 1:1 to performance of C++ Eigen code (some functions are slower in numpy, other are slower in Eigen).

So my conclusions for now are:

* It does not worth rewriting LinearSpline/CubicSpline/Spline2D classes in C++, as they are actually mostly written in C++ already (by usage of numpy). The main way to improve performance of nodes that use these is to process lesser amounts of data in most cases.
* It could worth rewriting some code that processes a lot of data with non-numerical algorithms, in C++. But I do not know such places in Sverchok.
* If you see some node / class in Sverchok, that does some numerical computations, and you want to improve it's performance - the step 1 is to use numpy.

Will write some other thoughts later...

zeffii:
- i'd love to see `bmesh.ops.boolean(bm_a, bm_b, operation='UNION', material=False)` implemented, get away from using python for that process.
- uv_connect nodes, these things seem like they could be written in C/C++ and be much faster.
- mesh join, mesh separate loose, maybe the math nodes (Scalar/Vector)..

The downside to moving the essential nodes to a compiled language is it reduces the number of people who can easily modify them (for their own use, or common use)..

zeffii:
we have a few script nodes that do cool stuff that's terribly slow, like 'random points on polygons' / 'best_candidate_rect_random_points`, 'topo sort for profile', '3d voronoi' (some SNmk1, some SNLite)

zeffii:
[...]

PR #2001, opened 2018-01-03
Contour 2D node and Sort By Connections

vicdoval:
This heavy node creates 2D iso-contours from points and edges.
It has to be feeded width points with the same Z coordinate
It can work with constant values:
isolines3
Or weighted distances:
isolines4

Is focused on creating a valid contour, to do so the algorithm contains the "Intersect Edges" node, with its behaviour, (also the "mask vertices" node, and the "Remove doubles")
It is pretty experimental but works on most of the cases
I also added the listMatcher function to switch between long_cycle and long_repeat (it may be a good practice for other nodes)

To sort the contour I created the function SortbyConnexions and i added it to the "Vector Sort" node:

vector_sort_by_connexion

I hope it is useful and enjoyable


nodes/modifier_make/isoLine2D.py
```diff
@@ -0,0 +1,694 @@
...
+from bpy.props import BoolProperty, IntProperty, FloatProperty, EnumProperty
+
+from sverchok.node_tree import SverchCustomTreeNode
+from sverchok.data_structure import (fullList, match_long_repeat,dataCorrect, repeat_last, updateNode, match_long_cycle)
+from sverchok.utils.sv_mesh_utils import mesh_join
+from sverchok.utils.cad_module_class import CAD_ops
+from sverchok.utils.sv_bmesh_utils import bmesh_from_pydata
+
+modeItems = [("Constant", "Constant", ""), ("Weighted", "Weighted", "")]
+listMatchItems = [("Long Repeat", "Long Repeat", ""), ("Long Cycle", "Long Cycle", "")]
```
zeffii: add a space between this line and the new unrelated function below it.

nodes/modifier_make/isoLine2D.py
```diff
@@ -0,0 +1,694 @@
...
+    rm_doubles = bpy.props.FloatProperty(
+        description = "Remove Doubles Distance",
+        min=0.0, default=0.0001, step=0.1, update=updateNode)
+    epsilon = bpy.props.FloatProperty(
+        description = "Intersection tolerance",
+        min=1.0e-5, default=1.0e-5, step=0.02, update=updateNode)
+    maskT = bpy.props.FloatProperty(
+        description = "Mask tolerance",
+        min=-1.0, default=1.0e-5, step=0.02, update=updateNode)
+    rad_ = FloatProperty(name='Radius', description='Radius',
```
zeffii: sort out the sudden weird formatting here.  dont have spaces between varname=varvalue inside a function call.

nodes/modifier_make/isoLine2D.py
```diff
@@ -0,0 +1,694 @@
...
+            listEdg.append((pointsL-1, 0))
+        return listEdg
+
+    def make_faces(self, Vertices):
+        listPlg = list(range(Vertices))
+
+        
+        return [listPlg]
+
+    def process(self):
```
zeffii: clean up the whitespace in this function especially :)

zeffii: i'd like to suggest using a pep8 checker or pylinter

zeffii: to get into the habbit of formatting code in a way that is generally easier to read (because it will match the general style of python code found in this repo )

nodes/modifier_make/isoLine2D.py
```diff
@@ -0,0 +1,694 @@
...
+    if not closest_points:
+        return True
+    if not isinstance(closest_points[0].x, float):
+        return True
+    if cm.num_edges_point_lies_on(closest_points[0], vert_vectors) < 2:
+        return True
+
+    # if this distance is larger than than VTX_PRECISION, we can skip it.
+    cpa, cpb = closest_points
+    return (cpa-cpb).length > cm.VTX_PRECISION
```
zeffii: the downside to using the static `cm.VTX_PRECISION` is that it will behave differently if the input geometry is thousands of units in scale.. ( range 1000, 10 000 ).. isntead of the usual scene scale (0.0... 10.0 ) -- in EdgesIntersection node I think i made a manual 'scale' parameter accessible from the `draw_buttons_ext` area.

zeffii:
those things aside, this looks like a cool new tool!

zeffii:
i see (after more careful reading) you are using the Epsilon (mutable tolerance) good!

portnov:
Looks very intriguing.

portnov:
@vicdoval ,
1. If I'm not wrong, more correct name for this is "equidistant line".
2. About "SortbyConnexions". Am I getting it right that this is a topological sorting (https://en.wikipedia.org/wiki/Topological_sorting)? I was inspecting this question in https://github.com/nortikin/sverchok/issues/1990#issuecomment-354599848 (you may be interested in reading full discussion there). I'm not suggesting you to write this in C++ (huh), but you may consider rewriting my implementation of Tarjan's algorithm (https://github.com/portnov/python_cpp_excersises/blob/master/benchmark.py#L31) without use of recursion. I'm not ready yet to say anything about your sortVerticesByConnexions method, maybe it is using some well-known algorithm already? Then it worths to add a reference in comments.

vicdoval:
@zeffii Thanks for your kind word and excuse me for my innocence,  :) I applied the pep8 standards as you suggested

Following the avise of @portnov  I changed the name to avoid confusions, I thing "Contour 2D" fits pretty well its function. 
I changed the file name and class name to match the new name.

About the @portnov ideas on "SortingByConnexions":
I didn't even know  how this was called, and I don't have any base on graph theory. I did this function to solve my problem, but probably there are many better ways to do it. I checked your gist, but I would need to study it deeply to understand what you are doing.
Anyway thinking about the non-iteration thing I came up with a better solution that turned down the time needed from 2.5 seconds to 1.5 on 6000 vertices. :D 

So feel free to check the code and thanks for your feedback!

zeffii:
> @zeffii Thanks for your kind words and excuse me for my innocence, :)

i think you'll learn fast that pep8 (with a few ignore flags) results in more pleasant code formatting in general and promotes consistent Python style. It doesn't address algorithm implementation/difficulty, just makes code visually more accessible (mostly, not always).

I look forward to using the node, I can already imagine many typographic uses.

nodes/modifier_make/contour2D.py
```diff
@@ -456,7 +410,7 @@ class SvContourNode(bpy.types.Node, SverchCustomTreeNode):
                              items=listMatchItems, default="Long Cycle",
                              update=updateNode)
 
-    rm_doubles = FloatProperty(name='R. Doubles', 
+    rm_doubles = FloatProperty(name='R. Doubles',
```
zeffii: format them like this.
```python
    modeI = EnumProperty(
        name="modeI",
        description="Constant or weigted distance when multiple radius are given",
        items=modeItems, default="Constant",
        update=updateNode)

    listMatch = EnumProperty(
        name="listMatch",
        description="Behaviour on diffent list lengths",
        items=listMatchItems, default="Long Cycle",
        update=updateNode)

    rm_doubles = FloatProperty(
        name='R. Doubles',
        description="Remove Doubles Distance",
        min=0.0, default=0.0001,
        step=0.1, update=updateNode)
    
    epsilon = FloatProperty(
        name='Int. Tolerance',
        description="Intersection tolerance",
        min=1.0e-5, default=1.0e-5,
        step=0.02, update=updateNode)
    
    maskT = FloatProperty(
        name='Mask tolerance',
        description="Mask tolerance",
        min=-1.0, default=1.0e-5,
        step=0.02, update=updateNode)

    rad_ = FloatProperty(
        name='Distance', description='Contour distance',
        default=1.0, min=1.0e-5, update=updateNode)

    vert_ = IntProperty(
        name='N Vertices', description='Nº of Vertices per input vector',
        default=24, min=4, update=updateNode)
```

nodes/modifier_make/contour2D.py
```diff
@@ -277,18 +275,20 @@ def CalcMidPoints(verts, edges):
...
 
 
-def maskEdges(edges,mask):
-    edges_out=[]
-    for m, ed in zip(mask,edges):
-        if m==True:
+def maskEdges(edges, mask):
+    edges_out = []
+    for m, ed in zip(mask, edges):
+        if m == True:
```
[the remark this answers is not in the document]
vicdoval: I left it like that for some kind of code superstition :D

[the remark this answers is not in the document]
zeffii: take a leap of faith for that kind of stuff.

zeffii:
for any functions that you take verbatim from other node files, you can import them instead. As long as those functions don't rely on some local storage to the source file. (here that's not the case)
```python
import sverchok.nodes.modifier_change.edges_intersect_mk2 as IX
from IX import remove_permutations_that_share_a_vertex
from IX import get_valid_permutations
```
etc...

vicdoval:
Great, but do i have to import de dependencies of that file? 
or by importing the file the dependencies will be automatically imported?

zeffii:
anything that those functions reference will be available to those functions _alone_ when imported, no namespace pollution.

zeffii:
take this short example of two files, one importing from the other.

#### file_a.py
```python
from sv_test import tk_dict_pusher, tk_dict_show

tk_dict_pusher("me")
tk_dict_pusher("too")
print(tk_dict_show())
```
##### sv_test.py
```python
some_dict = {}

def tk_dict_pusher(new_item):
    some_dict[new_item] = new_item

def tk_dict_show()
    return some_dict

```

zeffii:
if you can't get it working - that's OK. i'll consider turning those functions into a utility/module and guarantee that they are static

vicdoval:
I mange to import them 😄

nodes/modifier_make/contour2D.py
```diff
@@ -0,0 +1,599 @@
...
+
+    # optional correction, remove originals, add back those that are not intersecting.
+    edges_out = edges_out[trim_indices:]
+    edges_out.extend(add_back)
+    bm.free()
+    return verts_out, edges_out
+
+
+def ptInTriang(p_test, p0, p1, p2):
+    # Function taken from Ramiro R.C https://stackoverflow.com/a/46409704
```
zeffii: consider adding this to https://github.com/nortikin/sverchok/blob/master/utils/modules/geom_utils.py , and importing it.

nodes/modifier_make/contour2D.py
```diff
@@ -0,0 +1,599 @@
...
+                vL1b = [v1[0] + dLim1 * cos(beta[2+netOff]), v1[1] + dLim1 * sin(beta[2+netOff]), v1[2]]
+                vL2a = [v2[0] + dLim2 * cos(beta[1+netOff]), v2[1] + dLim2 * sin(beta[1+netOff]), v2[2]]
+                vL2b = [v2[0] + dLim2 * cos(beta[2+netOff]), v2[1] + dLim2 * sin(beta[2+netOff]), v2[2]]
+                A = ptInTriang(v, vL2a, vL1a, vL1b)
+                B = ptInTriang(v, vL2a, vL1b, vL2b)
+
+                if A or B:
+                    d = 1
+                    break
+        if d > 0:
```
zeffii:
```diff
-        if d > 0:
-            mask.append(0)
-        else:
-            mask.append(1)
+        mask.append(0 if d > 0 else 1)
```

nodes/modifier_make/contour2D.py
```diff
@@ -0,0 +1,599 @@
...
+            vm[2] += v1[2] * 0.5
+        midPoints.append(vm)
+    return midPoints
+
+
+def maskEdges(edges, mask):
+    edges_out = []
+    for m, ed in zip(mask, edges):
+        if m:
+                edges_out.append(ed)
```
zeffii: indentation ++

nodes/modifier_make/contour2D.py
```diff
@@ -0,0 +1,599 @@
...
+    def process(self):
+        inputs, outputs = self.inputs, self.outputs
+
+        # inputs
+
+        if inputs['Verts_in'].is_linked:
+
+            VertsAll = inputs['Verts_in'].sv_get(deepcopy=False)
+        else:
+            VertsAll = [[(0.0, 0.0, 0.0)]]
```
zeffii: consider using the `default` parameter, which will be used in case `.is_linked` returns False
```diff
-        if inputs['Verts_in'].is_linked:
-
-            VertsAll = inputs['Verts_in'].sv_get(deepcopy=False)
-        else:
-            VertsAll = [[(0.0, 0.0, 0.0)]]

+       VertsAll = inputs['Verts_in'].sv_get(deepcopy=False, default=[[(0.0, 0.0, 0.0)]])
or (not sure about the extra wrapping)
+       VertsAll = inputs['Verts_in'].sv_get(deepcopy=False, default=[(0.0, 0.0, 0.0)])
```

utils/modules/geom_utils.py
```diff
@@ -74,3 +74,22 @@ def mean(verts):
 
 def is_reasonably_opposite(n, normal_one):
     return dot_v3v3(normalized(n), normalized(normal_one)) < 0.0
+
+
+def ptInTriang(p_test, p0, p1, p2):
```
zeffii: `def pt_in_triangle` this module has lowercase function names with words separated by `_` , as is common practice in Python code.

nodes/modifier_make/contour2D.py
```diff
@@ -0,0 +1,526 @@
...
+        self.inputs.new('StringsSocket', "Nº Vertices").prop_name = 'vert_'
+        self.inputs.new('VerticesSocket', "Verts_in")
+        self.inputs.new('StringsSocket', "Edges_in")
+
+        self.outputs.new('VerticesSocket', "Vertices", "Vertices")
+        self.outputs.new('StringsSocket', "Edges", "Edges")
+
+    def draw_buttons(self, context, layout):
+        layout.prop(self, "modeI", expand=True)
+        r = layout.row(align=True)
```
zeffii: what's the assignment for?

nodes/modifier_make/contour2D.py
```diff
@@ -0,0 +1,526 @@
...
+
+    def process(self):
+        inputs, outputs = self.inputs, self.outputs
+
+        # inputs
+        vertsAll = inputs['Verts_in'].sv_get(deepcopy=False, default=[[(0.0, 0.0, 0.0)]])
+        radiusAll = inputs['Radius'].sv_get(deepcopy=False, default=[[self.rad_]])
+        verticesAll = inputs['Nº Vertices'].sv_get(deepcopy=False, default=[[self.vert_]])
+        verticesAll = [list(map(lambda x: max(2, int(x)), Vertices)) for Vertices in verticesAll]
+        edgesAll = inputs['Edges_in'].sv_get(deepcopy=False, default=[[]])
```
zeffii: : ) this is less chaotic

nodes/modifier_make/contour2D.py
```diff
@@ -0,0 +1,526 @@
...
+            n += 4
+        return Verts, Edges
+
+    def make_edges(self, Vertices, net, pointsL):
+        listEdg = [(i, i + 1) for i in range(pointsL - 1)]
+        if len(net) != 1:
+            listEdg.append((pointsL-1, 0))
+        return listEdg
+
+    def make_faces(self, Vertices):
```
zeffii: skip the variable..
```python
def make_faces(self, Vertices):
    return [list(range(Vertices))]
```
oh..hehe i see you aren't using this function.

nodes/modifier_make/contour2D.py
```diff
@@ -0,0 +1,526 @@
...
+
+        # inputs
+        vertsAll = inputs['Verts_in'].sv_get(deepcopy=False, default=[[(0.0, 0.0, 0.0)]])
+        radiusAll = inputs['Radius'].sv_get(deepcopy=False, default=[[self.rad_]])
+        verticesAll = inputs['Nº Vertices'].sv_get(deepcopy=False, default=[[self.vert_]])
+        verticesAll = [list(map(lambda x: max(2, int(x)), Vertices)) for Vertices in verticesAll]
+        edgesAll = inputs['Edges_in'].sv_get(deepcopy=False, default=[[]])
+
+        # outputs
+        if outputs['Vertices'].is_linked:
```
zeffii: it's fine to check `*socket.is_linked` like you do here, but i recommend  using that to `return` early, so you can drop an indentation level in the rest of the block.
```python
if not outputs['Vertices'].is_linked:
    return

family = listMatcher([vertsAll, radiusAll, verticesAll, edgesAll], self.listMatch)
...
```

nodes/modifier_make/contour2D.py
```diff
@@ -0,0 +1,526 @@
...
+
+                    net = buildNet(verts_in, edges_in, vLen, actualRadius[i])
+                    parameters = listMatcher([verts_in, Vertices, actualRadius[i], net], self.listMatch)
+                    parameters = [data[0:vLen] for data in parameters]
+
+                    points = [self.make_verts(vi, v, r, n) for vi, v, r, n in zip(*parameters)]
+                    totalPoints = 0
+                    for p in points:
+                        totalPoints += len(p)
+
```
zeffii: `total_points = sum(len(p) for p in points)`  ..just a suggestion.

nodes/modifier_make/contour2D.py
vicdoval: cool alternative :)

vicdoval:
Inside the Contour2D node:
the function Intersect Edges is an verbatim extract of IntersectEdges_mk2.py
the function MaskVertices  is an verbatim extract of vertices_mask.py

You think it would  be a good idea to extract the function in the original files and then importing it via  `import`?

portnov:
My personal "rule of thumb" for such questions is something like: if the block of code is used more than twice, OR if it is "complicated" in some sense — then it is usually better to factor it into separate method. There can be exceptions, mostly about code that uses a lot of variables from context (so that the extracted method would expect ten arguments or smth like that). But such code is anyway a candidate for refactoring :)

zeffii:
yeah i'm also leaning towards pushing more of these functions into shared modules
```
sverchok/utils/modules/<some_descriptive_name>_utils.py
```
if you need to customize the function you dupe it and append some form of indentifier like ..
```python
def original_func_name_<mod>(...):
    """
    this version of the function is used by node x, modified by y, for z reasons.
    """
```
just thinking out loud there.

vicdoval:
Images for docs
contour2d_head
multiple_distances_contour2d
multiple_objects_contour2d
repeat_last_or_cycle_contour2d

nodes/modifier_change/edges_intersect_mk2.py
```diff
@@ -134,6 +134,41 @@ def unselect_nonintersecting(bm, d_edges, edge_indices):
...
+
+    # optional correction, remove originals, add back those that are not intersecting.
+    edges_out = edges_out[trim_indices:]
+    edges_out.extend(add_back)
+    bm.free()
+
+    return verts_out, edges_out
+
+
+def remove_doubles_with_edges(verts_in, edges_in, distance):
```
zeffii:
```diff
- def remove_doubles_with_edges(verts_in, edges_in, distance)
+ def remove_doubles_from_edgenet(verts_in, edges_in, distance)
```
`*_width_edges` had me confused, but `*_from_edgenet` might be more descriptive

zeffii:
about the mode / enum property

```python
modeItems = [
    ("Constant", "Constant", ""),
    ("Weighted", "Weighted", "")]

listMatchItems = [
    ("Long Repeat", "Long Repeat", ""),
    ("Long Cycle", "Long Cycle", "")]
```

This describes the enum list _minimally_, but it is less favourable to upgrades. This why most of our enumeration lists have explicit indices.
```
modeItems = [
    ("Constant", "Constant", "", 0),
    ("Weighted", "Weighted", "", 1)]
```
that way if you want to add enum members you can be sure that their indices won't be taken over by another enum member. (without indices you will encounter surprises if you add members to the middle of that list, _order matters when you don't supply indices_)

an option is to use a lambda/ function to generate the items
```python
lambda items: [(n, n, '', idx) for n, idx in items]

# leaving unused indices between 0 and 10 allows you to add items in between at a later stage.
nodeItems = items([["Constant", 0], ["Weighted", 10]])   
```
or.. if you want to have less `[...]` then
```python
def enum_items(*item_list):
    return [(n, n, '', idx) for n, idx in item_list]

nodeItems = enum_items(["Constant", 0], ["Weighted", 10])
```
------------

whatever you do, do add indices.

zeffii:
the last thing i want to say is that the process function is quite hefty,, can I encourage you to try and abstract parts of that into one or more function calls, and let the name of the function describe the inner working of that block you are abstracting away.

nodes/modifier_make/contour2D.py
```diff
@@ -0,0 +1,492 @@
...
+        radiusAll = inputs['Distance'].sv_get(deepcopy=False, default=[[abs(self.rad_)]])
+        radiusAll = [list(map(lambda x: abs(x), radius)) for radius in radiusAll]
+
+        verticesAll = inputs['Nº Vertices'].sv_get(deepcopy=False, default=[[self.vert_]])
+        verticesAll = [list(map(lambda x: max(2, int(x)), Vertices)) for Vertices in verticesAll]
+
+        edgesAll = inputs['Edges_in'].sv_get(deepcopy=False, default=[[]])
+
+        # outputs
+        if not outputs['Vertices'].is_linked:
```
zeffii: this check can go directly at the top, stops all further processing if no outputs linked,

zeffii:
i think we can squash + merge this without too much further delay, if all nodes that have been edited exhibit no adverse effects.

zeffii:
@vicdoval  i like to get the process function down to almost nothing if possible, by abstracting 'actions' into functions which document themselves by having a reasonable function name. Here's step 1. notice that there's still a huge function which is definitely not easy to understand for an outsider/newcomer (me)

https://gist.github.com/zeffii/eaccc53befe30c54a76323dd74626cb1

**warning** i've not tested that - it might not work immediately. without persuading - i wish only to impress upon you the way we might abstract the more mundane stuff into functions.

step 2.  is going to break that `generate_outlines` function up, just enough to make it less imposing.

zeffii:
proof
```python
test_list = [[],[]]

def mutate_incoming(test_list, params):
    a, b = params
    list_a, list_b = test_list
    list_a.append(a)
    list_b.append(b)

a_list = [20,30,40]
b_list = ['huh', 'yes', 'ack']

for params in zip(a_list, b_list):
    mutate_incoming(test_list, params)

list_a, list_b = test_list
print(list_a, list_b)


>> [20, 30, 40] ['huh', 'yes', 'ack']
```

vicdoval:
When you talked about working with text I realized how slow was the node,  so i re-wrote everything upside down... at one point y stepped with the "intersect lines 2d" and that was an insane performance difference.
So I added the 2d routine to the intersect edges node...
I also made the changes  to empty de process function  :)

zeffii:
excellent @vicdoval ! did you see the changes @portnov made to the intersect_edges_mk2 node? (just because we don't see conflict warnings, doesn't mean...i can assume it's all going to be OK merging)

portnov:
https://gist.github.com/6d1433cf1825c447c5d6800fdbf53533

curve_contour_1

Interesting thing :)

zeffii:
@vicdoval  it's not clear to me why you remove the epsilon/min_dist ,  in the 2d version. Does the 2d version work for very large scale 10k+ bu mesh as well as much smaller than 1 bu scales?

zeffii:
am eager to merge this...

vicdoval:
It is a straight implementation of `mathutils.geometry.intersect_line_line_2d`, it is quite fast so there is no previous  edge removal.
The function returns `None ` in case of non intersecting pairs :)
Maybe more optimization could be made but  already hundred times faster than the 3d implementation and seems to work in all cases

zeffii:
cool. i'm going to merge unless you want to keep working on it before that..

vicdoval:
Last images for the docs
contour9d
contour5d
contour8d

vicdoval:
I don't know if I did well the rst file, when i see it on github web looks like code, not like the other rst.
Any hint?

zeffii:
@vicdoval it might need to be on a sverchok branch directly to show the .rst as html.. it is weird i admit.

vicdoval:
OK, then I think I'm ready for merging  :D

zeffii:
@vicdoval congratulations and thank you for this node :)

vicdoval:
Thanks @zeffii , I really believe this is a great project and I'm glad i could place my grain of sand on it :)

zeffii:
i have a few edits to make to the commit.  (haha... it never ends)

vicdoval:
there is always some last edits... :D 

And the rst still does not render... maybe it needs some time :D

zeffii:
leave the .rst to me. i'll report back when it resolves

zeffii:
@vicdoval  under the word "Parameter" in the table you had a AWOL char `-` instead of `=` :)

zeffii:
sorry for bulldozing the other rst formatting.

vicdoval:
great that now it is working! 👍

zeffii:
image

Durman:
@vicdoval I would like to ask is it possible to make the different modes working of this node such as make only inner contours or only outer contours or make contours to both directions how it works now? I already have tried to use this certainly useful node and I faced with problem to separate inner and outer contours.

Also it would not be bad to leave some comments about mode connect of vector sort. It seems does not work now.
https://gist.github.com/856dd6547cd0e7fed415d4ef33925c42

vicdoval:
Thanks @Durman  I'm glad you find it useful :)
The `Contour 2D` node works with edges, so there is not a concept like inside or outside.
You can use `Separate Loose Parts` to get each contour
contour2d_separate_contours

About the Vector Sort `Connect Mode`:
I've just made a pull request with the fix. 
It is meant to work with edges, specially with vertex with two connections. Check this gists https://gist.github.com/985bca8ca309c522b58108fd6368e08c
vector_sort_connex

https://gist.github.com/7e6c9cf4494fa88439213ae179914f53
vector_sort_connex1

Durman:
@vicdoval Thanks, if outer contour always is first in list of contours then there is no problem. We used `topo sort for profile script` in `SNL` before, @zeffii  should be it deleted now?

zeffii:
@Durman  i'd consider making a folder in `node_scripts` for `old_scripts`, to keep the old stuff around as a reference. Not likely to delete the SNL script so soon, it's useful as a teaching tool.

stansu:
i found out that vector sort by connection won't work when edges is not closed.
https://gist.github.com/2afb7c38a6b00798fc051f449fc9f5fc

enzyme69:
@stansu can try maybe a example Blend example and then we make new thread

stansu:
@enzyme69 we continue this topic in here #2100

PR #2015, opened 2018-01-05
prep profilenodemk2 for def storage_*

zeffii:


zeffii:
@portnov  this addresses profilenodemk2 iojson storage to the extent that I can test it.

Issue #1984, opened 2017-12-28
Vector in node, fullList to have a deepcopy alternative.


Durman:
![2017-12-28_14-36-41](https://user-images.githubusercontent.com/28003269/34408357-c391bc6e-ebdc-11e7-8197-c6dbc52092ca.png)

zeffii:
Im not certain this is wrong, you should also show the textbased output,

zeffii:
it's interesting however, that it's not symmetrical..  if you look at Vector In's source, maybe you can pinpoint why this is. Probably elements are extended to be as long as the first element, unless the current element is longer than the first or previous element.

Durman:
After first cycle For, Y and Z coordinate are filled and have 5 elements. :/

![2017-12-29_10-08-29](https://user-images.githubusercontent.com/28003269/34430640-8da143aa-ec80-11e7-9bac-d47c12674a84.png)

zeffii:
but is it _wrong_ or _unexpected_ or _undesired_. 

i understand the point you make, at one point we must have agreed that the current behaviour was desired, but I can completely agree that alternative behaviour might also be needed sometimes. It would need to be a MK2 with maybe a mode switch in `draw_buttons_ext` to go from `default mode` to `minimal mode`.

Durman:
@zeffii 
>but is it wrong or unexpected or undesired.

Of course it is wrong. Look:
![2017-12-29_22-18-45](https://user-images.githubusercontent.com/28003269/34444287-812f1226-ece6-11e7-8f3d-c97ed9090686.png)
```Python
        for i in range(max_obj):

            max_v = max(map(len, (X_[i], Y_[i], Z_[i])))
            print('Y_ -', Y_)
            fullList(X_[i], max_v)
            fullList(Y_[i], max_v)
            fullList(Z_[i], max_v)
            series_vec.append(list(zip(X_[i], Y_[i], Z_[i])))
```
We can see here that `fullList(Y_[i], max_v)` didn't fill values for item object but do it for all objects.

Durman:
@zeffii `fullList` function can work with nested lists correctly?

zeffii:
i've had a few beers :) i don't know what to think about this.

i didn't write fullList but it appears that we focus how it works by passing it the least nested list possible, when extending its length.

see implementation:

https://github.com/nortikin/sverchok/blob/69058d106528a671a74394751b12ed6911dabe1c/data_structure.py#L151-L157

[...]

Issue #1653, opened 2017-05-14
Custom Defaults for nodes. (already in master)


zeffii:
### Node Defaults

When we (as coders) submit a node, we decide on sane defaults. Sometimes those defaults are not optimal for you (the user). If you find yourself always tweaking the parameters of a node before using it, then you might want to define custom node defaults. 

We can now configure node defaults from a _.json_ file. 

##### datafiles/sverchok/node_defaults/node_defaults.json
```json
{

    "SvGenFloatRange": {
        "props": [
            ["stop_", 1.0],
            ["count_", 20],
            ["mode", "FRANGE_COUNT"]
        ]
    },

    "ViewerNode2": {
        "props": [], 
        "function": "view3d.init_vdmk2"
    }
}
```
All we have to do is create key, value pair for each node.

- key: ( the node's `bl_idname`) 
- value: Currently the `value` accepts two directives (but can be extended later)

    1. `props`  ( _push props in sequence with a list of prop_name, prop_value pairs._ )
    2. `function` ( _push a function:_ in the example below the function is referenced as `view3d.init_vdmk2`. read that as `filename.function` and filename shall be located in `datafiles/sverchok/node_defaults/` )

##### datafiles/sverchok/node_defaults/view3d.py
```python
import random
import colorsys

import bpy



def init_vdmk2(node):

    def fold_color(color, shift):
        h, s, v = colorsys.rgb_to_hsv(*color)
        return colorsys.hsv_to_rgb(shift, s, v)

    A = [0.938000, 0.948000, 0.900000, 1.000000]
    B = [0.500000, 0.752000, 0.899000, 1.000000]
    C = [0.030100, 0.488000, 0.899000, 1.000000]

    shift = random.random()
    node.vertex_colors = fold_color(A[:3], shift)
    node.edge_colors = fold_color(B[:3], shift)
    node.face_colors = fold_color(C[:3], shift)

    node.edge_width = 1.0
    node.vertex_size = 2.0

```

------------------

Each time you add a node, the `node_default_dict` is checked if the `bl_idname` has an entry. 

- this has no effect on nodes that you duplicate, they clone their properties according to `self.copy()`
- this doesn't offer a visual way to set the preferences yet, that would be the next step.

zeffii:
here's my current `node_defaults` dir / repo . it will be updated as i progress..
https://github.com/zeffii/node_defaults

zeffii:
[...]

PR #2019, opened 2018-01-07
Adding Deepcopy version of FullList

Durman:
object mode
#1984


zeffii:
"object_mode" doesn't strike me as a good name for that mode.  "deepcopy" mode, would be less ambiguous

also, this can be tidier:
```python
        if self.object_mode:
            fullList_deep_copy(X_, max_obj)
            fullList_deep_copy(Y_, max_obj)
            fullList_deep_copy(Z_, max_obj)
        else:
            fullList(X_, max_obj)
            fullList(Y_, max_obj)
            fullList(Z_, max_obj)
```
maybe
```python
        fullList_main = fullList_deep_copy if self.deepcopy_mode else fullList
        fullList_main(X_, max_obj)
        fullList_main(Y_, max_obj)
        fullList_main(Z_, max_obj)
```

interesting that you want to change the nestedness, (i'm not saying that's wrong.. just it might not be suitable as a replacement to vector_in with the same bl_idname.  old .blends may rely on "broken" behaviour)

Durman:
@zeffii I am not sure that deep copy is clear for users that don't know about Python. This new button lets to generate objects correctly so it seems more suitable name for this function but i can fix it of course.))

zeffii:
yes, please change it anyway.  `object_mode` is too much like Blender's own "Object Mode".

zeffii:
It might also be more appropriate to have that button in `draw_buttons_ext`, what do you think @portnov  @nortikin @DolphinDream   , or as the default behaviour of a `SvVectorInMk2` node, and give the draw_buttons_ext a toggle for the `non-deepcopy`.

portnov:
Can we set the property to be defaulting to False (old behaivour) for existing node, but to True for newly created nodes?

portnov:
For example, make BoolProperty(default=False), but set it to true in sv_init.

portnov:
And yes, I'd prefer this to be in draw_buttons_ext.

nortikin:
If we do such things, maybe to thing of generalize them?
I.e. 

1. Other nodes may need that too. 
2. Unify that function property also needed. 
3. Also if there is already such function we need deal with it also.

Durman:
@zeffii I have problem with switcher. Switcher does not appear when I add draw_buttons_ext function :/
Like this:
```Python
        layout.row().prop(self, 'object_mode')
```

Durman:
@nortikin Which nodes is needed to fix? I noticed that polar input node already works right.

nortikin:
not sure
beside of subject:
list split changes level at changing unwrap flag. it ruine IO.

zeffii:
> Switcher does not appear when I add draw_buttons_ext function :/
> Like this:
>    ```python
>    def draw_buttons_ext(self, context, layout):
>        layout.row().prop(self, 'object_mode')
>   ```

no reason that doesn't work @Durman  , check your error console?

image

Durman:
@zeffii Oh, I didn't notice that this function create switcher for N panel. :/

zeffii:
if you want you could add another boolean to the class `show_switcher`, and do
```
    ...
    ...
    if self.show_switcher:
        layout.row().props(self, "deepcopy_mode")
    ...
```

zeffii:
then 
```python
    layout.prop(self, 'show_switcher')  # <-- set here if you want to show it in draw_buttons
    layout.prop(self, 'deepcopy_mode')  # <-- always show this here.
```

zeffii:
with the new contextual right-click menu for nodes, @portnov might be able to show a way to add this "show_me" option as a rightlick menu entry, so you don't have to use the N-panel if it's inconvenient.

portnov:
Huh. I'm not sure, will layout.prop() work for menus? You can try anyway :)
After #2023 it can look like (in node class):

```python
  def rclick_menu(self, layout, context):
    super().rclick_menu(layout, context)
    layout.prop(self, "deepcopy_mode")
```

Or if it occurs that Blender doesn't support properties in menus, then you'll have to write an operator to be called from menu...

zeffii:
@portnov 

image

but as it receives a click event, the menu disappears. it will work, but ..

Durman:
Ok, I will wait.

@nortikin All generators nodes can't  generate mesh in the same object but we can do it with existing means:

vectorize generetion nodes

Durman:
I added rclick function but it does not work:

portnov:
Sorry then — rclick_menu(layout, context) :) See how it is defined in node_tree.py.

Durman:
I have looked ofcourse but it is dark forest for me.)) I think it can be merged in this condition.

zeffii:
@Durman  in nodetree.py each node gets this function by default
```python
    def rclick_menu(self, context, layout):
        """
        Override this method to add specific items into
        node's right-click menu.
        Default implementation calls `node_replacement_menu'.
        """
        self.node_replacement_menu(context, layout)
```
if you wanted to override that you would add this function to the node and make it something like this:

```python
    def rclick_menu(self, context, layout):
        layout.prop(self, "advanced_mode", text="use deep copy")

        # ignore this for now 
        # self.node_replacement_menu(context, layout)
```

Durman:
It works )

right click

zeffii:
the rightclick stuff is relatively new to us too.  but I think useful to users / devs alike

nortikin:
amazing. really it can handle one parameter for all nodes for some property. for example, separate flag

zeffii:
shall we merge this now @nortikin @portnov @Durman   ?

zeffii:
@Durman  remember you can make a macro for this node so it always ticks that box for you automatically when you add it.  See steps here : https://github.com/nortikin/sverchok/issues/1653

zeffii:
@Durman  your `../279/datafiles/sverchok/node_defaults/node_defaults.json` could look like this
```json
{
    "GenVectorsNode":  {"props": [["advanced_mode", True]]}
}

```

portnov:
I'd prefer "old school" checkbox in the N panel. For all other, I'm fine with merging this.

zeffii:
wait.. i thought the check was both in the N panel and appearing in the Rclick menu. it definitely needs to be in both.

zeffii:
well.. it _is_ in `draw_buttons_ext` too , i understand the reticence to the bool in a menu, but it's not a blocker for me.

portnov:
ah, so it is in two places. Ok then.

zeffii:
Fingers crossed :)

PR #2025, opened 2018-01-08
Generic text ops callback

zeffii:


zeffii:
i'm not going to register a new operator for each callback , just to get the tooltip. not today.

PR #2026, opened 2018-01-08
Add issue template and a PR template.

portnov:
This adds templates for Github issues and PRs. Obviously seasoned contributers can totally ignore them and write good issues/PRs without them, but for novices I think this is a good starting point.

@zeffii  Maybe some ideas?

### DIFF
--- a/docs/pull_request_template.rst
+++ b/docs/pull_request_template.rst
@@ -1,0 +1,19 @@
+## Addressed problem description
+
+Please describe what problem does this PR solve. If it was already was disussed in an issue, it is enough to give a link as #number.
+
+## Solution description
+
+Please describe how do you intend to solve the problem: change some method, or introduce new class, or whatever.
+
+## Preflight checklist
+
+Put an x letter in each brackets when you're done this item:
+
+- [ ] Code changes complete.
+- [ ] Code documentation complete.
+- [ ] Documentation for users complete (or not required, if user never sees these changes).
+- [ ] Manual testing done. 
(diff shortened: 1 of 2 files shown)


docs/pull_request_template.rst
```diff
@@ -0,0 +1,19 @@
+## Addressed problem description
+
+Please describe what problem does this PR solve. If it was already was disussed in an issue, it is enough to give a link as #number.
```
zeffii: discussed

portnov: fixed.

zeffii:
yes, i am all for a light introduction
   - show how to view the errors
   - show how to copy the error to the clipboard (instead of a screenshot)
   - describe the kind of screenshots that are helpful (not too large, cropped to the subject... )
   - if you have a code question, link it as a gist if it's a single node, or the the markup if it's a snippet.

yes, i have more to say about this for certain.

portnov:
@zeffii You are welcome to push to this branch :)

Though a full description of

> show how to view the errors
> show how to copy the error to the clipboard (instead of a screenshot)
> describe the kind of screenshots that are helpful (not too large, cropped to the subject... )

I think better locate in separate document (troubleshooting_guide.rst ?), and give a link to it in issue template. Because if we will have a lot of text in issue template it might look a bit weird.

zeffii:
i'll think about it.

portnov:
I think it's better to have these templates in current state than not have any at all. We can always update them when we feel the need.

PR #2029, opened 2018-01-08
Update formula_deform.py

Kosvor2:
Now it is also possible to get index of current object list evaluated as "I" variable.

So "i" for index of vertex, and "I" for index of object.


Kosvor2:
Wait, something is wrong.

PR #2030, opened 2018-01-08
Deformula index per object (and fixed imports in Exec Node)

Kosvor2:
Now it is also possible to get index of current object list evaluated as "I" variable.

So "i" for index of vertex, and "I" for index of object.

### DIFF
--- a/nodes/script/multi_exec.py
+++ b/nodes/script/multi_exec.py
@@ -197,7 +197,7 @@ class SvExecNodeMod(bpy.types.Node, SverchCustomTreeNode):
         append = out.append
 
         # locals() is needed for generic module imports.
-        exec('\n'.join([j.line for j in self.dynamic_strings]), globals().update(locals()))
+        exec('\n'.join([j.line for j in self.dynamic_strings]), globals(), locals())
 
         self.outputs[0].sv_set(out)
 
--- a/nodes/script/multi_exec.py
+++ b/nodes/script/multi_exec.py
@@ -197,7 +197,7 @@ class SvExecNodeMod(bpy.types.Node, SverchCustomTreeNode):
         append = out.append
 
         # locals() is needed for generic module imports.
-        exec('\n'.join([j.line for j in self.dynamic_strings]), locals())
+        exec('\n'.join([j.line for j in self.dynamic_strings]), globals().update(locals()))
 
         self.outputs[0].sv_set(out)
 
--- a/nodes/vector/formula_deform.py
+++ b/nodes/vector/formula_deform.py
@@ -48,13 +48,9 @@ class SvFormulaDeformNode(bpy.types.Node, SverchCustomTreeNode):
         Io = self.inputs[0]
         Oo = self.outputs[0]
         if Oo.is_linked:
-            out = []
             V = Io.sv_get()
             IndObj = range(len(V))
-            Value = "[("+self.ModeX+","+self.ModeY+","+self.ModeZ+") for (x, y, z),i in zip(L,range(len(L)))]"
-            for L,I in zip(V,IndObj):
-                out.append(eval(Value, globals().update(locals())))
-            Oo.sv_set(out)
+            exec("Oo.sv_set([[("+self.ModeX+","+self.ModeY+","+self.ModeZ+") for (x, y, z),i in zip(L,range(len(L)))] for L,I in zip(V,IndObj)])")
 
     def update_socket(self, context):
         self.update()
--- a/nodes/vector/formula_deform.py
+++ b/nodes/vector/formula_deform.py
@@ -49,8 +49,8 @@ class SvFormulaDeformNode(bpy.types.Node, SverchCustomTreeNode):
         Oo = self.outputs[0]
         if Oo.is_linked:
             V = Io.sv_get()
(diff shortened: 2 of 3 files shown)


Kosvor2:
[deformula4](https://user-images.githubusercontent.com/22656834/34698753-a1bd913a-f4fb-11e7-88a4-13a4f5cb3519.png)

@zeffii,  Exec node have numpy imported as np internally, but `locals()` not allow such imports.
So i used `globals().update(locals())` combined dict instead. Check if it works for you.

zeffii:
can't we do 
```python
exec( the_string, globals(), locals())
```
sorry I didn't extensively test my patch in ExecNodeMod.

Kosvor2:
`globals(), locals()` indeed works for `exec()` but not for `eval()` in deformula node.

`globals().update(locals())` seems to work well in both nodes.

portnov:
I don't think that `globals().update(locals())` is a good idea...

```python
var1 = 1

def func(x):
    var1 = 5 # this is a local variable with the same name as global one, nothing criminal
    globals().update(locals())
    return x*x

y = func(7)

print(var1) # prints 5
```

Kosvor2:
Maybe it will be better to not using `eval()` and convert deformula node to use of `exec()`.

nodes/vector/formula_deform.py
```diff
@@ -50,10 +50,10 @@ def process(self):
         if Oo.is_linked:
             out = []
             V = Io.sv_get()
-            Value = "[("+self.ModeX+","+self.ModeY+","+self.ModeZ+") for (x, y, z),i in zip(L, I)]"
-            for L in V:
-                I = range(len(L))
-                out.append(eval(Value))
+            IndObj = range(len(V))
```
zeffii: this looks incredibly awkward..
```python

        if Oo.is_linked:
            out = []
            V = Io.sv_get()
            IndObj = range(len(V))
            Value = "[("+self.ModeX+","+self.ModeY+","+self.ModeZ+") for (x, y, z),i in zip(L,range(len(L)))]"
            for L,I in zip(V,IndObj):
                out.append(eval(Value, globals().update(locals())))
            Oo.sv_set(out)
```

zeffii: (doesn't mean I immediately know a nicer way..)

Kosvor2:
Now, with `exec()`, deformula node works like a charm. No even need for `globals()` or `locals()`.

nodes/vector/formula_deform.py
```diff
@@ -48,13 +48,9 @@ def process(self):
...
         if Oo.is_linked:
-            out = []
             V = Io.sv_get()
-            Value = "[("+self.ModeX+","+self.ModeY+","+self.ModeZ+") for (x, y, z),i in zip(L, I)]"
-            for L in V:
-                I = range(len(L))
-                out.append(eval(Value))
-            Oo.sv_set(out)
+            IndObj = range(len(V))
+            exec("Oo.sv_set([[("+self.ModeX+","+self.ModeY+","+self.ModeZ+") for (x, y, z),i in zip(L,range(len(L)))] for L,I in zip(V,IndObj)])")
```
zeffii: does this work
```python
exec_string = "Oo.sv_set([[({n.ModeX},{n.ModeY},{n.ModeZ}) for (x, y, z), i in zip(L,range(len(L)))] for L, I in zip(V,IndObj)])"
exec(exec_string.format(n=self))
```
or..nicer..
```python
exec_string = "Oo.sv_set([[({n.ModeX},{n.ModeY},{n.ModeZ}) for i, (x, y, z) in enumerate(L)] for I, L in enumerate(V)])"
exec(exec_string.format(n=self))
```

nodes/script/multi_exec.py
```diff
@@ -197,7 +197,7 @@ def process(self):
         append = out.append
 
         # locals() is needed for generic module imports.
-        exec('\n'.join([j.line for j in self.dynamic_strings]), locals())
+        exec('\n'.join([j.line for j in self.dynamic_strings]), globals().update(locals()))
```
zeffii: i'm tempted to make it this way..
```
node_dict = {}


def update_import_locals(self, context):
    exec(self.imports)
    self.node_dict[hash(self)]['imports'] = locals()

def get_imports(self):
    try:
        return self.node_dict[hash(self)]['imports']
    except:
        print('imports failed')
        return {}
    

imports = StringProperty(name='defined imports', description='string to store imports', update=update_import_locals)


self.imports = "from math import pi, sin; import numpy as np"

    layout.row().prop(self, 'imports')
    if self.imports:
        import_lines = self.imports.split(';')
        for line in import_lines
            layout.row().label(line)


def process(self):

    ...

        locals().update(self.get_imports())
        exec('\n'.join([j.line for j in self.dynamic_strings]))
```

zeffii:
but for now...
```python
exec('\n'.join([j.line for j in self.dynamic_strings]), globals(), locals())
```
should be OK right?

zeffii:
cool.. i looked at this anyway..  for fun
image

this is probably overkill...  but interesting to abstract away the imports.. 
https://github.com/nortikin/sverchok/pull/2031

Issue #1728, opened 2017-06-25
Cross Section, Bisect nodes and objects.


Durman:
Is it possible to cut objects individually? I think this nodes unite objects in one inside itself. It will be better if one matrix cut first object and two matrix cut second object, for me in my case.))

![2017-06-25_14-42-44](https://user-images.githubusercontent.com/28003269/27515455-9491c1e4-59b4-11e7-8b26-293e91168122.png)
https://gist.github.com/8b592141eba2da47443ba4e0690fcf6a

zeffii:
@Durman  the node doesn't join the objects, but it does perform all matrix cuts on each incoming object.

zeffii:
look at the code to understand why
```
        for cut_mat in cut_mats:
            pp = cut_mat.to_translation()
            pno = Vector((0.0, 0.0, 1.0)) * cut_mat.to_3x3().transposed()
            for obj in zip(verts_ob, edg_pols):
                res = bisect(obj[0], obj[1], pp, pno, self.outer, self.inner, self.fill)
                if not res:
                    return
                verts_out.append(res[0])
                edges_out.append(res[1])
                polys_out.append(res[2])
```

zeffii:
it makes sense to add a toggle to use the above algorithm as "All matrices" vs a mode that does "individual matrices".

zeffii:
![image](https://user-images.githubusercontent.com/619340/27516683-0709c76c-59bf-11e7-95b9-564a4bbda6d9.png)

zeffii:
```python
# ##### BEGIN GPL LICENSE BLOCK #####
#
#  This program is free software; you can redistribute it and/or
#  modify it under the terms of the GNU General Public License
#  as published by the Free Software Foundation; either version 2
#  of the License, or (at your option) any later version.
#
#  This program is distributed in the hope that it will be useful,
#  but WITHOUT ANY WARRANTY; without even the implied warranty of
#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
#  GNU General Public License for more details.
#
#  You should have received a copy of the GNU General Public License
#  along with this program; if not, write to the Free Software Foundation,
#  Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
#
# ##### END GPL LICENSE BLOCK #####

import bpy
from bpy.props import BoolProperty
import bmesh
from mathutils import Vector

from sverchok.node_tree import SverchCustomTreeNode
[...]

PR #2040, opened 2018-01-13
Metaball out vectorized (WIP)

portnov:
The goal is to be able to create several META objects. For that, the Metaball Out node should be able to consume a list of lists of matrices: one list of matrices per Meta object. Or it in consume just a list of matrices (compatibility mode) and produce single object.


Durman:
@portnov How are you going to generate list of lists of matrices? I have received the same question from zeffi here: #1728 ))

Durman:
Can you give your example with monad node? I don't manage to create this.

portnov:
Well, I did not try yet :)

zeffii:
i'll port this version over to 2.8 today i think :)

zeffii:
i've ported this to sv 4 2.80

PR #2046, opened 2018-01-14
moves old class to old node where it belongs

zeffii:

orphan class takes up brainspace in nodetree.py - no longer.

## Solution description

moved to the viewer.py in old_nodes, where this class is used exclusively.

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -81,18 +81,6 @@ def process_from_socket(self, context):
     self.node.process_node(context)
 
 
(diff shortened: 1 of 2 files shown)


zeffii:
I doubt that there are many old files with this node/class, that aren't totally broken anyway.

zeffii:
`node.color_view` (for `ViewerNode` isn't tracked anyway for iojson.. and I won't be adding it)

PR #2048, opened 2018-01-14
text viewer mk3 haha!!

nortikin:
Node about viewer text PR. Old approach for trash.

## Solution description
MK3
vtmk3
and https://github.com/nortikin/sverchok/pull/2048/files#diff-0147edc4e7953b77a018b66924e77a37R48 was error giving. name unknown

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.


portnov:
comments.

Durman:
Maybe it will be interesting to have somewhere name of the label of a input node if the label is filled.

nortikin:
test left. how to make... there was issue i guess

nodes/text/viewer_text_mk3.py
```diff
@@ -0,0 +1,226 @@
...
+                if label:
+                    label = '; node ' + label.upper()
+                name = insert.name.upper()
+                # vertices socket
+                if insert.other.bl_idname == 'VerticesSocket':
+                    itype = '\n\nSocket ' + name + label + '; type VERTICES: \n'
+
+                # edges/faces socket
+                elif insert.other.bl_idname == 'StringsSocket':
+                    itype = '\n\nSocket ' + name + label + '; type EDGES/POLYGONS: \n'
```
portnov: Huh. @zeffii this is nearly the same as we were talking in #2058: text viewer nodes assumes that StringsSockets is for edges or polygons, but in fact it is for anything for now. @nortikin maybe put a bit more general phrase here, smth like "EDGES/POLYGONS/OTHER" ?

Also I'm not sure about caps :)

nodes/text/viewer_text_mk3.py
```diff
@@ -0,0 +1,226 @@
...
+    bl_label = "Sverchok viewer.mk1"
+    bl_options = {'REGISTER', 'UNDO'}
+
+    nodename = StringProperty(name='nodename')
+    treename = StringProperty(name='treename')
+    lines = IntProperty(name='lines', description='lines count for operate on',default=1000)
+
+    def execute(self, context):
+        node = bpy.data.node_groups[self.treename].nodes[self.nodename]
+        inputs = node.inputs
```
portnov: I'd propose to extract most logic of this execute() method into separate method, that could be called without operator. Something like `def get_text(input_socket)` or maybe even `def get_text(output_socket)`.

nodes/text/viewer_text_mk3.py
```diff
@@ -0,0 +1,226 @@
...
+from bpy.props import StringProperty, BoolProperty, IntProperty
+
+from sverchok.node_tree import SverchCustomTreeNode
+from sverchok.data_structure import levelsOflist, changable_sockets, multi_socket, updateNode
+
+
+
+class SverchokViewerMK1(bpy.types.Operator):
+    """Sverchok viewerMK1"""
+    bl_idname = "node.sverchok_viewer_buttonmk1"
```
portnov: Isn't this the same bl_idname as one that is used in older text viewer node? That could cause funny problems.

nodes/text/viewer_text_mk3.py
```diff
@@ -0,0 +1,226 @@
...
+            a.color[:] = color
+
+    def do_text(self,outs,node):
+        nTree = bpy.data.node_groups[self.treename]
+        #this part can be than removed from node text viewer:
+
+        if not 'Sverchok_viewer' in bpy.data.texts:
+            bpy.data.texts.new('Sverchok_viewer')
+
+        podpis = '\n'*2 \
```
portnov: Huh. At least some of Sverchok developers are not russian, so let's rename this to "footer", if you do not mind :)

nodes/text/viewer_text_mk3.py
nortikin: no

nodes/text/viewer_text_mk3.py
nortikin: i'll try

nortikin:
what to do with tests?
```bash
TypeError: bpy_prop_collection[key] = value: invalid, expected a StructRNA type or None, not a list
```

nortikin:
so i merge?

portnov:
@nortikin please see #2071 .

nortikin:
will we merge it?

portnov:
Okay now I'm for merging :)

Issue #2053, opened 2018-01-16
Branches that need user testing


zeffii:
- automated_dpi_nodeview_bgl
- master

zeffii:
anyone with privileges to edit posts can add branches to the first post

Issue #2052, opened 2018-01-16
A way to test branches for non git users.


zeffii:
from day 0 our issue with getting more eyeballs to test experimental features, is the extra hurdle for people to install `git` and learn how to use it. Routinely we've pushed code to master that had limited testing time. Pushing to _master_ just so people can test something isn't a sane development strategy. 

-----

I think sverchok needs "test mode", that instead of updating to the current master - would replace the download zip it points to and obtain the branch instead. The user doesn't need to know how to use git. This has the potential to open up testing to more people, far beyond our circle of `git` users.

https://github.com/nortikin/sverchok/blob/d5c1f4c8093a8fdf54a3760ffad3b5e4cd4be7b7/utils/sv_update_utils.py#L157-L162

an operator to download the list of branches to fill an enum of choices, once picked proceed to download/"cross-grade" (not upgrade), the enum will not need to list master, there will be a button to return to master.

https://api.github.com/repos/nortikin/sverchok/branches

![image](https://user-images.githubusercontent.com/619340/35006184-457046ce-faf6-11e7-8205-42c661f2b38b.png)

but that list is waaaay t0o long... so perhaps we need to take that into consideration.

----

( in prefs )
- [x] bool - "show_alternative_branch_box"
- [x] string - "selected_branch"
- [x] CollectionProperty - "branch_list" (names)

( operators )
- [x] node.sv_populate_branch_list
- [x] node.sv_pick_alternative_branch (text='switch to selected branch')

zeffii:
- or instead of auto-filling the list of branches we want people to test, we could have a  custom branch list, with 1...2..3 or 4 branches (a low number).

zeffii:
```python
import requests

r = requests.get('https://api.github.com/repos/nortikin/sverchok/branches')
json_obj = r.json()

for obj in json_obj:
    name = obj['name']
    if name.startswith("to_test"):
        print(name)

```
none will start with "to_test" yet,

zeffii:
we could even limit the choices to the content of a github issue specifically created for this.. The content of the issue would be:  https://github.com/nortikin/sverchok/issues/2053

zeffii:
`https://api.github.com/repos/nortikin/sverchok/issues/2053#issuecomment-1`

zeffii:
```python
import requests

comment_url = "https://api.github.com/repos/nortikin/sverchok/issues/2053#issuecomment-1"
r = requests.get(comment_url)
json_obj = r.json()

[...]

PR #2055, opened 2018-01-17
Allow users to try arbitrary branches (from a list we curate in issue 2053)

zeffii:
### warnings for gitusers:  don't use this feature, it's not for you.

Allow users to test arbitrary branches that we [list in this issue](https://github.com/nortikin/sverchok/issues/2053)

image

will show up like

image

then, when picked

image


portnov:
Maybe better maintain some .txt file in the git repo, instead of issue?

zeffii:
I quite like the idea of maintaining an issue .. or a gist (but issue is the easiest...) because it's not reliant on any branch .

nortikin:
maintain is time-consuming.

portnov:
Huh. We could instead have an agreement to name all such branches after certain pattern, for example "test_*". And code would just list all existing branches from repo by that pattern.

zeffii:
i mentioned what i thought were reasonable approaches with varying levels of sanity : https://github.com/nortikin/sverchok/issues/2052#issuecomment-358074035

- the downside to just a prefix is that it doesn't let users try a non `nortikin/sverchok/` branch, users might communicate among each-other to develop features, prior to getting commit access.
- allowing a user defined sverchok .zip path has the only drawback that it requires the path to be pasted/typed correctly.
- if we curate a list we can also theoretically allow users to propose non `nortikin/sverchok` archives.
    so rather than just 
    ```text
    - branch_name              # resolving to:   ....nortikin/sverchok/branch_name.zip
    
    # we could allow
    - user_name/branch_name    # resolving to:   ....user_name/sverchok/branch_name.zip
    ```

we could also just allow all of these options :)  with a tabbed interface

PR #2066, opened 2018-01-20
Init cleanup deeper

zeffii:

the `__init__.py` has been a slightly painful read for ages. I want to abstract away a lot that.

## Solution description

various lists can be relocated into the `__init__` that corresponds to their folder. Much of the logic plumbing for the initial registration and reload_event can be pushed into a dedicated file (inside `core.__init__.py` ) . This doesn't make the code any easier, but it compartmentalises blocks of code (in the hope that it's easier to understand the shorter functions )

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


zeffii:
@portnov   can you add an automatic test for "reload/f8"

zeffii:
and to check if `sverchok.nodes` exists after the reload

portnov:
Hmm. @zeffii can we programmatically trigger F8?

zeffii:
`bpy.ops.script.reload()`

zeffii:
i think...

zeffii:
works fine inside a script, but not when triggered from bpy console

portnov:
Hmm. @zeffii in update_system.py, `def process_tree()` we have

```python
   if data_structure.RELOAD_EVENT:
        reload_sverchok()
        return
```

So if I trigger reload in one test, and the following test does `tree.process()`, then it fails because no processing is done. I'm not sure how exactly this looks for user: first "tree update" after F8 does not work? Is it intended so?

portnov:
I commented out the "return" line and tests are ok...

portnov:
@zeffii see #2070.

zeffii:
it's a long standing "issue" for F8 needing the user to click at least 2 times onto something that triggers an update. As long as your patch results in the source of an edited modules/submodule (ui / node) to be re-loaded, then that's fine.

zeffii:
it's a long standing "issue" for F8 needing the user to click at least 2 times onto something that triggers an update.

portnov:
@zeffii what's the point in that "return" after "reload_sverchok()"? Are there situations when "reload_sverchok()" call leads to something in process_tree() not working?

zeffii:
judging by the history of that file  , i have never edited it -- i can't claim to have a deep understanding of the hidden subtleties . i'll dig in now. will continue here https://github.com/nortikin/sverchok/pull/2070

PR #2070, opened 2018-01-21
Simple test case for reload.

portnov:

See #2066.

## Solution description

Add simple testcase that triggers reload and then tries to import `sverchok.core`.
Do not abort first processing after reload.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/core/update_system.py
+++ b/core/update_system.py
@@ -440,7 +440,7 @@ def process_tree(ng=None):
 
     if data_structure.RELOAD_EVENT:
         reload_sverchok()
-        return
+        #return
     if not ng:
         for ng in sverchok_trees():
             process_tree(ng)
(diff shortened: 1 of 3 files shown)


zeffii:
best continue here

> @zeffii what's the point in that "return" after "reload_sverchok()"? Are there situations when "reload_sverchok()" call leads to something in process_tree() not working?

looking.

zeffii:
another test is to enable / disable the sv addon.

portnov:
I believe we can merge this? no point in hanging PRs...

zeffii:
do it!

PR #2071, opened 2018-01-21
Test case for TextViewer Mk3.

portnov:

It is better to have tests for new nodes. See also #2048.

## Solution description

Added a test. The test could be simpler if we could just call one method and check it's returned value. For now I'm invoking an operator and checking what it has written to text buffer.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.


tests/simple_viewer_text_tests.py
```diff
@@ -1,23 +1,31 @@
...
         self.tree.links.new(cyl.outputs['Vertices'], viewer_text.inputs[0])
 
-        # Trigger processing of NGon node
+        # Trigger processing of Cylinder node
         cyl.process()
-        # Read what ViewerText node sees
-        data = viewer_text.inputs[0].sv_get()
\ No newline at end of file
+        # Invoke "VIEW" operator
+        bpy.ops.node.sverchok_viewer_buttonmk1('EXEC_DEFAULT', treename=self.tree.name, nodename=viewer_text.name)
```
nortikin: thanks. it was not obvious for me

tests/simple_viewer_text_tests.py
```diff
@@ -1,23 +1,31 @@
...
+
+        # Read what the operator has written to the text buffer.
+        text = bpy.data.texts['Sverchok_viewer'].as_string()
+        
+        # Test that text in buffer equals exactly to the text saved
+        # in the file from the first test run.
+        with open(self.get_reference_file_path("text_viewer_out.txt"), "r") as f:
```
nortikin: wow!

PR #2072, opened 2018-01-21
Init cleaner final

zeffii:
finalizes the cleanup of `__init__`


zeffii:
I will not be touching `__init__` anymore. unless somehow I broke it without noticing.

PR #2073, opened 2018-01-21
pushes as much as possible to sensible functions

zeffii:

syntactic sugar for the iojson storage process.

## Solution description

perliminary stashing away logic into cohesive functions, and reusing self.storage_get_data more sanely.

## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/nodes/generators_extended/profile_mk2.py
+++ b/nodes/generators_extended/profile_mk2.py
@@ -956,6 +956,7 @@ class SvProfileNodeMK2(bpy.types.Node, SverchCustomTreeNode):
             local_storage['knotsnames'].append(outname.SvName)
         
         node_dict['profile_sublist_storage'] = json.dumps(local_storage, sort_keys=True)
+        node_dict['path_file'] = bpy.data.texts[self.filename].as_string()
 
 
 
--- a/nodes/scene/objects_mk3.py
+++ b/nodes/scene/objects_mk3.py
@@ -270,6 +270,10 @@ class SvObjectsNodeMK3(bpy.types.Node, SverchCustomTreeNode):
         outputs['Object'].sv_set([data_objects.get(o.name) for o in self.object_names])
 
 
+    def storage_get_data(self, node_dict):
+        node_dict['object_names'] = [o.name for o in self.object_names]
+
+
 classes = [SvOB3Callback, SvObjectsNodeMK3]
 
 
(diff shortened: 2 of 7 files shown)


portnov:
"oops" :) I think we can congratulate ourselves: Travis did what it is created for, it seems first time for Sverchok :)

zeffii:
heeh :) cool. yes. very helpful!

zeffii:
@portnov ,  i don't think i ever used the old groupnode,  but still a question arises about the code i just added to support this relic. I'm tempted to drop it.

Nonetheless, an interesting recursion question: my pylinter doesn't throw a hissy-fit, but i'm not sure if the function `handle_old_groupnode` can call `create_dict_of_tree` .  The function is defined before `create_dict_of_tree` is, but only called from within it.

zeffii:
yeah..that's not going to work.. :)

portnov:
Afaik python's scoping rules allow this:

```python

def test1(x):
    return test2(x)+1

def test2(x):
    return x*x

print(test1(2)) # prints 5
```

Python is dynamic language, so it is required that name `test2` is defined at the moment of execution of `test1`, not at the moment of compilation. But I'd like to avoid such usage, it's very unobvious and can lead to confusion while maintaining the code.

zeffii:
hmmm..

zeffii:
hah, upon reading/digging around.. this function is needed for supporting monads .. i think.

zeffii:
wtf.. anyway
```python
             handle_old_groupnode(node, k, v, groups_dict, create_dict_of_tree)            
```
has to work, because a few lines later there's 
```python
        if IsMonadInstanceNode and node.monad:
            pack_monad(node, node_items, groups_dict, create_dict_of_tree)
```
passing the function around works .

nortikin:
saying sincerely, me afraid of:
```python
import json
import sys
import os
import re
import zipfile
import traceback
from time import gmtime, strftime
import urllib
from urllib.request import urlopen

from os.path import basename, dirname
from itertools import chain

import bpy
from bpy.props import StringProperty, BoolProperty, PointerProperty
from bpy.utils import register_class, unregister_class

from sverchok import old_nodes
from sverchok.utils import sv_gist_tools
from sverchok.utils.sv_IO_monad_helpers import pack_monad, unpack_monad
from sverchok.utils.logging import debug, info, warning, error, exception
```

zeffii:
i intend to further modularize this file. agreed @nortikin   it's a frankestein's monster now.

PR #2079, opened 2018-01-22
add runtest bat

zeffii:
adds a windows version of the test, just to see if there's a difference.

##### run_tests.bat

```bat
echo off

:: blender executable path
set arg1=%1 

shift
%arg1% -b --addons sverchok --python utils/testing.py --python-exit-code 1 
```

usage: 
```
PS C:\Users\zeffi\Desktop\GITHUB\sverchok> .\run_tests.bat C:\fullpathto\b279\blender.exe
```
and it will start

### DIFF
--- a/run_tests.bat
+++ b/run_tests.bat
@@ -1,0 +1,7 @@
+echo off
+
+:: blender executable path
+set arg1=%1 
+
+shift
+%arg1% -b --addons sverchok --python utils/testing.py --python-exit-code 1 


zeffii:
not sure if we should have two `run_tests`  @portnov  .. maybe i'll rename to `run_tests_win.bat`

zeffii:
it does most of the tests and then fails

zeffii:
based on numerical anomalies it doesn't seem multi-platform tests are possible with simple equality assertions. That's OK for now.

portnov:
@zeffii ,
1. In *nix, it is not possible to run "run_tests.sh" as just "run_tests". In Windows, it is possible to run "run_tests.bat" as "run_tests". But, in Windows you can't run "run_tests.sh" as "run_tests" :) So there is no place to confuse "run_tests.sh" and "run_tests.bat". I think the platform for which the script is written is obvious from script's file name extension (bat/sh).

2. We have special assert methods in `SverchokTestCase` base class to compare numbers with certain precision (`assert_sverchok_data_equal()` for example; we can add some similar methods for special cases). We should use them with some sane precision (for example 8 decimal digits) in all cases when there are some random (but very small) floating-point errors.

zeffii:
so a test could fail at high precision, and be repeated at lower precision and pass -- that would be a single test case ?

portnov:
@zeffii I mean we should not write

```python
from math import sin

class SinTestCase(SverchokTestCase):
   def test_sin(self):
      self.assertEquals(sin(1), 0.8414709848078965)
```
because this test can fail on some different platform (32bit instead of 64 bit, or AMD instead of Intel, for example), because there sin(1) will be 0.8414709848078964, not 0.8414709848078965: floating point calculations are not so precise, last digits can vary on different processors. So instead we should write something like

```python

class SinTestCase(SverchokTestCase):
   def test_sin(self):
      self.assert_sverchok_data_equal(sin(1), 0.8414709848078965, precision=8)
```

This will pass on any platform where floats are precise enough to provide 8 exact digits after decimal point.

zeffii:
yeah, i get it. well, if you want a win `run_tests`, the initial file suggestion can still stand. but we don't reallllly need it asap.

PR #2087, opened 2018-01-24
propose at least one change.. more to come

zeffii:
adresses Label text on socket , this must be taken from prop_name , not socket "name" if prop_name is set.
There are other scenarios i'm sure.


zeffii:
but :)
```python

Issue #2088, opened 2018-01-24
Adding Nodes from Presets panel differs from the other panels


Ribel78:
## Problem statement

Created node groups and nodes in Presets panel in Tool shelf after clicking are added to some predefined location as opposed to the nodes on other panels which are in grab mode and move with the mouse.

## Steps to reproduce

1. Click on the node or node group in the Presets panel - the nodes are added to the node workspace but don't move with the mouse.
2. Click on node from any other nodes panel - the nodes are added to the work space but are not released yet from the mouse.

## Expected result

Nodes created through Presets panel should be in grab mode and move under the mouse pointer.

## Actual result

Nodes are placed always in predefined location which can be outside the current node view.
Still they are selected so if the user has enabled Continuous Grab mode by pressing G can move the node even if it is not in the view.
It could be a problem if a group node is added on top of existing node tree and accidentally deselected.

## Sverchok version

SV 0.5.9.6 (b546794)

portnov:
Yes, there is such kind-of-problem. I do not know how to solve it currently.
If anyone knows blender python API to initiate moving mode, please fill free to help :)

zeffii:
i think this might be two problems that compound eachother.

- when a node / set of nodes is stored as json, their locations are also stored. It's not clear whether it would be desirable to transform the nodes to 0,0, or make the select+active node the 0,0 of the nodes. 
     - let's not even discuss Framed nodes..
- the import from json code (import silent?) does not expect the user to need to grab/transform the nodes as you might expect from a singular step of adding a single node to the tree

it's broken in this sense, but it's not a bug... it's just an incomplete feature implementation.

zeffii:
somewhere in 
https://github.com/nortikin/sverchok/blob/de93d4ca72bf6520bd1e061f358983c73f5a70bb/ui/presets.py#L161-L163

you can pass several parameterse to the translate_attach operator ( `NODE_OT_insert_offset` .. might be handly.. )

zeffii:
hmmm
```
wm_macro_exec: 'NODE_OT_attach' cant exec macro
wm_macro_exec: 'NODE_OT_insert_offset' cant exec macro

```

zeffii:
relevant.. if we really want to do this.. :)  https://blender.stackexchange.com/questions/44971/how-to-use-bpy-ops-macro-defining-macro-steps

zeffii:
[...]

PR #2094, opened 2018-01-27
Presets: start moving when adding a preset.

portnov:

#2088

## Solution description

Invoke transformation operator when importing preset from file.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/ui/presets.py
+++ b/ui/presets.py
@@ -141,6 +141,9 @@ class SvPreset(object):
                 bl_label = "Add {} preset".format(self.name)
                 bl_options = {'REGISTER', 'UNDO'}
 
+                cursor_x = bpy.props.IntProperty()
+                cursor_y = bpy.props.IntProperty()
+
                 @classmethod
                 def poll(cls, context):
                     try:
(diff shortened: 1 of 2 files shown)


zeffii:
to do this correct i think we need at least two things. 
   - to be able to add an offset to the nodes, maybe centering the selected node cloud around the active node and setting that at 0,0, adjusting the other nodes relative to it .
   - fix io/json broken Frame implementation, nodes are not yet placed in identical locations.
     some of this problem is addressed here in a related issue, it's a head-ache
  https://github.com/nortikin/sverchok/blob/523fbf317cd1c351d5df69b446b5c38f04b891b8/ui/nodeview_rclick_menu.py#L85-L86

zeffii:
the latest `create_dict_of_tree` will now adjust the `node.location` to reflect the absolute location in the tree. The remaining issue is now to offer a method to offset the nodes / selected nodes relative to the average of selected node.locations or active node.location

portnov:
@zeffii I managed to write some code, that changes node locations in json before importing it, so that average node location would be placed at specified point. The question is, where should it be? Easiest is to place it to (0,0) always.

zeffii:
yes, deciding the 'epicentre' isn't necessarily. What I don't want is to have too many options for the user. So for now we keep it simple.. unless many people exclaim some kind of preference(s).

maybe :  
   - if no nodes selected, epicenter maybe average of all nodes
   - if nodes selected, epicenter could be the active+selected node.

zeffii:
i don't know. do whatever seems right.

portnov:
@zeffii what I do currently is (before actually importing json):
* Calculate average of all nodes (being imported) location
* Change locations of all nodes being imported so that average location becomes (0,0)

So that imported nodes appear at the (0,0) of node tree plane.

zeffii:
not sure that's right in the case the user is working in an existing tree, might have to offset that to the X,Y location being viewed in the active `area - view` of the nodetree where the operator was called.

portnov:
@zeffii okay, it seems I got your idea...

PR #2096, opened 2018-01-28
Exec node fixes

zeffii:
- [x] copying from a text file with fewer lines that are current in the CollectionProperty `< lines >` will result in incorrect copy
- [x] can not copy code from clipboard ( new feature )
- [x] add custom rclick for clipboard.
- [ ] can not use list comprehensions without getting unknown symbol for V1, V2
- [ ] can not extra wrap output.

### DIFF
--- a/nodes/script/multi_exec.py
+++ b/nodes/script/multi_exec.py
@@ -152,8 +152,7 @@ class SvExecNodeMod(bpy.types.Node, SverchCustomTreeNode):
         slines = bpy.data.texts[self.text].lines
         self.dynamic_strings.clear()
         for line in slines:
-            ds_line = self.dynamic_strings.add()
-            ds_line.line = line.body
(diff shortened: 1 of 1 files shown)


zeffii:
this is always depressing to read this kind of thread about exec() https://bugs.python.org/issue21161 .

so.. it might be neccessary to do something more invasive than a simple exec(), in snlite i compile the code using injected variables.. also sv_macros does some ast.compile stuff.  #todo...

zeffii:
so... we might be doing this in future
```python
exec_code = """
def exec_main(node, V1, V2, V3, out):
    append = out.append
    extend = out.extend
    {indented(code)}
"""
```
if there are `imports` needed, then they are passed in draw_buttons_ext..

wait.. didn't i do a branch for this? :)

zeffii:
will commit rest soon.

PR #2097, opened 2018-01-29
remove functions to more meaninful locations

zeffii:
`sv_io_panel_tools.py` is too damn big. Many functions inside rarely need any serious modification anymore, and it might be time to fully modularize (but I resist turning this into a class-system )

I'm electing to move many functions into their own (or existing, sensible location) and reduce the import if possible, i'd like `sv_io_panel_tools.py` to be no longer than 500 lines orso.


zeffii:
it's such a spaghetti file : /

zeffii:
found some possible bugs along the way.

PR #2098, opened 2018-01-29
Io json cleanup alt

zeffii:
stealthy non user facing file fixes.

`sv_io_panel_tools.py` is too damn big. Many functions inside rarely need any serious modification anymore, and it might be time to fully modularize (but I resist turning this into a class-system )

I'm electing to move many functions into their own (or existing, sensible location) and reduce the import if possible, i'd like `sv_io_panel_tools.py` to be no longer than 500 lines orso.

### DIFF
--- a/utils/__init__.py
+++ b/utils/__init__.py
@@ -105,7 +105,7 @@ utils_modules = [
     # UI text editor ui
     "text_editor_submenu", "text_editor_plugins",
     # UI operators and tools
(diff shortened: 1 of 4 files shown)


zeffii:
note: looking at the operators suggests that 3 of them could be reusing a utility function, at least for the first 5 lines of getting / setting a node tree.

zeffii:
there's nothing more I want to do to PR at the moment,  tests pass... and manual tests seem to be OK.

PR #2108, opened 2018-02-04
Add a testcase for importing json_examples.

portnov:

We have a bunch of templates in repo, but do we know if it possible to use them?

As you will see from travis results, some of templates are broken (either templates themeselves or our importing code).

## Solution description

Add a test case that tries to import each json file under json_examples.

## Preflight checklist

- [x] Code changes complete.
- [ ] Manual testing done. 
- [x] Unit-tests implemented.
- [ ] Ready for merge.


portnov:
Also I'm not sure about messages like

    Cannot find VectorMoveNode among old nodes

zeffii:
interesting, it's a part of the code i've not waded very deep into. I'll look again tomorrow!

portnov:
Later I want to enhance this testcase by adding a check that template does not use deprecated nodes.

zeffii:
I see,  loading templates / presets / jsons   should not happen after a reload event .. or disable/enable.

Those are scenarios where currently the old_nodes is not repopulated. Not very satisfying I know..

zeffii:
especially jsons that contain deprecated nodes...    hmmmm.  easy enough to fix I think.

zeffii:
i'd start with testing if `old_bl_idnames` is populated at the time of calling that function.. if not, then execute the collection code again

https://github.com/nortikin/sverchok/blob/47d6350bd9436098334b4af033376fe598905bcd/old_nodes/__init__.py#L36-L48

zeffii:
```python
old_bl_idnames = get_old_node_bl_idnames(path=os.path.dirname(__file__))


def is_old(node_info):
    '''
    Check if node or node.bl_idname is among
    the old nodes
    '''

    if not old_bl_idnames:
        old_bl_idnames = get_old_node_bl_idnames(path=os.path.dirname(__file__))

    if isinstance(node_info, str):
        # assumes bl_idname
        return node_info in old_bl_idnames
    elif isinstance(node_info, bpy.types.Node):
        return node_info.bl_idname in old_bl_idnames
    else:
        return False
```

zeffii:
i'll checkout this branch now

zeffii:
yes I think several example jsons must be updated, or removed. No examples should reference deprecated nodes, or feature sockets indexed by name, we moved over to int indices a few revisions ago (for sanity)

zeffii:
broken anyway:   (well.. it throws errors because of the script in sn_lite being a bit too error prone)

- 10.transformations_mk2.png.json

needs to be 
```python
"""
in   datain    s   .=[]   n=0
in   level     s   .=1    n=0
out  dataout   s
"""

import numpy as np
from numpy import array as ar

if datain:    # <--------------- this

    Data = ar(datain)
    print(Data.shape)
    shape = np.roll(ar([i for i in range(Data.ndim-1)]), level)
    shape = np.append(shape,ar([Data.ndim-1])).tolist()
    Data = np.transpose(Data, shape)

    dataout = Data.tolist()
```
else 
```python traceback
(0,)
Unexpected error: <class 'TypeError'>
on line:  47
  File "2.79\python\lib\site-packages\numpy\core\fromnumeric.py", line 57, in _wrapfunc
    return getattr(obj, method)(*args, **kwds)
TypeError: 'float' object cannot be interpreted as an integer

During handling of the above exception, another exception occurred:
```

zeffii:
yeah, so @portnov   I think it's better to try these example jsons first outside of the context of the tests, and  note any errors upon a fresh load. Then address those.  

not looking forward to doing that. the new node location code (for framed nodes) needs to be tested harder..

portnov:
My plan relating this PR is:
* Run tests and see which jsons are failing in tests. There are 4 such jsons for now.
* Check loading of these templates manually from blender. I tried some, not all; the ones I tried resulted with errors. Will check all 4 this week.
* Remove templates that cannot be loaded from repo.
* If it occurs that test fails to load some templates that can be loaded manually - fix the test.
* Add a check that template does not contain deprecated nodes to the test. I assume about 30-50% of templates will fail this check.
* Gradually update these templates one-by-one. This will take time; no rush, but does not worth linger too long, because otherwise nodes will be becoming deprecated faster than I'm updating templates :)
* When all tests are green - merge this PR.
* After that, if someone moves some node to oldnodes, he will immediately see if some tests are failing, and update templates.

zeffii:
@portnov  cool. :)   i'll still have a look at what causes the old nodes not to be loaded in `register_old`,
```python
Cannot find SvColorsInNode among old nodes
```
the bl_id _is_ in old_bl_idnames when that is printed.

zeffii:
- `wall_from_profile_node.json` and `wingflap4.json` was made before I added `node_ref['path_file']`
-  `KELLY_EGAN_spiral_disk.json`  still uses named sockets instead of indices
- `screw polygons.json` still uses named sockets instead of indices

the jsons that store socket names instead of indices, and are also dynamic will fail repeatedly. Technically,  i thought socket names was still supported, but apparently not.

zeffii:
ok @portnov  hopefully this speeds things up
```text
1.
2018-02-06 10:46:37,477 [INFO] tree_import_tests: Importing: 10.transformations_mk2.png.json
2018-02-06 10:46:37,575 [INFO] sverchok.utils.sv_IO_panel_tools: ############0.065  <-----------------     snlite code.

2.
2018-02-06 10:46:40,050 [INFO] tree_import_tests: Importing: 11.pepper.blend.json
2018-02-06 10:46:40,155 [INFO] sverchok.utils.sv_IO_panel_tools: ############0.065  <---- named sockets   "  `T 0'   "

3.
2018-02-06 10:46:45,577 [INFO] tree_import_tests: Importing: 13.transformations.json
2018-02-06 10:46:45,684 [INFO] sverchok.utils.sv_IO_panel_tools: ############0.065  <-----------------     snlite code.

4.
2018-02-06 10:47:09,999 [INFO] tree_import_tests: Importing: KELLY_EGAN_spiral_disk.json
2018-02-06 10:47:10,125 [INFO] sverchok.utils.sv_IO_panel_tools: ############0.062   <----- a little old Monad, and named sockets
2018-02-06 10:47:11,663 [ERROR] sverchok.utils.sv_IO_panel_tools: [['Script Generator', 'Verts', 'Monad', 'Vectors'], ['Monad', 'Vectors', 'Viewer Draw2', 'vertices'], ['Monad', 'Polygons', 'Viewer Draw2', 'edg_pol']]

5.
2018-02-06 10:47:38,479 [INFO] tree_import_tests: Importing: screw polygons.json
2018-02-06 10:47:38,607 [INFO] sverchok.utils.sv_IO_panel_tools: ############0.064  <-- named sockets
KeyError: 'bpy_prop_collection[key]: key "data 1" not found'
2018-02-06 10:47:41,503 [ERROR] sverchok.utils.sv_IO_panel_tools: [['Centers polygons 2', 'Origins', 'List Zip', 'data 1']]

6.
### I presume that due to the number of errors, some nodes/sockets are not created and thus their key doesn't exist
2018-02-06 10:47:47,274 [INFO] tree_import_tests: Importing: wall_from_profile_node.json
2018-02-06 10:47:47,373 [INFO] sverchok.utils.sv_IO_panel_tools: ############0.065
2018-02-06 10:47:47,625 [ERROR] sverchok.utils.sv_IO_panel_tools: 'path_file'   <--- biggest issue.
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\utils\sv_IO_panel_tools.py", line 421, in perform_profile_node_inject
    new_text.from_string(node_ref['path_file'])
KeyError: 'path_file'

7.
2018-02-06 10:47:49,337 [INFO] tree_import_tests: Importing: wingflap4.json
2018-02-06 10:47:49,398 [INFO] sverchok.utils.sv_IO_panel_tools: ############0.065
2018-02-06 10:47:49,916 [ERROR] sverchok.utils.sv_IO_panel_tools: 'path_file'   <---- biggest issue  (also a key error like above)
```

zeffii:
i'm happy to fix these, and do a PR into your Branch.  just give me the word :)

portnov:
After adding a check for deprecated nodes, there are 22 fails of 40 examples total. Well...

portnov:
In an hour I was able to half the number of failures. With remaining 11 jsons, I'd be glad if authors of these examples updated them, because probably only author knows exactly what the result should look like. It appears that all remaining examples were initially made by @nortikin :)

portnov:
Yoohoo. Tests are passing. I just need #2131 to be merged for this to work (because otherwise layouts which contains "text in" node will not be loaded correctly).

portnov:
Well, this currently includes #2131 . @zeffii don't you mind if I merge this one?

zeffii:
merge ahead!  but maybe include that last gist which stores the data correctly for `6.adaptiveSpread.json`
https://gist.github.com/9f41b7d8ea308185f142be54e1fd1e8a

portnov:
@zeffii it only worths if update all "text in" nodes in all examples, which is quite a lot. I think it actually does not matter what order of sockets will be used in our examples, it is only required that order is always the same. Sorting patch does this.

zeffii:
those gists that store links using  socket.name will work no matter how the node populates the sockets upon creation
those gists that store links using socket.index will have no added consistency, in the long term (sorted population or not)

it's OK. lets focus on new jsons/gists and hope the number of users that encounter incorrectly populated socket order  is insignificant (and not worth our continued discussion)

PR #2115, opened 2018-02-07
Examples test windows

zeffii:
- This removes the jsons that are complicated to fix. (ie not worth anybody's time... remake is faster)
- This adds modification to `register_old` code, avoids printing "cannot find among old nodes" when not appropriate.
- Fixes those json that could be salvaged.
- adds a windows version of the `run_tests(_win.bat)`


zeffii:
@portnov  i'd be OK with this. we can recreate the missing examples , or start using the .gz for more complicated layouts.  json was only ever intended for quick IO of simple stuff.

core/__init__.py
```diff
@@ -64,6 +64,13 @@ def import_modules(modules, base, im_list):
 def handle_reload_event(nodes, imported_modules, old_nodes):
     node_list = make_node_list(nodes)
     reload_all(imported_modules, node_list, old_nodes)
+
+    try:
+        from sverchok.old_nodes import old_bl_idnames
+        print('debug info --- old_bl_idnames:', len(old_bl_idnames))
```
portnov: Maybe replace with debug() ?

old_nodes/__init__.py
```diff
@@ -108,17 +108,21 @@ def register_old(bl_id):
...
         res = inspect.getmembers(mod)
+        # print('mod/res:', mod, res)
         for name, cls in res:
             if inspect.isclass(cls):
                 if issubclass(cls, bpy.types.Node) and cls.bl_idname == bl_id:
                     if bl_id not in imported_mods:
                         try:
                             mod.register()
                         except:
+                            print('failed mod register')
```
portnov: this ideally should be
```python
except Exception as err:
    exception(err)
```
This will print both exception message and stack.

old_nodes/__init__.py
```diff
@@ -108,17 +108,21 @@ def register_old(bl_id):
...
                             traceback.print_exc()
                         imported_mods[bl_id] = mod
                         return mod
-                    
+                    else:
+                        return imported_mods[bl_id]
+
     print("Cannot find {} among old nodes".format(bl_id))
```
portnov: I propose to replace this with error() call.

zeffii:
We could have a  ".blend import Node".

the node would store a location on disc or http(s) and have an button that triggers an operator to load the .blend file as the main file. (similar to what download zipped url downloading does)

This would allow us to have a small .json, which is easy to read in a text editor to determine the source location of the .blend. It would also make it easier to share complicated blends.

old_nodes/__init__.py
zeffii: yes, do it!

core/__init__.py
zeffii: yes, makes sense.

old_nodes/__init__.py
zeffii: i'll commit// i still had it open

portnov:
Eeh. Not sure if I understood correctly. What for do we need such a "small json", why not just share a link to blend file?

zeffii:
yeah maybe not. It would be simple enough to add multiple node trees to a .blend too.. :)

PR #2116, opened 2018-02-07
Create troubleshooting_nodes.rst

zeffii:

we can help people better debug their nodes and node trees. A significant number of people like to solve their own problems, and some of those people need a little help to understand the process and their options of finding solutions. Most of the time (and this is true for most new programmers) there comes a point where a user feels they understand something, but can't explain why a node has an error or won't work - the kneejerk reaction is to think there's a bug in the tool. Often, the bug is not in the tool. Most of the time we don't change any code to resolve an issue.

## Solution description

description of methods that we added and use to figure out what nodes are failing and why.

### DIFF
--- a/docs/troubleshooting_nodes.rst
+++ b/docs/troubleshooting_nodes.rst
@@ -36,12 +36,15 @@ A variety of reasons can cause exceptions, sometimes you have to solve a sequenc
  
 For most "unwanted" situations we can find what their cause is. We have quite a few ways to see the state of the output or input of a node.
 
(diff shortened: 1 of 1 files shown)


zeffii:
not being a fluent programmer should not be a serious impediment to understanding the process of location a problem / bug.

nortikin:
like this tongue

zeffii:
http://alumni.cs.ucr.edu/~ratana/PaneRatanamahatanaMyers00.pdf

```
           Studying the Language and Structure in
         Non-Programmers’ Solutions to Programming
                        Problems
```

portnov:
Do not see any reason for this to linger out of master branch :)

Issue #2126, opened 2018-02-10
Import of "Text In ++" is broken


portnov:
## Problem statement

https://gist.github.com/portnov/582de3f84a99210493882393a9137ef9

```

======================================================================
ERROR: test_import_examples (tree_import_tests.ExamplesImportTest) (file='5.Sverchok_Spread.json')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/tests/tree_import_tests.py", line 66, in test_import_examples
    import_tree(new_tree, path)
  File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/utils/sv_IO_panel_tools.py", line 729, in import_tree
    generate_layout(fullpath, nodes_json)
  File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/utils/sv_IO_panel_tools.py", line 711, in generate_layout
    place_frames(ng, nodes_json, name_remap)
  File "/home/portnov/.config/blender/2.79/scripts/addons/sverchok-master/utils/sv_IO_panel_tools.py", line 638, in place_frames
    ng.nodes[finalize(node_name)].parent = ng.nodes[finalize(parent)]
KeyError: 'bpy_prop_collection[key]: key "Text in+.001" not found'
```

## Steps to reproduce

1. Load example "5.Sverchok_Spread.json"
2. replace Text In nodes with Text In ++
3. Export result to json again
4. Try to import it again (from test or manually from blender, doesn't matter)

## Expected result

Loaded node tree

## Actual result

Exception

## Sverchok version

git master

zeffii:
[5spread_textin_test_2018_02_10_08_53.gz](https://github.com/nortikin/sverchok/files/1713210/5spread_textin_test_2018_02_10_08_53.gz)

zeffii:
https://gist.github.com/b0b21b5aa1af9c8b8a06bc24c84d7c58

zeffii:
hmmm..
```
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\utils\sv_IO_panel_tools.py", line 432, in perform_svtextin_node_object
    current_text = params['current_text']
KeyError: 'current_text'
```

zeffii:
```python
    params = node_ref.get('params')
    current_text = params['current_text']

```
looking at this json
```
    "Text in+": {
[...]

PR #2131, opened 2018-02-10
Text In mk2: sort output sockets, if order is not defined.

portnov:

#2126 

## Solution description

Order outputs by using sorted(), if "sockets_order" is not defined.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/nodes/text/text_in_mk2.py
+++ b/nodes/text/text_in_mk2.py
@@ -448,7 +448,7 @@ class SvTextInNodeMK2(bpy.types.Node, SverchCustomTreeNode, CommonTextMixinIO):
 
             socket_iterator = iterate_socket_order()
         else:
-            socket_iterator = json_data.items()
+            socket_iterator = sorted(json_data.items())
 
         for named_socket, data in socket_iterator:
             if len(data) == 2 and data[0] in {'v', 's', 'm'}:


portnov:
@zeffii please review.

zeffii:
whether you sort the sockets or not, it won't really fix all nodes. it might fix some older jsons ,,  like like a broken clock will give the correct time 2 times a day.

zeffii:
whether you sort the sockets or not, it won't really fix all older jsons (sans socket_order). it might fix some older jsons ,, just like a broken clock will give the correct time 2 times a day.

zeffii:
i'm all for including a fix for older nodes, but realistically all old jsons containing Text In node were fundamentally broken. If they worked, it seems they were accidentally ordered as expected.

zeffii:
!!!! oh I know why this is only now coming to the forefront!

 initially socket where linked using their socket `.  name`, i changed this socket `. index ` some time ago. This would have broken a few older exports :)

portnov:
Eeh. So, is it okay to merge this?

zeffii:
### Problem
a brief explanation.

I know why this is only now coming to the forefront!

initially io/json stored socket links using their socket `.  name`, i changed this (for good reasons) to socket `. index ` some time ago. This means that whatever the socket order was, didn't matter because links where remade using socket.name . even when the node recreated it's own sockets (however random that was) the relinking was correct.

When i switched to socket.index this `text_in` reloading/relinking was broken. 

### Solution

the proposed solution will sort keys alphabetically, for UI regenerating pre-linking


### My issue with the proposed solution

It will fix some older jsons, but as a node outputs more sockets the chances increase that even `sorted(outputs.items() )` will not place the sockets in the correct order.

### tldr

not sure there's any nice solution for old jsons prior to `socket_order`

portnov:
Huh. Let's look at the problem like so: does this change make things better or worse? As I see it:
* Before this change, order was random, so for example for node with 4 outputs there was about 1/24 chance that it would load correctly. For newly saved jsons, you could not know if it would load or not.
* After this change, order is fixed. For earlier existing jsons, there still is about 1/24 chance that it would load correctly (in case when initial order of sockets was exactly alphabetical). But for newly saved jsons, you know that it would load correctly.

zeffii:
you can of course commit this.  it doesn't affect me.

but, because it will still fail for some gists we try. 

So we can also offer a few steps for users to resolve their gists in the case sorting also didn't help.
```
import bpy


def get_socket_order(n):
    socket_order = [socket.name for socket in n.outputs]
    print('\"socket_order\":', str(socket_order))
        

n = bpy.data.node_groups['NodeTree'].nodes.active
get_socket_order(n)
```
this code could be executed 1 time for each text_in node, just to get something like this:
```python
"socket_order": ['Objects input:Edges', 'Objects input:Matrixes', 'Objects input:Polygons', 'Objects input:Vertices']
```
and you would use that to manually edit define the  `socket_order` , _after_ noticing that the no automatic sorting was producing the expected order.

zeffii:
this way
```diff
"socket_order": ['Objects input:Edges', 'Objects input:Matrixes', 'Objects input:Polygons', 'Objects input:Vertices']
"socket_order": ["Objects input:Edges", "Objects input:Matrixes", "Objects input:Polygons", "Objects input:Vertices"]
```

portnov:
Included in #2108 .

PR #2139, opened 2018-02-15
add archive path and name as sv_preferences member

zeffii:
i'd like to get these two properties in without a UI even... ;)


zeffii:
this also fixes the datetime case where an input is a float, but could be expressed as a `str(int(value))`, for example for `%Y` (1800)

PR #2140, opened 2018-02-16
Subdivide lite node

Kosvor2:
Have optional ability to chose edges by index instead of mask (faster when user wants to subdivide only some edge placed in the end of long list). Has one verts/polys socket instead of separate verts and faces sockets alowing to input polygons without the need to input also edges. Some other optimisations.

Deformation parameters no longer have sockets. Other nodes can be used instead if needed.

- [x] Code changes complete.
- [x] Manual testing done.
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


Kosvor2:
Current progress:
subdiv_lite57465

PR #2142, opened 2018-02-17
Line mk2 nestiness fix

Kosvor2:
It's been placing Edges output container list inside of additional tuple layer all this time. And many Sverchok nodes does not throws errors. But some of them do, so better lets fix it.


Kosvor2:
As a bonus, it is possible now to use several numbers for Num Verts socket:
line_inputs48567

Kosvor2:
On the other hand, users could use a "list split" node to create multiple objects. I'll check how the modified code is handled in this layout and close this pull request if everything is broken.

PR #2143, opened 2018-02-17
Plane mk2 fixes

Kosvor2:
Same as with Line node in here: https://github.com/nortikin/sverchok/pull/2142

But since Plane node have two Vert Number inputs instead of one, both must be feeded by same amount of values.

plane_inputs478346

Can be changed using `match_long_repeat()` but dont wanna slowing things too much. In addition, the old method of dividing the list into parts using "list split" node is still working (i'm not even sure that many knew that it was possible to do so) so that for users nothing will change.

### DIFF
--- a/docs/nodes/generator/plane_mk2.rst
+++ b/docs/nodes/generator/plane_mk2.rst
@@ -65,3 +65,5 @@ If **Separate** is True, the output is totally different. The grid disappear (no
 
 Example of usage
 ----------------
+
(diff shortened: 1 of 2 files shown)


DolphinDream:
A styling question : Is there a reason why all the empty lines separating the code were removed ? I know keeping vertical spread low is nice .. but not as nice if it affects readability. Typically I separate chunks of code with empty lines if it make sense to isolate them based on the local logic.

PR #2158, opened 2018-03-01
Added back Edges intersect Tolerance

vicdoval:
After using the Intersect Edges node (2D)  I observed tolerance was needed to avoid the creation of doubled vertices. 
So I added back intersection tolerance to intersect Edges 2d as it was suggested by @zeffi in the pull request https://github.com/nortikin/sverchok/pull/2001

I updated that change in  the Contour2D node as it was depending on it.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented. Could you point me where to check what this is?
- [x] Ready for merge.

### DIFF
--- a/nodes/modifier_change/edges_intersect_mk2.py
+++ b/nodes/modifier_change/edges_intersect_mk2.py
@@ -186,17 +186,20 @@ def intersect_edges_3d(verts_in, edges_in, s_epsilon):
 
 
 def edges_from_ed_inter(ed_inter):
-    '''create edges from intersecctions library'''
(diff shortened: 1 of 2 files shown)


vicdoval:
As I see it the "Remove Doubles" should be a double check, but the way it is right now it has to be used to avoid the doubled vertices (when one edge is  touching another for example).
Any thoughts on this @zeffii?  :D

zeffii:
i have no thoughts at the moment, bigger fish to fry :(

Issue #1663, opened 2017-05-22
Updating Node DocStrings


zeffii:
i'm going to start adding shortcut letter combos to the DoctString of all Nodes... this process will take a while and will involve editing most nodes. I will use this pass to reformat some akward pythonstyle if I find it. There will be no impact to the user, except that the extra-search feature will return your preferred nodes sooner if you type the new combos.

here's a sample of what I want 
```
ViewerDraw   |    vd View Geometry   |  shortcut is : vd

```
![image](https://cloud.githubusercontent.com/assets/619340/26299702/1449f44a-3edc-11e7-9643-14f8c6f5b2d6.png)

silly bit of code to get the list and cats

```python
import bpy

ddir = lambda data, filter_str: [i for i in dir(data) if i.startswith(filter_str)]
get_nodes = lambda cat: [i for i in getattr(bpy.types, cat).category.items(None)]

cycles_categories = ddir(bpy.types, "NODE_MT_category_SVERCHOK")
for cat in cycles_categories: 
    print(cat)
    for node in get_nodes(cat):
        print('- [ ] bl_idname: "{node.nodetype}", type: "{node.label}"'.format(node=node)) 
```

zeffii:
NODE_MT_category_SVERCHOK_Analyzers
- [ ] bl_idname: "SvBBoxNode", type: "Bounding box"
- [ ] bl_idname: "SvVolumeNode", type: "Volume"
- [ ] bl_idname: "AreaNode", type: "Area"
- [ ] bl_idname: "DistancePPNode", type: "Distance"
- [ ] bl_idname: "CentersPolsNodeMK2", type: "Centers polygons 2"
- [ ] bl_idname: "CentersPolsNodeMK3", type: "Centers polygons 3"
- [ ] bl_idname: "GetNormalsNode", type: "Calculate normals"
- [ ] bl_idname: "VectorNormalNode", type: "Vertex Normal"
- [ ] bl_idname: "SvKDTreeNodeMK2", type: "KDT Closest Verts MK2"
- [ ] bl_idname: "SvKDTreeEdgesNodeMK2", type: "KDT Closest Edges MK2"
- [ ] bl_idname: "SvMeshFilterNode", type: "Mesh filter"
- [ ] bl_idname: "SvEdgeAnglesNode", type: "Angles at the edges"
- [ ] bl_idname: "SvMeshSelectNode", type: "Select mesh elements by location"
[...]

PR #2159, opened 2018-03-03
Updated docstrings that appear in the browser

vicdoval:
The docstrings that appear in the search menu are sometimes missing, sometime useless due repeating the node name

I reviewed the menu and added some of them.
I tried to be as clear as possible
In some cases i added a shortcut as suggested  in https://github.com/nortikin/sverchok/issues/1663 
I tried to use different words to cover a bigger scope in the search.

I didnt change all the nodes, but i think 30 files is enough for this commit

## Preflight checklist

- [ x] Code changes complete.
- [x ] Documentation for users complete (or not required, if user never sees these changes).
- [x ] Manual testing done. 
- [ x] Unit-tests implemented.??
- [x ] Ready for merge.

### DIFF
--- a/nodes/analyzer/kd_tree_MK2.py
+++ b/nodes/analyzer/kd_tree_MK2.py
@@ -24,7 +24,7 @@ from sverchok.data_structure import (updateNode, match_long_repeat as mlr)
 
 
 class SvKDTreeNodeMK2(bpy.types.Node, SverchCustomTreeNode):
-    ''' KDT Closest Verts MK2 '''
+    '''Find nearest verts'''
     bl_idname = 'SvKDTreeNodeMK2'
     bl_label = 'KDT Closest Verts MK2'
     bl_icon = 'OUTLINER_OB_EMPTY'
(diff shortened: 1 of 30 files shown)


zeffii:
we have a better/new docstring convention.

https://github.com/nortikin/sverchok/blob/5cd48f56591b329537f8e6aca935bafcff04fff2/utils/docstring.py
https://github.com/nortikin/sverchok/blob/dc5a2bb8ed36cbf112060a88fe31e5d839af8d3c/nodes/vector/color_input.py#L27-L36

zeffii:
if you are willing to convert them to that.. that would be awesome :)

vicdoval:
I'm glad there is a new/better convention :) although it would have been better to discover it before. :D (The pitfalls of working before asking). 
I'm not sure when i'll be able to change them to the new convention, but I think it would  good to push this changes now, because despite of not being totally updated they are an small improvement compared to the actual state

zeffii:
ok : ) i don't have time to do it either

Issue #86, opened 2014-04-20
Edge graph to polygons


zeffii:
![edges_to_polies](https://cloud.githubusercontent.com/assets/619340/2751527/70dc2db6-c8d5-11e3-9273-551980c0f25b.PNG)

I'm investigating this now, automatically iterate through edges to find Edge Cycles with which to construct polygons. I think it makes sense to have a `min=3 max=n` parameter for the input. Hopefully will find the right graph theory paper to read, else I implement my own.

Suggestions welcome.

Durman:
It looks like not all holes are filled by this function. :/
https://gist.github.com/749b9604fc6361e93d313af3973123c8
![2018-02-20_09-00-11](https://user-images.githubusercontent.com/28003269/36408125-d055f05e-161c-11e8-899e-42745748c705.png)

Durman:
There is interesting artical on this theme, unfortunately in Russian:
http://www.e-maxx-ru.1gb.ru/algo/facets

zeffii:
yes, the nsides param doesn't expect many polygons in the incoming geometry, it works to a certain point, and then it will find polygons that you don't want - especially when nsides is larger than most of the easy to find polygons :).

i thought we had another fill holes... fill edgenet node..

Durman:
I have got something.)) I can now calculate outer edges but yet not for all graphs. I think it is important step on the way to creating this algorithm.

![outer edges](https://user-images.githubusercontent.com/28003269/36577808-04584b24-1873-11e8-94c7-6584ada6d146.gif)

Durman:
Little explanation about algorithm of searching outer edges:
![2018-02-24_18-26-15](https://user-images.githubusercontent.com/28003269/36631929-01322dda-1999-11e8-9e8a-64c8802953b8.png)

Durman:
It looks like nearly working.

![fill edges](https://user-images.githubusercontent.com/28003269/36711598-a0fc42c8-1b9d-11e8-87c0-7bb52ceca692.gif)

Durman:
![2018-02-27_10-53-00](https://user-images.githubusercontent.com/28003269/36714691-9fb8f0c8-1bac-11e8-962b-7688c8509fe9.png)

Durman:
Plane with randomly deleted edges was filled with new algorithm.
[...]

PR #2160, opened 2018-03-07
Planar edgenet to polygons

Durman:

#86
This will be base for other 2D manipulation with mesh.

## Solution description

This node fill all closed contours of edges graph but graph have to be planar.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


nortikin:
so, it is the same as fill holes node?

Durman:
not the same
https://gist.github.com/b0eb16271d33924457e443d74ac3c8d1

nortikin:
have to discribe it in docs

Durman:
ok, added this picture in documentation

zeffii:
add `planar_edgenet_to_polygons` to `sverchok/docs/nodes/alpha_nodes/alpha_nodes_index.rst` , then this can be pushed..

zeffii:
nice one Durman, my apologies for the delay.

enzyme69:
screen shot 2018-08-22 at 12 42 19 am

This is really weird, but with Planar Edgenet result, and manual edit selecting each and every face, I can basically Y operation and separate each face cleanly...

Durman:
New `Edges to faces 2D` node should replace this node but I have found that new node is less robust. There are still some cases which can crash the node. I'm too tired to fix this right now probably in future. So the `planar` node should be remained for a while.

PR #2164, opened 2018-03-08
Bevel only edges

vicdoval:

Bevel node does not work just with edges. The node does not update if there are no input polygons linked

I added (open) this feature.

This opens nice posibilities
beveledges

## Solution description

 When edges are inputed but not edges the propriety "Vertex only" is overwritten to True in order to work .

## Preflight checklist

- [x ] Code changes complete.
- [ x] Documentation for users complete (or not required, if user never sees these changes).
- [x ] Manual testing done. 
- [x ] Unit-tests implemented. Can you point me where to learn about this topic?
- [ ] Ready for merge.


zeffii:
ideally it would have a mode that took:
 
- verts+edges+vertmask+constant_radius
- verts+edges+vertmask+variable_radii

vicdoval:
I agree with you, 
For me it could be approachable to make the first option ( verts+edges+vertmask+constant_radius).
Right now if you input multiple radius you get multiple geometries
The implementation of variable radius in one mesh seems more complex with the actual bmesh.ops.bevel.  but i would be open to suggestions :)

zeffii:
yeah the variable radius would probably require rolling our own algorithm.. there's a few older script node examples where we use zmj1000's (or whatever his nickname was) algorithm to fillet corners.

vertmask+constant_radius would already allow a lot of cool stuff.

vicdoval:
screen
1- I've created the node, so the old "vertexOnly" propriety  (boolean) changes by and "mode" (enumpropiety), this may break old setups based on this node... 
    should i try to keep it somehow ?
    or should i create and mk2 ?
    or should i ignore this ?(the user may just have to change one button to make it work again)

2- Another question is that now I have BevelEdges and VerticesMask.
I think it would be more coherent to have EdgesMask and VerticesMask but this would make a bigger difference with the actual node (pushing more to the mk2..)
Any ideas on this?

vicdoval:
I twist it  again ... I will create a new clean pull request.
Thanks!
beveledges1

nortikin:
hilbert-bevel

PR #2167, opened 2018-03-12
dynamically generate submenu too

zeffii:
looks like this:

image


zeffii:
i'm hoping that categories in the examples will encourage creators to:

1. keep many examples simple, and try to cover few areas only
2. give more descriptive names to all examples, even those examples that are elaborate

zeffii:
@portnov I suspect the test would need to iterate over dirs too

portnov:
okay, will update the test.

zeffii:
i'll finish this first

zeffii:
@portnov  just this needs a change now.. want to do it or will I ?

https://github.com/nortikin/sverchok/blob/7c7a96834969f711f4035e9dc6a4b171db09fa1c/tests/tree_import_tests.py#L55-L74

zeffii:
i'd prefer if you did it :D

zeffii:
maybe something like this?
```python
from sverchok.utils.sv_examples_utils import examples_paths

@batch_only
class ExamplesImportTest(SverchokTestCase):
    def test_import_examples(self):

        for cat in examples_paths:

            if not examples_paths.get(cat):
                info("Dir named: %s" is empty, cat)
                continue

            info("Opening Dir named: %s", cat)

            for path in glob(join(examples_dir, "*.json")):
                name = basename(path)
                with self.subTest(file=name):
                    info("Importing: %s", name)
                    with self.temporary_node_tree("ImportedTree") as new_tree:
                        with self.assert_logs_no_errors():
                            # Do not try to process imported tree,
                            # that will just take time anyway
                            new_tree.sv_process = False
                            import_tree(new_tree, path)
                        for node in new_tree.nodes:
                            if is_old(node):
                                self.fail("This example contains deprecated node `{}' ({}). Please upgrade the example file.".format(node.name, node.bl_idname))
```

zeffii:
it's OK i'll do it.. it's almost done anyway.

zeffii:
looks like this now @portnov 

https://github.com/nortikin/sverchok/blob/3328f015f60aa848c2534cee6686bd350e3b340e/tests/tree_import_tests.py#L58-L78

is OK?

zeffii:
passes using testing.py

zeffii:
i'm anticipating that we will include, at some point, `.blend` files in the Example/Cats* folders. So perhaps eventually we will need a wrapper operator that decides what to do if .blend or .json are passed as filepath. (This is why i'm not doing any explicit filetype tests here)

PR #2168, opened 2018-03-14
List Math warp output

vicdoval:
The "List Math" output has a missing level.

During years it was necessary  to use a ·"list join warp" to input in a node with the output of "list math".
The "List Math" Output was [1] in stead of [[1]]
## Solution description

Added the missing brackets

## Preflight checklist

- [x ] Code changes complete.
- [ x] Documentation for users complete (or not required, if user never sees these changes).
- [ x] Manual testing done. 
- [x ] Unit-tests implemented.
- [ x] Ready for merge.


zeffii:
Best to avoid mk1 by adding a Boolean property to the UI , which defaults to false and doesn't autowrap, ..

 sv_set([out] if self.wrap else out)

PR #2171, opened 2018-03-18
Select mesh verts node

Kosvor2:


Kosvor2:
Now it's possible to choose between verts, edges and faces.

enzyme69:
@Kosvor2 Is there a documentation or in depth example of this node, I kind of dig this.

Kosvor2:
@enzyme69 try looking at this: https://github.com/nortikin/sverchok/blob/master/docs/nodes/beta_nodes/select_mesh_verts.rst

If something not clear enough, tell to me.

PR #2174, opened 2018-03-22
Formula colors node

Kosvor2:
Like formula deform but for rgba colors instead of vertices. Also have two input sockets.


Kosvor2:
colors_by_formula43475

nortikin:
json to examples

PR #2179, opened 2018-03-25
"Bmesh props" and "bmesh in" nodes reimagined.

Kosvor2:
bmesh_nodes_replacements234

Nodes which became unneeded now placed to old nodes.


zeffii:
'bmesh_list' sockets should be SvObjectsSocket I think...

am tempted to make a modification of this node with an interface for draw_buttons_ext to allow users to add output sockets "on demand" instead of presenting all when the node is created...

Kosvor2:
Using SvObjectsSocket for bmesh lists? If only it was possible to disable this automatic prop search on object socket...

zeffii:
Ok good point about the objects socket..

Issue #2182, opened 2018-03-29
"fixing/hating on" the default "add node" menu


zeffii:
this has bothered me for a long time... i vaguely recall going down this rabbit hole before..

```python
import bpy

import inspect
lines, line_idx = inspect.getsourcelines(bpy.types.NODE_MT_add.draw)
for l in lines:
    print(l, end='')

```

results in
```python
    def draw(self, context):
        layout = self.layout

        layout.operator_context = 'INVOKE_DEFAULT'
        props = layout.operator("node.add_search", text="Search ...")
        props.use_transform = True

        # actual node submenus are defined by draw functions from node categories
        nodeitems_utils.draw_node_categories_menu(self, context)

```
what does `nodeitems_utils.draw_node_categories_menu` look like..
```python
lines, _ = inspect.getsourcelines(nodeitems_utils.draw_node_categories_menu)
>>> for l in lines: print(l)
... 
def draw_node_categories_menu(self, context):
    for cats in _node_categories.values():
        cats[1](self, context)

```

zeffii:
then with little jumping through hoops we can do this.. to "monkey patch" this sucker.
```
import bpy
import nodeitems_utils


def draw_extra(self, context):
    """This draw function is overloaded by sverchok to revert, execute this:   """
    layout = self.layout

    ntree = context.space_data.node_tree
    show_sverchok = True if ntree else False
    
    if show_sverchok and ntree.bl_idname == 'SverchCustomTreeType':
        bpy.types.NODEVIEW_MT_Dynamic_Menu.draw(self, context)
    else:

        layout.operator_context = 'INVOKE_DEFAULT'
        props = layout.operator("node.add_search", text="Search ...")
        props.use_transform = True

        # actual node submenus are defined by draw functions from node categories
        nodeitems_utils.draw_node_categories_menu(self, context)


bpy.types.NODE_MT_add.draw = draw_extra
```

![image](https://user-images.githubusercontent.com/619340/38081281-eea1cfb2-3343-11e8-8400-8829fa93ae23.png)

that's much nicer than my original attempt..

zeffii:
now added a way to restore it ...
```python
>>> import sverchok
>>> sverchok.utils.perform_add_node_mp_restore()
```

PR #2183, opened 2018-03-29
adds option to monkey patch, proof of concept only

zeffii:
test if you want.. i probably will stick this behind a setting in prefs if needed.


zeffii:
https://github.com/nortikin/sverchok/issues/2182

zeffii:
the downside to this is.. it does assume no other `add-on` has taken any liberties with the draw function of `NODE_MT_add` and that might not be the case if AN is installed. so.... not sure.

PR #2194, opened 2018-04-07
Metaball lite

Kosvor2:
Main difference is use of `match_long_cycle` instead of `match_long_repeat` alowing for easier creation of patterns. But doesnt create meta objects by itself as it now uses object socket instead of textbox for new object name. Overall, the code has been very simplified.

meta_lite2353

### DIFF
--- a/index.md
+++ b/index.md
@@ -285,6 +285,7 @@
     SvBVHnearNewNode
     SvUnsubdivideNode
     SvLimitedDissolveMK2
+    SvMetaballOutLiteNode
     ---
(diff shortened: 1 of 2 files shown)


vicdoval:
Thanks, great addition!
I also really like `match_long_cycle` , I added to the "Crop 2d" node a propriety in the N-panel to change this behaviour because the user may expect the standard  `match_long_repeat`  and is nice to have both.
Maybe it would be a good practice  :)

vicdoval:
untitled

PR #2197, opened 2018-04-10
adds wad reader

zeffii:
adds a Basic Authorization storage/reader for gist uploads


nortikin:
nice. it will be benefit for sure. To store own collection.

zeffii:
ok if time later today will inspect and push

zeffii:
and add a gray-out for the upload button, when sverchok doesn't detect local wad file

zeffii:
internally sverchok will query the following code, and decide whether to show the `gist upload` button on the UI or not

```python
>>> sverchok.utils.sv_git_connection.login_found
True
```

zeffii:
ok this will work as follows; for basic authentication.

1. you need a github account, if yes - proceed to 2.  if no. byebye for now
2. you must convert your username:pw with a base64 encoder and obtains a string

    ```python
    import base64
    arp = '%s:%s' % ("your username", "your password")
    b43 = base64.b64encode(arp.encode()).decode('utf-8')
    print(b43)
    ```
    in this case

   > eW91ciB1c2VybmFtZTp5b3VyIHBhc3N3b3Jk

3. you store the string in a file called `sv_fx.wad` at this location on your HD `.../2.7x/datafiles/sv_fx.wad`
4. if blender was running, restart it and it will find a token and display the upload button again.

zeffii:
if people can confirm that they can figure out how to do this, then i'll maybe also add support for f8 (to avoid the restart)

portnov:
Any problems that prevent us from implementing that code (for creating a wad storage) in sverchok?

zeffii:
not really,  but i legitimately feel uncomfortable storing that b64-encoded string.

but of course. we could offer a popup windows with two text fields, and have "OK" generate the `.wad`, in the correct place.  .... but that's not really the point, and certainly isn't why i've avoided doing that.

zeffii:
if Sverchok could obtain a `secret` from github while generating the token, that would be cool.

portnov:
You mean you feel uncomfortable about storing non-encrypted passwords?... Well, I believe some other addons should of work with the same problem already...

zeffii:
i welcome proposals.

portnov:
It seems that "the correct way" of doing this is to ask users to generate github API access token manually (i think there is no way to generate it automatically from username/password, and should not be from security point of view), and enter that token in sverchok preferences.

zeffii:
if someone wants to make that happen / do a write up for users / cool

PR #2212, opened 2018-04-27
vector-drop-node-matrix-fix

Kosvor2:
or rather remake... I not exactly sure that `bmesh.ops.transform()` does same thing as old code.


nortikin:
right thing to do
but we can replace it with matrix math with inert mode... do we really need vector Drop?

Kosvor2:
I better not merge this pull request and do less dangerous fix directly in master.

Issue #1809, opened 2017-10-10
BMeshViewer to generate UV and retain between updates when topology is static.


enzyme69:
I have this dilemma with UV Map when working using Sverchok. I knew this matter will eventually arise.

1) I am generating some kind of random mesh using Sverchok and output it as BMesh. 
2) Once happy with random mesh, I am doing this UV projection and make UV nice.
3) Next step however, I want to be able to animate and modify the mesh, while preserving the UV. In the past, I just use Ctrl + L to transfer UV from mesh with good UV to final 3D object tweaked in Sverchok.

Blender has **DataTransfer Modifier** that actually do transfer UV on the fly, however this failed because whenever Sverchok updates the mesh, the UV disappear. So currently my "hack" is to use Exec Node to do smart_project(), but maybe not the best:

![prtscr capture_34](https://user-images.githubusercontent.com/3952687/31371940-47f45976-addf-11e7-85ae-b6617147547c.png)

Any tips on this? Thanks!

zeffii:
seems to me bpy.ops.uv.smart_project is a UI only operator? does it not take the 3dview viewing angle into account?

zeffii:
remember there's also the N-Panel settings for ViewerBmesh node, one mode may be useful here ( if your geometry is not adding / removing verts per update ) - I think this doesn't invalidate the UV ..but I could be wrong.

zeffii:
if the vert count and  edge/face indices change during animation is is only logical that the UV becomes invalidated by default and will need to be regenerated somehow.

I'm not sure if there's a bpy method to take a Mesh and pass it a Matrix (viewing angle) and get a UV projection.

zeffii:
o i see, smart_project doesn't need view..

[bpy.ops.uv.smart_project](https://docs.blender.org/api/blender_python_api_2_73_release/bpy.ops.uv.html#bpy.ops.uv.smart_project)
```python
uv.smart_project(angle_limit=66.0, island_margin=0.0, user_area_weight=0.0, use_aspect=True)
```

I think that operator only works in ExecNode when you have one object in the scene and it's active.

zeffii:
turns out this operator is coded in [pure python..](https://svn.blender.org/svnroot/bf-blender/trunk/blender/release/scripts/startup/bl_operators/uvcalc_smart_project.py)

enzyme69:
@zeffii Which means it can be easily become Script Node?

zeffii:
easily?  that depends who's coding :)

[...]

PR #2218, opened 2018-05-15
Set uv map

Durman:

This node is created to give opportunity do UV Map with power of Sverchok. I think best approach to solve the problem should be discussed in this theme.)) I already gave some examples here: #1809

## Solution description

The sense is to make opportunity create UV map and Unwrapping can be done with standard nodes of Sverchok.

[test_2018_05_15_17_12.zip](https://github.com/nortikin/sverchok/files/2006011/test_2018_05_15_17_12.zip)

## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/index.md
+++ b/index.md
@@ -337,4 +337,5 @@
     SvOffsetLineNode
     SvContourNode
     SvPlanarEdgenetToPolygons
+    SvSetUVMap
     
--- a/nodes/scene/Set_UVMap.py
+++ b/nodes/scene/Set_UVMap.py
@@ -18,23 +18,70 @@
 
 import bpy
 from bpy.props import StringProperty
+from mathutils import Vector, Matrix
 
 from sverchok.node_tree import SverchCustomTreeNode
-from sverchok.data_structure import updateNode
+from sverchok.data_structure import updateNode, repeat_last
+
+WARNING = []
 
 def create_uvmap(obj_in, uv_name):
     #Create basic UV map for object in scene with current name
     [obj.data.uv_textures.new(uv_name) for obj in obj_in 
         if not obj.data.uv_textures.get(uv_name, False)]
+        
+def remove_uv(obj_in, uv_name):
+    #Remove UV map from objects with current name of map
+    [obj.data.uv_textures.remove(obj.data.uv_textures[uv_name]) for obj in obj_in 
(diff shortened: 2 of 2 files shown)


nortikin:
ok, @Durman, write to me nikitronn (собака) yandex.ru

Durman:
It looks like `UVtextures` node distorts real UV map.

Durman:
I'm experimenting with uv operators and I have faced with such problem:

recursion bug

```
Smart Projection time: 0.00
2018-06-01 21:10:35,814 [ERROR] sverchok.core.update_system: Node Objects in mk3 had exception: maximum recursion depth
exceeded while calling a Python object
  File "C:\Users\User\AppData\Roaming\Blender Foundation\Blender\2.79\scripts\addons\sverchok-master\core\update_system.
py", line 325, in do_update_general
    node.process()
  File "C:\Users\User\AppData\Roaming\Blender Foundation\Blender\2.79\scripts\addons\sverchok-master\nodes\scene\objects
_mk3.py", line 231, in process
    with hard_freeze(self) as _:
  File "C:\Program Files\Blender Foundation\Blender\2.79\python\lib\contextlib.py", line 136, in helper
    return _GeneratorContextManager(func, args, kwds)
RecursionError: maximum recursion depth exceeded while calling a Python object
Smart Projection time: 0.00
```

Durman:
Experiments will continue in 2.8

gerroon:
Hi

How is this used? I am confused about how to make it work, also there is additional objects input and output there. Is this for creating new uvmaps ?

I would like to create some uvs procedurally in Sverchok.

thanks

vicdoval:
@gerroon check the docs :) https://github.com/nortikin/sverchok/blob/master/docs/nodes/object_nodes/set_custom_uv_map.rst

Durman:
@gerroon 2 and 3 inputs of the node expect 2D mesh of UV mapping. It's possible to use for this inputs the same mesh the object socket has. In this case the mesh will be projected into XY surface. 

To object output socket it's possible to connect other nodes like `assign material`. It lets to set order of applying attributes to an object.

gerroon:
thanks guys your answers were very helpful.

PR #2229, opened 2018-06-12
fixes presumption in changable_socket function that named socket exists

zeffii:
attempt to correct recent lackadaisical commits that break automated unit tests.


zeffii:
unfortunate that a few snlite example scripts have a 3rd party dependencies (conway, mcubes...) 

the options are:
- drop them
- write code in the auto tester to not test those scripts

even if the scripts include more elaborate try/except for the import, they won't be of any use and be degenerate and potentially not populate the tree anyway..

PR #2236, opened 2018-06-22
Fcurve sampler

zeffii:
slightly botchy fcurve sampler :)

several modes for this..(the node auto detects)

- default:  one value (take frame number, or if unconnected input..use `bpy.context.scene.current_frame`)
- multi_sample: take a sequence (a range, or ranges) and generate the evaluated range (or ranges)

### DIFF
--- a/docs/nodes/scene/FCurve_in.rst
+++ b/docs/nodes/scene/FCurve_in.rst
@@ -5,9 +5,9 @@ This node give you access to the FCurves bound to scene objects.
 
 Because this is an early version of the node, the UI is a bit rough. The Blender API is a little rough in this area too - don't expect huge future improvements. There doesn't appear to be a way to introspect the individual names of
 the standard keyframed properties like Location's `X Location Y Location Z Location`. ID props and 
-custom prop names can be introspected.
+custom prop names however _can_ be introspected, but i'm not adding it as a feature to the node.
 
-several modes for this..(the node auto detects)
+The node operates in several modes, it automatically attempts to process the input in these two ways.
 
 - default:  one value (take frame number, or if unconnected input uses `ctx.scene.current_frame`
 - multi_sample: take a sequence (a range, or ranges) and generate the evaluated range (or ranges)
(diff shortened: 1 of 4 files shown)


zeffii:
state 1 , single frame or current frame
image

state 2: multiple frames, including subframe
image

zeffii:
right now it only samples fcurves[0]  , which is probably whichever kind of property you first set keyframes for. In the example image above that's Location of `[Object Action]`

zeffii:
@durman test it.

it wont work is a json/gist because it assumes scene data

zeffii:
it could also accept custom properties, rather than hijacking a location keyframe. Maybe an enum between `Default |  custom` .
 - where default expects location keyframes, and either xyz  
 - and custom takes a float property (so no array_index needed)

Durman:
looks working

fcurve


But if object hasn't any actions we have got an error. :/
```Python
  File "C:\Users\User\AppData\Roaming\Blender Foundation\Blender\2.79\scripts\ad
dons\sverchok-fcurve_sampler\nodes\scene\FCurve_in.py", line 42, in draw_buttons

    action = bpy.data.objects[self.object_name].animation_data.action
AttributeError: 'NoneType' object has no attribute 'action'
```

Also what do you think about UI like choosing material of Bmesh weaver?

I am positive that we can want to use many instances of f-curve and it can be a bit confusing to use index only. List could feature all actions that an object has and with plus button we can create custom property with key on the first frame. I think it can become easier for newcomers to sort out in this node.

Also I noticed that it is possible to hide object and still have f-curve available for editing with pressed this button.

Durman:
fcurve test

Durman:
F-curve has loads of interesting modifiers ))

Durman:
fcurve test3

zeffii:
thanks for testing @Durman  yeah, the UI isn't sorted out yet :)

error handling and UI tweaks will come soon.

zeffii:
the simplest solution to "picking objects that don't have animation data / keyframes" @Durman  is to add a button that adds a location keyframe (standard property)

it's maybe worth going with that option first, because adding keyframes for a custom ID property has wasted an hour of my time already.

zeffii:
regarding the UI ugliness of picking an array index.. well i agree, named property selection would be awesome. Just the code is a little bit more boring/involved at this early stage.

zeffii:
image

Issue #2238, opened 2018-06-27
Vector polar output


Durman:
## Problem statement

Little problem with `vector polar output`.

## Steps to reproduce

- Connect `Z`/`theta` socket. 
- Switch `Cylinder`/`Sphere` modes. 
- Do the same with disconnected `Z`/`theta` socket.

## Expected result

`Z` and `theta` sockets should replace each other when `cylinder` and `sphere` modes are being changed.

## Actual result

In `sphere` mode `Z` socket cannot give any value.

![vector polar output](https://user-images.githubusercontent.com/28003269/41966259-b4a6c52a-7a0f-11e8-94b4-519b368f1ff2.gif)

Durman:
It looks like the socket cannot be hidden if it has a connection.

https://github.com/nortikin/sverchok/blob/f053deb6bb1024667c610ee2d02e5d88718b7754/nodes/vector/vector_polar_out.py#L56-L59

zeffii:
i added a method last year called `socket.safe_hide` or `.safe_hide`, nodes that don't use this will behave the way you show.

PR #2239, opened 2018-06-28
Update vector_polar_out.py

Durman:

#2238

## Solution description

As to me, it is not convenient to hide sockets because user can hide and unhide sockets by himself.

## Preflight checklist

- [x] Code changes complete.
- [x] Ready for merge.

### DIFF
--- a/nodes/vector/vector_polar_out.py
+++ b/nodes/vector/vector_polar_out.py
@@ -54,9 +54,14 @@ class VectorPolarOutNode(bpy.types.Node, SverchCustomTreeNode):
     ]
 
     def coordinate_changed(self, context):
-        if self.coordinates != self.outputs[2].name:
-            self.outputs.remove(self.outputs[2])
-            self.outputs.new('StringsSocket', self.coordinates)
+        # changing name of third output socket
+        replaceable_socket = self.outputs[2]
+        mode_name = self.coordinates
+        
+        if mode_name in ['z', 'theta']:
+            replaceable_socket.replace_socket('StringsSocket', new_name = mode_name)
+        else:
+            raise Exception ("Unexpected mode - {}".format(mode_name))
 
     coordinates = EnumProperty(items=coord_modes, default='z', update=coordinate_changed)
 
--- a/nodes/vector/vector_polar_out.py
+++ b/nodes/vector/vector_polar_out.py
@@ -54,9 +54,9 @@ class VectorPolarOutNode(bpy.types.Node, SverchCustomTreeNode):
     ]
 
     def coordinate_changed(self, context):
-        self.outputs["z"].hide = self.coordinates == "theta"
-        self.outputs["theta"].hide = self.coordinates == "z"
-        updateNode(self, context)
+        if self.coordinates != self.outputs[2].name:
+            self.outputs.remove(self.outputs[2])
+            self.outputs.new('StringsSocket', self.coordinates)
 
     coordinates = EnumProperty(items=coord_modes, default='z', update=coordinate_changed)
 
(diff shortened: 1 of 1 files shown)


Durman:
We have now always only three sockets instead four. It doesn't matter for the rest part of the code.

https://github.com/nortikin/sverchok/blob/f053deb6bb1024667c610ee2d02e5d88718b7754/nodes/vector/vector_polar_out.py#L110-L111

zeffii:
i added `hide_safe` to allow for consistency, i'd like that to be used rather than keeping the socket when it outputs nothing. 

or do something like `unused_socket.sv_set([["this socket not used in this mode"]])` , at least that informs of the UI inconsistency

zeffii:
but wait.. 

💡 

we can also use `socket.replace_socket` to rename the socket between Z and tetha depending on the mode. That would leave any connected socket unchanged and not create dead sockets :)

nodes/vector/vector_polar_out.py
```diff
@@ -54,9 +54,9 @@ class VectorPolarOutNode(bpy.types.Node, SverchCustomTreeNode):
     ]
 
     def coordinate_changed(self, context):
-        self.outputs["z"].hide = self.coordinates == "theta"
-        self.outputs["theta"].hide = self.coordinates == "z"
-        updateNode(self, context)
+        if self.coordinates != self.outputs[2].name:
```
portnov: I do not like magic numbers :) And, the idea of relying on node's property being equal to output name is not great imho. Just because it is not explicit.
I'd prefer more explicit (even if more lines of code):
````
if self.coordinates == "z":
...
elif self.coordinates == "theta":
...
else:
    # Just in case we will add something later and forget to add a check here
    raise Exception("Unsupported coordinates type: {}".format(self.coordinates))
````

Comment (portnov) [reply nodes/vector/vector_polar_out.py]: Huh, I just discovered I used the same idea in another part of this code :) Well, I still do not like it.
As an option, we could do
````
def get_third_coordinate_name(self):
    # Here we are relying on self.corodinates being equal to name of output, but now it is explicit
    return self.coordinates
````
and then use this method , for example, in line 110 and similar places:

````
 if self.outputs[self.get_third_coordinate_name()].is_linked:
````

File: nodes/vector/vector_polar_out.py [hunk 2]
```diff
@@ -54,9 +54,9 @@ class VectorPolarOutNode(bpy.types.Node, SverchCustomTreeNode):
...
+            self.outputs.remove(self.outputs[2])
+            self.outputs.new('StringsSocket', self.coordinates)
```
portnov: I agree with @zeffii , that it is better to rename the socket than to remove and create new one, because this way existing connections will remain. Probably this is more expected behaivour from the user's point of view.

zeffii: @portnov  .. i think replace_socket keeps track of the old onnection and reconnects..

this goes into: 

    def replace_socket(self, new_type, new_name=None):
        """Replace a socket with a socket of new_type and keep links,
        return the new socket, the old reference might be invalid"""
        return replace_socket(self, new_type, new_name)

this:

https://github.com/nortikin/sverchok/blob/master/data_structure.py#L611

Durman: So much comments for so small change. ^_^
According all that was suggested, I also have an idea. What about such code?

```Python
    def coordinate_changed(self, context):
        self.outputs[2].name = self.coordinates
```
It looks short and clear, isn't it?
And connections are keeping.

zeffii: please use `.replace_socket(....)` as documented in *data_structure.py*, it handles also the freezing of the node tree. That code is there for a reason :)

Durman: @portnov @zeffii Is it okay? As I see there is no necessary in `if` function at all.
```Python
    def coordinate_changed(self, context):
        replace_socket(self.outputs[2], 'StringsSocket', new_name = self.coordinates)
```

zeffii: no. use `self.outputs[2].replace_socket("self.coordinates")`  
- if socket type stays the same, it doesn't need to be passed
- if only thing that changes is the name, that's all you pass :)

zeffii: correction? this should work :)  
```python
self.outputs[2].replace_socket("StringsSocket", self.coordinates)
```

portnov:
now it's better, imho :)

zeffii:
lets do this

Issue #2244, opened 2018-07-10
3pt Arc node, type of vector input UI


Durman:
## Problem statement

As I see it, there is no point to have input where it is possible to set coordinates for only one point because the node expects at least three coordinates.

![2018-07-10_08-45-05](https://user-images.githubusercontent.com/28003269/42489709-a4bb1900-841d-11e8-80ac-07181d45a6b3.png)

zeffii:
Agreed this default UI makes no sense.on 3pt node.

zeffii:
What about a quicklink that adds 3 vector nodes and a list join, sets the vectors to a sane default then links them.

zeffii:
or indeed, just a vector quicklink :)

Durman:
Great idea. If it possible we can connect just `list input` node in `vector` mode.

Durman:
might so?

```Python
vector_prop.prop_name = 'at least 3 points'
```
![2018-07-10_12-53-08](https://user-images.githubusercontent.com/28003269/42499646-5c3f345e-8440-11e8-9d29-28774000bdbd.png)

zeffii:
no. we dont do that. i'll push an acceptable change :) soon!

zeffii:
not ready yet:  https://github.com/nortikin/sverchok/pull/2245

PR #2245, opened 2018-07-10
3pt update

zeffii:
the default UI for this node doesn't make a whole lot of sense, we can use the quicklink mechanism to inject a vector node into the 3pt input socket. It might be interesting and beneficial for other nodes to also offer a way for nodes to override the exact kind of  quicklink behaviour.

The initial intent of the quicklink handler was to allow specific behaviour that makes sense to individual nodes. A quicklink for the Rotation node isn't the same behaviour you might want for the 3pt_node, proof is the [issue posted](https://github.com/nortikin/sverchok/issues/2244) by @Durman .

This pr updates the 3pt node  and now will offer a custom function (defined in the node) to set up your own quicklink behaviour.

 - new general `socket property` called "custom_quicklink_function_name", it will (obviously?) store the name of the function (located on the node) that will be used to draw the `socket ui row`.
- when the above property is not "", then the function will be called in a try/except by the quicklink pathway in `node_tree.py` -  

```python
     def draw_quick_link(self, context, layout, node):
         ...
```
- if there are errors in the UI draw call, then the default quicklink for that sockettype will be drawn.

no ETA.

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -89,6 +89,8 @@ class SvSocketCommon:
     use_quicklink = BoolProperty(default=True)
     expanded = BoolProperty(default=False)
 
+    quicklink_func_name = StringProperty(default="", name="quicklink_func_name")    
+
     @property
     def other(self):
         return get_other_socket(self)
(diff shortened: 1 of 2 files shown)


zeffii:
@Durman  you can try the branch

Durman:
If quick link should add vector in node so this works well.

zeffii:
The new code replicates the standard behaviour of quicklink, and places it inside the node's own class. Now, because, there's no "perfect" or "ideal" goal here i'm going to keep it at passing a Vector In node.

What i'd consider doing (maybe later today) is write a new operator for this node based quicklink definition, that also lets us define the entire quicklink operator code (like a callback). This would mean we don't need to keep making operators..just functions

zeffii:
remember there are two other ways to get automatic node creations
 - add a method to the right click feature (each node can have a few right click options)
 - custom macros that operate on the currently active node

all of which have examples, so there are fairly extensive options here. 

> A default will satisfy most people some of the time, and no-one all the time.

zeffii:
so now if you are a node developer and you want a clever way to provide an automatic set of input nodes, then you create an operator and write your own draw_quicklink function :)

PR #2246, opened 2018-07-16
Upgraded HexaGrid MK1 to Polygon Grid and Contour2D to beta

vicdoval:

The hexaGrid node is great working with hexagons but  does not  handle any other polygon.

## Solution description

I expanded the node so it can work also with triangles and squares.
Also added:
Possibility to define polygon by radius or side length.
A smart Remove Doubles (when scale is 1 and "separate is unchecked) 
List handling. Now if you click "separate" you will get the unpacked polygons unless you are creating many grids simultaneously.

Images for the reference:
polygongridpaav
poligongrid

## Preflight checklist

- [ x] Code changes complete.
- [ x] Code documentation complete.
- [x ] Documentation for users complete (or not required, if user never sees these changes).
- [x ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ x] Ready for merge.


Durman:
It is strange that this node is not green as other generation nodes. May it worth to fix? 
Also, as I see this node could replace `plane` node if several function would be added to the node.

vicdoval:
Now the  `Generators Extended` are coloured as the standard `Generators`.

screen1

About replace de `plane`  node:
Although generally speaking a "Plane" is a infinite planar surface, in the Blender World a plane is a rectangle.. I think is nice to have a standard "Plane" Node for basic usage.
Anyway I could add some more functions to the node.
The XY, YZ and ZX are quite practical but easy to achieve with a 'Vector Rewire'
Maybe the "Normalize" option could worth to be implemented. It would demand also the option of define normalization by radius or by side length of the layout. I'm not sure the complexity vs the obtained the functionality.
Also this `polygon grid` node could be expanded to hold rectangles an even more interesting "irregular pentagons". Rectangles and pentagons would need more parameters to define the shape
What do you think about this?

Also i slightly change the help panel, just the format :)
I would remove the capital letters but I also wanted to respect last @nortikin changes.

screen1

portnov:
* I'd like to ask to make separate PRs for separate changes: 1) new node, 2) move of Contour node to another category, 3) changes in the help panel.
* Am I getting it right that you did not touch existing HexaGrid node, and just added a new node?

DolphinDream:
Do we want all extended generators to be the same color as the main generators? Perhaps make them a different color may be more suitable ?

DolphinDream:
Even though the node is called Hexa Grid it actually generates a Triangular Grid suitable for generating hexagonal tiles (or triangle tiles). I’m not sure why this would be called “polygonal” grid though. The hexagonal and triangular tiles lie on a triangular grid, while the square tiles lies on a rectangular grid which can be easily done with a plane node (minus the specific triangle/rhomb/hexa layouts).

Durman:
@DolphinDream 
How would you advise to divide generators node for dyeing them in different colours? 
If i see green coloured node I understand that it is generator. What else should I to know about generator nodes?

DolphinDream:
@Durman about the genX color.. my initial idea was to color them a slight shade of green to distinguish them from the main generators. Ultimately it may not be that important, but I kept a different color considering they had different colors to begin with. I made them different than the default gray color to disginguish them form the rest of the gray nodes but  kept the greenish color to make them similar to the generator colors.

Here’s the implementation I came up with a while ago:
https://github.com/DolphinDream/sverchok/tree/generatorXNodeColors

sv-generatorxcolors

portnov:
@vicdoval To do not break existing layouts, you have to
* move existing HexaGrid node to old_nodes directory (without changing it's bl_idname)
* remove it from menu
* give your new node a new bl_idname.

portnov:
@vicdoval also, you can allow users to easily replace existing HexaGrid nodes with your new node via right-click menu; you need to add just one line like https://github.com/nortikin/sverchok/blob/master/old_nodes/matrix_in.py#L31 to old node code.

vicdoval:
@portnov  Ok I will close this PR and open some others with the changes separated. Also i will move the hexagrid to the old nodes and establish the right click link

@DolphinDream indeed it is possible to create triangular tiles with the HexaGrid node but if you try to create a big grid you would need to do quite an extra calculations to get the triangular tiles (maybe with Kdt Closest edges + planar edges to poligons) because  the triangular tiles have alternate rotations  so the can't be placed just by its location.
This upgrade  just make it faster using the same concepts who where inside the HexaGrid MK1 to create different tilings.

Mi initial idea was to call the node "grid" but "polygon grid" talks about the polygon array it generates so i find it more complete and could be expanded in the near future to hold pentagonal grids
About the colours I would go with the same colours between generators and generators extended because I don't se clearly the difference between this categories

vicdoval:
@portnov  should I move the HexaGridMk1.rst to the docs/old folder or just delete it?

portnov:
As far as I remember, we do not have exact agreement on this matter. Maybe @zeffii will suggest better.

zeffii:
> @portnov should I move the HexaGridMk1.rst to the docs/old folder or just delete it?

old node help files can always be read using gits history, we have not been in the habit of keeping old docs around. Usually the doc is updated (renamed..updated...) or removed when the node is entirely superseded by an alternative node.

PR #2253, opened 2018-07-31
start operator for dupunique

zeffii:
new duplicate unique operator, specifically for monad nodes.

### DIFF
--- a/core/monad.py
+++ b/core/monad.py
@@ -67,19 +67,14 @@ def make_new_classname(monad_node_group):
 def monad_make_unique(node):
 
     """
-    create a new version of the monad class, with a copy of the node tree linked to .monad
+    Create a new version of the monad class (duplicate but unique)
 
-    this will attempt to store the duplicate in a json, like is used for IO/json. The upside is 
-    that this will test the pack/unpack routine continuously. the downside is that this will likely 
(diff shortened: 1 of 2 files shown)


zeffii:
this is close to being testable at the moment. just a few final tweaks.

zeffii:
todo..

    # at this point record all incoming and outgoing connections
    ...

    # remove the old node (this is probably a duplicate anyway)
    ...

    # reconnect old connections.
    ...

zeffii:
and i'm pretty sure we've already written very very similar code already

zeffii:
in essence this is the functionality i wanted to provide.

Q - how do i make a unique duplicate?
A - steps: 
    1. copy the monad node instance you want to make unique
    2. hit spare bar to get the list of operators
    3. type "dup" and selected "Duplicate Unique".
    result:
    this will 
    - copy the instance
    - register a new version of the monad class
    - place new monad class over the copy

does not yet remove the existing instance... will add later. (today?)

zeffii:
i've not dedicated a kb shortcut to it yet. .. but am considering adding a custom menu item to the monad node rightclick. just more work than I want right now.

zeffii:
i've renamed the operator and bl_idname to 
```
bl_label = "Make Unique (Monad)"
```

PR #2260, opened 2018-08-18
add notion to return new node from make_monad_unique function

zeffii:
thinking about making various version of the operator. some of these modes/scenarios may overlap

- [ ] ensure fake user of new nodegroup

- [ ] mode 1
use an existing unattached instance of monad to make a unique duplicate, remove existing instance - replace with unique duplicate

- [ ] mode 2
if  existing instance is connected to other nodes, add a unique version and remove the existing instance
remap any previously existing connections

- [ ] mode 3
allow dup unique operator to drag from existing instance and create the unique duplicate with no attachments.

### DIFF
--- a/core/monad.py
+++ b/core/monad.py
@@ -85,8 +85,10 @@ def monad_make_unique(node):
 
     # the new tree dict will contain information about 1 node only, and 
     # the node_group too (at the moment) but the node_group data can be ignored.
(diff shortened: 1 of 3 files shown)


zeffii:
i had to merge this unfinished because it resolves some seriously incomplete implementation of the initial commit. Will require more work soon.

PR #2265, opened 2018-08-28
adds size socket dynamically

zeffii:
addresses the lack of Size socket for LineMK2


zeffii:
this can not be pushed with the same classname as MK2. so don't,

zeffii:
tested , seems OK .

PR #2285, opened 2018-09-24
use transform for monadclone unique

zeffii:
monad clone worked, but didn't behave as Blender's own AddNode operator (which attaches the new node to the mouse cursor/ relative to it)

### DIFF
--- a/core/monad.py
+++ b/core/monad.py
@@ -98,6 +98,8 @@ def monad_make_unique(node):
     node_items['monad'] = new_monad_group.name
     node_items['cls_dict']['cls_bl_idname'] = new_cls_name
 
+    pre_nodes = set(nodes)
+
     # place new empty version of the monad node
     import_tree(node_tree, nodes_json=layout_json)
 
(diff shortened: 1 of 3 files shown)


zeffii:
a slight note on usage..

1. if invoked from the node's own rightclick menu, then the attach_translate will detach from the node fairly soon after triggered. Not sure why this
2. invoke the operator from the space menu, does not exhibit this problem. The node remains attached.

Durman:
@zeffii 
Clone of monad with custom name can be done one time only. 
Names of several monads skip numbers, for example: monad => monad.001 => monad.003

zeffii:
interesting, i will look at it. soon.

zeffii:
it's very tempting to code something like sverchok as a stand-alone program :)  sans-blender

Durman:
Is it possible?)) I think Svechok cann't survive without Blender.

zeffii:
there would be a 'bridge' to _output to_ and _update_ the scene. I think it would be supremely doable. with enough time :)

PR #2286, opened 2018-09-27
Random Num Gen Weighted Int and distributed Float 

vicdoval:
"Random Num Gen" offers uniform distributions, but it could be expanded (using numpy) it to create Weighted distribution of Integers and non-uniform distribution of floats

## Solution description

I added most of the methods described in de numpy.random reference https://docs.scipy.org/doc/numpy-1.14.0/reference/routines.random.html

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/nodes/number/random_num_gen.py
+++ b/nodes/number/random_num_gen.py
@@ -32,7 +32,6 @@ class SvRndNumGen(bpy.types.Node, SverchCustomTreeNode):
     Triggers: Random thru a range
     Tooltip: Generate a random number (int of float) thru a given range (inclusive) .
     """
-    ''' '''
     bl_idname = 'SvRndNumGen'
     bl_label = 'Random Num Gen'
     bl_icon = 'OUTLINER_OB_EMPTY'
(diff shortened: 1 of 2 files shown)


vicdoval:
Images for docs
screenrandom
screenrandom1
screenrandom2
screenrandom3

zeffii:
nice work!  but i'm a little uncomfortable with the retention of of the classname. an MK2 seems more appropriate.

another small issue i have is with adding/remove sockets. please look at examples of using `.hide_safe`

vicdoval:
Thanks! 

I will change the `.hide_safe`,  it looks great
About the MK2 thing:  with the updates it  does not break old setups as the default is the same as how it use to be (Uniform distribution)
¿Any way would you be fine with changing the class but leaving the name un-touched?
(I don't like very much the MK2 on the menu)

vicdoval:
Also not changing the class is more like a automatic update 😃

Durman:
Great.)
test1
test2

vicdoval:
long life to .hide_self 😃

zeffii:
i'm more confident about this Pull Request now :)

zeffii:
code looks pretty nice btw.

nodes/number/random_num_gen.py
```diff
@@ -27,7 +28,11 @@
 
 
 class SvRndNumGen(bpy.types.Node, SverchCustomTreeNode):
-    ''' Generate a random number (int of float) thru a given range (inclusive) '''
+    """
+    Triggers: Random thru a range
+    Tooltip: Generate a random number (int of float) thru a given range (inclusive) .
+    """
+    ''' '''
```
zeffii: empty string?

vicdoval:
> code looks pretty nice btw.

I've been learning from the best (the sverchok developers) 😃

zeffii:
good man!

PR #2290, opened 2018-10-16
Smooth splines node

zeffii:
## adds a "filleted polyline" Node. I call it smoothed line

```
        b  .                                     ..y..
       .        .                          .        . 
      .              .            .                  .
     .                    c                           .
    a                                                ..z..
```
in individual weight mode you pass a `location` and `weight` for all points ( when weights is connected )
in homogeneous weight mode all weights are the value of the slider. (currently not smart)

progress
- [x] relative mode ( weight is relative to the distance between the `a-b b-c`
- [x] absolute mode ( weight is real world value dependant)
- [x] clamping, automatically constrict values that are not possible in absolute mode
- [x] cyclic open / closed
- [x] generate edges
- [ ] drop consecutive doubles (optional, and with epsilon) 
- [ ] forgetting other modes...

image

image

### DIFF
--- a/index.md
+++ b/index.md
@@ -42,6 +42,7 @@
     SvTorusKnotNode
     SvRingNode
     SvEllipseNode
+    SvSmoothLines
 
 ## Analyzers
     SvBBoxNode
--- a/nodes/generators_extended/smooth_lines.py
+++ b/nodes/generators_extended/smooth_lines.py
@@ -84,12 +84,12 @@ class SvSmoothLines(bpy.types.Node, SverchCustomTreeNode):
         items=smooth_mode_options, description="offers....",
         default="absolute", update=updateNode)
 
-    close_mode_options = [(k, k, '', i) for i, k in enumerate(["cyclic", "open"])]
-    close_selected_mode = EnumProperty(
-        items=close_mode_options, description="offers....",
+    type_mode_options = [(k, k, '', i) for i, k in enumerate(["cyclic", "open"])]
+    type_selected_mode = EnumProperty(
+        items=type_mode_options, description="offers....",
         default="cyclic", update=updateNode)
 
-    n_verts = IntProperty(default=5, name="n_verts", min=0) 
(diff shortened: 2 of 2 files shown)


zeffii:
finding it hard to concentrate on this, but it's a nice distraction anyway

image

undecided about how cyclic mode should work, whether it should 
- auto drop the last vert
- auto drop the last vert if it is identical to the first vert

zeffii:
image
cyclic mode added,

zeffii:
i may push this as it without implementing the clamping. clamping will introduce extra processing time per update because each weight/radius would need to be checked against available free "line" to fillet from.

it's not the prettiest node code, but the user doesn't have to know that :)

zeffii:
this is the kind of node where you can get desired results if you feed it "non garbage" input, and you might get some interesting / unexpected (but cool) results if you just throw exaggerated values at the weight socket

zeffii:
docs forthcoming :)

zeffii:
let me know what you lot think.

Durman:
For what is `attributes` socket need?

zeffii:
not implemented yet, i could hide it..  but if I don't it's likely i will implement it sooner :) 

> eventually....

attributes should accept a `list(dict)`, where the dict contains something like
```python
{
0: [weights, mode (for each corner), type: closed/open]
1: [weights, mode (for each corner), type: closed/open]
...
n: [weights, mode (for each corner), type: closed/open]
}
```
can be passed by `execnode mod`, i think.

zeffii:
note to self : 
image

zeffii:
almost got arcs :)

image

zeffii:
got arcs !  pushed already

image

zeffii:
now.. clamping isn't yet implemented, and is also not essential for productive use of this node. so might be a while

Durman:
It also can be interesting to have option to prevent to set radius of circle more than distance to closest neighbour point.

zeffii:
yeah, that's what i call "clamping", it will be optional because slightly slower

PR #2291, opened 2018-10-29
Add fillet to smoothline

zeffii:
image

adds arc :)  - the image shows a slight artificial offset for illustration purposes only, the node doesn't produce the offset.

### DIFF
--- a/docs/nodes/generators_extended/smooth_lines.rst
+++ b/docs/nodes/generators_extended/smooth_lines.rst
@@ -4,7 +4,7 @@ Smooth Lines
 Functionality
 -------------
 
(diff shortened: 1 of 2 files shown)


nortikin:
faces + laplacian smooth crashes blender with no debug output
[smooth.zip](https://github.com/nortikin/sverchok/files/2538743/smooth.zip)

zeffii:
this is not intended for faces...  but it should not crash

zeffii:
that's a different node @nortikin    ,  "smooth vertices"

zeffii:
image

regarding the smooth vertices, it might be that input vertices that all have the same 0.0 for z, might make the algorithm slow if it's smoothing out x y and z, when only x and y make sense.

nortikin:
smooth

Issue #2309, opened 2018-11-13
fix location + scale of easing node graph


zeffii:
come on @DolphinDream  :)  while you're at it

zeffii:
keep in mind the drawing code will change substantially in `sverch for 2.8`

Issue #2284, opened 2018-09-22
Stethoscope info out of place & Viewer index font too small


DolphinDream:
I noticed that on current master the stethoscope info is displayed overlapped with the node.. and the font is smaller than usual.. Also (may be related) the viewer index text in the view port is way smaller than it used to be (unreadable). Any ideas what the problem may be ? 

 sverchok version: 0.5.9.6 - master (0bdf0a18) 

<img width="1344" alt="screen shot 2018-09-21 at 9 58 51 pm" src="https://user-images.githubusercontent.com/2083719/45912053-933bce80-bde9-11e8-9a0d-81bfa9a79558.png">

enzyme69:
A know issue 😏 https://github.com/nortikin/sverchok/issues/1969

@zeffii Has been trying to fix it but I think it's more like Apple thing...

zeffii:
yep, feel free to suggest a solution for Mac, and test the crap out of it. If that works i will integrate it into the nodes in a way that works for both OS.

alessandro-zomparelli:
Actually I have the same problem on windows 10... the position is always unpredictable.
![immagine](https://user-images.githubusercontent.com/6708848/46906849-34f69d00-cf0a-11e8-8922-56b656d18e2f.png)

DolphinDream:
Yeah. The node is almost entirely unusable for me (high sierra, macbook pro 2018).

DolphinDream:
It seems that other nodes that render in node view have their rendering location messed up.
Here’s the easing and texture nodes renders.

<img width="788" alt="sv-nodeview-render-bug" src="https://user-images.githubusercontent.com/2083719/46907888-dae4e000-cee7-11e8-924c-335f6603b399.png">

DolphinDream:
Originally I though this may have something to do retina display in OSX having messed up the font DPI somehow (as stethoscope renders text).. but apparently the artifact may not be related to fonts at all.

zeffii:
what kind of DPI settings guys? .. i also use win10

zeffii:
The texture and easing draw code doesn't take into account the Blender DPI/resolution settings. Unlike the stethoschope node. Look at how it decides where and at what scale to draw
```
                    scale = prefs.stethoscope_view_scale
                    location_theta = prefs.stethoscope_view_xy_multiplier
```
[...]

PR #2310, opened 2018-11-14
Fix render scale and location of easing and texture viewer nodes

DolphinDream:
Easing and texture viewer nodes suffered form the same x2 resolution issue as the index viewer and stethoscope nodes due to resolution doubling on some machines (e.g. mac book with retina display). This PR fixes the issues by using the same render scale / xy multiplier used already by the stethoscope node to adjust the rendered location/scale.

- made scale and xy multiplier pref settings generic to be used by multiple nodes
- add scale/multiplier adjustments to the texture viewer and easing node and update stethoscope node to use the same scale/multiplier

Addressing issues discussed in #2309 and #2284

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/nodes/number/easing.py
+++ b/nodes/number/easing.py
@@ -23,6 +23,7 @@ from bpy.props import FloatProperty, EnumProperty, StringProperty, BoolProperty
 import blf
 import bgl
 
+from sverchok.utils.context_managers import sv_preferences
 from sverchok.data_structure import updateNode, node_id
 from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.ui import nodeview_bgl_viewer_draw_mk2 as nvBGL2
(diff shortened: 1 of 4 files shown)


DolphinDream:
sv-renderscalefix

zeffii:
also looks fine, feel free to merge if you're happy.

DolphinDream:
Btw, since the name of the "xy multiplier” preference setting was changed (to make it generic and used by other nodes needing such multiplier) any previously set preference for the “stethoscope xy multiplier” is no longer effective. Users who set this multiplier to a value other than the default will need to reset it to their desired value in the SV preferences again. This time it is called the “Render scale” and “Render location xy multiplier”.

Note: the new “render scale” setting is used currently by the easing and texture viewer nodes. I intentionally kept the stethoscope and the index viewer node scale settings in place in case the user may want to tweak these scales independently.

@enzyme69 make note of this before starting your next livenoding :) so your stethoscope output is showing in the right place.

DolphinDream:
Well.. I thought this battle was kind of won.. but nope.. there are some issues still. 

Texture Viewer Lite node

sv-textureviwerlight-outputoffset

zeffii:
that reminds me, i should port the lite version to b2.80 :)

DolphinDream:
@zeffii while porting it.. note the new issue #2342  .. there may be some problems with the GL rendering code in the TVL node.

zeffii:
thanks, i see. yes, shouldn't be a problem.

DolphinDream:
@zeffii I fixed the offset/scale in this PR #2343

zeffii:
as for the 'stale drawing' issue, that's because sverchok nodes generally don't have a reason to use the 'update' function. The last time a node would ever be sent a message that it no longer has any input data is if the node implements the `update` function, and from that function calls the disable callback.

See here like in stethoscope
```
    def update(self):
        if not ("Data" in self.inputs):
            return
        try:
            if not self.inputs[0].other:
                nvBGL.callback_disable(node_id(self))        
        except:
            print('stethoscope update holdout (not a problem)')
```



The reason there's extra code in that update function (f.ex to test if a socket exists) , is because that function will get called like 5 orso times before it ever hits the `process` function for the first time.

zeffii:
the dirtier implementation details :)

PR #2314, opened 2018-11-17
Added "BW" Grayscale and "A" Alpha channel support

alessandro-zomparelli:

Based on the suggestion of @enzyme69 I added two more output sockets for grayscale and alpha channel

## Preflight checklist

- [x ] Code changes complete.
- [ x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ x] Ready for merge.

### DIFF
--- a/nodes/analyzer/evaluate_image.py
+++ b/nodes/analyzer/evaluate_image.py
@@ -34,8 +34,9 @@ class EvaluateImageNode(bpy.types.Node, SverchCustomTreeNode):
 
     boundary_modes = [
         ("CLIP", "Clip", "", 0),
-        ("CYCLIC", "Repeat", "", 1),
-        ("MIRROR", "Mirror", "", 2)]
+        ("EXTEND", "Extend", "", 1),
+        ("CYCLIC", "Repeat", "", 2),
+        ("MIRROR", "Mirror", "", 3)]
 
     shift_modes = [
         ("NONE", "None", "", 0),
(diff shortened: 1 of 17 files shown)


nodes/analyzer/evaluate_image.py
```diff
@@ -183,10 +186,16 @@ def process(self):
...
                     green[0].append(imag[index+1])
                     blue[0].append(imag[index+2])
+                    alpha[0].append(imag[index+3])
                 else:
                     red[0].append(0)
                     green[0].append(0)
                     blue[0].append(0)
+                    alpha[0].append(0)
+        if outputs['BW'].is_linked:
+            bw[0] = [(r+g+b)/3 for r,g,b in zip(red[0], green[0], blue[0])]
```
zeffii: `(r+g+b)/3`  ?
not to be picky but...
image
https://en.wikipedia.org/wiki/Grayscale

zeffii:
adding sockets at the top of the socket list (either inputs or outputs) will break layouts that contain versions of the node before the update, best practice is to add sockets to the end of the socket lists.

UI coding is the painful part of nodes :)

nodes/analyzer/evaluate_image.py
alessandro-zomparelli: You are not picky, it's my bad. I was to hurried :P, I'll fix that ;)

alessandro-zomparelli:
> 
> 
> adding sockets at the top of the socket list (either inputs or outputs) will break layouts that contain versions of the node before the update, best practice is to add sockets to the end of the socket lists.
> 
> UI coding is the painful part of nodes :)

Thanks for pointing out, I didn't know that.

nodes/analyzer/evaluate_image.py
zeffii: : ]

nodes/analyzer/evaluate_image.py
```diff
@@ -133,11 +135,13 @@ def process(self):
...
 
         # outputs
+        bw = [[]]
         red = [[]]
+        alpha = [[]]
         green = [[]]
         blue = [[]]
 
-        if outputs['R'].is_linked or outputs['G'].is_linked or outputs['B'].is_linked:
+        if outputs['R'].is_linked or outputs['G'].is_linked or outputs['B'].is_linked or outputs['A'].is_linked or outputs['BW'].is_linked:
```
zeffii:
```python
if any(socket.is_linked for socket in self.outputs):
    ...
```

alessandro-zomparelli: That's magic :D 
Unfortunately I'll be on vacation for two weeks (Mexico with no PC). I will be able to handle that only at beginning of December. If you want to fix that in the meanwhile, fell free to modify it. Otherwise I'll take care of it when I'll be back.

zeffii:
ok. i'll leave it for now. merged anyway.

zeffii:
done.

PR #2294, opened 2018-10-30
line mk3 with AB and OD modes

vicdoval:
## Addressed problem description

The line node does not accept multiple inputs on the size socket
Also if you want to make a line from A to B in Sverchok you have to deal with vector evaluation or another tricks.
This also happens if you want to create a line from one origin point with one direction.
The new node tries to be friendly for new users and to mimic the classical definitions of a line

## Preflight checklist

Put an x letter in each brackets when you're done this item:

- [x] Code changes complete.
- [x] Code documentation complete.
- [x] Documentation for users complete (or not required, if user never sees these changes).
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.

PR #2318, opened 2018-11-22
Add MK4 version of the line node

DolphinDream:
This is based on MK3 version features (AB/OD modes) except that it proposes a slightly different approach:

- the XYZ are no longer modes.. but simply presets to set the V1 and V2 vectors to their proper values as to align the line along XYZ. This way the V1,V2 sockets always stay visible and do not need to be hiden/shown based on the mode. This helps with the connected nodes to the V1/V2 sockets to not be disconnected when the user aligns the line along XYZ.
- the node has only two modes : AB and OD
- simplify the code (the MK3 code got to be a bit too hard to follow after the additions of the new features)
- vectorized the size .. so each line now can get a different size if needed.

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

sv-linemk4-1


DolphinDream:
sv-linemk4-demo1

DolphinDream:
sv-linemk4-demo2

vicdoval:
Some thoughts about this PR:

1_The X,Y and Z buttons became useless on the AB mode? If they do maybe they should be hidden when using the AB mode and they should be under the AB/OD buttons.

2_The upgrade_if_needed function (which was already removed in de MK3) was useful when the "size" socket was added in the MK2, i think is nonsense to keep it in a new version of the node

3_The MK3 version already allowed multiple inputs on the size socket. 

4_At a first glance I don't see this code simpler compared to the MK3, you even added  a full operator. Some of the complexity on the mk3  appears from the intention to keep the node as fast (and as simple for user) as possible on the xyz modes, not parsing the vectors to to the make_line function

5_Personally I don't se the point in exposing the MK version in the node name (expect if both are in use) it makes the menu look dirtier. Also, the caps on MK4  call the attention of the eye but it is not important for the final user, it is only useful for developers. Hiding this info has been already done on the viewer draw node ( for example)

DolphinDream:
@vicdoval 
1. Yes.. the XYZ presets are not effective if the V1-V2 input sockets are connected since the inputs are taken from the sockets not from the prop. In MK3 the sockets are disconnected once you select XYZ “modes” which is kind of annoying. In MK4 the sockets are kept in place. I agree, we can hide the XYZ buttons once the sockets are connected to unclutter the UI.
2. Agree. I wasn’t sure what this code was used for so I left it in place. This should be removed if not needed.
3. OK. 
4. We may have to agree to disagree on this one. I was going over the MK3 code many time trying to understand what it does to make sure it didn’t break anything, and it was still not clear to me. I had to add debugging prints to see what data is being generated at what point based on the sockets input etc. Eventually I gave up and created the new functionality starting from MK2. Yes, the operator class for XYZ presets adds to the “complexity” of the node, but the generation of the prameters for making the line I think it’s a lot less intricate than MK3. Besides MK3 adds functionality to other files in order to complete its own functions while MK4 doesn’t. In general it’s important that the code is kept easy to understand by the developers for any future maintenance and expansion.
5. Agree. I’m not stuck on MK in names.. I got zeffi’s point on that, so this can definitely be removed.

DolphinDream:
Here’s the update with the XYZ buttons hidden once V1/V2 inputs are connected. 


sv-linemk4-demo3

nodes/generator/line_mk4.py
```diff
@@ -0,0 +1,207 @@
...
+        name='V2', description='Point V2 (ending point or direction)',
+        size=3, default=(1, 0, 0), update=updateNode)
+
+    def sv_init(self, context):
+        self.inputs.new('StringsSocket', "Num").prop_name = 'num'
+        self.inputs.new('StringsSocket', "Step").prop_name = 'step'
+        size_socket = self.inputs.new('StringsSocket', "Size")
+        size_socket.prop_name = 'size'
+        size_socket.hide_safe = True
+        self.inputs.new('VerticesSocket', "V1", "V1").prop_name = "point_V1"
```
zeffii:
```diff
-self.inputs.new('VerticesSocket', "V1", "V1").prop_name = "point_V1"
+self.inputs.new('VerticesSocket', "V1").prop_name = "point_V1"
```
etc :) and all the other duplicates too please.  It offers generally no benefit at the moment.

DolphinDream:
Btw, @zeffi, is there an easier way to get / set the caller nodes on operators? I borrowed then join/split code from another none but I thought it would be beneficial to add this feature at a lower level so that any node can use it. Not sure what the best place and implementation would be. Any thoughts / suggestions ?

vicdoval:
@Durman 
Now i remember i did some tricks in that node in order to keep it minimal... maybe it went to cryptic... it seemed clear... 😄 
The other question is that i keept the xyz mode as they where because they are faster than the vector interpolation method. Although it is a fast node there are some performance differences.
I tested the mk4 as it is now and when using the AD and OD modes your implementation is 15% faster than the mk3, but around a 20-30% slower compared to the xyz basic modes..... (probably faster would be numpy 😄 )

About depending on other files I find a good practice to use functions already written in other files so the code is not duplicated, also completing with some functions files like geom_utils.py can be useful for others...

BTW at this point the size socket does not appear on normalize maybe you can borrow this
https://github.com/nortikin/sverchok/pull/2294/files#diff-af13320fc5ed5a93a726cef57b2b69eeR63

DolphinDream:
Opps.. in last commit I accidentally removed the show/hide the size socket which broke the normalization. I fixed that.

DolphinDream:
@vicdoval 
I optimized the vector interpolator to take care of the cases when lines is aligned with x/y/z directions.. and incidentally this also optimizes the cases when the line is within xy / yz / xz planes. 

Could you please run your performance tests and see how this version performs ? 
Tough, keep in mind that this may still be slightly slower than the x/y/z in MK3 since it does an addition of the V1 to the lines. So, in order to compare apples to apples, you’d have to do a vector addition to the lines generated with MK3.

DolphinDream:
@vicdoval I didn’t mean to be picky on the geom_utils additions. If they are useful for other nodes it’s good to have.

Thanks for running some performance tests. I was wondering whether it was faster or slower. I expected that the xyz “modes” would be slower in MK4 since these are no different than the AB mode, which does interpolation instead of just setting the points to one of the x,y,z single component vectors as the MK3/2 does.

DolphinDream:
@vicdoval 
I optimized the vector interpolator to take care of the cases when lines are aligned with x/y/z directions.. and incidentally this also optimizes the cases when the lines are within xy / yz / xz planes. 

Could you please run your performance tests and see how this version performs ? 
Tough, keep in mind that this may still be slightly slower than the x/y/z in MK3 since it does an addition of the V1 to the lines. So, in order to compare apples to apples, you’d have to do a vector addition to the lines generated with MK3.

zeffii:
> Btw, @zeffi, is there an easier way to get / set the caller nodes on operators? I borrowed then join/split code from another none but I thought it would be beneficial to add this feature at a lower level so that any node can use it. Not sure what the best place and implementation would be. Any thoughts / suggestions ?

most of the time the callback can get enough information simply from within the operator's context like here:

https://github.com/nortikin/sverchok/blob/065af54b3a5b9e0aecc619c81533420a79f13d7b/nodes/script/multi_exec.py#L47-L63

There exists already a generic callback Mixin that you can base your operator on. Which means adding the functionality of the operator as member functions to the Lines Node class instead.

https://github.com/nortikin/sverchok/blob/065af54b3a5b9e0aecc619c81533420a79f13d7b/utils/sv_operator_mixins.py#L22-L61

you'll find that the operators that need treename and nodename are usually operators that can be called from other sources (like bakeall...is triggered from a panel which has no direct link to the node, but will call bake on all instances of nodes that have a bake function..)

```
            opera = row.operator('node.sverchok_mesh_baker_mk2', text="B A K E")
            opera.idname = self.name
            opera.idtree = self.id_data.name
```
I think the reason `image_decompose` node uses that awkward `split(`|><|`)` is down to my inexperience with the API.. in 2014. These days i would write that operator using a mixin and add member functions to the node class to handle the specific operation. Something like:

something like:

```python
from sverchok.utils.sv_operator_mixins import SvGenericCallbackWithParams

class SvSetLineDirection(bpy.types.Operator, SvGenericCallbackWithParams):
    bl_label = "Set line direction"
    bl_idname = "node.set_line_direction"   # dont use sv.
    bl_description = "Set the direction of the line along X, Y or Z"

    direction = StringProperty(default="X")
```

inside the node class:
```python
    def set_direction(self, operator):
        node.direction = operator.direction
        node.mode = "OD"
        return {'FINISHED'}
```

DolphinDream:
@zeffii the reason why I disregarded the enum implementation for XYZ directions in MK4 (which keeps the V1/V2 inputs and XYZ are no longer special modes) is that when you select lets say X and V1/V2 get updated to their corresponding direction, but then you modify the V1/V2 values, the X enum would still show selected which is no longer appropriate since the V1/V2 could be a different direction. I didn’t want to have the XYZ enum buttons selected once you modify the values of V1/V2, hence I opted to make them preset operators instead.

zeffii:
ok..well, i should probably try the node to get a better understanding. keep it as you see fit.

zeffii:
(but use the mixin please)

zeffii:
I would prefer you don't make this kind of change at the moment @DolphinDream  .   when the user hovers over those vectors I think they show the full name anyway..  so..

DolphinDream:
@zeffii no problem. Nevertheless, I think it would look more polished if you showed specific short names than whatever the first letter of the name happens to be. As you know, the description is currently simply set to the name  (if not explicitly set in the new calls) and it is currently not used, so I thought this would be a good chance to put it to good use.

zeffii:
in future that makes sense.   i'd like to freeze nodetree for the time being.

nodes/generator/line_mk4.py
```diff
@@ -0,0 +1,216 @@
...
+    for s in [0.0] + steps:
+        l = l + s
+        add_vert(vec(l))
+    edges = [[i, i + 1] for i in range(len(steps))]
+
+    return verts, edges
+
+
+class SvLineNodeMK4(bpy.types.Node, SverchCustomTreeNode):
+    ''' Line '''
```
zeffii:
```python
    """
    Triggers: Line  l4
    Tooltip: blablablabla
    
    blablablablablam use it to describe the node to the reader of code, keep short?.
    """
```

vicdoval:
@DolphinDream  
About the UI: in the mk3 I created 4 vector proprieties to correctly display  the names and keep separated the modes information 
I used this function to avoid the socket to unplug and keep everything clearer, maybe it was not a bad idea to implement it in the mk4
https://github.com/nortikin/sverchok/blob/065af54b3a5b9e0aecc619c81533420a79f13d7b/nodes/generator/line_mk3.py#L70-L90 


About the performance:
I was comparing the implementations and the mk4 is in most of the cases almost doubles the speed of the lineMk3 vector interpolation. Congratulations!
Compared to the XYZ is still slower but only a 15-20%
Thinking a about this i did a implementation in NumPy.... the tricky thing about NumPy is this tolists() bottleneck.. so depending on the input parameters it can be 4 times faster or 3 times slower.
But it would be able to expose the result as NumPy array and becoming 5-10 times faster...
Maybe it would be nice to offer the option
```
def make_line_numpy(steps, size, v1, v2, center, normalize, mode, as_numpy):


    # get the scaled direction (based on mode, size & normalize)
    if mode == "AB":
        nx, ny, nz = [v2[0] - v1[0],  v2[1] - v1[1], v2[2] - v1[2]]
    elif mode == "OD":
        nx, ny, nz = v2

    nn = sqrt(nx * nx + ny * ny + nz * nz)
 
    np_steps = np.array([0]+steps, dtype=np.float64)
    np_stepsLength = np.sum(steps) # length of the non-normalized steps
    eval = np.cumsum(np_steps)
    if normalize:
        scale = 1 if nn == 0 else (1 / nn / np_stepsLength * size)  # scale to given size
    else:  # not normalized
        if mode == "AB":
            scale = 1 / np_stepsLength  # scale to AB vector size
        elif mode == "OD":
            scale = 1 if nn == 0 else (1 / nn)  # scale to steps size

    nx, ny, nz = [nx * scale, ny * scale, nz * scale]

    np_vec=np.zeros((len(eval),3), dtype=np.float64)
    np_vec[:,0] = np.interp(eval,[0,np_stepsLength],[v1[0],nx*np_stepsLength])
    np_vec[:,1] = np.interp(eval,[0,np_stepsLength],[v1[1],ny*np_stepsLength])
    np_vec[:,2] = np.interp(eval,[0,np_stepsLength],[v1[2],nz*np_stepsLength])
    verts = np_vec if as_numpy else np_vec.tolist()
    
    ed=np.zeros((len(steps),2), dtype=np.int64)
    ed[:,0]=np.arange(len(steps))
    ed[:,1]= ed[:,0]+1
    edges = ed if as_numpy else ed.tolist()

    return verts, edges
```
the  use_numpy and as_numpy  and  could be exposed in the N-panel
```
    use_numpy = BoolProperty(
        name='Use NumPy', description='use NumPy to interpolate',
        default=False,
        update=updateNode)

    as_numpy = BoolProperty(
        name='As NumPy', description='on means output a NumPy Array, off means output regular list',
        default=False,
        update=updateNode)
```

DolphinDream:
Thanks for the profiling. Interesting. Honestly I did not expect MK4 would be twice as fast in general cases than MK3. For the time being I don’t see how I can make the special xyz cases go faster unless I create an 64 list of lambdas for all cases when any of ox,oy,oz,nx,ny,nz are zero. The optimal interpolator currently only optimizes over the cases when nx,ny,nz are zero. 

As for the socket names.. for MK4 I was trying to stay away from having to swap sockets in/out just so that the name appears to match the mode. Too much code for very little benefit (IMO).  In the end the code logic will use the input to the vector sockets to decide whether they are AB or OD vectors so it doesn’t really matter how we label them, hence the V1/V2 generic labels. If you guys like to have the names match the modes then yes, we can put this in MK4 too. @zeffii on this topic btw, I was wondering whether (in the future) we could have a socket display name/description be separate from its string ID (currently defined to be its name). A socket could be created with a string ID (name), but be given a different display name and display description. If the socket type stays the same, but its semantic changes, the node would only have to update the sockets display name/description without having to swap in/out with other sockets (id/prop_name).

As for the nuympy I don’t think we can safely switch to it unless it is faster in all cases which would probably require the entire node system to propagate numpy input/outputs, otherwise converting to list, which the current implementation uses, will compromise any speedup we gain from using numpy.

zeffii:
> I was wondering whether (in the future) we could have a socket display name/description be separate from its string ID (currently defined to be its name). A socket could be created with a string ID (name)

yeah, i'd like to reserve a property name in svSocketCommon, maybe `.display_name` , which can be used when that property is set to anything other than the empty string.

DolphinDream:
Here’s a case study of various interpolators.  I added two other vector interpolators and some component based interpolators to see which one performs better. I used a simple timing/averaging profiling technique, but I hope is good enough to get an estimate using various node setups (lines aligned along X/Y/Z, zero origin etc).  If you want to profile it you’ll have to uncomment the lines for each case and reload the addons. 

Based on my measurements the fastest case is the vector interpolator2. This is the same as the previous interpolator (intepolator1) but instead of defining an array of lambdas and picking one using an index based on some bit operation on the nx,ny,nz values, I did a simple if/else instead. The set of lambdas is the same among the first two interpolators. For reasons that are beyond me the interpolator2 performs better. Why is it that returning a lambda from an array is slower than returning a lambda based on if/else decisions is not quite clear to me.

The component based (XYZ) interpolators  in principle should perform faster or the same as the vector interpolators 1/2  in the special cases when ox,oy,oz,nx,ny or nz have zero values since the component base interpolators covers all special cases while the interpolators 1/2 only covers cases when nx,ny, nz have zero values (and don’t handle the ox,oy,oz = zero cases). Again, for some reason the component based interpolators seem to be actually slower than the vector based intepolators, my guess is being that python needs to make 3 calls to 3 lambdas vs of one call to a lambda in the vector interpolators.

The case a3 covering all 64 lambdas where ox,oy,oz,nx,ny or nz have zero values (stored in an array) also seem to be slower than the vector interpolators 1/2. Not sure why.

portnov:
You may wish to use a decent profiler as well :) #1920

DolphinDream:
> You may wish to use a decent profiler as well :) #1920 , #1933

Thanks for the tip @portnov . This is sweet! I can see much better now where the problems are :)

Btw, for anyone following my “case study” comments, I figured out why the get_interpolators using array of lambdas are slower than the if/else get_interpolators. :) The entire array of lambda is generated every time as they depend on the input go the get_interpolator method, and this is apparently is a costly operation. The if/else version will narrow down to the point where only one lambda is generated.

Also, by using the profiler I noticed that the edge generation (the typical list comprehension: edges = [[i,i+1] for i in range(n)] is surprisingly slow. In some cases about 25% of the time spent in the make_line is only to generate the edges. :(  which led me to some ideas of how we can go about optimizing this (to come shortly in a different PR). Essentially we can cache the linear edge list and not have to regenerate the same list over and over in any node. This linear edge list generation, btw, is used in many nodes so once we optimize the edge generation multiple nodes will be able to take advantage of this speedup. 

Below is the profiling for line MK4 node with and without the edge generation optimization. The first statistics are for the line node using the optimized edge generation (using an edge cache), the other is for the node using the typical list comprehension. As you can see, the percentages of the edge generation time (get_edge_list) from the total time spent in making the line (make_line) are significantly different:

```
 0.420 / 30.962 = 1.36% - for the optimized edge generation
20.392 / 50.969 =   40% - for the non optimized edge generation 
```
I measured similar results in other nodes btw (e.g. spiral  node).

```
  
   Ordered by: cumulative time

(optimized edge generation)

     2126    0.395    0.000   35.379    0.017 line_mk4.py:350(process)
   257246    0.450    0.000   33.981    0.000 profile.py:89(wrapper)
   257246    9.885    0.000   30.962    0.000 line_mk4.py:186(make_line)
 24023800   16.645    0.000   16.645    0.000 line_mk4.py:81(<lambda>)
   257246    0.719    0.000    2.516    0.000 profile.py:45(is_profiling_enabled)
 29598172    2.317    0.000    2.317    0.000 {method 'append' of 'list' objects}
   514492    0.168    0.000    0.943    0.000 {built-in method builtins.next}
   257246    0.103    0.000    0.915    0.000 contextlib.py:57(__enter__)
   514492    0.463    0.000    0.774    0.000 context_managers.py:29(sv_preferences)
  2402380    0.602    0.000    0.602    0.000 line_mk4.py:70(<lambda>)
  2402380    0.595    0.000    0.595    0.000 line_mk4.py:76(<lambda>)
   257246    0.188    0.000    0.530    0.000 contextlib.py:134(helper)
   257246    0.275    0.000    0.420    0.000 data_structure.py:968(get_edge_list)
     6378    0.022    0.000    0.411    0.000 socket_data.py:85(SvGetSocket)
     4252    0.015    0.000    0.370    0.000 node_tree.py:299(sv_get)
   257246    0.221    0.000    0.351    0.000 contextlib.py:63(__exit__)
   257246    0.298    0.000    0.342    0.000 contextlib.py:37(__init__)
     4252    0.153    0.000    0.333    0.000 data_structure.py:94(match_long_repeat)

(not optimized edge generation)

     2089    0.433    0.000   55.702    0.027 line_mk4.py:350(process)
   252769    0.485    0.000   54.237    0.000 profile.py:89(wrapper)
   252769   10.016    0.000   50.969    0.000 line_mk4.py:186(make_line)
   252769    0.290    0.000   20.392    0.000 data_structure.py:968(get_edge_list)
   252769   20.102    0.000   20.102    0.000 data_structure.py:978(<listcomp>)
 23605700   16.376    0.000   16.376    0.000 line_mk4.py:81(<lambda>)
   252769    0.794    0.000    2.730    0.000 profile.py:45(is_profiling_enabled)
 29083058    2.308    0.000    2.308    0.000 {method 'append' of 'list' objects}
   505538    0.178    0.000    1.014    0.000 {built-in method builtins.next}
   252769    0.111    0.000    0.986    0.000 contextlib.py:57(__enter__)
   505538    0.510    0.000    0.836    0.000 context_managers.py:29(sv_preferences)
  2360570    0.770    0.000    0.770    0.000 line_mk4.py:76(<lambda>)
  2360570    0.666    0.000    0.666    0.000 line_mk4.py:70(<lambda>)
   252769    0.201    0.000    0.576    0.000 contextlib.py:134(helper)
     6267    0.023    0.000    0.420    0.000 socket_data.py:85(SvGetSocket)
     4178    0.014    0.000    0.377    0.000 node_tree.py:299(sv_get)
   252769    0.326    0.000    0.376    0.000 contextlib.py:37(__init__)
   252769    0.236    0.000    0.375    0.000 contextlib.py:63(__exit__)
     4178    0.155    0.000    0.339    0.000 data_structure.py:94(match_long_repeat)

```

vicdoval:
@DolphinDream About the edge cache sounds good, but im not sure how it would handle if every item has different length, probably you already solved it 😄 

About the numpy implementation:
I was running some more tests and I got that in my computer it was faster when working between 2000 verts.  and 60000 verts. the peak was around the 6000 doubling the actual implementation and with 100000 gets to be 30% slower.
I still think it would be nice to have it as a option in the N-panel, I believe is a good practice to move little by little to numpy when possible pointing to a numpy node-tree, What do you think about this @zeffii , @nortikin @portnov

DolphinDream:
@vicdoval the edge cache stores the list of edges generated by a given node (e.g. [[0,1], [1,2] ... ) during a call to get_edge_list(n), and this cache grows the list as edge lists with larger number of edges are requested. In case when less edges are needed a call to get_edge_list will simply get a subset of the latest edge cache list.

e.g. say the edgeCache is currently: [[0,1], [1,2], [2,3]] (from a previous calls) 
then a call to generate 5 edges :

`edges = get_edge_list(5) 
`
will expand the edgeCache list to [[0,1], [1,2], [2,3], [3, 4], [4, 5]] and return this to caller.

a later call (by the same node or different node) to get a list with fewer edges will simply return a subset of the current edge cache:

`e.g. edges = get_edge_list(4) 
`
will return edgeCache[:4] = [[0,1], [1,2], [2,3], [3, 4]] without having to augment the cache (since it was already augmented to more (5) edges in previous calls).

zeffii:
i'm a social liberal, but computational conservative. 

i bet there are 10's of ways to improve linear edge list generation..  personally I would consider doing a few initial cachings of np.arrays `n = 10. 100, 500, 1000, 2000, ....`  or whatever gradient of increase in `n` is economical, and then do a numpy slice on the array that's bigger than the needed n.  numpy slices are not copies but "views" of an np.array, so no duplication.

zeffii:
or it might be interesting to pass symbolic info (ie: this is a linear list of n verts) / or lazy evaluate the indices only when needed (generator functions).

some nodes however do expect all that data to be present,, but some nodes really don't care as long as they can iterate over it.

zeffii:
isn't something like this already ridiculously fast?
```python
import numpy as np

def get_edges_A(n, as_list=False):
    edge_array = np.array((np.arange(n), np.arange(1, 1+n)), np.int32).transpose()
    if as_list:
        edge_array = edge_array.tolist()
    return edge_array

def get_edges_B(n, as_list=False):
    array_one = np.arange(n + 1)
    array_two = np.roll(array_one, -1)
    edge_array = np.array((array_one, array_two), np.int32).transpose()[:-1]
    if as_list:
        edge_array = edge_array.tolist()
    return edge_array


print(get_edges_A(4))
print(get_edges_B(4))
```

portnov:
@vicdoval From a performance perspective, we must use numpy whenever it is possible. See also #1990.
I'm not so sure however about what should we transfer between nodes. For now, most of nodes expect list or tuples, and you can see a lot of tests like "if isinstance(x, list)" across the whole codebase. So for compatibility reasons, I think we still have to make all calculations in numpy, then transform result to list, then pass it to next node, then transform list to numpy array again...

to move the whole processing to numpy was one of ideas behind "Sverchok 0.6" project, as far as I remember. Nobody had enough time to make that alive, though...

DolphinDream:
Latest UI demo

sv-linemk4-demo4

nortikin:
so?

DolphinDream:
> so?

So, if there are no objections we could merge this into master.

One thing left to do is to update the documentation "Example of Usage" section. Any hints as to how to update the images associated with the documentation? Where should these images be posted so that the documentation generated based on the *.rst files finds them ? Any particular recommendation/guidelines for the SV theme to be used to capture the node trees. Is there a need for consistency of theme for these examples?

Durman:
after this merging #2815 this PR is outdated I think

PR #2321, opened 2018-11-28
Optimize the linear edge list generation using caching

DolphinDream:
Profiling the edge generation in some nodes indicated that it is a slow operation and since this edge list is generated over and over in various nodes it makes sense to cache it. By using caching a speedup of almost two orders of magnitude in generating the edges was measured compared to generating the edges using list comprehension (e.g. edges = [[i, i+1] for i in range(N)].

With the proposed optimization, the list of edges are stored in an edge cache (e.g. [[0,1], [1,2]... [n-1,n]] .. and the cache is extended as longer edge lists are requested/generated. Any call to get_edge_list will return a subset of this edge cache thus not having to re-generate the same list over and over.

Various nodes like the line, spiral, torus knot, ellipse etc, which generate lines with linear list of edges can benefit substantially from this speedup.

example code:

```
edges = get_edge_list(n)
vs
edges = [[i, i+1] for i in range(n)]
```
Additionally, if multiple calls are going to be made to get_edge_list with different numbers (of which max value is known), one could further optimize the edge generation calls by pre-caching the largest edge list by first calling update_edge_list(maxN), this way the subsequent calls to get_edge_list with different n values (less than maxN) will not have to extend the cache since the cache is large enough to generate edges for all n values.

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/data_structure.py
+++ b/data_structure.py
@@ -927,3 +927,52 @@ def std_links_processing(matcher):
         return real_process
 
     return decorator
+
+
+# EDGE CACHE settings : used to accellerate the (linear) edge list generation
+_edgeCache = {}
+_edgeCache["main"] = []  # e.g. [[0, 1], [1, 2], ... , [N-1, N]] (extended as needed)
+
+
+def update_edge_cache(n):
+    """
+    Extend the edge list cache to contain at least n edges.
+
+    NOTE: This is called by the get_edge_list to make sure the edge cache is large
+    enough, but it can also be called preemptively by the nodes prior to making
+    multiple calls to get_edge_list in order to pre-augment the cache to a known
+    size and thus accellearate the subsequent calls to get_edge_list as they
+    will not have to augment the cache with every call.
+    """
+    m = len(_edgeCache["main"])  # current number of edges in the edge cache
+    if n > m: # requested #edges < cached #edges ? => extend the cache
+        _edgeCache["main"].extend([[m + i, m + i + 1] for i in range(n - m)])
+
+
+def get_edge_list(n):
+    """
+    Get the list of n edges connecting n+1 vertices.
+
+    e.g. [[0, 1], [1, 2], ... , [n-1, n]]
+
+    NOTE: This uses an "edge cache" to accellerate the edge list generation.
+    The cache is extended automatically as needed to satisfy the largest number
+    of edges within the node tree and it is shared by all nodes using this method.
+    """
+    update_edge_cache(n) # make sure the edge cache is large enough
+    return _edgeCache["main"][:n] # return a subset list of the edge cache
+
+
+def get_edge_loop(n):
+    """
+    Get the loop list of n edges connecting n vertices.
+
+    e.g. [[0, 1], [1, 2], ... , [n-2, n-1], [n-1, 0]]
+
+    NOTE: This uses an "edge cache" to accellerate the edge list generation.
+    The cache is extended automatically as needed to satisfy the largest number
+    of edges within the node tree and it is shared by all nodes using this method.
+    """
+    nn = n - 1
+    update_edge_cache(nn) # make sure the edge cache is large enough
+    return _edgeCache["main"][:nn] + [[nn, 0]]


data_structure.py
```diff
@@ -927,3 +927,37 @@ def real_process(node):
...
     return decorator
+
+
+# EDGE CACHE settings : used to accellerate the (linear) edge list generation
+_edgeCache = {}
+_edgeCache["main"] = []  # e.g. [[0, 1], [1, 2], ... , [N-1, N]] (extended as needed)
+
+
+def update_edge_cache(n):
+    '''
```
zeffii: these look like backticks?

DolphinDream: I checked. they are not. I can replace them with normal quotes if that’s  more desirable.

zeffii: no matter, docstring syntax is unfortunately mixed in data_structure.py. so asking you to use double quotes would be almost arbitrary.

DolphinDream:
> (zeffi) isn't something like this already ridiculously fast?
```

import numpy as np

def get_edges_A(n, as_list=False):
    edge_array = np.array((np.arange(n), np.arange(1, 1+n)), np.int32).transpose()
    if as_list:
        edge_array = edge_array.tolist()
    return edge_array

def get_edges_B(n, as_list=False):
    array_one = np.arange(n + 1)
    array_two = np.roll(array_one, -1)
    edge_array = np.array((array_one, array_two), np.int32).transpose()[:-1]
    if as_list:
        edge_array = edge_array.tolist()
    return edge_array


print(get_edges_A(4))
print(get_edges_B(4))
```

It seems neither of these two are any faster than even the list comprehension.. here are some profiling results:

```
Ordered by: cumulative time

     2383    0.588    0.000   72.575    0.030 line_mk4.py:352(process)
   288343    0.679    0.000   70.499    0.000 profile.py:89(wrapper)
   288343   13.154    0.000   65.959    0.000 line_mk4.py:186(make_line)
   288343    0.731    0.000   26.097    0.000 data_structure.py:933(get_edges_A)
   288343   23.613    0.000   23.613    0.000 {method 'tolist' of 'numpy.ndarray' objects}
 26927900   21.312    0.000   21.312    0.000 line_mk4.py:81(<lambda>)
   288343    1.115    0.000    3.798    0.000 profile.py:45(is_profiling_enabled)
 33176126    2.927    0.000    2.927    0.000 {method 'append' of 'list' objects}
   576686    0.233    0.000    1.387    0.000 {built-in method builtins.next}
   288343    0.145    0.000    1.345    0.000 contextlib.py:57(__enter__)
   576686    0.720    0.000    1.154    0.000 context_managers.py:29(sv_preferences)
  2692790    0.982    0.000    0.982    0.000 line_mk4.py:76(<lambda>)
   288343    0.958    0.000    0.958    0.000 {built-in method numpy.core.multiarray.array}
  2692790    0.872    0.000    0.872    0.000 line_mk4.py:70(<lambda>)
   288343    0.305    0.000    0.819    0.000 contextlib.py:134(helper)
   576686    0.702    0.000    0.702    0.000 {built-in method numpy.core.multiarray.arange}
     7149    0.033    0.000    0.618    0.000 socket_data.py:85(SvGetSocket)
     4766    0.021    0.000    0.554    0.000 node_tree.py:299(sv_get)
   288343    0.332    0.000    0.519    0.000 contextlib.py:63(__exit__)
   288343    0.442    0.000    0.514    0.000 contextlib.py:37(__init__)
     4766    0.221    0.000    0.493    0.000 data_structure.py:91(match_long_repeat)


     2527    0.613    0.000   78.833    0.031 line_mk4.py:352(process)
   305767    0.766    0.000   76.767    0.000 profile.py:89(wrapper)
   305767   13.525    0.000   71.786    0.000 line_mk4.py:186(make_line)
   305767    0.997    0.000   30.353    0.000 data_structure.py:939(get_edges_B)
   305767   25.037    0.000   25.037    0.000 {method 'tolist' of 'numpy.ndarray' objects}
 28555100   22.338    0.000   22.338    0.000 line_mk4.py:81(<lambda>)
   305767    1.221    0.000    4.146    0.000 profile.py:45(is_profiling_enabled)
 35180894    2.999    0.000    2.999    0.000 {method 'append' of 'list' objects}
   305767    0.935    0.000    2.811    0.000 numeric.py:1335(roll)
   611534    0.248    0.000    1.514    0.000 {built-in method builtins.next}
   305767    0.158    0.000    1.478    0.000 contextlib.py:57(__enter__)
   611534    0.798    0.000    1.265    0.000 context_managers.py:29(sv_preferences)
  2855510    1.025    0.000    1.025    0.000 line_mk4.py:76(<lambda>)
   611534    0.996    0.000    0.996    0.000 {built-in method numpy.core.multiarray.array}
   917301    0.906    0.000    0.906    0.000 {built-in method numpy.core.multiarray.arange}
  2855510    0.894    0.000    0.894    0.000 line_mk4.py:70(<lambda>)
   305767    0.320    0.000    0.886    0.000 contextlib.py:134(helper)
     7581    0.032    0.000    0.601    0.000 socket_data.py:85(SvGetSocket)
   305767    0.577    0.000    0.577    0.000 {built-in method numpy.core.multiarray.concatenate}
   305767    0.484    0.000    0.566    0.000 contextlib.py:37(__init__)
   305767    0.367    0.000    0.562    0.000 contextlib.py:63(__exit__)
     5054    0.021    0.000    0.539    0.000 node_tree.py:299(sv_get)
     5054    0.210    0.000    0.473    0.000 data_structure.py:91(match_long_repeat)
```

The percentage of time to generate edge / time to make the line:

```
26.097 / 65.959 = 39.56% - for get_edges_A
30.353 / 71.786 = 42.28% - for get_edges_B 
```

which is close to the list comprehension results. 

Honestly, I suspect it’s gonna be hard to beat the edge cache implementation. I’m not bragging, but just thinking that any conversion to other types would necessarily incur some overhead which is going to be more than just slicing a cached list (yet, I can’t claim I know enough python to say this for sure. Just my own hunch). Sometimes sacrificing some memory to gain speed may be the only way. Besides, memory is not even a big concern since even if we cache a 1 million edge list (which would likely way and beyond what any node tree would ever need to create), such a list hardly imposes any memory burden (about 8MB). A growing cache per node tree necessity I think it’s better though, than pre caching some arbitrary size.

https://stackoverflow.com/questions/43404210/how-much-memory-will-a-list-with-one-million-elements-take-up-in-python

zeffii:
by all means cache the living snot out of the things that can be cached :)

DolphinDream:
If there are no objections this is ready to land.

zeffii:
do it

PR #2324, opened 2018-12-03
New PolygonSort node

DolphinDream:
This new nodes allows to sort the polygons based on a distance to a given point or distance along a given direction. The Input is the verts/polys needed to be sorted and the reference (either a point P or a direction D). The node outputs the sorted polygons, same (input) vertices, centers of the sorted polygons, distances for the sorted polygons. The node has two modes P & D for distance to point and distance along a direction. Also there is an option to toggle the sorting ascending/descending. And yes, the node also has 3 presets X/Y/Z to preset the sorting direction to be one of the xyz axis.

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

sv-polygonsort-ui1

Example 1 : Sorting by distance to point P

sv-polygonsort-example1

Example 2 : Sorting by distance along distance D

sv-polygonsort-example2

### DIFF
--- a/docs/nodes/list_mutators/polygon_sort.rst
+++ b/docs/nodes/list_mutators/polygon_sort.rst
@@ -1,0 +1,76 @@
+Polygon Sort
+============
+
+Functionality
+-------------
+
+Polygon Sort node sorts the input polygons by various criteria: distance, area and normal angle.
+
+Inputs
+------
+
+All parameters except **Mode** and **Descending** are vectorized and can take single of multiple input values.
+
+Parameters
+----------
+
+All parameters except **Mode** and **Descending** can be given by the node or an external input.
+
++----------------+-----------+-----------+--------------------------------------------------------+
+| Param          | Type      | Default   | Description                                            |
++================+===========+===========+========================================================+
+| **Mode**       | Enum      | "D"       | The sorting direction mode:                            |
+|                |           |           |  P  : Sort by the the distance to a point P            |
+|                |           |           |  D  : Sort by the projection along a direction D       |
+|                |           |           |  A  : Sort by the area of the polygons                 |
+|                |           |           |  NP : Sort by the normal angle relative to point P     |
+|                |           |           |  ND : Sort by the normal angle relative to direction D |
++----------------+-----------+-----------+--------------------------------------------------------+
+| **Descending** | Bool      | False     | Sort the polygons in the descending order.             |
++----------------+-----------+-----------+--------------------------------------------------------+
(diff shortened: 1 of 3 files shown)


DolphinDream:
And a simple live demo.. 

sv-polygonsort-demo1

DolphinDream:
And a more intricate example / demo..

sv-polygonsort-example3

sv-polygonsort-demo2

nortikin:
what about area, other qualities? because we already have list sort with key user defined, for poligons can be added more specific options?

DolphinDream:
> what about area, other qualities? because we already have list sort with key user defined, for poligons can be added more specific options?

Sure thing. that can be done (area, normal angle etc). Btw @nortikin , can you show me an example of how you would sort the polygons by distance to point/direction using list sorting node?

DolphinDream:
Behold.. the sorting fiesta.. now sorting also by area, normal angle to point and normal angle to direction.
sv-polygonsort-ui2

DolphinDream:
Anybody knows who created the area node ? I was wondering why the unit_normal is calculated via determinant instead of simply via crossproduct of the poly first edges + normalization. What’s the catch ? I notice the normals generated via both methods are the same, albeit the cross product should be faster.

enzyme69:
I am totally intrigue by this and all the example demos...!

DolphinDream:
Here’s another demo with sorting by “Normal Angle Point”. The red indices are the original (unsorted) polygon indices, the blue indices are the sorted indices. The numbers above the poly normals are the angles between the normals and the vectors connecting the centers to the moving point. (hence the sorting name)

sv-polygonsort-demo4

DolphinDream:
And sorted by area:

sv-polygonsort-demo5

DolphinDream:
If there are no objections.. this PR is ready to land.

DolphinDream:
Final UI layout.

sv-polygonsort-ui3

zeffii:
@DolphinDream   who?
image

This is one of the older nodes :) i don't see where it would handle (concave) ngons. I would do a rewrite. I'll port it over to `sv4b280` immediately if you do.

DolphinDream:
@zeffii The polygon sort node already does area calculation so it would really be replicated code to have it in two nodes. I was wondering if it’s worth factorizing the polygon related calculations into a polygon utilities file so that its  functions can be used in multiple nodes.  And the same could be said about edge related calculations.

zeffii:
yes, factorize it out if you wish. (`sverchok.utils.modules.polygon_utils.py`) or wait till these nodes appear in sv4b28, no need to do it immediately.

PR #2333, opened 2018-12-13
Added Deformation Node

vicdoval:
Added new ~~Tension~~ Deformation node to measure variations between different vector states, it can be used to create a tension map or to create new deformations using the outputed info

tension
tension1
tension2

# Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/nodes/analyzer/deformation.py
+++ b/nodes/analyzer/deformation.py
@@ -172,7 +172,7 @@ class SvDeformationNode(bpy.types.Node, SverchCustomTreeNode):
     '''
     bl_idname = 'SvDeformationNode'
     bl_label = 'Deformation'
-    bl_icon = 'SNAP_NORMAL'
+    bl_icon = 'MOD_SIMPLEDEFORM'
 
     output_numpy = BoolProperty(
         name='Output NumPy', description='output NumPy arrays',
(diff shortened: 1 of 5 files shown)


zeffii:
neat node/code :)

vicdoval:
Im not sure if i should call this node "Tension" on "Deformation" ¿Any thoughts on this?

zeffii:
to me, it seems `deformation` might be appropriate because the idea is to pass two states of the same mesh?
If one were to just provide a single object, and calculate the stresses based on joint interaction and gravity, then I think 'Tension' would be appropriate?

up to you.

nodes/analyzer/deformation.py
```diff
@@ -107,85 +111,93 @@ def area_calc(np_verts, area_params):
...
         sinw('VerticesSocket', "Rest Verts")
         sinw('VerticesSocket', "Distort Verts")
         sinw('StringsSocket', "Edges")
         sinw('StringsSocket', "Pols")
 
-        sonw('StringsSocket', "Edges_Tension")
-        sonw('StringsSocket', "Pols_Tension")
-        sonw('StringsSocket', "Vert_Edge_T")
-        sonw('StringsSocket', "Vert_Pol_T")
+        sonw('StringsSocket', "Edges_Def")
```
zeffii: underscores not necessary here

enzyme69:
It's probably more like Tension node if it has attribute like Stretch etc. In the past I think there was an addon for Blender which kind of doing this and I use SV also to use it with.

PR #2341, opened 2018-12-23
Add Quaternion nodes

DolphinDream:
To start with, this adds 3 nodes (quaternion in, quaternion out and quaternion math nodes).

**Quaternion In**:
Provides conversion to quaternion from various format inputs (selectable via the mode option):
- components wxyz
- angle + axis
- euler angles
- matrix

**Quaternion Out**:
Provides various components (conversion) from quaternion to various components:
(w,x,y,z) components, matrix, axis + angle, Euler angle

**Quaternion Math**:
Provides the essential quaternion arithmetic operations:

ADD
SUB
MULTIPLY
DIVIDE
ROTATE
DOT
DISTANCE
INVERT
NORMALIZE
CONJUGATE
NEGATE
SCALE
MAGNITUDE

Note: some operations take multiple quaternion inputs (input list grows as new sockets are connected), as well as single quaternion inputs, or quaternion + scalar input.

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

All three nodes :
sv-quaternionnodes-ui1

Quaternion IN node:
sv-quaternionin-01

Quaternion OUT node:
sv-quaternionout-01

Quaternion MATH node:
sv-quaternionmath-01

### DIFF
--- a/docs/nodes/quaternion/quaternion_in.rst
+++ b/docs/nodes/quaternion/quaternion_in.rst
@@ -1,0 +1,64 @@
+Quaternion In
+-------------
+
+Quaternion In node constructs quaternions based on various input components provided for a selected mode.
+
+
+Modes
+=====
+
+The available **Modes** are: WXYZ, EULER, AXIS-ANGLE & MATRIX.
+
++============+================================================================+
+| Mode       | Description                                                    |
++------------+----------------------------------------------------------------+
+| WXYZ       | Converts W, X, Y, Z components into a quaternion. [1]          |
++------------+----------------------------------------------------------------+
+| EULER      | Converts X, Y, Z Euler angles and an order of rotation         |
+|            | into a quaternion. [2,3]                                       |
++------------+----------------------------------------------------------------+
+| AXIS-ANGLE | Converts an Axis & an Angle of rotation into a quaternion. [2] |
++------------+----------------------------------------------------------------+
+| MATRIX     | Converts an orthogonal 4x4 rotation matrix into a quaternion.  |
++============+================================================================+
+
+Notes:
+[1] : For WXYZ the node provides a "Normalize" option to generate a normalized quaternion.
+[2] : For EULER and AXIS-ANGLE modes (which take angle input) the node provides an
+angle unit conversion to let the angle values be converted to Radians, Degrees or Unities (0-1 range).
+[3] : For EULER mode the node provides the option to select the Euler rotation order:
+"XYZ", "XZY", "YXZ", "YZX", "ZXY" or "ZYX".
+
+
+Inputs
+======
+
+The node takes a list of various components, based on the selected mode, and it
+constructs the corresponding quaternions. The node is vectorized so the inputs take
+a value or a list of values. When multiple lists are connected the node will
+extend the length of the connected input lists to match the longest one before computing the list of output quaternions.
+
+Based on the selected **Mode** the node makes available the corresponding input sockets:
+
++============+================================+
+| Mode       | Input Sockets (types)          |
++------------+--------------------------------+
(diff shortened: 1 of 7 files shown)


DolphinDream:
As you notice.. I decided to make the Quaternion In node mode based (to switch input formats).. where the input given to the node to convert into a quaternion can have various formats (wxyz components, euler angels, axis+angle and matrix). It can be argued (based on recent conversations) that is preferred to have simpler nodes that do one, simple function, and thus perhaps make 4 “Quaternion In” nodes that generate quaternions based on different inputs. The second school of thought (which i prefer in this case) is to keep the number of nodes to a minimum by making one node with 4 modes (as proposed). If anybody feels strongly about this proposal feel free to comment, brainstorm etc.

DolphinDream:
For the Quaternion Output node I decided to show all possible components that users may want to get out of a quaternion with all outputs visible. The components are calculated of course only for the connected outputs. An alternative implementation could provide a “mode” similar to the QIN node and have the user select the kind of output they desire to read (Matrix, Components, Axis-Angle, Euler, Magnitude etc).  (though perhaps magnitude can be made part of the QMath node a separate function, similar to what the vector math does). In this case, the visible output nodes will only the ones corresponding to the selected mode. Personally I don’t see a benefit of hiding output sockets but if anybody feels strong agains it I can cleanup the output interface by adding a mode. Thoughts ?

DolphinDream:
> What about using data structure of quaternions from mathutils library without conversion to list, like we do with matrix now?

@Durman I was going to ask about it :) it is beneficial to have it as mathutils Quaternions directly an I can surely remove all the tuple()’s around the output :) it would save having to convert back and forth from list to quaternions to do Q operations. The only reason why I didn’t do it was because a Quaternion is closer relative to a Vector than a Matrix (structure wise).. and for vertices we don’t have output as Vector(0,0,0). But to your point, maybe we should have vertices also output mathutils Vector instead of list? If the Stethoscope output becomes too verbose having the extra type displayed (Vector, Quaternion, Matrix etc) then perhaps the Stethoscope node could be smarter about it and have an option to not display the type but only the value.

DolphinDream:
Some updates to all 3 nodes.. updated UI and output  is now Quaternion instead of tuples.

sv-quaternionnodes-ui2

nortikin:
ok, updated from github, find out this nodes, searched how to use it, find book on market, bought it. This chain anyway not solving my understanding of quaternions. Also book is more as math history, not math itself. But i see, we can rotate anything anyhow now.
Thanks

DolphinDream:
@nortikin Sorry for your frustration :) It's not clear yet the real use of these nodes, but I suspect we'll learn as we go and discover if they were as useful as anticipated. What book on quaternions did you get, btw? 

Also, note that there is a whole subset of nodes currently missing that are needed to complete the quaternion (4D Vector) domain, such as the 4D Matrices (5x5) which are to operate on 4D vector (quaternions) space the same way as the 3D Matrices (4x4) operate on 3D vector space. Some I already started developing (emerged from the hypercube PR I started a while back) and still need some more developing before I can push them in a PR (the nodes marked with a green bullet). Perhaps once you can operate on 4D space with more power (4D Matrices) the quaternion nodes could become more useful. 

In the meantime, of course, since normalized quaternions map to 3D rotation matrices they can definitely be used to do 3D rotations (as well as SLERP in conjunction with the MixInput nodes). 

sv-3d-4d-space

nortikin:
this book is Ian Stewart 'calculating the cosmos' and 'professor stewart's cabinet of math curiosities', not quaternions and ordered his 'symmetry' book

PR #2343, opened 2018-12-23
Add scale/multiplier to TextureViewerLite node to fix rendering

DolphinDream:
Port the same preference settings for xy/scale into the TVL node to fix the location and the size of the rendering.

- [x] Code changes complete.
- [x] Ready for merge.

Before..

sv-tvl-renderfix-before

After...

sv-tvl-renderfix-after

### DIFF
--- a/nodes/viz/texture_viewer_lite.py
+++ b/nodes/viz/texture_viewer_lite.py
@@ -24,6 +24,7 @@ from bpy.props import EnumProperty, StringProperty, IntProperty
 from sverchok.data_structure import updateNode, node_id
 from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.ui import nodeview_bgl_viewer_draw_mk2 as nvBGL2
+from sverchok.utils.context_managers import sv_preferences
 
 
 gl_color_list = [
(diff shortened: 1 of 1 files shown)


zeffii:
quite possibly we'll need to add an even handler for the `undo`, to go through the nodes in ng.nodes and see if any viewer nodes (`node_id(self)`) are present in the bgl callback pool. and drop if there are.

DolphinDream:
Btw, is it just me or the two colors (the selected color in the Color input node and the output of the TVL node) are different ? Could it be the BGL rendering is not doing the same gl composition as the CI node ? The TVL output seems a bit darker.

zeffii:
in 2.80 it seems to swing the other way, it's waay brighter. Not sure what to do about that @DolphinDream

DolphinDream:
> in 2.80 it seems to swing the other way, it's waay brighter. Not sure what to do about that @DolphinDream

I wonder if it’s related to the color management module in Blender.

DolphinDream:
any objections to merging this into master?

zeffii:
nope. work away!

PR #2347, opened 2018-12-31
Restructure MixNumber into MixInputs node 

DolphinDream:
Based on PR #1440 discussion the MixNumber node was restructured into a MixInput node that handles, in addition to the  Int/Float types, also: Vectors, Colors, Quaternions and Matrices. The UI was also slightly updated and new features added.

- The node now has a *Mode* drop-down to select from the available types: Int, Float, Color, Vector, Matrix and Quaternion. 

- The easing in/out/in_out dropdown was grouped with the interpolation dropdown to simplify the UI.

- The new SWAP option allows the inputs A/B to be swapped.

- The new MIRROR option allows the interpolation factor to be mirrored around 0.5 (e.g. a factor 0->1 will translate to 0->1->0 thus interpolating the inputs A&B form A->B->A.

- The output value name matches the selected type.

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

Mix Numbers vs Mix Inputs UI:

sv-mixinputs-mixnumbers-ui

sv-mixinputs-ui1

All mixing types:

sv-mixinputs-alltypes


DolphinDream:
A demo of the new features and UI (for **float** mixing):

sv-mixinputs-demo1

DolphinDream:
@zeffii about the old MixNumber node.. what’s the proper approach here, since the node was not only upgraded to include more type, but it was also renamed (so it’s not quite a MK# situation). Should I keep the old MixNumbers node and demote it to the "old nodes" section while adding this **new** version of the node?

zeffii:
i like it. nice work.

The IPO icons look a bit naf on that node theme, in 2.80 that will look different.

DolphinDream:
And a fun Demo :) mixing Quaternions, which is essentially SLERP (spherical linear interpolation) a nice way to interpolate rotations and do some smooth camera transformation interpolations. 

This examples shows the camera start orientation looking in +Y direction, +Z = up direction, transitioning to the end orientation looking down (-Z direction) with -Y = up direction. The Input quaternions define the camera orientation (e.g. axis & angle inputs to the quaternions). The duplicate mixing nodes are just so that the camera path is also displayed along with the interpolating the camera transformation.

The node setup:

sv-mixinputs-quaternions-slerp-setup


And the camera transition result (while changing node settings swap/mirror/easing/interpolation):

sv-mixinputs-quaternions-slerp-demo

DolphinDream:
> The IPO icons look a bit naf on that node theme, in 2.80 that will look different.

Agree... that icon was temporary..
Here’s a (hopefully) better one:

mixinputsarrows

sv-mixinputsnode-icon

PR #2352, opened 2019-01-09
Added Path Length Node

vicdoval:
Added new **Path Length** node to measure edges length, I have done a NumPy implantation that is over a 30% faster than the old scripted node "path_length" it can also work without edges.

length
length1

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


DolphinDream:
Sweet. 
I would call this Edge Length to better hint to what the node is doing. (distinguish from the List Length for instance).

Some other suggestions:
* allow poly/edge or both edge and polygon input 
* perhaps use “Separate” instead of “by edge”  to give lengths of separate edges

nodes/analyzer/length.py
```diff
@@ -0,0 +1,128 @@
...
+        if len(edges_in[0]) < 1:
+            edges_in = edges_aux(vertices)
+
+        return match_long_repeat([vertices, edges_in])
+
+    def process(self):
+        '''main node function called every update'''
+        si = self.inputs
+        so = self.outputs
+        if not (so[0].is_linked and si[0].is_linked):
```
DolphinDream: You really only care to NOT do any operation if the output is not connected. This code will actually let the node compute if the output is not connected by the input is.

nodes/analyzer/length.py
```diff
@@ -0,0 +1,128 @@
...
+
+        sonw('StringsSocket', "Length")
+
+    def get_data(self):
+        '''get all data from sockets'''
+        si = self.inputs
+        vertices = si['Verts'].sv_get(default=[[]])
+        edges_in = si['Edges'].sv_get(default=[[]])
+        if len(edges_in[0]) < 1:
+            edges_in = edges_aux(vertices)
```
DolphinDream: You could use the get_edge_list method to speed up linear edge generation :)

edges_in = get_edge_list(len(vertices))

and even (only read the input if connected) ..

if si[“Edges”].is_linked:
   edges = si['Edges'].sv_get(default=[[]])
else:
   edges = get_edge_list(len(vertices))

vicdoval: Surprisingly after some tests the actual implementation (creating the edges in Numpy) is faster than using get_edge_list and the turn the result inside one array...
Maybe if the cache was already in numpy it would be faster but i'm not sure it worth to add it to the data_structure because is already pretty fast

nodes/analyzer/length.py
vicdoval: As it is written the node will skip calculations if there is not verts_in linked. If I remove the "input check" the node will return  an error if you plug first the output until you connect the input, is a minor error but I find this way is cleaner

vicdoval:
@DolphinDream  thanks for the feedback!
About the name:
As in can work with or without edges the name "Edges Length" doesn't seem the right name for the node. The previous script doing this task was called "Path Length" witch could be a good name, but when you check the Analyzers category, next to the "Area", "Volume"," Bounding Box" seems that "Length" is more appropriate. Doesn't it?
About Polygons input:
That wound mean including the pols_to_edges function that could be handy but I don't know if it makes a lot of sense to measure the length of a polygons mesh and you could always use de "pols to edges" node
About the "by Edges":
On my view "Separate" can mislead to split in different lists, but as it can  work without edges maybe it could be more appropriate something like "by Segments" or "Segment"

nodes/analyzer/length.py
DolphinDream: I see your point. However, shouldn’t there BE an error if the inputs are not connected and you output some values for the next node? In some cases errors may be helpful in order to see what is not yet connected that should be connected.

DolphinDream:
> @DolphinDream thanks for the feedback!
> About the name:
> As in can work with or without edges the name "Edges Length" doesn't seem the right name for the node. The previous script doing this task was called "Path Length" which could be a good name, but when you check the Analyzers category, next to the "Area", "Volume"," Bounding Box" seems that "Length" is more appropriate. Doesn't it?

Even if you don’t provide an edge input the node **does** create a list of edges implicitly connecting the edges linearly, so in essence the node computes **edge lengths** whether you provide an edge list or not. Sure, you can argue that one could feed a list of random vertices that are not necessarily in a “path” order hence the automatic (somewhat arbitrary) linear edge generation (in absence of edge input) is not as obvious how this would represent path/edge length (though it actually does, albeit a very specific path made by edges connecting vertices in the order). In this case you can say there is no obvious “edge length” meaning for what the node outputs so why call it an “edge length” node.  I think the user of the node will quickly learn (via documentation and/or experimentation) that an implicit list of edges is constructed from the input vertices, and **these edges** are what the node will then use to compute the length/lengths for. You’re not computing lengths of vertices if only vertices are connected. In my view honestly I don't see a big problem with calling the node Edge or Path length. It’s up to you how you want to call it, but I would not recommend to call it just “Length” as this is too generic / ambiguous. 

And btw, there is already another node with the same filename (length.py) for the list length. I think the very least these two node files should be called list_length.py and edge_length.py or path_length.py or something even though the files are located in different paths.

> About Polygons input:
> That wound mean including the pols_to_edges function that could be handy but I don't know if it makes a lot of sense to measure the length of a polygons mesh and you could always use de "pols to edges” node

Sure you can use another node. I think there are some nodes that are takin PolyEdge as input indicating that both edges and polys would work and it’s understood that if polys are provided the node will extract the edges from the polygons, without necessarily needing to use polygons per se (e.g. normals etc).

> About the "by Edges":
> On my view "Separate" can mislead to split in different lists, but as it can work without edges maybe it could be more appropriate something like "by Segments" or "Segment"

nodes/analyzer/length.py
DolphinDream: hm.. that is surprising indeed.. as the caching should be faster than any conversion.  Did you call update_edge_cache with a max number of edges you need to get from the cache? For a vectorized case when you have multiple lists of edges you can update the cache size once with the size of the largest list, then the cache will simply return the slice of an array which should be faster than generating arrays. You can even augment the cache to an arbitrary size (e.g. 1000 edges) so any calls to get_edge_list will never augment the cache.

vicdoval: I did as you explain, calling the function once per update calling for the maximum length needed, and is still slower... maybe NumPy does some kind of memory optimisations under the hood... 🏹

vicdoval:
I renamed it to "Path Length" taking your point @DolphinDream. 😄  
About taking "polys" I left it out because when extracting the edges the user may want *unique edges* or repeated ones and adding that option to the node seems out of the nodes scope

Theses are the new images for docs:
length2
length1
length

nodes/analyzer/path_length.py
```diff
@@ -0,0 +1,131 @@
...
+
+        result.append(length if gates[0] else length.tolist())
+
+    return result
+
+
+class SvPathLengthNode(bpy.types.Node, SverchCustomTreeNode):
+    '''
+    Triggers: Path / Edges length
+    Tooltip: Masseuses the length of a path or the length of it's segments
```
DolphinDream: minor typo: its segments

vicdoval: thanks!, it was also written Masseuses in stead of measures 😆

PR #2353, opened 2019-01-12
B28 nodes fix

nantille:
## Nodes fix

Fix 1, Scale mk2: fixed operation on vectors, my shortest commit. :)

Fix 2, Objects in mk3: fixed 'Get selected' button for a nasty case. When you've got an object in the viewport and delete it, it stays in bpy.data.objects and its attribute select_get() is still set to True and so when you select another object in the viewport and hit 'Get selected', this function sometimes gets a previously deleted object that we can't access in the UI anymore. I added a check on users_scene and the new users_collection.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/nodes/scene/objects_mk3.py
+++ b/nodes/scene/objects_mk3.py
@@ -108,7 +108,7 @@ class SvObjectsNodeMK3(bpy.types.Node, SverchCustomTreeNode):
             groups = bpy.data.groups
             names = [obj.name for obj in groups[self.groupname].objects]
         else:
-            names = [obj.name for obj in bpy.data.objects if obj.select_get()]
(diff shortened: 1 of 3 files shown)


zeffii:
you might be right about the edits in move_mk2 and scale_mk2 , but i'm not sure. 

`@` is now used most `M x V` calculations:   see https://wiki.blender.org/wiki/Reference/Release_Notes/2.80/Python_API

zeffii:
cool. accepting.  ` m` is a variable name we should be reserving for matrices only. :) thanks for the PR.

nantille:
Yeah the variable was already named like that and that's why I initially thought a matrix was passed but it's just a scalar, this should be renamed. You're welcome!

zeffii:
done and pushed. Thanks!

zeffii:
try to stick to one topic per issue

PR #2359, opened 2019-01-25
Mesh Switch node

DolphinDream:
The node allows to connect multiple mesh inputs (verts/edges/polys) and to select one of them to output. Useful for changing the input mesh model to be processed by a node tree. 

The input sockets are added as new meshes are connected to allow the next mesh (verts/edges/polys) to be connected.

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

sv-meshswitchnode-ui

sv-meshswitchnode-demo1

### DIFF
--- a/docs/nodes/modifier_change/mesh_switch.rst
+++ b/docs/nodes/modifier_change/mesh_switch.rst
@@ -1,0 +1,35 @@
+Mesh Switch
+-----------
+
+This node allows to switch among an arbitrary number of mesh inputs (verts, edges, polys) connected to the node.
+
+Inputs
+======
+
+**Verts A**
+Takes a Vertex list from outside nodes.
+
+**EdgePolys A**
+Takes either an Edge or a Poly list from outside nodes.
+
+The node starts with one set of mesh input sockets ("Verts A" and "EdgePolys A") and will create new mesh sockets whenever the last Verts input socket is connected, thus allowing additional mesh inputs to be connected to the node. Whenever the last Verts input socket is disconnected the node will remove the all disconnected inputs at the end of the input list except the last disconnected set of inputs, thus always allowing one extra mesh to be connected to the node.
+
+Parameters
+==========
+
+The **Selected** parameter accepts single input values directly from the node or from an outside node. The value is sanitized to be bounded to the range of the connected inputs (0, N-1) and it is also converted via a modulo-N function to wrap around within (0, N-1) range for values larger than N. In other words, as the **Selected** value increases the node essentially cycles through the connected meshes.
+
++---------------+---------+---------+-------------------------------------------------+
+| Param         | Type    | Default | Description                                     |
++===============+=========+=========+=================================================+
+| **Selected**  | Int     | 0       | The index of the selected mesh to output.       |
++---------------+---------+---------+-------------------------------------------------+
+
+Outputs
+=======
+
+**Verts**
+**EdgePolys**
+
+When an output socket is connected the node will output the mesh corresponding to the selected index.
+
--- a/index.md
+++ b/index.md
@@ -86,6 +86,7 @@
     PolygonBoomNode
     Pols2EdgsNode
     SvMeshJoinNode
+    SvMeshSwitchNode
     ---
     SvBevelNode
     SvSubdivideNode
--- a/nodes/modifier_change/mesh_switch.py
+++ b/nodes/modifier_change/mesh_switch.py
@@ -1,0 +1,103 @@
+# ##### BEGIN GPL LICENSE BLOCK #####
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software Foundation,
+#  Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# ##### END GPL LICENSE BLOCK #####
+
+import bpy
+from bpy.props import IntProperty
+
+from sverchok.node_tree import SverchCustomTreeNode
+from sverchok.data_structure import updateNode
+
+ABC = tuple('ABCDEFGHIJKLMNOPQRSTUVWXYZ')
+
+
+class SvMeshSwitchNode(bpy.types.Node, SverchCustomTreeNode):
+    """
+    Triggers: Mesh, Switch, Select
+    Tooltips: Switch among multiple Verts/Edge/Poly inputs
+    """
+    bl_idname = 'SvMeshSwitchNode'
+    bl_label = 'Mesh Switch'
+
+    selected = IntProperty(
+        name="Selected", description="Selected Mesh",
+        default=0, min=0, update=updateNode)
+
+    def update(self):
+        inputs = self.inputs
+        if inputs[-2].links: # last Verts socket linked ? => create more sockets
+            name = ABC[int(len(inputs) / 2)]  # pick the next letter A to Z
+            inputs.new("VerticesSocket", "Verts " + name)
+            inputs.new("StringsSocket", "EdgePolys " + name)
+        else:  # last Verts input unlinked ? => remove all but last unlinked
+            # get the list of Verts socket labels (linked or unlinked)
+            vertSockets = filter(lambda s: "Verts" in s.name, inputs)
+            socketLabels = [s.name[-1] for s in vertSockets] # (ABCD)
+            # get the labels of last unlinked Verts sockets in reverse order
+            unlinkedSocketLabels = []
+            for label in socketLabels[::-1]: # in reverse order (DCBA)
+                if inputs["Verts " + label].is_linked:
+                    break
+                else:
+                    unlinkedSocketLabels.append(label)
+            # delete all unlinked V-EP inputs sockets (except the last ones)
+            for label in unlinkedSocketLabels[:-1]:
+                inputs.remove(inputs["Verts " + label])
+                inputs.remove(inputs["EdgePolys " + label])
+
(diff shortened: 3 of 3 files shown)


DolphinDream:
When testing a particular node tree on a given mesh I often times needed a way to be able to change the mesh to see how the nodetree works on different models. For this I usually had to connect and disconnect the input verts/edges/polys.. hence the need for such a node where I could connect all the different meshes to the node once.. and be able to select through the meshes via a single input :) 

Any thoughts, suggestions, critique are appreciated as usual. 

It is perhaps arguable whetheer such node would necessarily need both the edges and polys sockets or whether a unified poly/edge socket would be sufficient. Having 3 sockets created for every connected mesh could grow the node tall pretty fast.

DolphinDream:
Latest UI (EdgePoly input socket version)

sv-meshswitchnode-ui2

DolphinDream:
And a frame driven switch demo :) 

sv-meshswitchnode-demo3

DolphinDream:
Btw, I added this node to the "modifier change" category. If there's a better category this should go into let me know.

enzyme69:
screen shot 2019-01-26 at 15 33 32

Kind of wish it can do this but then again .... List Join can do it?

enzyme69:
[mesh_switcher_004.blend.zip](https://github.com/nortikin/sverchok/files/2798786/mesh_switcher_004.blend.zip)

enzyme69:
At the moment Monad seems to be the answer :D Still pretty magical every time I do this.

screen shot 2019-01-26 at 15 46 55

DolphinDream:
@enzyme69 you mean.. if Mesh Switch node can do this? 😃 

sv-meshswitchnode-demo4

Well.. now it does :) The *Selected* input needed to take multiple inputs. I didn't think it needed to output more than one selection, but it makes sense for this sort of applications.

vicdoval:
@DolphinDream  Cool node!
 An idea could be to expand/ generalise the node  to be the new switch node.  Right now the Switch node in the  'single' mode you can do the same task but just with two inputs groups. It would be great to combine the socket flexibility expanded to more groups as this node does and with vectorized selection
switch

DolphinDream:
@vicdoval Hm.. This switch node is a bit confusing to me. Why is it "single" has multiple set of inputs and "multiple" has only one set of inputs? :) 
You're right.. this node could be generalized to take arbitrary type of inputs. As a matter of fact even now you can feed any type of inputs to the nodes (though perhaps it may not work fine with matrixes I suspect). I'll give it another try to make it more general.

DolphinDream:
Here are some other improvement ideas:  (feel free to add yours)

* Change the name of the outputs to the name o the selected input (e.g. Alpha, Beta etc), unless the Selected parameter is a list of selections in which case the output socket names can stay "Data"
* Color the input socket nodules differently better visually separate the input sets 
* automatically link the output node sockets of a node to a set/bunch of input sockets of the switch node. Whenever the first output socket of a node is linked to the first input socket in a bunch, the second socket of the input node could also be automatically connected to the second input in the switch node in the same bunch (assuming the bunch number is at least two. The generators, in particular, could benefit from this since they usually have verts, edges, polys output sockets in this order.
 
Btw can we bet the Node Wrangler type of functionality with SV nodes?  Like Ctrl+Cmd+LMB drag from one node to the switch node could automatically ask what output sockets to link (V,  VE, VP, VEP) etc.

DolphinDream:
Here's what it looks like for each input set/bunch to get its own socket color. (tried to keep the colors a bit different than the ones already designated for the other node types).

sv-meshswitchnode-demo6

DolphinDream:
Btw, how does the "right click menu" command links the node to the viewer draw nodes ? I get the option when i CMD+click on the node but it only spawns the viewer draw node without connecting the sockets. Does this option look for a vertex socket in the input node to connect to the VD input verts socket?

DolphinDream:
Btw.. I discovered that VDMK2 node has some issues .. when you feed it a list of meshes that are not all vert/poly or vert/edge. :) I will post that as a separate thread. And it's not a switcher node related, I even replicated that with regular list join nodes.

DolphinDream:
Anyone knows (@zeffii) how to insert new input sockets without disconnecting the already connected sockets? E.g. If the switch node has two inputs per set (bunch = 2) and the first inputs in each set is connected (e.g. verts and edges inputs are both connected) when increasing the number of inputs to 3, each set would need to have a third socket added to the set, but I'd like to be able to keep the already connected sockets (1&2) connected. 

Essentially go from:
```
Alpha 1  * (connected)
Alpha 2  * (connected)
 Beta 1  * (connected)
 Beta 2  * (connected)
```

to
```
Alpha 1  * (connected)
Alpha 2  * (connected)
Alpha 3    (unconnected, newly created)
 Beta 1  * (connected)
 Beta 2  * (connected) 
 Beta 3    (unconnected, newly created)
```


One way I can think of doing this (if possible) is to first query the nodes/sockets connected to each input sockets of the switch node.. then delete all input sockets (of the switch node), recreate them with larger number of inputs per set, and then reconnect the previously connected sockets of the external nodes to their sockets they were previously connected to. 

Another way, though I don't think this is possible, is to directly INSERT sockets in between other sockets already created in the input sockets variable (which seems to be a prop collection).

DolphinDream:
Example of the desired behavior when increasing the set size:

sv-inputswitchnode-demo2

zeffii:
i do not like the idea of arbitrary colors at all. for input switch... so.. without lookingthe implementation, i'm not in favour. There must be some other way to differentiate the socket grouping.

DolphinDream:
> i do not like the idea of arbitrary colors at all. for input switch... so.. without lookingthe implementation, i'm not in favor. There must be some other way to differentiate the socket grouping.

OK.. any other ideas how to better differentiate between sets?

And how about the last issue i'm having about inserting sockets without disconnecting others?

zeffii:
because we have limited options to control the layout of a socket ui, beyond color.. you could make a new dummy-like socket with an entirely transparent nodule, that would act as a faux-separator between sets. 

regarding inserting sockets, it might be that the pynodes API is limited in our high need for dynamic ui.. i do not have a fast answer :/

DolphinDream:
With a dummy fully transparent socket you'd still be able to connect to it, right? If so that would be a bit confusing, I think.

I would be cool if you could change the background color of a socket to make the group of sockets be like alternating row colors of sort.

zeffii:
you can implement code that automatically disconnects any input into the 'separator' socket,, if that's a real issue...  or automatically connect the link to the next appropriate socket.

it would be nice to be able to set the socket+label background colors, sadly it's not an option in the layout engine. but imagine for a moment..

```
  | switcher                        o |
  |                                 o |
  |                                 o |
  | [ link selected node ]            |     <-- operator in red, using "layout.warning"
  | o                                 |
  | o                                 | 
  | o                                 |
  | [ link selected node ]            |    
  | o                                 |
  | o                                 | 
  | o                                 |
```

zeffii:
```python
    row = self.layout.row()
    row.alert = True
```
anything operator buttons on that row will be a different "warning" color.

zeffii:
there's also `socket.link_limit` , which can be set to `0` - never tried it, but seems useful

DolphinDream:
@zeffii cool. Thanks for the suggestions. Unfortunately, I don't find this hack appealing. I tried to see if layout.separator() inserted in the socket drawing level can space the sockets apart but I can't see any effect. maybe I 'm not putting it in the right place. 

Another idea would be to add custom icons before socket labels? (alternating between white/black or something).

sv-socketgrouping-01


Ideally, I would like to have something like this (if possible) from left to right in the increasing order of preference.

inputswitch-backgroundcolors123

DolphinDream:
Would alternating green colors for sockets, with slightly different shades of green be acceptable? 

sv-socketgrouping-02

DolphinDream:
@zeffii how could I remove a link automatically? I'm creating a faux SvSeparatorSocket and i want the socket to be disconnected if you try to connect to it.

```
class SvSeparatorSocket(NodeSocket, SvSocketCommon):
    ''' Separator Socket used to separate groups of sockets '''
    bl_idname = "SvSeparatorSocket"
    bl_label = "Separator Socket"
```

Also.. can you make the socket completely transparent? Including the stroke? 

sv-separatorsocket

DolphinDream:
@zeffii I added this to my node's update method and it does the job of detaching any connection to Separator socket however I wonder if this could/should be done instead in the node_tree in the SvSeparatorNode class itself (if so what method?)

```
       for s in self.inputs:
            if s.bl_idname == "SvSeparatorSocket":
                if s.is_linked:
                    print("separator socket is linked")
                    for link in s.links:
                        s.id_data.links.remove(link)

```

zeffii:
the more i think about this, the more it feels more appropriate to just keep it simple and use conventional sockets. which are sufficient to act as indicator of the group.

```
VerticesSocket   | verts a
StringsSocket    | edges a
StringsSocket    | faces a 
VerticesSocket   | verts b
StringsSocket    | edges b
StringsSocket    | faces b 
VerticesSocket   | verts c
StringsSocket    | edges c
StringsSocket    | faces c 
```
but i get your position, for this particular node the default ui/layout features are showing their sparsity.

zeffii:
and if you really have to, insert the SvSeparatorSocket inbetween sets. 

as for the update function, you could stick most of the working code for the disconnection into the socket definitino, and call it from within the node update function.

DolphinDream:
I like the idea of putting the disconnection code in the SvSeparatorSocket class but unfortunately this is not going to be a lot of code since in my node I would still have to search for each socket and only call the disconnection code (two lines of code) if the socket is a separator type and is it’s connected. A more effective way (if possible) would be if the socket received a call whenever it’s connected and let the socket class automatically handle the disconnection without being prompted by the node. I assume this would be done at the lower level infrastructure.

zeffii:
since the socket class doesn't have an `update` function, you might settle for 

```python
# ... socket class...
def remove_links(self):
    for link in self.links:

# ... in the node class ....
def update(   ):
    ... other stuff ...
    _ = [s.remove_links() for s in self.inputs if s.bl_idname == '....']
```

nortikin:
right side without colors rainbow. only boxes shading
> inputswitch-backgroundcolors123

portnov:
@DolphinDream , @Durman In which way this is supposed to be in 2.8x? Considering there is #2671 issue.

DolphinDream:
This node was replaced by the Input Switch node So i don’t think it has to be ported to 2.8

Issue #1538, opened 2017-04-18
Macro examples


zeffii:
```
"""

This file demonstrates a few features that can be used in a macro

"""
import bpy

def domo_esta(self, context):
    """ info about domo esta /// will ignore this"""
    active_node = context.active_node
    print('mouse:', context.space_data.cursor_location)
    if active_node:
        print(active_node.name)
        print(active_node.location)
        print(active_node.bl_idname)


def domo_esta4(self, context, a=10, b=140):
    """ info about domo esta4 ///"""
    print(a, b)


def windows_console(*args):
    """ show / hide console ///"""
    bpy.ops.wm.console_toggle()

```

## Some tips

in macros.py you can do any regular importing, but i recommend not importing functions directly unless you want them to be visible in the extended search menu.

```diff
- from math import floor, roof, swimmingpool
+ import math   # math won't show up in the menu because it isn't 'callable'
```
so
```diff
def some_macro(self, context):
    """ some short description ///  
    longer description if needed
    """
-    print(floor(45/2))
+    print(math.floor(45/2))
```

zeffii:
@kalwalt  i'll show how to switch the current vdmk2 node into a bmeshviewer node..

zeffii:
```python
def swap_vdmk2_bmeshview(self, context):
    """ hotswap vd <---> bmview """
    active_node = context.active_node
    if active_node and active_node.bl_idname == 'ViewerNode2':
        loc = active_node.location[:]

        tree = context.space_data.edit_tree
        nodes, links = tree.nodes, tree.links

        bmv = tree.nodes.new('SvBmeshViewerNodeMK2')
        bmv.location = loc

        a_inputs = active_node.inputs  
        links.new(a_inputs[0].links[0].from_socket, bmv.inputs[0])   #verts
        links.new(a_inputs[1].links[0].from_socket, bmv.inputs[2])   #edgepol
        links.new(a_inputs[2].links[0].from_socket, bmv.inputs[3])   #matrix

        nodes.remove(active_node)
        
```
![freisland](https://cloud.githubusercontent.com/assets/619340/25150874/c2298850-2483-11e7-851b-98f028da1490.gif)

zeffii:
that's the general idea, a lot more potential error checking can be done , and it could even detect if the vdmk2 edgpol connection is coming from a socket called edges, it would hook that into the edges socket of bmviewmk2.   Currently the macro is doing the bare minimum.

zeffii:
[...]

PR #2367, opened 2019-02-09
New Input Switch node (WIP)

DolphinDream:
Add new InputSwitch Node (WIP)

This node is based on the MeshSwitch node (PR #2359) and generalizes it to take **any input type** (not just verts/polyEdge). The new node also provides additional features:

- variable input set size
- keeps prior connections when changing the set size
- add separators between input sets
- add distinct labels to input sets (greek letters. alpha, beta etc)
- vectorizes the selected parameters (generates multiple output sets if selected gets an array of integers)

- New SeparatorSocket was also added to the set of socket types to help with the UI socket visual separation (e.g. group multiple sockets into sets).

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

sv-inputswitch-ui1


DolphinDream:
And a lil demo...

sv-inputswitchnode-demo5

DolphinDream:
And another demo.. driving the selection with a timer.

sv-inputswitchnode-demo3

DolphinDream:
And a vectorization demo...

sv-inputswitchnode-demo6

DolphinDream:
Can anybody recommend where should I learn more about how to create macros? I want to be able to select a bunch of generator nodes and trigger a macro that connects all generator VEP outputs to the input switch node.

DolphinDream:
**Wishlist** (I'd appreciate any recommendation how I can do any of these):
 
* be able to hide the SeparatorSocket 100% (I don't know how to make the socket outline transparent)
* make macro to connect selected generators to this node 
* right click menu to connect to a viewer draw

DolphinDream:
@enzyme69 can you check out this node and see what you like, not like, works, not works? gracias

Any feedback is appreciated as usual.

zeffii:
remind me why you aren't using verts sockets ?

zeffii:
i suppose it's for switching any kind of data then?

i think this looks reasonable, given the UI layout limitations. You could if you really wanted to streamline the UI make the node hide the separator sockets when the node is not active.  The ExecMod (in scripts) node does this.

DolphinDream:
A would be macro demo mockup ... (I Sill need some hints where to look for starting to create such a macro)

sv-inputswitchnode-demo7

zeffii:
see https://github.com/nortikin/sverchok/issues/1538  ,

zeffii:
i'd be happy to implement the macro, after you commit this.
(i'll even push it to b28 for you)

zeffii:
how about "sw" as the macro shortcut..

zeffii:
but, you try first :)  check out the default macros (issue tracker shows some good links if you search  macro  )

DolphinDream:
Thx @zeffii . I already started adding this to the macros.. :) It's looking good. I agree with the shortcut. I'll have to figure out how to do that. So far I use the search and "> gen switch".

Here's the demo so far

sv-inputswitchnode-demo8

zeffii:
cool `> blah  ` is necessary . and the other of macro connection is determined by Y position order? or X (left to right order)

DolphinDream:
@zeffii You're right. The order of connection should be smarter. I tried sorting the selected nodes by Y (reverse order since +Y is bottom to top and it seems to work better. 

Perhaps an even smarter algorithm would sort by +Y first, then by +X if the nodes happen to be at the same Y location (e.g. horizontally distributed and top aligned nodes). 

sv-inputswitchnode-demo9

DolphinDream:
Some considerations opened for brainstorming & debate related to the macro:

How to connect input nodes to Input Switch node considering that you would want to connect multiple single output nodes (eg.  scalar values like math nodes, or vectors like the vectors math nodes etc) or maybe you want to connect multiple generator nodes which have 3 relevant outputs (VEP). How should the macros handle these situations or mixed nodes situations ?

* For generator nodes (VEP) should the macro ask for which outputs to connect? (maybe a modal dialog asking for a number or list of numbers ?) so that it can connect either V & E, V & P or all three? 
Right now I'm only connecting outputs 0 & 2 from selected nodes (assumed to be generators) to the input 1&2 in the next set of the InputSwitch node.

* What if you select mixed nodes? generators and other  type of nodes that may not have VEP outputs? Should the macro fail to connect? should it try its best to connect them (even though it may not make sense to have VEP and scalar inputs mixed.

* Should the ViewerDraw node be created and connected **only** if all selected nodes are generators thus guaranteeing to have the VEP values needed by the VD?

* Is it possible for the macro to be triggered by a shortkey? so you don't have to do a search for "> sw" ? 

* Should the macro handle the situation in which new set of generators need to be connected to an existing InputSwitch node which may already have other generators connected ? i.e. if the user selects new generator nodes AND one InputSwitch node the macro can detect that an IS node is selected so it will not generate a new one but append the generator nodes outputs to the IS node.

* other considerations anyone else may have ...

zeffii:
>  * Is it possible for the macro to be triggered by a shortkey? so you don't have to do a search for "> sw" ?

i don't consider it worth the extra hassle to bind a kb combo to an operator to a macro trigger.  it's useful to weigh the effort required to do shift+A -> `> sw` , versus hooking all the sockets manually.

> * Should the macro handle the situation in which new set of generators need to be connected to an existing InputSwitch node

consider if you need that, that you can use the `node.active` vs `node.select`, and have variations of the macro according to these qualities.

DolphinDream:
Pushed the "> sw" macro .. check it out! :)

enzyme69:
@DolphinDream Do I need to change to test branch or download latest and reinstall the zip? Will check...

DolphinDream:
I’ll push it to the same branch upstream so you can just change branches in the main repo instead of my repo. (Assuming you have your SV addon installed as a repo and you can change branches in the terminal ). Downloading the zip and installing works too.

DolphinDream:
BTW, I limited the input set size to 7 :) Increasing the set size can potentially grow the node really tall if multiple sets are connected. I don't anticipate input sizes larger than 7 to be needed, but who knows. Do you guys think this cap is rather artificial and should be removed?

enzyme69:
screen shot 2019-02-10 at 10 53 09

So far so good! 

I am testing with just random numbers with different length. I understand the "Selected" but "Set Size" option needs further investigating. I will read your example above first...

enzyme69:
Ok I see, this is pretty cool "Switch" node.
screen shot 2019-02-10 at 10 57 46

DolphinDream:
> screen shot 2019-02-10 at 10 53 09
> 
> So far so good!
> 
> I am testing with just random numbers with different length. I understand the "Selected" but "Set Size" option needs further investigating. I will read your example above first...

The "set size" is the number of inputs in a set. If set size is 2 there will be Alpha 1, Alpha 2, then next set is Beta 1, Beta 2 ..etc.. for a set size of 3 you get Alpha 1, Alpha 2, Alpha 3 in the first set, etc.

DolphinDream:
Btw, I wonder if this node is versatile enough to obsolete the Switch and Switch MK2 nodes. If not, maybe it can be further augmented to do more switching stuff. 

 Switch MK2 is not yet integrated as I understand from Durman?  https://github.com/nortikin/sverchok/blob/ef5de98b4a32548e56994cf69b444361af7cd853/docs/nodes/alpha_nodes/switch_MK2.rst

utils/sv_default_macros.py
```diff
@@ -71,7 +74,7 @@ class DefaultMacros():
     @classmethod
     def ensure_nodetree(cls, operator, context):
         '''
-        if no active nodetree
+        if no active_no nodetree
```
zeffii: ??

DolphinDream: I noticed too. i'll fix in next push. thx.

zeffii:
what program/settings do you use to make these animated gifs?  they really crush my machine.

DolphinDream:
> what program/settings do you use to make these animated gifs? they really crush my machine.

I use LICEcap v1.28 .. sometimes I use 12fps sometimes 24 fps. no particular settings (It doesnt' have any really). Do all GIFs I make crash your system? or just the ones on this PR ?

vicdoval:
Another feature could be implemented the autotransformation of the input and output socket type when connected, as in the "list join" node or the "list match" node. I really like this feature because helps also to visually sorts the outputs.

vicdoval:
Personally l like to use:
        '''draw buttons on the N-panel'''
To add advanced features.
Some of them could be:
-Hide/Show separators.
-VEP/Greek naming.
...

zeffii:
>  Do all GIFs I make crash your system? or just the ones on this PR ?

some images in the PR do drastically cripple the browser, especially when most comments contain a gif.   the issue is not gifs in general, but some gifs are encoded in a way that maximizes compression (heavier to process by the viewer) over file size (larger download, but lighter to process).

DolphinDream:
> Another feature that could be implemented is the autotransformation of the input and output socket type when connected (as in the "list join" node or the "list match" node).

Thanks for the feedback, @vicdoval 

Are you suggesting the sockets to change their types based on what input is connected to? Right now the input/output sockets are all StringsSockets so they can take any input and output it as is, without changing the socket type. Is there a real need for the sockets to change type? Other than the visual indicator the only other situation in which I can think the socket type may be important is for the socket autoconversion (e.g. matrix input can take vectors and thus it can auto-convert the input from a vertices socket to matrix, and it may fail to convert if the incoming input socket  is a strings socket.

One caveat though is when you feed the **selected** a list of integers (not just one value). in this case the output may have a list of mixed values. E.g.

for this input:

Verts ->  Alpha 1
Edges -> Alpha 2

Floats -> Beta 1
Colors -> Beta 2

selected = [0, 1] 

the output is:

Data 1 -> [[Verts], [Floats]]
Data 2 -> [[Edge], [Colors]] 

in this case, you cannot have the output sockets be of a particular type since the lists contain mix type data.

Any ideas about what to do in this case ?

DolphinDream:
@vicdoval 
> Some of them could be:
> -Hide/Show separators.
> -VEP/Greek naming.

We could definitely add these to the Property Panel.  Some consideration I see:

The hide/show separators should be fine to add (with the default set to show them). Not having separators helps with the vertical spread of the node, but it also makes it hard to differentiate between sets, a feature which I think it's particularly important when you manually connect inputs to the node and you need to know which set you 're connecting to. You really have to try it to know what I'm talking about. :) 

As for the VEP/Greek naming, this may be a problem since this node is a generalized switch node (compared to the previously introduced MeshSwitch) and may not necessarily take just VEP inputs. Naming the inputs VEP could be misleading if the user connects non VEP inputs, or if already connected non-VEP inputs and decides to switch the labels from Greek to VEP.  Also, the user may want to have a size **two** sets with V & P. .. or V & E only, but not necessarily VEP. Originally when I started this node I chose these inputs labels to be just "Data" .. but I found it hard to differentiate between sets. I also needed to differentiate between inputs within sets (via numbers), so one way would have been to have the labels: [Data A1, Data A2].... [Data B1, Data B2], but this approach is kind of hard to differentiate given that all inputs start with word Data and you have to check the letter-number combo at the end to know which set you're in when you connect, hence my decision to go with the Greek letters. 

I'm not opposing the VEP / greek swap feature.. but we need to clearly understand its limitations.
What if the set size is 4? what would be the "VEP" labels in this case?
1 Verts
2 Edges
3 Polys
4 SomethingElse? 

What if the set size is 2?  What should be the "VEP" labels in this case?
1 Verts
2 Edges ?

or 

1 Verts
2 Polys  ?

vicdoval:
The naming feature may be too much hassle, maybe it is better to leave with the Greek letters or maybe just A1, A2...B1, B2... wich is shorter.
About the socket conversion:
I can't think of any useful case of having different lists types on A1 and B1 if the output is going to be connected to the same place  (except with pols and edges).
As I see it once you plug the first A1, all the N1 sockets should be converted to this socket type and the first Output socket too.
Maybe this closes some of the possibilities but seems clearer and I think it helps to keep things tidy.

DolphinDream:
@vicdoval  Using Latin letters instead of Greek letter labels it's easy to do, though I must say, I did try this too and (IMO) it's a lot easier to differentiate between words than it is between letters.. See for instance this list:

```
A 1
A 2
A 3
B 1
B 2
B 3
```

versus

```
Alpha 1
Alpha 2
Alpha 3
Beta 1
Beta 2
Beta 3
```

I personally find it easier to distinguish between Alpha 3 and Beta 1 to see where the set boundary is compared to A3 B1. It's a pattern recognition thing I believe; more letters in the label from one label to the next makes it easier to see the contrast compared to one letter comparison. Again, this may not be true for everybody. My goal here with the Greek labels (and separators) was really to make it as obvious as possible to distinguish between sets. Maybe if we keep the separators having ABC labels would not be as bad.

DolphinDream:
Here's a side by side comparison:

latin-vs-greek-labels-separators

DolphinDream:
One more thing about the separators that I found useful is that it spreads the link lines vertically instead of all of them being packed together in a closed space (e.g. if you don't have the separators the connections are closer to one another). Having separators makes it easier to tell the link lines apart. But again, this may be just my personal impression based on my playing with the node in various setups and with/without separators.

DolphinDream:
Note that the last push adding the latin/greek label PP feature does not work correctly (it doesn't preserve prior connection). I'll be fixing this shortly. 

Also, it seems that if there are gaps in the connected sets, changing the set size also doesn't handle preserving connectivity well (it drops connections) (that's because when set size changes the sockets are first all removed and then recreated and due to the way new sets are being automatically created on socket connection it's not possible to create unconnected sets within other connected sets).

DolphinDream:
Btw, @zeffii about animated GIFs.. I wish github would allow linking mp4 video formats as well. (that way you would only start to view the animation you care about instead of the browser automatically starting to play them. I like the movie format better also for the quality (in some cases desirable over the animated gifs).

DolphinDream:
Another demo illustrating the show/hide separators and greek/latin labels. I also fixed connectivity preservation when changing the set size.  In addition, when changing the set size, show/hide separators and toggle latin/greek labels any unconnected set in the middle of other connected sets are removed from the node.

sv-inputswitch-demo10

DolphinDream:
I'm considering changing the behavior of spawning a new set only when the first input to a set is connected.. and instead have a new set be spawned if any input to the previous set is connected (e.g. if any input into a set is connected the set is considered to be connected, hence a new set is spawned at the end). Any objections to this?

DolphinDream:
@zeffii when I disconnect input sockets (cut it or pull it out) the update method is being called twice.. why?

zeffii:
update gets called an excessive amount of times, and depends on the UI interaction. That's why we moved away from using `update` for most of the node/nodetree processing. Best you can do is to devise a way to reduce the amount of work you do blindly in `update`.

don't get too upset about the occasional double update call, remember it's not happening constantly only infrequently and only when you manually interact with the socket. (which is not frequent...in the lifetime of the nodetree)

zeffii:
but you could possibly freeze the nodetree prior to removing the links, it think that might be causing an additional update

DolphinDream:
> but you could possibly freeze the nodetree prior to removing the links, it think that might be causing an additional update

How do you "freeze the node tree" ?

zeffii:
@DolphinDream  grep the repo for the use of `freeze`

```python
node.id_data.freeze(hard=True)
try:
    # do stuff that might trigger updates here
except:
    # print some message

# definitely make sure it unfreezes
node.id_data.unfreeze(hard=True)
```

vicdoval:
@DolphinDream  Did you left out the socket auto-conversion? Many of the List category nodes does it and i think it would be more consistent.
Maybe in the cases when there are different data types it could stay green, but it would be great for vector an matrix difference. :)

vicdoval:
> Btw, I wonder if this node is versatile enough to obsolete the Switch and Switch MK2 nodes. If not, maybe it can be further augmented to do more switching stuff.
>  
In order to obsolete Switch the node would need a` level ` slider as in the `multi `mode the switch is done by element (level 2). I think this feature would be very handy
The improvements from the mk2 that this node does not cover are (form [docs](https://github.com/nortikin/sverchok/blob/ef5de98b4a32548e56994cf69b444361af7cd853/docs/nodes/alpha_nodes/switch_MK2.rst ) ): "Flatten list"," Generation of bool" ,and "Using as filter."
On my view they are cool but not so related to the node.
Also mk1 and mk2 had also output socket auto-conversion.

zeffii:
`changable_socket` is cool, but isn't a must ...

search
```python
changable_sockets(node, inputsocketname, outputsocketname)
````

**Comment by DolphinDream** [comment]
> `changable_socket` is cool, but isn't a must ...

Thanks or the tip, however, this method seems to reorder the sockets.. if I have 2 output sockets and do this on the first socket, the first socket gets added to the end of the socket list. I actually ended up using replace_socket instead.

**Comment by DolphinDream** [comment]
> 
> In order to obsolete Switch the node would need a`level` slider as in the `multi `mode the switch is done by element (level 2). I think this feature would be very handy.
> 
> The improvements from the mk2 that this node does not cover are (form [docs](https://github.com/nortikin/sverchok/blob/ef5de98b4a32548e56994cf69b444361af7cd853/docs/nodes/alpha_nodes/switch_MK2.rst) ): "Flatten list"," Generation of bool" ,and "Using as filter."
> On my view they are cool but not so related to the node.
> 
> Also mk1 and mk2 had also output socket auto-conversion.

Thanks for the reminder, @vicdoval . I'm working on a socket auto-conversion. It's a lot more complicated than it seems. The sockets need to be updated on connect/disconnect the inputs as well as when changing the selection.. and also on connect/disconnnect selection socket (since in this case the output is a potentially a mix type and the sockets must be made generic/StringSockets). It's almost working right now aside from some insidious segfaults I'm trying to track down which I suspect have something to do with the replacing sockets (to change type) while perhaps not properly deactivating updates (the freeze/unfreeze doesn't seem to help though). We'll need to do a thorough stressing of this node to make sure is bug free.

**Comment by DolphinDream** [comment]
if anyone wants to pull the upstream inputSwitchNode and try it out that would be awesome. :) I attached the blend I used to trigger the segfault. I usually swipe the SetSize back and forth continuously for a while.. then try to cut the last input link, or all input links (CTRL + mouse drag to cut links). This helps trigger some race condition that triggers the segfault. The code has a lot of prints added to help identify where the problem is so it's kind of messy right now. If you feel lucky you can dig into the code to see if you can spot an obvious mistake (you can cleanup the prints locally to see the actual code easier).

sv-inputswitch-bug

[inputSwitch-Dev4.blend.zip](https://github.com/nortikin/sverchok/files/2873082/inputSwitch-Dev4.blend.zip)

**Comment by DolphinDream** [comment]
Is the SV node code single thread? can I assume that only one method in the node code is executed at any one time? (I'm having a hard time trying to figure out why the  code seg-faults). It seems to execute everything properly. :frow  There doesn't seem to be any out of bounds array access or exceptions thrown. Maybe this needs to be debugged with a GDB debugger in conjunction with blender code. Any ideas how can I go about digging into finding the root cause of this segfault?

**Comment by enzyme69** [comment]
Yes, I am getting segmentation fault also...
```
+++++++ TIC inputs_in_set: 2
++++++++++++++ TOC inputs_in_set: 2

++++++++++++ TOC is_set_linked: 2  :  True


++++++++++++ TIC is_set_linked: 3

++++++++++++++ TIC inputs_in_set: 3
++++++++++++++ TOC inputs_in_set: 3

++++++++++++ TOC is_set_linked: 3  :  True


++++++++++++ TIC is_set_linked: 4

++++++++++++++ TIC inputs_in_set: 4
++++++++++++++ TOC inputs_in_set: 4

++++++++++++ TOC is_set_linked: 4  :  True


++++++++++++ TIC is_set_linked: 5

++++++++++++++ TIC inputs_in_set: 5
++++++++++++++ TOC inputs_in_set: 5

++++++++++++ TOC is_set_linked: 5  :  False

++++++++++ TOC number_of_linked_sets: 5

++++++++ set_size: 2
++++++++ inputs: <bpy_collection[19], NodeInputs>
++++++++ outputs: <bpy_collection[2], NodeOutputs>
++++++++ input size: 19
++++++++ output size: 2
++++++++ index =  0

++++++++++ TIC: label_of_set: 0
++++++++++ TOC: label_of_set: 0  :  Alpha

++++++++ inLabel:  Alpha 1
++++++++ outLabel:  Data 1
++++++++ inSocketType:  SvColorSocket
++++++++ TIC replacing: 1  of  2
++++++++ replace : Data 1
++++++++ replacing out socket type with:  SvColorSocket
++++++++ TOC replacing: 1  of  2
++++++++ inLabel:  Alpha 2
++++++++ outLabel:  Data 2
++++++++ inSocketType:  VerticesSocket
++++++++ TIC replacing: 2  of  2
++++++++ replace : Data 2
++++++++ replacing out socket type with:  VerticesSocket
++++++++ TOC replacing: 2  of  2
++++++++ TOC update_socket_types

++++++ TOC update

Writing: /tmp/inputSwitch-Dev4.crash.txt
Segmentation fault: 11
logout
Saving session...
...copying shared history...
...saving history...truncating history files...
...completed.

[Process completed]

```

**Comment by enzyme69** [comment]
Crash report:
[inputSwitch-Dev4.crash.txt.zip](https://github.com/nortikin/sverchok/files/2877757/inputSwitch-Dev4.crash.txt.zip)

**Comment by DolphinDream** [comment]
Maybe I'll deactivate the socket-type update for now.. until we can figure out why this is not working. Let me know if anyone has a clue how to track this.

**Comment by zeffii** [comment]
segfault suggests access to (or operation on) a reference which doesn't exist anymore. I got many segfaults experimenting with monads. maybe search the issue tracker for "closes to desktop".

In the meantime I would also just push the node in the state where it works without predictable segfaults, and periodically experiment with the code in a separate pull request dedicated to finding the cause.

**Comment by zeffii** [comment]
or https://github.com/nortikin/sverchok/search?q=crash+blender&unscoped_q=crash+blender&type=Issues

**Comment by nortikin** [comment]
@DolphinDream why not use boxes shading rectangle? it will give simple and clean interface without extra sockets.

**Comment by nortikin** [comment]
ok @DolphinDream  i see that extra socket exists. Too extra as for me. Lovely would be to find.
2.80 has something strange D.worlds['World'].node_tree.nodes['Background'].inputs['Color'].draw_shape = 'DIAMOND' ('SQUARE', 'CIRCLE')
but it gives nothing

**Comment by DolphinDream** [comment]
Maybe my code is doing something weird though it is possible that it exposes some bugs in the sverchok infrastructure code or even in the blender code itself.

**Comment by DolphinDream** [comment]
Also it seems that when swiping the set size the output links are dropped sometimes, in spite of using replace_socket. I wonder if the replace_socket is doing something shady (possibly related to the crash?).

**Comment by Durman** [comment]
related theme
#2198

**Comment by DolphinDream** [comment]
Got the auto socket type working without crashing. Here's the latest version. Pull the latest upstream branch to check it out. 

sv-inputswitchnode-example1


Here's a little live demo.. showing the auto socket type.

https://youtu.be/9hKJgpppHpc

**Comment by zeffii** [comment]
good stuff @DolphinDream 

can I pester you into 
- using python style variable names instead of camelcase
- using f string formatting instead more?

```diff
- inLabel = selectedSetLabel + " " + str(i + 1)  
+ in_label = f"{selected_set_label} {i + 1}"
```

DolphinDream:
Sure. I will convert this node code and try to stick to the_ugly_lower_case_filled_with_underscore_and_ahere_to_the_annoying_policy_that_every_language_has_to_have_its_own_ridiculus_naming_convention. That being said, I prefer camelCase style (It's more readable to me for sure, and it's more condensed as it doesn't have gthe extra underscores all over). Python should have made this an option and leave it to the user style preference. Readability is to a great extent subjective.

Thanks for the feedback btw. I will fix these two bullet points before closing this PR.

zeffii:
>the_ugly_lower_case_filled_with_underscore_and_adhere_to_the_annoying_fact_that_every_language_has_to_have_its_own_ridiculus_naming_convention_policy

that variable name is too long. It's python, it has some useful consistencies. variable naming conventions is one we do follow generally in the codebase, because to python heads that's easier to read.

regarding `updateNode`, yes. no doubt.  However, it is sufficiently pervasive and has been "grandfathered in". In 2.80 i would consider we fix that.... but there are waay more important things.

DolphinDream:
@zeffii  any ideas why i'm getting this syntax error:

```
out_label = f"Data {i + 1}"
                              ^
SyntaxError: invalid syntax
```

zeffii:
oh   right.    for some reason i thought blender2.79 was also py3.6 (needed for f-string)

zeffii:
relevant discussion :  https://developer.blender.org/T54355   i've only ever used builder.blender.org, so adapted early.  

hold-off the f-string stuff then :)  sorry. (but f-string is awesome)

DolphinDream:
yeah, I like the f-string syntax too.

DolphinDream:
One more example.. a more complex node tree this time. 

Yeah.. you can group sets of nodes together via an InputSwitch node and link them to another InputSwitch Node. :) this way you can control the input from one set of generators separately from the input from another set of generators :) the master switch, in this case, selects the winner of both groups :) I guess you could go even crazier than this with what this node would allow you to do. I can't wait to see how users end up using it. 

sv-inputswitchnode-example2

And the animated version:

https://youtu.be/h0n2H2fOLlY

DolphinDream:
I added more switch macros to help various connections to the InputSwitch node. 

The new switch macros are:
"> sw1" : connects the output 1 of the selected nodes
"> sw12" : connects the outputs 1&2 of the selected nodes
"> sw13" : connects the outputs 1&3 of the selected nodes
"> sw123" : connects the outputs 1,2&3 of the selected nodes



This should cover a good amount of cases, but it may miss some important ones that we can add as needed. In the future, a more generalized solution involving perhaps a popup (triggered after invoking the switch macro) letting the user select dynamically what outputs of the selected nodes to connect to the InputSwitch node is desirable, but this is out of the scope for the time being. Such popup could also let the user select whether the macro should also connect the InputSwitch node to a ViewerDraw node or any other useful node.

For now, I decided to spawn a ViewerDraw node whenever the selected nodes have a VerticesSocket node type as the first output node. For generators, this makes sense, though for other vector nodes (non-generators) this may not be useful. 

https://youtu.be/M6eP7b0zrvA

DolphinDream:
If there are no objections I can merge this PR.

zeffii:
no objections

DolphinDream:
Thanks for all the feedback, folks! 
This big boy is ready to be put to work :)  Have fun!

@enzyme69 your move! ;) (please read the node documentation if something is unclear)

vicdoval:
Works great! 
One late idea: Wouldn't be more logical to place this node on the "Logic" category?

DolphinDream:
Thanks @vicdoval Yeah. I don't have any objections to relocating this to the more appropriate category. Feel free to relocate it if all agree.

PR #2379, opened 2019-03-06
Formula node mk3

portnov:
Our current "Formula" node has a couple of problems:

* Inconvenient syntax with only variables being available "X" and "n[0]" .. "n[150]"
* weird code with Lenin and some other strange people being involved for some reason >_<

It is also not very convenient to use it to produce curves or surfaces, because you have to write something like "[sin(n[0]), cos(n[0]), 0]".

## Solution description

Re-implement Formula node:

* Allow to write arbitrary Python expressions with arbitrary variable names, so that you can just write `R*sin(phi)`. Use the same technique as in "mesh evaluate" node to manage input sockets.
* Add possibility to write not only single formula, but up to 4. So to generate an ellipse, you have to write `A*cos(t)` in one input box, `B*sin(t)` in another, and that's it, no brackets and so on. By default, show only one input box to save screen space for most cases.
* Support a flag to enclose each vector produced by several formulas into separate sub-list. It is usually required when you have more than one formula; it is usually not required if you have only one formula; but there may be other options.
* Support a flag to enclose the whole output into additional pair of brackets, which other nodes may want to see.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Migration from old node supported.
- [ ] Ready for merge.

### DIFF
--- a/docs/nodes/number/formula3.rst
+++ b/docs/nodes/number/formula3.rst
@@ -1,0 +1,88 @@
+Formula Node Mk3
+================
+
+Functionality
+-------------
+
+This node allows one to evaluate (almost) arbitrary Python expressions, using inputs as variables.
+It is possible to calculate numeric values, construct lists, tuples, vertices and matrices.
+
+The node allows to evaluate up to 4 formulas for each set of input values.
+
+Expression syntax
+-----------------
+
+Syntax being used for formulas is standard Python's syntax for expressions. 
+For exact syntax definition, please refer to https://docs.python.org/3/reference/expressions.html.
+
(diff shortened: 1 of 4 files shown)


portnov:
screenshot_20190306_235158

portnov:
screenshot_20190307_190530

portnov:
screenshot_20190307_200145

portnov:
screenshot_20190307_202842

nortikin:
спасибо, Илья.

Issue #2380, opened 2019-03-10
speed up update system


Durman:
The idea is to protect us from extra calculations. 

For example we have spiral node that generate huge data (700 000 points particularly). There is no need to generate this data each time when update system is calling. 
It is very simple to avoid recalculations. All that we have to do is to memorize properties of the spiral node and next time when update system will call to update the node, we will compare current properties and memorized properties, if nothing is changed do nothing.

Next step is do something with this data for example move points somewhere. But if this needs only once there is also no need recalculate this each time. 
In this case we should check not only properties of the move node but also was data coming from inputs sockets changed. For this goals I created new property of output sockets that returns True if parent node was recalculated.

I made speed test for old and new update system with next layout:
![2019-03-10_10-12-50](https://user-images.githubusercontent.com/28003269/54081347-18c22500-431d-11e9-8117-f0d103f83ed3.png)

Spiral node generates 700 thousand points. New update system shows increasing speed to 15-75 times. It looks strange that with the same conditions new update systems shows different increasing speed. And I think it should work even faster.
Old update system:
![old_update_system_700000points](https://user-images.githubusercontent.com/28003269/54081399-d816db80-431d-11e9-83ef-9a8c7db7a56a.png)
New update system:
![new_update_system_700000points](https://user-images.githubusercontent.com/28003269/54081400-e36a0700-431d-11e9-80c4-db578a57e2f9.png)

Related theme: #121

Durman:
I have understood that stethoscope node is guilty in decreasing speed. Test without this node. This still seems too slower.
![new_update_system_700000points_wthout stethocsope](https://user-images.githubusercontent.com/28003269/54081603-c97df380-4320-11e9-97c7-910dec013a85.png)

portnov:
See also #1934.

Durman:
If output or input data is big it is take a lot of memory for memorizing more than one result. And it is difficult to predict how a node will be used. So I think in most cases there is no sense to memorize more than one (last) result.

| input | output | complexity | memory | times |
| --- | --- | --- | ---:| ---:|
| big | big | hard | a lot of | 1 |
| big | big | easy | a lot of | 1 |
| big | small | hard | a lot of | 1 |
| big | small | easy | a lot of | 1 |
| small | small | hard | low | 100 |
| small | small | easy | low | 10 |

By times I meant how much times to memorize result of a node.

One more reason against memorizing more than one result is that we should do calculate something like a hash number of input data. 
[...]

PR #2393, opened 2019-03-28
Update system mem

Durman:
This is an attempt to remake update system. Main goal is to make such system that would calculate node only when this is really need. I think this can become really important update of Sverchok that will make nodeing a little bit easier.
There are 3 update events:

| event | old US | new US |
| --- | --- | --- |
| update node | updating only changed nodes | updating only changed nodes |
| update tree | updating every node in tree | updating only changed nodes |
| animation update | updating every node in tree | update only animation nodes and next after them |

In new update system there are 3 conditions that determine should node be recalculated or not:
- All sockets (inputs and outputs) should have the same links as previous update. 
- All data passing from input sockets should have the same data as previous update event. 
- All parameters of node should be the same as previous update.
In most cases this approach should work faster and even much more faster. 

This branch is ready to test. In this moment monads is not supported yet the same as nodes which takes data from blender database excluding frame info node. 

### Testing instruction:

- First of all old update system should be disable in Svechok preferences.

After that you can start testing. In node tree changed nodes will have red color. If you will notice some nodes that should not be updated but nevertheless has red color please make bug report in this topic.
update colors

It is possible to get additional information about process of update system:
- For this you should enabled debug level of logging in Svechok settings and toggle log to text buffer.

 - Now in "Sverchok log" text file updating information is available.

For better perfomance of animation I would recommend disable logging.

#2380

## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


vicdoval:
This looks awesome! 
Just some questions:
- When animating shouldn't be updated all keyframed nodes also?
- In the Pulga Physics node I have created "Accumulative" mode. When activated the node stores the output and uses it as the input in the next update. To perform many updates I play the animation to make the shape evolve. This behaviour wont be possible with this new update system except if there where some way to create exceptions (if something then update always)
You think that could be included?   
- Because of the size of this update wouldn't you prefer to aim to the 2.8 branch?

Durman:
- Only frame info nodes are taken in account during animation at present. 

- Well, you can create additional property with value of random number for example. When your node is updating first time change this value if the node should be updated next time. Also you should keep in mind that there is at least three types of events of update system. This way will work only when the node tree was changed (click on any socket of any node in the tree).

- At present I stucked on how to detect fastly was input object from scene updated. Probably this branch will remain experimental. I created this pull request only to share my thoughts about improving the update system. If it is closer to workable version I will consider about 2.8

PR #2299, opened 2018-11-02
B28 prelease master

zeffii:
image

## What is this?

This branch aims to get sverchok loading in Blender 2.80. Each node has been modified to account for changes to the new API. Most modifications to existing nodes are superficial, but you will encounter a few new nodes that are complete rewrites. Mostly the behaviour should be the same, our work process hasn't changed.

I'd like to point out that this is experimental, and still changing hourly/daily, and will be for some time to come. If you want stable, use B2.79 and the last Sverchok release ( That's going to be the LTS version ). The hope is to keep this version more forward looking for a while, and not call it stable until Blender 2.80 itself moves to full release status.

There's still much stuff missing, some stuff might not come back, some stuff will be new or implemented in a different way. The branch is also an opportunity to ditch a lot of vestigial code (old nodes directory is empty for the time being)

## Missing nodes

Blender 2.80 is still a moving target, this means for sanity's sake it's necessary to hold-off trying to get all nodes working so soon. As a result some nodes are gone at the moment, because it's unclear how to replace their internal workings. Additionally some nodes that use `bgl` or interact with Blender `Objects` are rewritten. This means layouts that contain these nodes can encounter initial failure.

- old sverchok .blends will not work
- old sverchok `gists / json` will not work correctly (also most `examples` won't )

eventually those nodes that have approximate replacements will be able to automatically find a replacement. 

## How to install?

Installation procedures are described in #2326. This description will remain actual until separate branch for 2.8 would be merged into master.

## Bug reporting

That's tricky, i'm only really interested in Bugs encountered while trying to install the software, or if a node goes red then the error message from console is interesting. Or if you get a Blender crash using a specific node (if you can't show the crash with 3 nodes, i'm not interested). 

All other issues can be ignored for now. Eventually of course we / i will be more receptive to user experiences.

## tracking work done, or to complete

Another attempt at porting sverchok, this time i'll dedicate a few days to it. 

- [x] update `sv_update_utils.py` temporarily,
    maybe just explain how set sv_prefs  , so people 'can' update
    ```python
        # you could run this in the text editor, then press save preferences in the Preferences
        # then maybe update will work :)
        
        import sverchok
        from sverchok.utils import sv_prefs

        sv_prefs.set_vals(
            dload_archive_path="https://github.com/nortikin/sverchok/archive/",
            dload_archive_name="b28_prelease_master"
        )
    ```
- [x] use keyword arguments where appropriate
- [x]  some form of nodeview update using updateNode
[...]

PR #2394, opened 2019-03-29
Pulga Physics Node

vicdoval:
From its docs :D
> This node creates simulations from input parameters, it is meant to be used in form-finding purposes.
> It creates the simulation using the inputed vertices as spheric particles that react to applied forces.
> The node is a basic NumPy implementation of basic physics system heavily inspired in "The Nature of Code" by Daniel Shiffman and the Kangoroo Plugin for Grasshopper. Due the nature of the algorithm it can get very intensive, handle with responsibility...

Pulga Physics
Pulga Physics
Pulga Physics
Pulga Physics
Pulga Physics Example
tensileCover
attractorsField
catenary_modeling
reactionsOnPins

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/alpha_nodes/alpha_nodes_index.rst
+++ b/docs/nodes/alpha_nodes/alpha_nodes_index.rst
@@ -8,3 +8,4 @@ Alpha Nodes
    offset_line
    contour2D
    planar_edgenet_to_polygons
+   pulga_physics
--- a/docs/nodes/alpha_nodes/pulga_physics.rst
+++ b/docs/nodes/alpha_nodes/pulga_physics.rst
@@ -1,0 +1,100 @@
+Pulga Physics
+=============
+
+Functionality
+-------------
+
+This node creates simulations from input parameters, it is meant to be used in form-finding purposes.
+It creates the simulation using the inputed vertices as spheric particles that react to applied forces.
+
+
+Input & Output
+--------------
+
+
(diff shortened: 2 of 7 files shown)


portnov:
Hello.
Wow! :)

* where is pulga_physics_core  module? was it merged before?
* am I correct that this module implements physics by itself, and does not rely on external libraries?
* Maybe it is possible to split such a big node into several? It would be easier to create different combinations of effects.

vicdoval:
Thanks @portnov!

_ I added the pulga_physics_core  to sverchok/utils mainly to split the code between UI and the calculation process.
It has never been published or merged before anywhere :)

_ You are right, this module implements physics only relying in NumPy.

_ Many smaller nodes could came out of this node by taking out features but they wouldn't be able to produce the same results by mixing them (that's why I added the top toggles, so the user can choose between the features combination).
It could be splitted with a bunch of new socket types in to a set of dedicatted nodes connected to a main one. This approach could be interesting... maybe for the 2.8...

zeffii:
superb! i look forward to using it.

nortikin:
ahaha!!!
Much of fun ahead.

vicdoval:
It was ported to 2.8 in https://github.com/nortikin/sverchok/pull/2299/commits/d11ca09476b430190944ee4cc68afc72b269529c :)

PR #2395, opened 2019-03-30
Added RGB Curves Link Node

spamrakuen:

## Solution description

This node allows to sample data from a Material's RGB Curve node.

In order to work, a material that use nodes should be created, then just drop a RGB Curves node.

This new node will sample values from that curve(s)

## Preflight checklist

- [x] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


spamrakuen:
Broken, please forgive me. I'll fix this later.

Durman:
Probably what you are looking for already implemented. 


This is "UI curves manual" script for SNL node. 
Also have a look here: #2236

nortikin:
maybe make this snl to complete node?

PR #2396, opened 2019-03-30
Added new RGB Curves Link Node (II)

spamrakuen:

## Solution description

This node allows to sample data from a Material's RGB Curve node.

In order to work, a material that use nodes should be created, then just drop a RGB Curves node.

This new node will sample values from that curve(s)

## Preflight checklist

- [x] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/index.md
+++ b/index.md
@@ -45,6 +45,7 @@
     SvEllipseNode
     SvSuperEllipsoidNode
     SvSmoothLines
+    SvRGBCurvesLink
 
 ## Analyzers
     SvBBoxNode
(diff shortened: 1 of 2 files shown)


zeffii:
I have mixed feelings about this. we've already had the ability to load rgb curve UI element via a mode in SNLite (and that works with IO/JSON..and requires a bit more work to serialize the curve data/state).

If you aren't familiar with the "Fcurve sampler" node yet, it is probably worth looking at, as it doesn't require a fake-material for the UI element.

nortikin:
maybe convert snl script to complete node?
this interface is usefull much

spamrakuen:
Created a new Pull Request, closing this one. 
I've rewriten the node, this time it also works with ColorRamp.

I didn't know SNL could do the same, but seems more hidden and limited than this node.
"Fcurve sampler" on the other hand, seems more powerful and precise, but IMHO more tedious.

As nortikin suggested, I think the best would be a complete node. 
IMHO the best would be that blender developers create a new CurveProperty. The material trick is a bit dirty...

PR #2397, opened 2019-03-30
Added new Material Link Sampler node (III)

spamrakuen:

Sverchok already supports curve UI via SNLite and "Fcurve sampler", this is just another way to do the same. **SNLite needs node name to be unique per material but this node does not. "Fcurve sampler" is very powerful but quite different.**

## Solution description

This node allows to sample data from a Material's **RGB Curve** or **ColorRamp** nodes.

In order to work, a material that use nodes should be created, then just drop a RGB Curves or ColorRamp node.

This new node will sample values from that curve(s). Just choose the material and type the node's name.

## Preflight checklist

- [x] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


spamrakuen:
Sorry for this third try of pull request.

I've been testing SNL curve manual, but I can only have one curve working at the same time. If I add a second SNL it shows the same curve. With my node, there is no limit of curves, and each node supports 4 curves at the same time.

Also I've added support for ColorRamp node.

I'm still not happy with node's name... since it's used for sampling curves, somethink like "material curve sampler" would be better. I don't know...

nortikin:
@spamrakuen 
Lets do it together. you only grab curve itself. but not control it. What the reason for that?
Lets enhance.
https://github.com/nortikin/sverchok/blob/master/node_scripts/SNLite_templates/UI_curves_manual.py#L7-L12 two lines of code to define new rgb curve. We have to make blender veriable to handle names to autocreate new RGB curves.
https://github.com/nortikin/sverchok/blob/master/node_scripts/SNLite_templates/UI_curves_manual.py#L16-L20 here we use this in UI draw()
Not understand reason for CRGB inputs and outputs. We are working here with one float series (or vertices as xyz). what the reason to dictate inputs of CRGB? User not need to know what is on every socket, it makes him crazy mostly.
Lets do that thing easier.
Also we need extend this file https://github.com/nortikin/sverchok/blob/master/utils/snlite_utils.py#L47-L61 to support our nodes new naming approach. It is reached in our node from this https://github.com/nortikin/sverchok/blob/master/node_scripts/SNLite_templates/UI_curves_manual.py#L23-L25 point
Naming itself can be based on this https://github.com/nortikin/sverchok/blob/master/nodes/viz/viewer_bmesh_mk2.py#L275-L277 but extended as sv_<NAME> for material and rgb curve

nodes/generators_extended/matl_link_sampler.py
```diff
@@ -0,0 +1,111 @@
...
+                outputs[0].sv_set([data_c])
+
+            if outputs[1].is_linked:
+                data_r = []
+                for i in range(0, count_r):
+                    data_r.append(mnode.color_ramp.evaluate(i / (count_r-1))[0])
+                outputs[1].sv_set([data_r])
+
+            if outputs[2].is_linked:
+                data_g = []
```
nortikin: this is repeating part can be as definition also inside node

nortikin: and iteration for inputs and outputs.
Anyway, we need other approach in that

nortikin:
Снимок экрана в 2019-04-01 23-50-37
here is adventures. But as you can see, we have to open material window, and also create material and curve node in current state.

zeffii:
for b2.8  i'm tempted to implement a version of the RGB curve node much like the SNL version, which 
- automatically generates the material and required nodes.
- stores state / points of the rgb curve data in the node (used for IO/JSON )
- implement `def copy` to clone the RGB curve node too, and link that.

the node is only really complicated because of the ugly hacks needed to get it working  : /  , I don't think anyone would disagree that the functionality is cool ( that's why I made the SNLite script initially )

nortikin:
this SNL node used to produce my wedding ring, so yes, all for complete solution

spamrakuen:
Improved UI and code

As nortikin suggested I've added:
- some definitions for removing some repeated code.
- UI curve/color ramp interface.

I'm using  CRGB inputs and outputs because with only one node I've up to 4 curves. Using only 1 curve for each node, means creating more nodes in the material.

nortikin:
next step - sv init should initiate creating material or add if exists rgbcurve node with next name from global counter.
than propose define one thing. If we put float range inside input RGBC sockets, node should understand it and evaluate RGBC curve and change input floats accoarding to curve. If node see integers input, than it acts as you mentioned. Moreover, it can be vectorized, but i not sure we need it.

spamrakuen:
New commit

Button for creating material and node
Button for refreshing/updating result after curve changed (removed checkbox)
Support for using array of floats as input

Durman:
#3064

PR #2403, opened 2019-04-26
Import the edge list caching from 2.79 to 2.80

DolphinDream:
Import edge list cache (will be needed by other nodes soon to be ported to 2.80)

### DIFF
--- a/data_structure.py
+++ b/data_structure.py
@@ -160,11 +160,11 @@ def fullList(l, count):
     return
 
 def fullList_deep_copy(l, count):
-    """the same that full list function but 
+    """the same that full list function but
     it have correct work with objects such as lists."""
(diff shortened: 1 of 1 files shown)


DolphinDream:
@zeffii If no objections, I can integrate this in b28_prelease_master.

zeffii:
yes go for it.   but i caution against retrofitting existing nodes/classes initially.  get it working first, then become confident that it is implemented correctly. then (later) draw up a plan / protocol to implement it in any node where it would reasonable be a net positive.

DolphinDream:
@zeffii any preference to the merge type? Create a merge commit / Squash and Merge / Rebase and Merge.

zeffii:
squash and merge. please

PR #2405, opened 2019-04-26
CI travis test fixing

vicdoval:
Strange changes in decimals and edges output of intersect_edges_mk2


kalwalt:
Why this happens? Travis for me is something obscure... what does kind of checks it makes?

vicdoval:
It makes the test we supply, usually works like charm but in this case something truly obscure, an inexplicable change in edges order in one node and  0.000001 differences in some calculations.... :)
Now seems fixed

kalwalt:
thank you for the answer @vicdoval

PR #2408, opened 2019-05-01
Import inputSwitch node from 2.79 to 2.80

DolphinDream:
The port seem to work except the macro connections or auto connection to VD node. I'm not sure if there is something missing a the socket / node_tree level, or whether there is something not quite working with the VD node.

The anomalous behavior I notice:
1. right click to connect to VD creates the VD node but it doesn't connect the sockets to it
2. the sw macros work except the connection to the VD node (e.g. sw123 only connects V and E but not P socket).
3. once connecting the outputs of the switch node to the vD node the viewport doesn't draw anyting unless i change the frame.

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/docs/nodes/logic/input_switch.rst
+++ b/docs/nodes/logic/input_switch.rst
@@ -1,0 +1,72 @@
+Input Switch
+------------
+
+This node allows switching among an arbitrary number of sets of any type of inputs connected to the node.
+
+Inputs
+======
+
+When first created, the node has a single **set** of input sockets (e.g. ["Alpha 1", "Alpha 2"]) and it will automatically create new sets of input sockets (e.g. ["Beta 1", "Beta 2"], ["Gamma 1", "Gamma 2"] etc) whenever any socket of the last set is connected, thus always allowing additional set of inputs to be connected to the node. Also, when all the sockets of the last connected set are disconnected the node will remove all but the last disconnected sets, thus always allowing one extra set of inputs to be connected to the node.
+
+The node takes any type of inputs from the external nodes for its sets.
+
+The sets are labeled based on the Greek alphabet letters from **Alpha** to **Omega** and the inputs within each set are suffixed with an increasing number starting from 1 (e.g. Alpha 1, Alpha 2, etc).
+
+**Alpha #**
+This first set of inputs always exists, no matter whether any of its input sockets are connected to or not.
+
+**Beta #**
+...
+**Omega #**
+These sets of inputs are created automatically, one at a time, only as the last set is connected.
+
+Note: Be advised that while it is possible to connect any arbitrary mixture of input types to any sets (e.g. Vertex & Edge to the Alpha set, and Float & Color to Beta set etc) this combination of mixed sets of inputs may likely be of little to no use. This is because the output of the Input Switch node is likely going to be connected to a node tree that expects specific types of inputs, and therefore when changing the selection of the set, the Input Switch node will output a set of values that may be incompatible with the ones expected by the nodes it connects to. There are exceptions to this rule, of course, for instance: you can connect Vertex & Edge to Alpha set and Vertex & Poly to Beta set and have the output of the Input Switch node connected to a node that expects Vertex and EdgePoly as inputs. In this case, switching the set selection from Alpha to Beta will not break the node-tree since the final node can take either VE or VP as inputs.
+
+
+Parameters
+==========
+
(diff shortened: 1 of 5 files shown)


DolphinDream:
It seems that the VD node does not draw vertices unless the edges and/or faces inputs are also connected.

zeffii:
your macro must assume the socket names are different for the VD node in 2.8

DolphinDream:
@zeffii there seems to be an issue with the VD node. Try connecting a box generator to the VD node one socket at a time.. if you connect just the vertices it draws nothing until you also connect either the edges and/or the faces. 

The VEP colors could be better chosen I think. The current edge color is almost invisible on the gray background. I always think the VD doens't draw the edges..until I realize they are hidden in plain sight :)

DolphinDream:
@zeffii Here's an issue with (apparently) socket data not being updated properly. As I connect the output of the switch node to the VD nothing happens.. until I change the frame. Note how the labels of the output sockets of the switch node don't have the data count numbers next to the label (Data 1 .. vs Data 1.1) . These numbers appear once the frame changes. I'm not sure if this is a switch node issue or is caused by some other changes in SV b28 at a lower level/infrastructure. 

SV-InputSwitchNode-B28-demo1

zeffii:
the issue with the VD node is that I didn't entirely fix the input logic, it's nothing overly complicated - i just didn't get around to it.

it's almost weekend and rainy, i might have more time for this in the next few days.

zeffii:
the best way to test the switch node conversion in b.28 is possibly by using the DebugPrint node, which accepts multiple inputs.     it's possible that the use of the `macro` causes the tree (and all nodes involved) to be updated less times than needed.  There's all kinds of update limiting going on  the update system/event handler, and it is different substantially than 2.79 in that respect.

zeffii:
remember i implemented custom defaults for nodes some time ago. check the issue tracker for how to set defaults for example the VD node.  I happen to use a bright background

zeffii:
btw, using the arrow keys up/down is how i update the frame when stuff like this happens. Naturally one should not have to do any of this, but it's a faster ui trick.

DolphinDream:
> btw, using the arrow keys up/down is how i update the frame when stuff like this happens. Naturally one should not have to do any of this, but it's a faster ui trick.

Thanks for the tip. Yes, this is what I normally do for chaning the frame, though for the purpose of the gif capturing I chose to click on the timeline to capture the frame change action since left/right keys are not going to show the frame is changed.

DolphinDream:
Also, the default color features is nice to have. thanks for reminding me about it. However, IMO for the new users we should probably come up with a default colors that look best on blender's default theme.

DolphinDream:
I imported this node to 2.8 but I can't figure out why the updates don't work correctly, until changing the frame a couple of times. If anyone wants to take a hit at it, be my guest. I kind of lost hope.

For what is worth.. here's another wierd behavior... initiating a socket connect and drop it before connecting it changes the inputs to the other sockets (apparently, if that even makes sense).

SV-InputSwitchNode-B28-demo2

InputSwitchNode-B28-Issue

zeffii:
weird, i'll have a look later.

zeffii:
there's a few things going on there, none of which I have a satisfactory fix for at the moment. i'm not sure where there's a `cache-miss` as a result of some non-event.

i'm merging so it's easier to check.

zeffii:
i'm slowly rewriting this node @DolphinDream -- to the best of my ability, with less superficial features.. just in an attempt to see if a completely alternative approach lets us identify the problem.

DolphinDream:
@zeffii the MOD version of the InputSwitch node seems to have some issues:
* Once new set is created it doesn’t get removed if the input sockets are disconnected.
* The switch macros do not work since the label_of_set method (used to figure out the next set to connect the input nodes to) does no longer exist.

zeffii:
1. yep, there's no automatic removal if last visible set is unconnected 
    - (or last two sets is what you probably mean, would be the criterion. If last two sets visible are unconnected, then hide last set )
   - i think this would be acceptable to do at the top of the process function, either automated ..but simpler is an operator to "prune" socket sets.

2. of course switch macros is useless on the modified version, first things first .

my goal was really to see if this would be acceptable / useful / helpful as an example / alternative approach.

zeffii:
https://github.com/nortikin/sverchok/commit/8760419e48b497845095a3774087cd2f425b5dbe
pushed a working version (at least 123 works ) of the macro.

this is not massively tested yet, i make no claims of its stability. I also don't know how the node behaves in iojson :) haha probably not great

Issue #2376, opened 2019-03-03
project point to line


Durman:
We have a line that consists with many edges and we have one or several points. The goal is to find find closest to given point (or points) point lying on the line. I made small algorithm that can reach this goal. 

The algorithm can be divided on two parts. First part consist from finding closest point of the line to given point. I used KDTree from mathutils but faced with next problem.

![problem1](https://user-images.githubusercontent.com/28003269/53693118-f5dfc000-3db4-11e9-89c2-a846d104ca03.gif)

Unexpected solution of this problem was to subdivide the line to smaller edges. I expect that this approach can gives wrong result time to times but it should work in most cases.

![problem2](https://user-images.githubusercontent.com/28003269/53693254-553ecf80-3db7-11e9-89c6-ccd294767cd9.gif)

By default subdivision is equal to smallest edge in the line but also I made parameter that don't let to default subdivision have too small value.

Second part of the algorithm to find projection given point to edges that coinciding with found point on the line. For this task there are all that we need in mathutils. Product vector to vector and scalar multiplication of vectors. The scalar multiplication gives as a sign that says projected vector has same direction or not with another vector.

![projection_vector](https://user-images.githubusercontent.com/28003269/53693445-eca52200-3db9-11e9-95cd-eb4cc7ac7885.gif)

And final result.

![final](https://user-images.githubusercontent.com/28003269/53693546-68539e80-3dbb-11e9-82da-77b30c7edd79.gif)

Durman:
I was thinking about is it worth try to memorize only part result of a node. And I have made experiment on this node. 
![2019-03-14_19-32-54](https://user-images.githubusercontent.com/28003269/54369805-ff5c0880-468f-11e9-85bb-32fcadcea632.png)

In this case I expect that first input socket can get big data but second only a few points. So it is possible to prepare line and cache this result in memory. If only project points are changed  the end result will be get much faster.
![memoryzing parg of process](https://user-images.githubusercontent.com/28003269/54369838-0d118e00-4690-11e9-8378-e11c4c27ee20.gif)

PR #2416, opened 2019-05-12
New Distance Point Line Node

vicdoval:
Basic and super useful distance point-line node.

screen
screen2

screen1
screen1

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/data_structure.py
+++ b/data_structure.py
@@ -182,7 +182,22 @@ def sv_zip(*iterables):
             result.append(elem)
         yield result
 
-
+list_match_modes = [
+    ("SHORT",  "Match Short",  "Match shortest List",    1),
+    ("CYCLE",  "Cycle",  "Match longest List by cycling",     2),
+    ("REPEAT", "Repeat Last", "Match longest List by repeating last item",     3),
+    ("XREF",   "X-Ref",  "Cross reference (fast cycle of long)",  4),
+    ("XREF2",  "X-Ref 2", "Cross reference (fast cycle of short)",  5),
+    ]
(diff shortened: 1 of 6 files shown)


Durman:
Something similar to this: #2376 ?

vicdoval:
Similar but simpler.
The line is defined by 2 vertices. (first and last) I think it can be useful in many  parametric/mathemaical definitions.
#2376 would be more "project point on edges"

nortikin:
will do video for this one.
cool node

Durman:
@nortikin I'll switch attention to this node but still can't say how long does it take.

Durman:
I'm slightly confuse. What does line mean? 
As I thought Line is shape that consists from several edges. But also there is boundless line, probably straight line.

nortikin:
@Durman Good question. there is straight line. Not curved or segmented one. Would be great to deal both. straight for simple cases and yours for complex.

portnov:
Any reason why this is not merged yet?

vicdoval:
@Durman this node refers with Line to the endless straight line, it also gives some info about the relationship with the supplied segment.

@portnov i had some minor details i wanted to change. Now it is done.

More details in the docs file 😄

Durman:
> I'm slightly confuse. What does line mean?
> As I thought Line is shape that consists from several edges. But also there is boundless line, probably straight line


I found evidence that line is something consisted with several segments according Blender API.))

PR #2418, opened 2019-05-17
Project point to line

Durman:

#2376

## Solution description

My modest experience in coding shows that the smaller and simpler the functions the easier they are to use and the more ways to reuse them. And I'm starting to think that it is fair also and for nodes. So I changed my mind about adding extra features in the node. It will be better to create additional nodes instead.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


Durman:
<details> 
  <summary>project sphere to circle</summary>  


sphere to circle

</details>

<details> 
  <summary>monkey</summary>  

monkey

</details>

curve organizes random vectors along itself


<details> 
  <summary>nodetree</summary>  



</details>

Durman:
Sorting projected points along line:

<details> 
  <summary>nodetree</summary>  



</details>

Connect 2 loops with different number of vectors:

<details> 
  <summary>gif</summary>  

2 loops

</details>

Looks close to builtin "bridge edge loops" function of Blender:

Durman:
@zeffii 
I have a question about data structure of `Belonging` socket. 
The idea of output data of this socket is to show where projected point hit line. In most cases output is an edge like (1,2) but sometimes it can be a point like (0). 
I tried to build output list in this way - `[[1,2],[2,3],[3],[3,4]]`. But it looks like Sverchok unpuck lists if they consists with only one value - `[[1,2],[2,3],3,[3,4]]`. 
Is there a way to fix this how I intended to do this? 
Or if I leave it like this will this cause trouble?

zeffii:
that's tricky, i don't know off the top of my head. Maybe wrap the output one more time?

zeffii:
you are outputting either/both 
- the two vertex indices ( that make up the edge )
or 
- the single vertex index ( if the hit is directly on a vertex )

?
 maybe it makes sense to have two output sockets, one for edges hit ( a mask ) and one for verts hit (also a mask) 

i don't know i'm just asking :)

zeffii:
> I tried to build output list in this way - [[1,2],[2,3],[3],[3,4]]. But it looks like Sverchok unpuck lists if they consists with only one value - [[1,2],[2,3],3,[3,4]].

yeah , I think that will depend on the node reading the socket data from `Belonging`,  because sv_get() does not (to my knowledge) to any modifications of its own.

i would try outputting ` [  [[1,2],[2,3],[3],[3,4]]     ] ` instead, to match the convention.

Durman:
@zeffii My fault.)) I thought that this `(1)` will return `tuple` but it's not. Right way for this is `(1,)`. Now looks like all work as I expected.

Durman:
I have finished. Welcome to read documentation. I will wait with merging if there are some suggestions about code or someone wants to make a tests.

nortikin:
this is fantastic node!!!
hahahaha!!!

Durman:
@nortikin Thanks.)

I was thinking how the idea of the node can be spread out not only for lines but also for normal mesh.
The idea of this node is that the 2 meshes are taken. One them is base for modify another mesh. Second mesh is modified by the node and returns modified mesh with the same length and order. 
After I have came to this I have one more feature for this node and one question: 

- It will be not bad to have to input mask for project vectors in case there is no need to project tham all. But I think It is not too much important at this stage. So I'm going to do this in next iteration. 

- The question is was I right to potted the node into analyzers categorie. According my description above it fits better to modifiers change. Any ideas?)

portnov:
@Durman not sure what exactly do you have in mind. see "bend along surface" node?

Durman:
@portnov Yes, I saw. 
The idea of developing of principles of this node is to create node which will be able to project one mesh to another. For example it will be possible to project some flatten object on surface of sphere or another 3d object. 

It should works close to builtin modifier `shrinkwrap` of Blender. 
If `bend along surface` do the same work I don't know how to use the node for achievig such result.

nortikin:
@portnov  @Durman yes, bend over object is built on UV rows-cols, but this will be usefull too. shrink wrap  over non-ordered surface

portnov:
The same question: any reason why this is not merged?

Durman:
@portnov Well, I hope there was no.))

Durman:
> I was thinking how the idea of the node can be spread out not only for lines but also for normal mesh.

It's funny but the node with such functionality already exists. 😕

portnov:
That one can consume only blender's objects, as far as I can see, not sverchok's data.

nortikin:
@Durman  no excuses to not merge

PR #2421, opened 2019-05-23
Linear approximation node.

portnov:
The new node tries to approximate provided set of vertices by either a plane or a straight line. In other words, it searches for such a plane / line, that all provided vertices have the minimum distance to it.

CodeCogsEqn

This PR also contains a set of utility functions/classes in `geom.py`, which may be useful for other nodes, included scripted ones.

## Preflight checklist

Put an x letter in each brackets when you're done this item:

- [x] Code changes complete.
- [x] Code documentation complete.
- [x] Documentation for users complete (or not required, if user never sees these changes).
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.

PR #2425, opened 2019-05-25
New Distance Point Plane Node

vicdoval:
Keeping on with the basics

distance_point_plane_sverchok_blender_project_on_plane

distance_point_plane_sverchok_blender_project_on_plane

distance_point_plane_sverchok_blender_nodes

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/analyzers/analyzers_index.rst
+++ b/docs/nodes/analyzers/analyzers_index.rst
@@ -9,6 +9,7 @@ Analyzers
    bbox
    distance_pp
    distance_point_line
+   distance_point_plane
    deformation
    edge_angles
    kd_tree
--- a/docs/nodes/analyzers/distance_point_line.rst
+++ b/docs/nodes/analyzers/distance_point_line.rst
@@ -30,7 +30,7 @@ Inputs / Parameters
 +---------------------+-------------+----------------------------------------------------------------------+
 | **Line Vertices**   | Vertices    | It will get the first and last vertices's to define the line segment |
 +---------------------+-------------+----------------------------------------------------------------------+
-| **Tolerance**       | Float       | Minimal distance to accept one point is intersecting.               |
+| **Tolerance**       | Float       | Minimal distance to accept one point is intersecting.                |
 +---------------------+-------------+----------------------------------------------------------------------+
 
 Advanced Parameters
--- a/docs/nodes/analyzers/distance_point_plane.rst
+++ b/docs/nodes/analyzers/distance_point_plane.rst
@@ -1,0 +1,80 @@
+Distance Point Plane
+====================
+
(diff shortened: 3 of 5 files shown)


portnov:
Huh :)

I was thinking about a series of nodes that work with generic equations of line and plane. For example, a node that takes two planes and returns line... For now it's just API in #2421.

portnov:
Probably we should somehow merge two versions of API, it has similar functionality (but not exactly the same).

vicdoval:
@portnov :) 
I was also perceiving the convergence of our ideas.
When I was doing the previous node i found the mathutils.geometry functions and I thought they could be pretty useful for sverchok.
I did all the nodes and they are in this branch (in the analyzer category https://github.com/vicdoval/sverchok/tree/geometry_nodes
I decided to split them in different pull request to expose them one  by one.
On this three  #2416 #2425  and #2426  I have done a NumPy implementation that usually perfoms faster
I have not pulled but coded:
- intersect_cirlce_circle
- intersect_line_line
- intersect_line_sphere
- intersect_plane_plane

The nodes just match the data and call to the mathUtils function. Probably would be faster a NumPy implementation but I'm not sure if i will do it before making the pull request because they already work and it could take a while to crush the math part.
There is missing intersect_line_plane but it could be very similar to the others

You could use them as a base or something to compare or improve.

This node offers already two different implementations but if you find another which can be faster at any case it could be added :)

Also i have started to add the images of the docs to this branch https://github.com/vicdoval/sverchok/tree/docs_images so they don't loose the images names if you want to use it to just create do a PR with your changes

portnov:
NumPy approach has a lot of performance potential in case when you are computing a lot of data at once (for example, compute distances from many points to the plane): it can effectively vectorize computations (afaik numpy uses SIMD processor instructions when they are available). When dealing with single values, I don't think numpy is any faster than mathutils.
It has to be reflected in API; for example, you can write a method like distances_from_points(points_list), which will be much faster than `[distance_from_point(point) for point in points_list]`.

portnov:
Also, I have a request. Could you please add some comments to your code, especially to parts where you are doing numpy calls? Something like `vp0 = p_triang - v0[np.newaxis, :]` is not obvious for one who does not know what algorithm you are implementing. It might help to describe mathematical background of such code in comments, maybe add wikipedia links if necessary.

vicdoval:
You are right, NumPy could be even slower when you calculate just a couple of points, but the performance problems came when you need to do many calculations, and that is when Numpy shines.

I added this [link](https://www.scratchapixel.com/lessons/3d-basic-rendering/ray-tracing-rendering-a-triangle/barycentric-coordinates) in the numpy implementation and expand some variable names :D

nortikin:
https://vk.com/wall-35076122_10454

PR #2432, opened 2019-06-02
Calculate Mask node.

portnov:

There are nodes, which output, for example, "All faces" and "Processed faces" or smth like that. To do something with that output, it would be usually more effective to deal with "All faces" and "Processed faces mask" instead.

## Solution description
The new node takes two sets of objects (numbers, vertices, edges, faces, whatever) and returns a mask: True if item is in subset, otherwise False.

The node can work on different levels of data trees. For example, given Subset = `[[1, 2], [3,4]]` and Set = `[[1, 2], [3, 4], [5, 6]]`:

* with level = 0 it will output `[True, True, False]`
* with level = 1 it will output `[[True, True], [True, True], [False, False]]` 

Given Subset = `[[1], [5,6]]` and Set = `[[1, 2, 3], [7, 8, 9]]`:

* with level = 0 it will output `[False, False]` (because, for example, there is no `[1, 2, 3]` in the `[[1], [5,6]]`)
* with level = 1, it will output `[[True, False, False], [False, False, False]]`

There is also new API `data_structure.calc_mask`, which may be used by other nodes, including scripted ones.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/data_structure.py
+++ b/data_structure.py
@@ -365,6 +365,42 @@ def describe_data_shape(data):
     nesting, result = helper(data)
     return "Level {}: {}".format(nesting, result)
 
+def calc_mask(subset_data, set_data, level=0, negate=False, ignore_order=True):
+    """
+    Calculate mask: for each item in set_data, return True if it is present in subset_data.
+    The function can work at any specified level.
+
+    subset_data: subset, for example [1]
+    set_data: set, for example [1, 2, 3]
+    level: 0 to check immediate members of set and subset; 1 to work with lists of lists and so on.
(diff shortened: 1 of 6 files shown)


portnov:
Example: iterative use of the Inset Special node:
Screenshot_20190602_110622

Durman:
Useful with remove doubles node

nortikin:
i can use it in my current job!!!!

PR #2442, opened 2019-06-09
Bevel Curve node.

portnov:
This is more or less an analog of standard Blender's "bevel a curve" functionality.
Taper object can be optionally provided.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


portnov:
Screenshot_20190609_151949
Screenshot_20190609_152524
Screenshot_20190609_153333

portnov:
Screenshot_20190609_154344

portnov:
Screenshot_20190609_163332

portnov:
Screenshot_20190609_164709

portnov:
Screenshot_20190609_183617
Screenshot_20190609_183728

Durman:
Great node.))
I see we have our own mesh data structure now. :rocket:
I have asked before about using named tuple for vectors before. Zweffii says that this will slow down computation. Is it possible that with such data structure we can have similar effect?

portnov:
Well, actually, it does...
But I think it's better to optimize this structure, than to have more verbose code in the node...

Durman:
In the circumstances I think there is a point to start to talk about new socket - Mesh. It could decrease extra connections between nodes, and probably to decrease complexity for newcomers.

portnov:
Well, that definitely goes to "after 2.8" milestone :)

Durman:
And one more question about own class of mesh. Would not it be repeating of already existed Bmesh class of blender?

https://docs.blender.org/api/blender_python_api_current/bmesh.types.html#module-bmesh.types

portnov:
Yes, it does partially repeat the same functionality. My hope was that, since my class implements only very basic functionality, it will be faster than bmesh (each method of bmesh is a call from python to C part of blender).
This hypothesis still requires confirmation from profiling results. Will check.

portnov:
Checked: no, bmesh is actually faster. Will rewrite to bmesh.

Durman:
I have extra question about Taper Verts socket. I haven't understood what data expected for running with scaling. I try to use radiuses of curve for scaling. Is it possible?

zeffii:
ported just now, because i wanted to try it :)

zeffii:
while bmesh structure is fast, it does not guarantee face/edge index consistency - and this is why we do not use that pervasively or as a first class socket.

zeffii:
( i didn't port over the docs yet ..)

Durman:
> This is more or less an analog of standard Blender's "bevel a curve" functionality.

It looks like more then less.))

portnov:
@Durman , it expects a curve elongated along the orientation axis (for example, Z, if your bevel object is in XY).
List of radiuses is not enough, since it is not known which Z coordinate corresponds to which radius.

Durman:
<details> 
  <summary>node tree</summary>  



</details>

bevele curve node

zeffii:
image

Durman:
Using f-curve for scaling

nortikin:
it will work with SN toe node generator

PR #2447, opened 2019-06-10
Added Export/Import feature for tree properties in 3D view panel + 2 bug fixes

dimanjy:
[fix] Bug at ui.sv_panels line 235
[fix] Scip missing nodes in tree.SvShowIn3D
[add] Added Export/Import feature for tree properties in 3D view panel

### DIFF
--- a/ui/sv_panels.py
+++ b/ui/sv_panels.py
@@ -19,18 +19,15 @@
 import bpy
 from bpy.props import StringProperty, BoolProperty, FloatProperty
 
-
 import sverchok
 from sverchok.utils.sv_update_utils import version_and_sha
 from sverchok.core.update_system import process_from_nodes
 from sverchok.utils import profile
 
-
 objects_nodes_set = {'ObjectsNode', 'ObjectsNodeMK2', 'SvObjectsNodeMK3'}
 
 
 class Sv3DViewObjInUpdater(bpy.types.Operator, object):
-
     """Operator which runs its self from a timer"""
     bl_idname = "wm.sv_obj_modal_update"
     bl_label = "start n stop obj updating"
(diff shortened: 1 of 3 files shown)


dimanjy:
3D Panel Export/Import

utils/sv_panels_props_IO_operators.py
```diff
@@ -0,0 +1,161 @@
...
+    """Import current tree settings from .JSON file"""
+    bl_idname = "node.tree_props_importer"
+    bl_label = "Import settings"
+
+    target_node = bpy.props.StringProperty(name="Target node ID")
+    filepath = bpy.props.StringProperty(subtype='FILE_PATH')
+    filter_glob = StringProperty(default="*.json", options={'HIDDEN'})
+
+    def execute(self, context):
+        file = open(self.filepath, 'r')
```
zeffii: use 
```python
open(self.filepath, 'r') as file:
    ...rest of your file writing/reading code here 
```
that way you know the file is always closed when the code exits the scope ( with-as does that automatically )

zeffii:
We prefer to have bugfixes and new features in separate pullrequests. It's procedural.

ui/sv_panels.py
```diff
@@ -177,14 +174,16 @@ def draw(self, context):
                     for item in tree.Sv3DProps:
                         no = item.node_name
                         ver = item.prop_name
+                        # BUG: item could be in tree.Sv3DProps for some reason, even if it is deleted from tree.nodes!
```
zeffii: good point, i think we should pop deleted node props from sv3DProps in the tree update handler, to make sure we don't need to add code like this. This will be fixed a different way in svb2.8.

dimanjy:
> We prefer to have bugfixes and new features in separate pullrequests. It's procedural.
@zeffii, I have no idea how to make it :(

zeffii:
this time it isn't a big deal, but keep it in mind in future / bigger PR : )

zeffii:
in svb2.8 i added code in the update handlers that tests iif the update changed the node count of the tree. I might add a short function call there to pop the removed node property out of `tree.Sv3DProps` .

zeffii:
if no-one has objections, i'll merge this @nortikin @portnov @DolphinDream

portnov:
ok with me

zeffii:
congratulations @dimanjy on your first sverchok commit

dimanjy:
Thanks, @zeffii! :)

nortikin:
great

PR #2450, opened 2019-06-12
Profile node mk3

portnov:

Current Profile node has the following problems:

* Very restricted expression syntax - for example, one cannot even use `sin` or `abs`
* Restricted variable names - only one-letter names
* Comments can start only in the beginning of the line
* Knots/KnotsNames outputs work only when "from selection" feature is used, nor for any user-provided profile.
* Old-style "a la Basic interpreter" implementation makes it hard to maintain and add new features.

## Solution description

This is in fact a re-implementation of most of the Profile mk2 node:

* I wrote a very simplistic parsing framework, based on combinatorical parsing. This makes parsing code much simpler to read and maintain. Framework is in `sverchok.utils.parsec` module. It can be used by other parts of Sverchok if someone wants to implement another kind of DSL for specific node.
* Split parsing and interpretation to different classes and stages of work.
* Trailing comments are supported.
* Input sockets handling is implemented more like "mesh expression" node.
* Each segment provides data about it's knots, so Knots/KnotNames outputs work for any profile, not only when "from selection" is used.
* I added some profile examples into `profile_examples` directory, and an import operator to node's N panel (similar to scripted nodes).
* As for "from selection" operator... well, for now I can not understand how exactly it works. It is mostly copy-pasted from mk2 implementation.

## Parsing framework

This framework implements the approach known as "combinatorial parsing",
see https://en.wikipedia.org/wiki/Parser_combinator for starters.

A parser is a function, that takes a string to be parsed, and
yields a pair: a parsed object and the rest of the string. It can
yield several pairs if the string can be parsed in several ways.

We say that parser *succeeds* or *applies* if it yields at least one
pair.

We say that parser *fails* if it does not yield anything.

If a call of `parser("var = value")` yields, for example, a pair
`(Variable("var"), "= value")`, we say that the parser *returned* a
`Variable("var")` object; it *consumed* the `"var "` string and *retained*
`"= value"` string to be parsed by subsequential parsers.

The common pattern

```
    parser = ...
    for value, rest in parser(src):
        if ...
            yield ..., rest
```

means: apply parser, then analyze returned value, then yield something and the rest of the string.

A parser combinator is a function that takes one or several parsers,
and returns another parser. The returned parser is somehow combined
from the parsers provided. Parser combinator may, for example, apply
several parsers sequentionally, or try one parser and then try another,
or something like that.

This module provides minimalistic set of standard parsers and parser combinators.

It still has poor error detection/recovery. Maybe some day...

## Profile description DSL

It did not change much since mk2. The most notable changes are:
* Long variable names are supported
* Large part of Python's expression syntax is supported, including some restricted subset of standard functions
*  Expressions which are more complex than just variable name or variable name with negation, have to be enclosed in curly brackets, for example `{a+1}` or `{rho * cos(phi)}`.
* New commands supported: H/h, V/v, S/s, Q/q, T/t.
* C/c, S/s, Q/q, T/t commands support multiple segments, as in SVG.
* Possibility to set default values for variables ("default" statement) and to introduce temporary variables, expressed via others, for later usage ("let" statement).
* Statements may optionally be separated with semicolons (`;`). For some commands (namely: H/h, V/v) trailing semicolon is *required*!

In general, I had the following ideas in mind while creating new version of this DSL:
* Compatibility with previous version would be good where possible, but is not exactly necessary, since there will be some migration process anyway.
* Better compatibility with SVG specification - we are still far from it, but nearer then before.
* More developer-friendly: long names, more lenient spaces and comments, more functions available in expressions, temporary variables ("let") and default values.

Our DSL has relatively simple BNF:

```
    <Profile> ::= <Statement> *
    <Statement> ::= <Default> | <Assign>
                    | <MoveTo> | <LineTo> | <CurveTo> | <SmoothLineTo>
                    | <QuadCurveTo> | <SmoothQuadCurveTo>
                    | <ArcTo> | <HorLineTo> | <VertLineTo> | "X"

    <Default> ::= "default" <Variable> "=" <Value>
    <Assign> ::= "let" <Variable> "=" <Value>

    <MoveTo> ::= ("M" | "m") <Value> "," <Value>
    <LineTo> ::= ...
    <CurveTo> ::= ...
    <SmoothCurveTo> ::= ...
    <QuadCurveTo> ::= ...
    <SmoothQuadCurveTo> ::= ...
    <ArcTo> ::= ...
    <HorLineTo> ::= ("H" | "h") <Value> * ";"
    <VertLineTo> ::= ("V" | "v") <Value> * ";"

    <Value> ::= "{" <Expression> "}" | <Variable> | <NegatedVariable> | <Const>
    <Expression> ::= Standard Python expression
    <Variable> ::= Python variable identifier
    <NegatedVariable> ::= "-" <Variable>
    <Const> ::= Python integer or floating-point literal
```

## Profile Mk2 Backwards Compatibility

In some simple cases, profiles made for mk2 will work with mk3 without modifications. However,  there are the following incompatibilities:

* In Mk2, C/c commands had two last additional parameters — "num_verts" and "even_spread". In Mk3, "even_spread" is completely removed, and "num_verts" is now optional, and is specified after "n=", if you need it. The simplest solution: remove last two parameters of each C/c command.
* In Mk2, A/a commands had last additional argument — "num_segments". In mk3, this parameter is optional and specified after "n=", if you need it. The simplest solution: just remove last parameter of each A/a command.
* In Mk2, in "c" (relative curveTo) command, handle2 was calculated relative to handle1, and knot2 was calculated relative to handle2. This contradicts SVG specification, so it was changed in Mk3. In Mk3, as per SVG spec, handle1, handle2 and knot2 are all calculated relative to knot1. There is no such simple fix for this: you have to recalculate your parameters of "c" commands. I hope that there were not so much people that used "c" command, as it is not so easy to use (and was even harder in mk2).

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [ ] Ready for merge.


portnov:
Screenshot_20190612_235107

```
M 0,0
C 1,1 3,1 4,0 20 0
s 3,-1 1,1 20 
s 3,1 1,-1 20 
s 1,-3 -1,-1 20 
```

portnov:
@zeffii  ping :)

I also plan to:
* add Q/q, T/t commands from SVG spec.
* Remove last argument of C/c ("even_spread") - it was never implemented, it is not in SVG spec, and what is that anyway? :)
* Try to make "num_verts" argument in curve commands optional - not sure if it is possible (if it will not make the syntax ambigous).

zeffii:
COOOL :)

i think Q was supposed to provide an homogeneously spaced vertex list (constant interpolation?), especially for beziers. it was an afterthought, and no-one has needed it.

I look forward to reading your much cleaner/proper solution the the parser :)

zeffii:
> Try to make "num_verts" argument in curve commands optional - not sure if it is possible (if it will not make the syntax ambigous).

maybe as a comment?

portnov:
Since we invent our own DSL here, which is only based on SVG, but is not bound to be exactly SVG, we do not have to add things like "additions in comments", we can invent our own elements of syntax here.
Proposition: use "n = N", for example:

```
C {-0.5+a},{0.5+a} {0.0+a},{0.0+a} {1.0+a},{0.0+a} n=7
```

One can still pass variable or expression as `n`, and even `n=n` is valid (it will mean: number of segments will be equal to the `n` variable).

portnov:
Screenshot_20190613_222812

```
Q 3,H 6,0
t 6,0
t 6,0
t 0,-6
t -6,0
t -6,0
t -6,0
t 0,6
```

portnov:
Screenshot_20190614_210700

portnov:
Screenshot_20190615_110943
```
C 1,1 2,1 3,0 4,-1 5,-1 6,0
s 1,2 0,3 -1,5 0,6
S 1,7 0,6 -1,-1 0,0 n=40
X
```

Screenshot_20190615_131838
```
Q 2,1 4,0
t 1, -3 -3,-3
T 0,0
X
```

portnov:
Screenshot_20190615_133546
```
M radius,0
h {size-radius} ; v {size-radius} ;
q 0,radius -radius,radius
h {-size+radius};
v {-size+radius} ;
q 0,-radius radius,-radius
X
```

portnov:
@zeffii , I think about adding possibility to assign default values to variables (somewhat similar to "mesh expression" node), and to express one variables via other. It could look something like

```
default n = 10; # This is just default value of variable, it will be replaced by value from the input, once it is connected
default radius = 7.5;
default size = 15;

let diff = size - radius; # This variable will not appear as the input, it will be always calculated by this expression

L size, size ...
```

What do you think?

zeffii:
yes, that makes a lot of sense. it's why I have defaults in the scriptnode too

zeffii:
i'm pleased you're doing this. i'll implement the text input for idxviewer in 2.8 asap.

portnov:
penta_profile

portnov:
Screenshot_20190615_185552

```
default straight_len = 1;
default radius = 0.4;

let rem = {radius / tan(phi/2)};

H straight_len ;
a radius,radius 0 0 1
  {rem * (1 - cos(phi))}, {rem * sin(phi)}
  n = 10
l {- straight_len * cos(phi)}, {straight_len * sin(phi)}
```

portnov:
@zeffii 
The node file became a bit fat (1600+ lines); I think I should split the code in several files (one for parsing, one for interpretation, one for node itself). The question is: where to put auxiliary modules? The hierarchy under `nodes/` directory seem to be quite fixed. What if I place them as `utils/nodes/profile_mk3/parsing.py`, `utils/nodes/profile_mk3/interpreter.py`?

zeffii:
i'd love to see `utils/modules/...` folder used more

portnov:
So `utils/modules/profile_mk3/parsing.py`?

zeffii:
yes, looks good!   more directories in `utils` would make that folder more oversee-able.  but that's off topic :)

zeffii:
just a warning about the `F8` , may not include a refresh of the content of `utils/modules` when you are live coding/changing `parsing.py`.

portnov:
Well, it seems to me I implemented everything I want here :)
@zeffii , @nortikin please review and test. I think this will require some testing before we can move mk2 into old_nodes (and remove it in 2.8 branch).

portnov:
Merging. Will wait for bugreports in the master branch :)

portnov:
Screenshot_20190617_212750

```
let dx = {radius + sqrt(radius*radius - y*y)}

M {-length/2},0;
a radius,radius 0 0 0
    dx,y;
h {length - 2*dx};
a radius,radius 0 0 0
    dx,-y;
```

zeffii:
this is certainly the upgrade profile node deserves : )   thank you @portnov .

nortikin:
one of best nodes, proven by time

nortikin:
i wish, from selection to have by default:
```python
default b=1
let c={b*a}```
so other text will contain only c in selected vertices/knots

nortikin:
it not creates A number node for a and not create index viewer also with from selection

portnov:
Personally I do not think that is a good idea... maybe have another operator for that?

nortikin:
> Personally I do not think that is a good idea... maybe have another operator for that?

than need to add this button and for mesh expression also than.
because i personally use it mainly in workflow.

PR #2457, opened 2019-06-16
Bevel Curve node update

portnov:
Introduce more precise mode for taper object interpolation. See N panel
-> "Taper Metric" -> "Orientation Axis".

Default value "Same as Curve" leads to imprecise results in many cases, but I left it for compatibility reasons. Maybe when merging to 2.8 branch, we should set "Orientation Axis" as default option.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


nortikin:
metric = metric
love python

Issue #2420, opened 2019-05-19
Handler for objects in node


Durman:
I have an idea how it is possible to track changes of objects in scene. 

Instead of creating global event like "Some object have been changed" it is easier to bind such handler to a node in node tree. In particularly I suggest create additional button for `objects in` node named like "track changes":

- When the button is pushed function tracking changes is added to handlers. 
- Additional pushing on the button will remove function from handlers. 
- Pushing on `get selection` button alse will remove function. 

I think button for creating handler is most appropriate because it gives understanding that it is not cheap functionality. I already experimented with such handlers during working under alternative update system.

<details> 
  <summary>code</summary>  

```Python
@persistent
def changes_of_objects_handler(scene):
    select_objs = bpy.context.selected_objects
    active_obj = scene.objects.active
    if select_objs:
        if select_objs[0].is_updated:
            # this means object was moved move by user but setting new material also can pass this condition
            debug('{} - is updated'.format([ob.name for ob in bpy.context.selected_objects]))
            tree_updater.changed_object_in_scene.update(select_objs)
            tree_updater.update(props='OBJECTS_UPDATE')
            return
    if active_obj:
        if active_obj.is_updated_data:
            # setting new vertex group, changing geometry and etc.
            debug('{} - is updated data'.format(active_obj.name))
            tree_updater.changed_object_in_scene.add(active_obj)
            tree_updater.update(props='OBJECTS_UPDATE')
            return
        if active_obj.data.is_updated:
            # setting new vertex group, changing geometry and etc.
            debug('{} - DATA is updated'.format(active_obj.data.name))
            tree_updater.changed_object_in_scene.add(active_obj)
            tree_updater.update(props='OBJECTS_UPDATE')
            return
        if active_obj.data.is_updated_data:
            # did not catch this trigger
            debug('{} - DATA is updated data'.format(active_obj.data.name))
            tree_updater.changed_object_in_scene.add(active_obj)
            tree_updater.update(props='OBJECTS_UPDATE')
            return
```

</details>

This function tracks changes of selected objects in scene but can be easily changed for tracking changes of named objects.

What benefits it can achieve. 

- Firstly we can stop using animation during changing objects for getting updating node tree immediately. 
- In second the function can call update from node event that is as fast as possible.

Problems:
Main problem as I see is possibility to create loop in update events. For example, object from scene is gotten, next its matrix is changed and set to this object back that will rise update event again. 

[...]

PR #2458, opened 2019-06-16
live update node - detection feature

Durman:

#2420

## New features of live update mode

- Update `object` nodes only when objects was actually changed. 
- Update only those `object` nodes which objects was changed. 
- Does not update whole node tree during changing of objects.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/ui/sv_panels.py
+++ b/ui/sv_panels.py
@@ -21,8 +21,9 @@ from bpy.props import StringProperty, BoolProperty, FloatProperty
 
 import sverchok
 from sverchok.utils.sv_update_utils import version_and_sha
-from sverchok.core.update_system import process_from_nodes
+from sverchok.core.update_system import process_from_node
 from sverchok.utils import profile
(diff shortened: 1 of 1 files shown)


nortikin:
now blender hanging on sverchok updates.. not tested yet

zeffii:
i think 2.80 API has not yet calmed down for this feature

Durman:
I have noticed problems in detection changes if they happens momentarily like subdivision operation. I do not know what can be done for fixing this. Increasing speed does not give any appreciable effect.

Durman:
I was wrong about speed. The faster the more update events of mesh are catched. But even with speed 1/100 not all events are catched. It looks like using of operator is not appropriate in this case. Scene_update_post handler should be used instead.

PR #2465, opened 2019-06-19
Selection graber

Durman:

Sometimes there is need in selecting part of mesh by hand. 
Short instruction: 
- Generate object via Sverchok. 
- Bake the object on stage when topology won't be changed. 
- Select required elements of mesh. 
- Bake selection in new node.
- Continue working with selected part of the object.

Also this node gives all features in selection which there are in Blender.
Downside - topology of object should be constent.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/object_nodes/selection_graber.rst
+++ b/docs/nodes/object_nodes/selection_graber.rst
@@ -1,0 +1,50 @@
+Selection graber
+================
+
+.. image:: https://user-images.githubusercontent.com/28003269/66333434-06a67480-e948-11e9-98d3-b32130db68d2.png
+
+Functionality
+-------------
+
+The idea of node is to save selection mask from selected object in blender scene. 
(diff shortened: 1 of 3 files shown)


Durman:
Any suggestion about node name? 😕

portnov:
Also check out  discussion in #1320.

Durman:
@zeffii By the way according my observations in Houdini such approach is widely using. I mean creating masks by manual selection of elements. I think it meant that such approach could be quite convenient in procedural modeling.

enzyme69:
Vertex with Selection is one example that would well quite well, if index remains stable. I use it with Skin Modifier. The output is actually "dynamic" because Skin Modifier continuously trying to adapt. 

Screen Shot 2020-04-04 at 11 29 36 pm

PR #2470, opened 2019-06-22
New Boolean 2D node (WIP)

vicdoval:
The code is pretty convoluted but works (almost always)
untitled2

untitled
untitled1

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


Durman:
Great node.))
It looks like difference mode does not work correctly

boolean

vicdoval:
Thanks @Durman  I was thinking if the node should cover that case...
When you perform that difference the node would need to perform a hole creating arbitrary edges to connect it with the other perimeter...
Right now the node does not create any arbitrary edges or vertices as the GGS Boolean does, it keeps the topology creating ngons...
I don't have a perfect idea... so I was thinking to leave it like this and explicit its behaviour in the docs..

Durman:
Well this is better than nothing. Then one question. Is it possible to fix corners? On the picture I see normal corner and cut.

vicdoval:
Are you using the "ngons tesselation" in the viewer draw node proprieties?

Durman:
Yes, it have helped.
It looks like I have found real unexpected behaviour.))

<details> 
  <summary>node tree</summary>  



</details>

boolean

Durman:
And it would be nice to see mask of polygons in Int + Diff mode for dividing them on two groups: polygons which are related with object A and polygons which related with object B.

vicdoval:
> Yes, it have helped.
> It looks like I have found real unexpected behaviour.))
> 
> node tree
> 
> 
> boolean

In this example the problem is the same as in the first case, the first difference outputs a geometry with overlapping polygons wich freaks out the second difference operation

> 
> Also it can't find intersection between coincided edges.
> 
> 

The problem in this case is that the coincident edge in not inside the cutting geometry so it shouldn't be deleted... but is also not outside... so shouldn't be keeped... maybe i should add some propriety to let the user choose...

Durman:
> The problem in this case is that the coincident edge in not inside the cutting geometry so it shouldn't be deleted... but is also not outside... so shouldn't be keeped... maybe i should add some propriety to let the user choose...

As I understand the boolean operation in difference mode should delete from base mesh all parts of mesh which are overlaping by another mesh. So deleting of overlapping polygon which lies on boundary of base mesh is exactly what is expecting from this node in this mode. I don't see usefulness of having any properties about this. If the node could do this job it would be nice.))

vicdoval:
@Durman as @zeffii  says (https://github.com/nortikin/sverchok/pull/2494#issuecomment-512239549) the coincident edges are a tricky business, I started this node very long ago and I have been trapped in solving the coincidences till my personal limit many times...
I solved it for the Intersect mode but the Difference mode (which seems the same) is insane...when I fix one case I break another...again
Now I'm thinking I won't be able to solve it totally and I face two alternatives:
- 1 Leave it like this with some issues on coincident edges/verts in the "Diff mode"
- 2 Split the node into the solved "Intersect mode" and leave the "Difference mode" as an alpha node

Anyway maybe in the development of  #2494  you get to a solution that fixes my troubles...

zeffii:
maybe a useful solution is to have a mode that checks coincident geometry and slightly offsets those verts in input Mesh A that coincide with Mesh B, if the offset distance is relative to the overal Epsilon i imagine the result will generate a few really thin slivers of mesh (which could be removed as a postprocessing step).  --- i can't be certain the output would be useful, but .. it might.

vicdoval:
@zeffii  crazy idea... but works pretty well!!!, as you said it creates some tiny faces and weird flag-looking ngons 🎏  but i think it can be useful in many cases, obviously increases computing time so i will leave it as an option toggled off by default
Im thinking about renaming the node "Polygonal Cutter 2D" or Crop 2D or just Cutter 2D because is not a full boolean node, (misses union)
(screenshots using the node as a 2D remesher) screen1
screen2

zeffii:
better to have something to kind of works in some subset of unfavourable cases, rather than something that works in no subset of unfavourable cases. cool! 

I imagine union would be the less-headache-enducing mode to do, of all modes.

zeffii:
( i say that without ever having tried to write any boolean code ...so..haha colour me grossly misinformed )

portnov:
This should be rebased to b28 branch I think?

portnov:
Also I'm not quite sure what is exactly the difference between this one and #2494. @vicdoval @Durman maybe these two nodes can be merged? Or even if not, I think both can gain from using technologies from each other.

vicdoval:
You are right @portnov. Whenever i find the strength I will face the update and refinement,  and I will probably embed #2494 inside this node...  :D

portnov:
https://jamiethompson.me/posts/GJK-Collision-Detection/ /
https://habr.com/ru/post/472404/
:)

Durman:
I was thinking about an issue with name like 'computational geometry and algorithms library' where links to interesting articles could be collected. This one can be one of them.))

PR #2474, opened 2019-06-24
Turning live update mode from Operator to Handler

Durman:

#2458 
Previous pull request have revealed that Operators are not appropriate for monitoring objects changes. 
I decided that having two different modes of `live update` mode is too much in spite of this approach is completely different. So I have deleted old approach. If it will be found that it had some benefits we can return it back. There is need in testing new approach. I expect that it can become part of update system in 2.8. 

Also i deleted options of enabling this mode in preferences of Sverchok. I don't see that 3D panel is overweight now and can't keep button of this node constantly.

If there will not be any objections against such changes I merge it to master. 

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -82,11 +82,14 @@ def sv_main_handler(scene):
 
 @persistent
 def sv_object_live_update(scene):
+    # This handler is created especially for objects in nodes. When live updated is enabled in 3d panel this handler
+    # detects user changes of objects and call process of appropriate objects in node.
     if not scene.SvShowIn3D_active:
         return
 
     obj_nodes = []
     for ng in bpy.data.node_groups:
+        # search of objects in nodes
         if ng.bl_idname == 'SverchCustomTreeType':
             if ng.sv_process:
                 nodes = []
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -6,6 +6,8 @@ from bpy.app.handlers import persistent
 from sverchok import old_nodes
 from sverchok import data_structure
 from sverchok.core import upgrade_nodes, upgrade_group
+from sverchok.utils.context_managers import hard_freeze
+from sverchok.core.update_system import process_from_node
 
 from sverchok.ui import (
     viewer_draw,
@@ -17,7 +19,7 @@ from sverchok.ui import (
     color_def
 )
 
-
+objects_nodes_set = {'ObjectsNode', 'ObjectsNodeMK2', 'SvObjectsNodeMK3'}
 _state = {'frame': None}
 
 def sverchok_trees():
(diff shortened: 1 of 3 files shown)


zeffii:
You might want to revert the behaviour to like how it was for a number of years, which worked fine for me and others. 

as for the handlers, they are different in 2.8 - and by no means a priority (for Blender Devs). it's not that this isn't an interesting issue (it clearly is.. ).  but the handlers and scene update information is not granular enough imo ( or show me how it is )

Durman:
I understood you. 
In this case I see two ways of developing those features:
- It have to become separate mode. 
- Or it can replace old code but after enough testing. 

Which way will be more welcome?

zeffii:
i think what might happen when updates from `edit_mode` / `object_mode transforms` are obtained by a handler, is the time between tree update request may become dramatically smaller than the time needed by the python interpretter to do the update. Complexity of nodetree downstream and computer speed will of course play a massive roll.  This is another reason why the operator is fine for functional usage,  i don't really get the real need to have instant feedback,.it's not a live performance tool where latency must be as minimal as possible.

> Which way will be more welcome?

i don't know

Durman:
I don't understand idea with timer. 

I'm imagining working of Blender like video game. There is such important parameter like FPS. Translating to Blender language updates events per second. Blender like video game can generate for example 60 FPS. We can reduce the number but what purpose for? If you can computer unable to calculate 60 events per second this number will be reduced automatically. 

If you have timer 13 FPS it won't be work if performance of computer can't handle even with such number of update events. By the way timer can be implemented to handler. There is no difference here with Operator.

Complexity of node tree cause the same effect on performance with Operator and Handler implementation. They have similar code in this part.

> i don't really get the real need to have instant feedback,.it's not a live performance tool where latency must be as minimal as possible.

Here I'm really suprised. 😕
What was all this discussion about? #2420
In this case probably the name of this mode should be renamed. It looks that `live update` is not suitable enough.

zeffii:
> I don't understand idea with timer.

with timer set to something low, it's likely easier for any computer to handle - depending on complexity (but equally as easy to grind the machine to a halt by `unlucky input` / `heavy processing` steps)

i don't know if updates handler events are queued while waiting for python interpretter, but i assume they are.
------

i suspect there are two very different types of Sverchok users. 

1.  set up layout, control everything algorithmically ( sort of static input, but parametric requiring no live user input )
2.  some users mix the features of the tool as a live-interface that depends at times on their input, and processes that input constantly.

with the set of people in group 2 i'm not super interested in entertaining / spending time on. yet again, i guess i'll better leave this to person(s) who do need this live interaction.

Durman:
Okay, I will revert all changes which I done to this mode and won't do any others.

PR #2480, opened 2019-07-08
intersection edges 2d analyzer node

Durman:

http://algorist.com/problems/Intersection_Detection.html

## Solution description

Implementation of finding intersection between set of 2d edges algorithm described in this [book ](https://www.amazon.com/Computational-Geometry-Applications-Mark-Berg/dp/3540779736)based on [swiping line](https://en.wikipedia.org/wiki/Sweep_line_algorithm) approach. 

[Bentley–Ottmann algorithm.](https://en.wikipedia.org/wiki/Bentley%E2%80%93Ottmann_algorithm)

I decided to code on pure python for simplicity without using any library. Actually in some cases it shows better performance than existing `intersect edges` node. Performance of this algorithm is `(n+m)*log n` where `n` is number of input points and `m` is number of intersection. So usage of this node is when there are not too much intersection.

[AVL tree](https://en.wikipedia.org/wiki/AVL_tree) is being added to Sverchok with this node. Was taken from [here](https://github.com/pgrafov/python-avl-tree). Useful data structure which can be used in many geometric algorithms.

[AVL tree visualization](https://www.cs.usfca.edu/~galles/visualization/AVLtree.html)

Another highlight of this node is that it can find intersection when point of one node lies on boundary of another edge or when two edges are overlapped each other.

intersection 2
intersection 1

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/analyzers/intersect_edges_2d.rst
+++ b/docs/nodes/analyzers/intersect_edges_2d.rst
@@ -1,0 +1,42 @@
+Intersect edges 2D
+==================
+
+.. image:: https://user-images.githubusercontent.com/28003269/66367695-49973500-e9a6-11e9-9ff2-a97d68ac8510.png
+
+Functionality
+-------------
+This node just finds intersections of edges.
+
+There are some limitation in implementation of this node:
+
+- According the name of the node it can work in 2d space particularly on 'XY' surface. Z coordinate just ignores by the node.
+- It can handle big number of edges but if there are too much intersection the performance can slow down dramatically. For example if each input edge intersect with every others. In this case only few edges can be handled by this node.
+- Overlapping of edges will lead to stopping of node working.
+- Overlapping of points and point with edge does not lead to error but also such case does not takes in account by the algorithm and does not recognize as intersection.
+
+Inputs
+------
+
+- **Verts** - just vertices
+- **Edges** - just edges
+
+Outputs
+-------
+
+- **Intersections** - points of intersection of edges
+- **Involved edges** - list of edges which produce the intersection in order of intersection points
+
+Further developing of the node
+------------------------------
+
+There are some improvements which can be done for the algorithm:
+
+- increase robustness of algorithm in case when edges overlapping to each other
+- take in account overlapping of points
+- add accuracy parameter (because float numbers are compear with each other)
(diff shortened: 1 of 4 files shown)


portnov:
Maybe just add this feature to existing node?

zeffii:
is the MIT license compatible with GPL3?  i'm uncertain that it is. This _is_ the problem with including external libraries.

cool implementation tho :)

portnov:
MIT is GPL-compatible in sense that you may include MIT-code into GPL-code (by such action, you will be re-licensing MIT-code under GPL; but MIT is very permissive, so it allows such action).

zeffii:
if you do make this into it's own node, and not a mode of the existing intersection node, then we should add a switch/replace feature in the rightclick menu, to hotswap between the nodes.

Durman:
> cool implementation tho :)

Thanks =)

Unlike of existing `intersection edges` node this one does not merge points with initial mesh. But of course output information enough for doing that. All intersection point go out in order from leftmost upward direction to rightmost downward direction and has links to edges which produce intersections. 

But I thought it is worth to have node which only searching intersection without merging them. And it is more explicit to keep such node in analyzer category. 

I think we can have simultaneously this analyser node and extra mode for existing `intersection edges` node which will merge intersection points according this algorithm. As well It will let to keep code in better order if different algorithm located in different files. Existing node can just import function of searching intersection from analyser node and using output of this function for merging intersection in initial mesh.

Durman:
According that Sverchok for 2.79 is not under development anymore I'm going to stop developing of this node on this stage. 
What else could be done here: 
- support of overlapping edges, 
- add new mode to existing `intersect edges` node, 
- create unit test for the node, 
- improve the code

vicdoval:
@Durman  This node was never ported to 2.8. Wouldn't you like to do it?

Durman:
If there is interest I can port it.

vicdoval:
I thought it could be useful for #3599

JanFilipec:
Yes, please port it, this is more or less exactly what I need.

Durman:
I'm not sure that it will help too much. Because you have merge all meshes into one object before the node and it only will say which edges indexes to blame in an intersection points. And after that you somehow should to project this information into initial object what is definitely possible but I'm afraid too tricky for users because it will require python scripting for sure.

PR #2494, opened 2019-07-17
merge mesh 2d - new node

Durman:
## Solution description

Node with really simple functionality of merging two 2d meshes in one with finding all intersections between them.

But it also has output of mask which represents belonging new faces to input meshes.

boolean

It is has the same basic algorithm as intersection edges node: #2480
Actually it still remains a lot of work to do. And I'm not even sure that the node work properly at this stage. I would appreciate any help in testing it. It will be more easier to fix bugs at this moment while the node did not come more complicated.

### Index mask

Since last commit index mask is available. I would say this mask is compromise between flexibility and simplicity. 

This mask gives separately for each output mesh faces index of initial face from mesh A and B. If mesh A and mesh B does not have self intersection then for each output face there is only one related face from A and B. If mesh A or mesh B has self intersection this figure can be increased. This mask ignores such cases and just takes smallest index of related face from A and B. 

Why this mask? This intended to help bind information of initial mesh to final result after the node process. I will show on simple example how this should work.

<details> 
  <summary>example</summary>  

Consider intersection between four squares and circle. In the example below mask index A marks faces with indexes of input mesh A. I decide to mark faces which have not related faces from initial face by -1 value. It is easier to handle.

Suppose the initial squares had information about their height for extrude node. With such mask we can apply such values to new mesh. But output mesh has also some polygons which are related with mesh B only. For such faces the value of height should be generated. Suppose squares will have height 1 and rest of mesh 0. For this we should generate simple mask which will mark only faces related with mesh A. And rest of work will be done by switch node.

This just simple example. With such mask there are plenty of room for creativity.

</details>

New mask can be found on N panel.

To do:
- [x] Handling of coincidence edges case
- [x] Hole detection
- [x] Monotonization of polygons
- [x] Advanced mask

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Ready for merge.


zeffii:
looks cool :)

coincident edges are tricky business, and it some cases too ambiguous to expect a sane output.

Durman:
<details> 
  <summary>tree</summary>  



</details>

untitled

Durman:
untitled

Durman:
untitled1
panel

Durman:
I think the node should not only merge mesh but and any data related with this mesh:

merge face data

Durman:
untitled
untitled1
a+b2

Durman:
I have given some explanation about new index mask in start message. 
One more possibility of using such mask:

searching position

Durman:
untitled

town

nortikin:
just wow!!!

Durman:
Test of showing polygons with holes:
face

zeffii:
coool :)

Durman:
toor

toor

enzyme69:
@Durman Is this node for SV Blender 2.79 only?

Durman:
@enzyme69 Of course no. When I will finish with this PR I'm going to switch to developing 2.8 and add all nodes which was done for 2.79 by me. 

But this will take for a while. This node has quit tricky algorithm and issue with debugging is quite sharp. So in spite of that functionality nearly completed the code should be rewritten in more accurate way for avoiding troubles in future.

Durman:
<details> 
  <summary>tree</summary>  



</details>

Durman:
dcel data visualization for debugging purposes

Durman:
It is difficult to believe but it looks like the node is going to work with overlapping edges.))
coincidence edges

zeffii:
beautiful work!

Durman:
just for examples:

Durman:
I'm proud to say that all functionality that I would like to implement into this node is done. I dived in computational geometry a beat.

I hope this node will become quite useful for most Sverchok users. I think that this node is so important to have that I'm going make advertising of it somewhere like blenderartists. 

Nevertheless there is still a lot work which I would like keep doing on this node. If not Sverchok for 2.8 I would continue to work but I'll push on this stage anyway.

What can be improved in the PR:
- unit test for the node and new modules (avl tree, dcel data structure)
- switch to `dcel` module (actually it implemented now without usage)
- avoiding using any indexes for points
- switching from `avl tree` to `heapq` standard python module where it possible

Durman:
Complex example of cropping a pattern with keeping initial face information:
<details> 
  <summary>tree</summary>  



</details>

crop pattern

portnov:
Is my understanding correct that port of this to 2.8x is #2623 ?

Durman:
yes

PR #2498, opened 2019-07-30
Partitioning polygon

Durman:

http://algorist.com/problems/Polygon_Partitioning.html

## Solution description

Simple node which takes polygon (ordered set of points) split it on Y [monotone pieces](https://en.wikipedia.org/wiki/Monotone_polygon). 

Actually I have done this node only for algorithm testing which I am going to use in `merge mesh` (#2494) node but finally I decided that it can be useful to have such node in educational purpose at least. 

monotone

Great benefit of this node is that it can make polygon monotone even with holes, what can be used for representation such polygons in Blender.

monotone2

This node does not checks whether your polygon and holes have self intersections. If they have result can be wrong or node can return an error.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.


Durman:
monotone3

enzyme69:
This node is intriguing, but I am confused on how to achieve what you did above... 

I tried:
Screen Shot 2019-12-17 at 9 23 57 pm

enzyme69:
However, I seems to figure out:
Screen Shot 2019-12-17 at 9 29 32 pm

enzyme69:
This is interesting:
Screen Shot 2019-12-17 at 9 36 07 pm

And the algorithm failed if the holes are overlapping.

enzyme69:
Slightly strange when it is failing on a Simple Plane Area but works with Circle Polygon Area.

[sverchok_split_004.blend.zip](https://github.com/nortikin/sverchok/files/3972855/sverchok_split_004.blend.zip)
Screen Shot 2019-12-17 at 9 54 54 pm
Screen Shot 2019-12-17 at 9 54 01 pm

Durman:
@enzyme69 Actually I had added this node more for educational purpose than practical.)) I would recommend to use `crop mesh` node if you are interested in hole generation. I soon will add fast mode for this node like I have added for `merge mesh light` node recently.

Issue #2355, opened 2019-01-13
Particles in b28


zeffii:
I add it here as I found an issue that seems to be related to Blender 2.8, not Sverchok.

Issue: empty particle data

How: 
1. add an object on scene
2. add a particle system and tweak it to see particles in the viewport
3. now go to Sverchok and add a particle node
Expected: the particle node gives you the locations of the particles

Test: 
1. select your object and go in Console, 
2. write len(C.active_object.particle_systems[-1].particles) and it returns 0 (!...)

Done with latest January update of blender master repo.

So I suppose we have to wait for either an explanation on how to access particle data or a fix.

_Originally posted by @nantille in https://github.com/nortikin/sverchok/pull/2353#issuecomment-453829503_

zeffii:
yeah, I don't know how permanent the particle API is in 2.80

zeffii:
i will reopen this when there is more clarity from 2.80 devs

nantille:
Well, it seems the workaround turned to be the way we have to access particle data from now on. They just closed now the tickets as they reached a consensus about that.
Bastien Montagne said "this should remain an exception (and a documented one)."
More details here https://developer.blender.org/T60083.

zeffii:
if i'm not mistaken, it is for reading only.. not setting..?

nantille:
I did a quick test and you can set values like the location of the particle *but* the position is not updated in the viewport. As long as you don't change frame in the timeline, the particle location remains the same as what you set in Python (I tried Psys with physics None and also with Newtonian). Now I don't know how to refresh the viewport. Calling bpy.context.scene.update() does nothing.

zeffii:
OK. seems similar to the 2.79 'debacle' then :)  i'm sure there's a way to trigger a viewport update,   maybe `bpy.context.depsgraph.update()`

nantille:
Yes, the particle system is Blender's orphan. It can do a lot of good things but it has issues that haven't been addressed in years, such as wrong rotations when emitting using face normals. Maybe after the release of 2.8...
Depsgraph has no update method so it's probably something else.

zeffii:
maybe
```python
def tag_redraw_all_3dviews():

    for window in bpy.context.window_manager.windows:
        for area in window.screen.areas:
            if area.type == 'VIEW_3D':
                for region in area.regions:
                    if region.type == 'WINDOW':
                        region.tag_redraw()
```

from  ui/callback_view3d

niarchou:
[...]

PR #2511, opened 2019-08-27
Node access to dependency graph - ParticlesMK2 particle data

niarchou:

ParticlesMK2 node returns empty particle data, issue #2355. This happens because the way to access particle data has changed.

## Solution description

The ParticlesMK2 node requires access to the evaluated dependency graph in order to retrieve the particle data. We add a reference to the graph in the node tree, so that the node can get the particle data via the graph (discussed in  #2355).

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -146,6 +146,8 @@ def sv_post_load(scene):
     sv_types = {'SverchCustomTreeType', 'SverchGroupTreeType'}
     sv_trees = list(ng for ng in bpy.data.node_groups if ng.bl_idname in sv_types and ng.nodes)
     for ng in sv_trees:
+        if hasattr(ng, "deps_graph"):
+            ng.deps_graph = bpy.context.evaluated_depsgraph_get()
         ng.freeze(True)
         try:
             old_nodes.load_old(ng)
--- a/node_tree.py
+++ b/node_tree.py
@@ -172,6 +172,18 @@ class SverchCustomTree(NodeTree, SvNodeTreeCommon):
             self.tree_link_count = link_count
             return True
 
+    @property
+    def deps_graph(self):
+        if getattr(self, "_deps_graph", None) is None:
+            print('initializing depsgrah reference')
+            setattr(self, "_deps_graph", bpy.context.evaluated_depsgraph_get())
+        return self._deps_graph
+    
+    @deps_graph.setter
+    def deps_graph(self, value):
+        print('resetting depsgrah reference')
+        setattr(self, "_deps_graph", value)
+
     def update(self):
         '''
         Tags tree for update for handle
--- a/nodes/scene/particles_MK2.py
+++ b/nodes/scene/particles_MK2.py
@@ -45,7 +45,8 @@ class SvParticlesMK2Node(bpy.types.Node, SverchCustomTreeNode):
     def process(self):
         O, V, L, S = self.inputs
         outL, outV = self.outputs
-        listobj = [i.particle_systems.active.particles for i in O.sv_get() if i.particle_systems]
+        evalO = [i.evaluated_get(self.id_data.deps_graph) for i in O.sv_get()]
+        listobj = [i.particle_systems[0].particles for i in evalO if i.particle_systems]
         if V.is_linked:
             for i, i2 in zip(listobj, V.sv_get()):
                 i.foreach_set('velocity', np.array(safc(i, i2)).flatten())


zeffii:
@vicdoval  @portnov @nortikin  how do we feel about this?

i'm tempted to accept the PR and ask for forgiveness later if it breaks stuff :)

zeffii:
i don't think we can safely use `setattr(nodetree, "some_new_property", some_value)` in the scenario of the classes we generate in Blender / sverchok. You'll see very few instances of properties in any Python/Blender classes that don't use the standard set of `bpy.props.StringProperty/BoolProperty....etc`.

we might have to use a `dictionary`

```
class NodeTree:

    deps_dict = {}

```

where we reference the current tree by using the hash of self

```
    # set depsgraph
    self.deps_dict[hash(self)] = bpy.context.evaluated_depsgraph_get()

```

i don't like being so vague, and it might even work as is.. but I think the depsgraph will never be truely cached without using a `deps_dict`

zeffii:
i'm not sure we can hold multiple `dependency graphs`, ie as soon as one is requested it's logical to think that any previously obtained dependency graph becomes invalidated. ( But i have no proof of this )

We can add a dict to the `sverchok` that can be set/get ultimately using 
```
dg = sverchok.blender_deps_graph.get(node_tree.name, None)
if not dg:
    sverchok.blender_deps_graph[node_tree.name] = bpy.context.get_evaluated_depsgraph()
````
we avoid storing the deps_graph inside the instance of node_tree, but rather in the sverchok monolith.

```

vicdoval:
I still dont understand how you overcome the recursion that happens when you call `bpy.context.get_evaluated_depsgraph()` from` bpy.app.handlers.depsgraph_update_post` (or `bpy.app.handlers.depsgraph_update_pre`)

niarchou:
I think dynamically setting the attribute would work for the lifetime of the object (`id(depsgraph)` remains the same when manipulating nodes in the tree), but there might be some other workflow or interaction with other scripts that could cause issues, so I understand that this pattern is undesirable.

I experimented with creating a lookup both at the class and sverchok level and did not encounter any issues. In the former case, we have:

```python
class SverchCustomTree(NodeTree, SvNodeTreeCommon):
    . . .
    depsgraph_registry = {}
    . . .
    @property
    def deps_graph(self):
        dg = SverchCustomTree.depsgraph_registry.get(hash(self), None)
        if not dg:
            self.register_depsgraph()
        return dg

    def register_depsgraph(self):
        SverchCustomTree.depsgraph_registry[hash(self)] = bpy.context.evaluated_depsgraph_get()
```
and in the latter, assuming we add `depsgraph_registry` in `sverchok/__init__.py`:

```python
class SverchCustomTree(NodeTree, SvNodeTreeCommon):
    . . .
    @property
    def deps_graph(self):
        dg = sverchok.depsgraph_registry.get(hash(self), None)
        if not dg:
            self.register_depsgraph()
        return dg

    def register_depsgraph(self):
        sverchok.depsgraph_registry[hash(self)] = bpy.context.evaluated_depsgraph_get()
```
In both cases we need to set the records in `sv_post_load(scene)`:

```python
@persistent
def sv_post_load(scene):
    . . .
    for ng in sv_trees:
        if ng.bl_idname == 'SverchCustomTreeType':
            ng.register_depsgraph()
    . . .
```
I am getting an exception from the ParticlesMK2 node when loading a file, but this seems to happen because the node is processed before the depsgraph is properly set up (when dynamically setting the attribute, I would get an exception from VD Experimental at this stage).

I used `hash()` as a lookup key when adding the dictionary to sverchok as well, because using the tree name would lead to stale entries if a user decides to change it.

If you could let me know which of the two approaches is preferable (class-level or sverchok-level dictionary), I will push the changes so that you can experiment.

Regarding the recursion problem, I have never come across it. I also tried adding a `depsgraph_update_post` callback that only calls `bpy.context.evaluated_depsgraph_get()` and it runs without any issues. Perhaps this used to happen in an earlier blender build? My guess would be that `update()` is not called when the depsgraph is up-to-date.

vicdoval:
Right now the sverchok's main handler is triggered from  `bpy.app.handlers.depsgraph_update_pre` which is great for updating the Bmesh viewers after changing a value.
with the 2.80 if you add `bpy.context.evaluated_depsgraph_get()` from there you get a infinite recursion crash. 
In the 2.81 Alpha it stops before crashing but generates recursion that could be very bad in almost any case.
With `bpy.app.handlers.depsgraph_update_post` this is solved but bmesh viewers update one depsgrah_update later...
Are you creating another handler? or you found another path?

zeffii:
more info so we are all on the same page, off the top of my head so it might be wrong... but it's my working interpretation for a long time. Some of this will restate what has been mentioned here already. Maybe this deserves it's own wiki page that we can reference/maintain

## tree . update()
```
node_tree.update()
```
will be triggered by direct UI interactions, and triggered often (like mouse movements over a UI element in the Node Tree ).  We stopped using this update to trigger downstream updates because it is called many more times than we need, and in scenarios that aren't easily filtered.

## Handlers

The assortment available in 2.8+ differs from 2.79, includes but is not limited to
```
- handler post load: triggered by loading a blend file)
- handler pre save
- handler depsgraph_update_pre
- handler depsgraph_update_post
- handler frame_change_pre
- handler frame_change_post
```
### post_load
Because Sverchok generally does an good job at filtering attempts to recalculate the mesh, we sometimes need to kick it into action by handling `post_load`. Although (if i'm not mistaken) we
generally do perform some form of manipulations in this handler for all .blends loaded.

### pre_save
uhmm..i'd have to look it up

### depsgraph_update_(pre / post)

this sucker (these suckers) is new in 2.8+ and fom what i can tell, it is triggered by things that modify the depsgraph in some way. This might be done by obvious changes, but also some less obvious changes. 

- It is triggered by calling the `context.get_evaluated_depsgraph()`  needed for things like `Object.to_mesh(...with modifiers / with shapekeys / anything)` 
- it is triggered by writing new data to the scene (ie.. by updating BMeshViewer)

### Frame Change

if you take a look at the handler we employ for that, i suggest putting down your coffee, swallow whatever you just drank, and place it away from your computer so you don't accidentally spit it out.
This handler's implementation grew as demands grey for editmode interaction with `Objects In` node. There exists a bit of defensive coding to test if the event was called even if the frame isn't really changed: _ie there was a scenario where the tree would update twice in an animation scenarios with both a `depsgrah_update` and `frame_change` event, some throttling was needed_ 

## node.updateNode()

This is what we use to "cascade" changes in the node output caused by node property changes. (slider update, bool / switch / enum some / UI-string) . This is the most stable / simple update "path"  to trace and contemplate. It is generally not triggered by things outside of Sverchok, unless those external things are consciously/actively calling a node's `updateNode` (or aliases)

zeffii:
the new requirement for some Object methods (like `to_mesh` ) to use `context.get_evaluated_depsgraph()` does make the `Objects In` node of limited usefulness for some users (both static and animated).

But................ the core functionality of Sverchok remains intact, as an add-on for generating new geometry procedurally. It was easy and more fun to build new Sverchok nodes than trying to make nodes that interact with Blender objects/ and complicated idioms.

niarchou:
I see, I reproduced the infinite recursion issue by adding a call to `bpy.context.evaluated_depsgraph_get()` in `sv_main_handler(scene)` and changing the number of particles of the cube. I had previously added the call to a `depsgraph_update_post` handler only, which did not trigger the recursion.

One way to break the cycle would be to add flags that indicate whether one of this handlers is already running and returning immediately if this is the case. In `sverchok/handlers.py`:

```python
pre_running = False
post_running = False

@persistent
def sv_main_handler(scene):
    """
    Main Sverchok handler for updating node tree upon editor changes
    """
    global pre_running
    if not pre_running:
        pre_running = True
        print('DEPSGRAPH UPDATE PRE')
        dg = bpy.context.evaluated_depsgraph_get()
        for ng in sverchok_trees():
            # print("Scene handler looking at tree {}".format(ng.name))
            if ng.has_changed:
                print('depsgraph_update_pre called - ng.has_changed -> ')
                # print('sv_main_handler')
                # print("Edit detected in {}".format(ng.name))
                ng.process()
        pre_running = False

@persistent
def sv_dummy_depsgraph_update_post(scene):
    global post_running
    if not post_running:
        post_running = True
        print('DEPSGRAPH UPDATE POST')
        dp = bpy.context.evaluated_depsgraph_get()
        post_running = False
```
(also, add this entry: `'depsgraph_update_post': sv_dummy_depsgraph_update_post` to `handler_dict`).

I am also attaching the file I am experimenting with.
[particles-test.blend.zip](https://github.com/nortikin/sverchok/files/3568860/particles-test.blend.zip)

vicdoval:
@niarchou  your solution to the recursion problem is so simple makes me feel a little weird 😄 
About where to store the depsgraph i would go for storing it in the sverchok-level, (seems like a global thing) but i don't know if there are any performance differences...

zeffii:
relevant.. https://developer.blender.org/rB411adadf6753715b2c84e41707383f145842a36c

zeffii:
i've committed essentially the same stuff to particles mk2 directly. closing this pr. you are welcome to test :)

zeffii:
sorry about the massive delay @niarchou

niarchou:
Many thanks :)

PR #2512, opened 2019-08-27
New "List Item Insert" and "Mask To Index" nodes

vicdoval:
List Item Insert: The reverse of List Item that can also replace items
List Item Insert
Mask to Index: 

In this PR also:
Renamed "item" input to "index" in List Item Node
Moved "Index to Mask" Node from Alpha to List Mask category/folder
Docs for "Index to Mask"
Icons for the four affected nodes

- [X] Code changes complete.
- [X] Code documentation complete.
- [X] Documentation for users complete (or not required, if user never sees these changes).
- [X] Manual testing done. 
- [ ] Unit-tests implemented.
- [X] Ready for merge.


Durman:
By the way `list mask (out)` node also has `true index` and `false index` output.

vicdoval:
I think you are right @Durman 
Those outputs don't bother me but maybe it would be cleaner to remove them.
What do you think @zeffii , @nortikin ?

zeffii:
don't remove the outputs. List Mask out rocks the way it is.

vicdoval:
Regarding breaking compatibility:
Apart of renaming one of the inputs of the List Item Node I renamed `ListItem2Node` to `SvListItemNode`  following the actual conventions.
This will break the compatibility of this node so there are three solutions.
1 Leave it like this (this is a break point so it would be understandable)
2 Move the old file to old_nodes
3 Undo the changes i have done and leave that node stay as it was.

Any thoughts on this?

zeffii:
upgrade it with the Sv prefix, and stick the old one in the old nodes.

using the 'index' socket does feel more pythonic.

zeffii:
also..   index_insert  seems like a cool node :)

vicdoval:
thanks! @zeffii.   
(I guessed that more pythonic means better also in this case 😄)

Now is ready to be merged :)

nortikin:
all nodes are cool. No need to remove something

PR #2513, opened 2019-08-28
Completed N-panel and right-click menu of Math_mk2

vicdoval:
Just added the function menu to the N-panel and the inputs type (Float or Int) to the right-click menu
This enhancements are particularly useful in big nod-trees because you don't have to zoom in to change the parameters  
Also I changed the order of the list. First the Operators, after trigonometry and constants at the end. I think add/sub/multiply/divide should be the first as they are used more often. Also now in the menu of the node "A" is for Add, "S" for subtract, "M" for multiply... :D 

math_mk2
math_mk22

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/index.md
+++ b/index.md
@@ -178,7 +178,7 @@
 
 ## Number
     SvNumberNode
-    SvScalarMathNodeMK2
+    SvScalarMathNodeMK3
     GenListRangeIntNode
     SvGenFloatRange
     SvListInputNode
--- a/nodes/number/scalar_mk2.py
+++ b/nodes/number/scalar_mk2.py
@@ -34,51 +34,51 @@ from sverchok.utils.sv_itertools import (recurse_fx, recurse_fxy)
 #     2) only add new function with unique number
 
 func_dict = {
-    "--------------TRIG" : "#-------------------------------------------------#",
-    "SINCOS":      (0,   lambda x: (sin(x), cos(x)),      ('s ss'), "Sin & Cos"),
-    "SINE":        (1,   sin,                              ('s s'), "Sine"),
-    "COSINE":      (2,   cos,                              ('s s'), "Cosine"),
-    "TANGENT":     (3,   tan,                              ('s s'), "Tangent"),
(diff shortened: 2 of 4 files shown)


portnov:
I'd suggest to add a call of `super().rclick_menu(context, layout)` to the end of `rclick_menu` implementation. Just in case something interesting will be added into base implementation later.

zeffii:
seems like a useful enhancement : ) 

but... i'm weary of changing the enum numbers... does this play nice with stored jsons / .blends ?

vicdoval:
@portnov  you mean in the global rclick menu of sverchok?

@zeffii  You are right It messes the blend files and works fine on the jsons. If we are going to re-arrange it the moment is now :) , other possibility would be turning this into mk3 and start filling the "old nodes" folder

zeffii:
i intentionally resisted trying to make svb28 compatible with svb279.

but i realize that the math node is one of the more frequently used nodes that will kick people in the painful parts if they suddenly open svb28 test files that aren't working.. Maybe mk3 is justified.

zeffii:
might be a nice opportunity to implement a "replace this node with new version" operator, stick it in the draw_buttons   "upgrade node to mk3".

yeah i might do that tomorrow !  don't move mk2 to old just yet.

portnov:
@zeffii "replace with new version" is more or less already there, see default implementation of `rclick_menu` node method.

zeffii:
> @zeffii "replace with new version" is more or less already there, see default implementation of `rclick_menu` node method.

yeah, OK. I welcome @vicdoval to add the upgrade option to the rclick menu then

zeffii:
@vicdoval  consider 
https://github.com/nortikin/sverchok/blob/94686bb56bc45981b9853a2c5176bad843e87e3a/node_tree.py#L335

vicdoval:
Done :)
By the way, it would be very cool if the "replacer operator" could remove the red frame when done :)

PR #2518, opened 2019-09-04
List input Mk2 with Edges and Pols mode

vicdoval:
Expanded list input to create edges and pols
list_input

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/index.md
+++ b/index.md
@@ -181,7 +181,7 @@
     SvScalarMathNodeMK2
     GenListRangeIntNode
     SvGenFloatRange
-    SvListInputNode
+    SvListInputNodeMK2
     SvRndNumGen
     RandomNode
     Float2IntNode
--- a/nodes/number/list_input_mk2.py
+++ b/nodes/number/list_input_mk2.py
@@ -25,26 +25,46 @@ from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode
 
 
-class SvListInputNode(bpy.types.Node, SverchCustomTreeNode):
+class SvListInputNodeMK2(bpy.types.Node, SverchCustomTreeNode):
     ''' Creta a float or int List '''
-    bl_idname = 'SvListInputNode'
+    bl_idname = 'SvListInputNodeMK2'
     bl_label = 'List Input'
     bl_icon = 'OUTLINER_OB_EMPTY'
 
     defaults = [0 for i in range(32)]
     to3d: BoolProperty(name='to3d', description='show in 3d panel', default=True)
-    
+    def update_max(self, context):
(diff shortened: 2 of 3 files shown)


vicdoval:
I feel this node shouldn't be in the `Number `category.
It could be in `List` or in `Generator `
Any thoughts on this?

zeffii:
probably in `List` : )

zeffii:
personally I think i'd be more inclined to use a `formula node` for manual edge/polygon dictation, out of convenience..  that reminds me,  portnov made `formula mk3` but it wasn't yet ported to svb28

zeffii:
it is not my intention to discourage you, but after consideration i find that approach not general enough. i propose an alternative node for manual topology..

A node similar to List Input, maybe call it Topology Input, with a default set of Outputs (Verts, Edges, Faces). The node could have various modes, "Manual", "From File", "From Object"

## Manual

`Manual` would add `Vertex / Edge / Polygon` categories and input fields, that automatically grow once the last field in each category is no longer empty (or with a `+` sign.. like the execnode). We could allow more liberal parsing of the String Input fields to allow us to drop punctuation, so instead of defining an edge like `[0, 1]`, it could be just `0 1` , same for Faces `0 1 2`

## From File

would let the user pass the name of a internal / external text file with a simple geometry format, and populate the sockets that way. This mode could optionally populate the interface of Manual mode to let the user take it from there. (tho... if autofilling-manual mode this way, there would be a limit to the geometry size)

## From Object

Not unlike the Object In, this too could populate the Manual mode, with the same size limitations

-----
everything can always be stored in a node_dict, or simple StringProperty , as json_from_string if we want.

anyway, these are my thoughts on it.

zeffii:
but.. there's also the `mesh_eval` node..

vicdoval:
I was also thinking about one operator that copy the data from the 3d view, but right now everything is stored in a VectorProperty with a limited size...
My original idea was to be able to add input sockets to the node but after studding the code I thought it was nice to have such a simple node with a little more (Edges and Pols). 
I understand it does not cover so many cases but it does cover many of the basics.
And we have `mesh expression` node....

zeffii:
I sympathize, let me make a simple `topology node` and you tell me if you still want to commit changes to list_input.

zeffii:
https://github.com/nortikin/sverchok/pull/2522

vicdoval:
I see. Your topology node is much more flexible and easier to use, probably is harder to understand at first sight but i guess that is part of the nature of powerful tools..

For the time being I will close this PR and leave `List Input` as it is.

PR #2522, opened 2019-09-07
Topology node

zeffii:
simple topolgy input

image

### DIFF
--- a/index.md
+++ b/index.md
@@ -355,3 +355,4 @@
     SvContourNode
     SvPlanarEdgenetToPolygons
     SvPulgaPhysicsNode
+    SvTopologySimple
--- a/nodes/script/topology_simple.py
+++ b/nodes/script/topology_simple.py
@@ -25,10 +25,10 @@ class SvTopologySimple(bpy.types.Node, SverchCustomTreeNode):
     bl_label = 'Topology Simple'
     bl_icon = 'GREASEPENCIL'
 
-    _edges = StringProperty(default='', name='edges', update=updateNode)
-    _faces = StringProperty(default='', name='faces', update=updateNode)
-    simplified = BoolProperty(name="simplify input", update=updateNode)
-    wrap = BoolProperty(default=True, name="wrap output", description="wraps outputs with a set of [  ]", update=updateNode)
+    _edges: StringProperty(default='', name='edges', update=updateNode)
+    _faces: StringProperty(default='', name='faces', update=updateNode)
+    simplified: BoolProperty(name="simplify input", update=updateNode)
+    wrap: BoolProperty(default=True, name="wrap output", description="wraps outputs with a set of [  ]", update=updateNode)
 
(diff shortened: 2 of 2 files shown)


vicdoval:
I couple of thoughts on this node:

The name seems a little confusing to me. Maybe `Topology Input` or "Manual Topology" or "Topology Dictator"

I don't see this node fitting in any of the actual SV categories. Maybe in the CAD (which is not very clear because everything Sverchok does is CAD) maybe that's why you opened #2523 .... Have you thought already a place for it?

zeffii:
I want a `topology` category  ( for UV Connection, disjoint_meshes ) -- but yeah, `Topology Input` Node seems an appropriate name.
  
The CAD category is an attempt to group those nodes that are close to singular CAD commands, like 
`Fillet/Chamfer, Radial Array, Bisect, CrossSection, Offset, Inset, Extrude, Bevel, Loft.....` (these are very specific commands) ..

i'm not sure I understand the categorical difference between `Modifers make / change` , or the various List subcategories..

Best I can do is attempt to provide a menu (it could be an optional preference in toolbar ) that gives a different perspective on the categories.

zeffii:
some of the new menu items could be Macro calls, for example  we have no `radial array` node, so we could spawn a `circle node + matrix apply` -- maybe.

vicdoval:
> some of the new menu items could be Macro calls, for example we have no `radial array` node, so we could spawn a `circle node + matrix apply` -- maybe.

This kind of behaviour would be very cool

I got the concept of the CAD category, but i thought the name was too wide..
The tools you refer are usually in the `Modify` section of many CAD tools and they usually include what you mentioned and 
In 2D  like LibreCad Move, Rotate, Scale and Mirror, Explode (could be Polygon Boom)
In ·3D like FreeCad Shear and Boolean Extrude, Matrix along Path
In some proprietary software `3d modeling` (sweep, boolean, extrude) is a subcategory of `Modify`..

In Blender 2.79 we had Transforms separated and Mesh Tools with subcategories Deform, Add and Remove
In B28 there are some tools in the sidebar and after they are subdivided in Mesh, Vertex, Edge, Face with some commands repeated with subtle modifications (for example Bevel is in vertex and edges)


The B28 approach i guess would be better integrated with the environment but I think it would require the idea of having macros/operators in the menu for some items (Bevel Vertices in Vertex and Bevel Edges in Edges)

The B279 could be nice and easy to implement but I'm not sure everything will have a clear place. 

After all this thinking a don't have a clear proposal maybe with "Modifier Sanitate" the "Modifier Change" would be cleaner and the CAD category could get back to the modifiers tab.

Modifier Change/Make as I understand them means: If there is geometry added then it goes in the Make. which brings me again to the idea of adding the Sanitate/Remove category. (as in 2.7x)

The List Categories aren't clear for me too but I don't have any reference or expertise in this topic so i don't have anything to add.

Just thinking out loud...

Issue #2523, opened 2019-09-07
New categories for svb28


zeffii:
just thinking out loud here.

To make this more intuitive for new and current coders, we should make menu categories correspond with the file structure (without exceptions). Ideally it would be obvious which categories to use, but some nodes perform very unique tasks, while other nodes perform a great variety - this means it's not always clear in which category to put a node for maximum logical coherency. 

Doing that in the Sverchok app structure is not trivial, but it will happen eventually - What i want to do is have a `Lite Menu`, which is a subset of the main Menu (maybe `Shift + ]`... ) , where a start can be made to re-categorize the nodes, and hide a few nodes

vicdoval:
I have been thinking about this lately, not so much thinking in the coders but in the noders..
Some ideas that maybe could be done straight in the standard menu:
_Y propose moving:
   - Random Vector to the `Vector `Category
   - Bend object along path and along surface to `Transforms`
   - KDT closest edges and KDT closest Path to `Modifier Make`..

_Categories that could be added: 
   - Modifier Sanitate: Delete Loose, Remove Doubles (recalling it Merge by Distance), Limited Dissolve, Recalc Normals...
   - Randomness: Noise, Turbulence, Random Num Gen...
   - Script: SNL, SN Fuctor, Formula, Excec Node Mod, Deform by Formula....

_ Maybe we could offer more than one menu arranging things in a different way for example one by topic and the other by place in the node tree ( for example text_in could be in `text` category in one menu and in the other in `Inputs` category) or offer partial menus (one Inputs menu and one Outputs menu)

zeffii:
i also think Stethoscope should be in Viz node.

i'll add Text socket to indexviewer. sooon.

zeffii:
"sanitize" would be the correct term I think?..

anyway,,, categories are hard.. so expect progress to be slow -- we might not even reach consensus.  for instance i'm not settled on the categories and i've been thinking about this a long time.

but something has to be done.  We could also just have two menus, and put all Blender related interaction nodes in one menu , and all pure sverchok nodes in their own menu.

vicdoval:
I know is a hard topic but as you said something has to be done
I think #2528 is one step and to me it looks far more organised than what we have right now   

We could have a master menu with everything and smaller menus with parts of it (Inputs, Outputs...)

zeffii:
for the time being I would like to suggest keeping that `archaic index.md` derived menu, for simplicity and something to fall back on.

[...]

PR #2528, opened 2019-09-08
Menu restructuration change-make-sanitize

vicdoval:
Restructuring Add Menu based in #2523 and thoughts in #2522
Avoiding double-depth menus
Three categories:
Change: modifies current geometry
change
Make: creates new geometry from existing one
make
Sanitate: Reduces  or simplifies existing geometry
sanitate

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/index.md
+++ b/index.md
@@ -8,7 +8,7 @@
 >
 > Failing to follow these points will break the node category parser.
 
-## Generator
+## Generate
     SvLineNodeMK3
     SvPlaneNodeMK2
     SvNGonNode
@@ -29,7 +29,7 @@
     SvSNFunctorB
     ImageNode
 
-## Generators Extended
+## Generate Extended
     SvBoxRoundedNode
     SvBricksNode
     SvPolygonGridNode
@@ -46,7 +46,7 @@
     SvSuperEllipsoidNode
     SvRegularSolid
 
-## Analyzers
+## Analyze
     SvBBoxNode
     SvVolumeNode
     SvAreaNode
@@ -59,6 +59,7 @@
     CentersPolsNodeMK3
     GetNormalsNode
     VectorNormalNode
+    Pols2EdgsNode
     SvIntersectLineSphereNode
     SvIntersectPlanePlaneNode
     SvKDTreeNodeMK2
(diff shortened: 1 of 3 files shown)


vicdoval:
sanitize

portnov:
I have yet another suggestion: to use YAML file instead of custom-parsed MD. For example,

`menu.yaml`:

```
- category: Generator
  description: Mesh generation nodes
  icon: OBJECT_DATAMODE
  nodes:
    - SvLineNodeMK3
    - SvPlaneNodeMK2
    - SvNGonNode
    - SvBoxNode
    - SvCircleNode
    - CylinderNode
    - SvCylinderNodeMK2
    - SphereNode
    - SvIcosphereNode
    - SvTorusNode
    - SvSuzanneNode
    - ---
    - BasicSplineNode
    - svBasicArcNode
    - RandomVectorNodeMK2
    - SvScriptNodeLite
    - SvSNFunctor
    - SvSNFunctorB
    - ImageNode
  children:
    - category: Extended
      description: More complex mesh generation nodes
      icon: PLUGIN
      nodes:
        - SvBoxRoundedNode
        - SvBricksNode
        - SvPolygonGridNode
        - HilbertNode
        - Hilbert3dNode
        - HilbertImageNode
        - SvProfileNodeMK2
        - SvProfileNodeMK3
        - SvMeshEvalNode
        - SvGenerativeArtNode
        - SvImageComponentsNode
        - SvTorusKnotNode
        - SvRingNode
        - SvEllipseNode
        - SvSuperEllipsoidNode
        - SvRegularSolid
- category: Analyzers
  description: Mesh analysis nodes
  icon: VIEWZOOM
  nodes:
    - SvBBoxNode
    - SvDiameterNode
    - SvVolumeNode
    - SvAreaNode
    - DistancePPNode
    - SvDistancePointLineNode
    - SvDistancePointPlaneNode
...
```

This would allow us:
* get rid of custom MD file parser
* Incorporate category names, icons, description in one file
* Declare the structure of menu decoratively.

I'd suggest to have a convention that in panels we would show top-level menus only (merging all sub-categories into their parent categories), but in menus show sub-categories as sub-menus. 
Another possible convention is to get rid of sub-categories at all, and thus get rid of sub-sub-menus. But in such a case it may appear that we have too many categories and too long menu.

@zeffii  @vicdoval  your thoughts on this?

zeffii:
i'm not ready for that :) -- I'm first leaning towards making all node directories correspond `1-1` with the menus.

i don't know yet how i feel about relying on yaml.  but i'm not against it.

vicdoval:
I started this PR thinking more in the "What" more that the "How" but i didn't find a clear path of acceptation for such a big change.
I could keep on trying to organise the menu to avoid double-depth menus for after pairing the menu with the folder structure but i'm not sure if that is everyone's desire

zeffii:
I think we all agree that the menus/categories are a mess, but we're finding it difficult to state exactly what is wrong with them, and even more murky about what would be better.

this is a classic problem with Nodesystems that have many Nodes. In Jeskola Buzz ( a modulR tracker software for music ) people can script their own "index". This means that people could drop a tonne of nodes they don't like from their menu experience, and even rename the menu labels.

it's why i started LiteMenu, but haven't managed to get around to finishing it. I think it will make sense once you use it.

portnov:
In Blender 2.8, there is a "quick menu". You can add a node to the menu by right-clicking on it's icon or item in the menu and selecting "add to quick favorites". The menu is assigned to the Q key by default.

vicdoval:
I like the idea of the custom menu but anyway we have to supply a basic menu even if it is not the perfect one

zeffii:
in jeskola buzz, the community (thousands of users) eventually converged on the custom menu of a small set of users, and they would supplement/prune the custom menu to fit their exact need. 

once i've added dashed lines to vd experimental, i'll focus on the `menu / directory structure`.

Durman:
The `merge mesh 2d` node should be in `CAD` menu or `modifiers`?

zeffii:
i think CAD

Durman:
Actually I don't understand how describe the CAD category and which functions can be categorized to this category. 😕

Durman:
And should be new folder with name `CAD` be created in nodes folder?

zeffii:
i find the "modifier" category also unsatisfactory, it's too vague.  So is CAD.

zeffii:
@Durman  for the time being place the node in the category that seems most appropriate to you .

zeffii:
can we close this, sverchok has drifted a lot since the PR was made.   it's not a solved problem yet though :)

zeffii:
although short menus using 1..5 hotkey does solve the problem of ever increasing "big menu" size - big menu is an artifact, not a solution. The more we can convince people to use extrasearch and short menus the better their experience will be.

PR #2530, opened 2019-09-10
Litemenu contexthop

zeffii:
This demonstrates the slightly obscure way to dynamically populate menus, there are pros and cons to this approach. I will focus on the positive.. this lets us totally hot-evaluate a new menu and submenus. 

to test, switch to this branch

1.  hit `f3` and type `pop`....ulate , it will show the `populate lite menu` operator
     behind the scenes this populates a CollectionProperty with items, all of which contain a StringProperty called `item.heading`, these headings are used as a submenu lookup.
2.  hit ctrl+shift+`]`  (right bracket) this will display a menu, with submenus (that are just `layout.labels` for now.

----------------

the cons of this approach are, the menu must be populated once per active node_tree.   I suspect this could be done in the builtin nodetree `update` function...


zeffii:
here the `short_menu` dict could be the product of a parsed `litemenu.markdown` file, or something.

zeffii:
this would be my tentative menu, note the unsorted categories...
https://gist.github.com/zeffii/a3c53d151b233b721305310d4eed6455

zeffii:
check it out @vicdoval    the keycombo is now ctrl+shift+A  so you can test :)

zeffii:
am merging this so more people can try it. will soon add a setting in preferences to allow users to set the location of their lite_menu to `/datafiles/... `  `lite_menu.md` would only be a suggestion, and example.

PR #2535, opened 2019-09-15
start pr for textrendering in idxviewer

zeffii:
adresses longstanding regression in idxviewer - unable to draw arbitrary strings on vertex locations.

- [x] allow text input  
- [x] auto repeat, in a meaningful way

"background" feature will not be implemented in this PR.

### DIFF
--- a/nodes/viz/viewer_idx28.py
+++ b/nodes/viz/viewer_idx28.py
@@ -227,8 +227,8 @@ class SvIDXViewer28(bpy.types.Node, SverchCustomTreeNode):
                     for idx, vpos in enumerate(final_verts):
                         concat_vert((idx, vpos))
                     if geom.text:
-                        # autolong text_items to final_verts...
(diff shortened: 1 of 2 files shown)


zeffii:
image

zeffii:
you should feed this text socket, a list which contains a string of text for each input vertex, the structure of the text socket data should match that of the vertex socket data. Else, the node cuts the text input, or auto resizes with last element.

Durman:
Is it possible to do the same thing with faces? =)

zeffii:
not at the moment, use centers polygons mk3 :)

but there is ofcourse no _real_ reason why the polygon centers couldn't be implemented... it's just.. work.

zeffii:
restoring the "text background" feature is higher on my todo list

Durman:
> not at the moment, use centers polygons mk3 :)

Yes, I did not think of such approach, this solution will solve the problem.

PR #2536, opened 2019-09-15
More icons for sv_b28

vicdoval:
menuG
menuTransforms
menuNumber
menuVector
menuMatrix
menuLogic
menuList

- [x] Ready for merge.


zeffii:
i'm concerned about the low contrast on the Scalar math F icon.. but I like the idea of those dots representing the types of sockets the node uses, a lot.

vicdoval:
Here they are!!!
(Added more contrast to Scalar Math as requested)

genrateExtended

analyze
make
change
cad
number
vector
matrix
logic
mask
main
struct
viz
text
scene
layout
bpyData
network
beta
alpha

portnov:
Icons with white lines do not look very good with "Blender Light" theme...

portnov:
If we add icons to all or almost all nodes, we may draw a nice toolbar :)

Screenshot_20190929_152012

Screenshot_20190929_151900

zeffii:
it's amazing what can be communicated just with intelligently placed dots :)  zipnode is cool

vicdoval:
> Icons with white lines do not look very good with "Blender Light" theme...

@portnov 
Maybe the way to go would be trying to load them as svgs as the blender icons and to have custom colors. At this moment they only appear on the shift+A Menu that is dark in both cases. Anyway if it is necessary I can review then to work with both backgrounds

> If we add icons to all or almost all nodes, we may draw a nice toolbar :)

I totally agree 😄 , in this PR there are unique icons for every node (except list descompose because i don't understand it clearly) till the branching time and I will update the new ones soon (if it is not done by someone else).
 Also I would like to polish some  a little bit more but if no one is against it I will merge this PR for easier testing.

@zeffii  Is true, I have learned a lot by making the icons. Its funny how they explain concepts and sticks to the memory at the same time 😄  I suspect that users will get more from Sverchok with a visual guide especially with with the waterfall effect they have when you first open the menu 😂

zeffii:
`list decompose`, is similar in nature to `Vector Out`/`Color Out`.. admittedly i haven't used it for a long time.

zeffii:
merge ahead!

vicdoval:
Annotation 2019-09-29 170424
All the custom icons sort by name

portnov:
@vicdoval 
Screenshot_20190929_205727

Durman:
> If we add icons to all or almost all nodes, we may draw a nice toolbar :)

Totally agree. T panel can get second life.

vicdoval:
@protnov i will have to work on it if i want to get some coherence along all icons and all themes...

The icon panels quite pretty handy
Maybe a fixed buttons with the categories icons on top would be faster than the drop-down menu?

portnov:
You mean a separate panel for each category?
The problem is they would take a lot of vertical space, so you would have to scroll for minutes :)
We could make the panels collapsed by default, but then you will have to open and collapse them manually.

zeffii:
i wish extended search could show the icons... there has to be a way :)

vicdoval:
I mean a top panel with the category icon (Number, Vector, Matrix..) that control which nodes' icons are visible.
I could work as a EnumProperty (when you click one you change the panel) or as a BoolVectorProperty (you could choose many categories at the same time)

portnov:
Interesting idea. Will try.

portnov:
@vicdoval unfortunately it's not so easy to make icon panel for categories, since we actually do not have icons for node categories. We have icons for submenus, but one submenu contains several node categories...

zeffii:
https://blender.stackexchange.com/a/73878/47

> EnumProperty does not have a way to display custom icons.

PR #2463, opened 2019-06-18
Profile mk3 update

portnov:
This adds "n=" option for straight line commands — L/l, H/h, V/v.

## Preflight checklist

Put an x letter in each brackets when you're done this item:

- [x] Code changes complete.
- [x] Code documentation complete.
- [x] Documentation for users complete (or not required, if user never sees these changes).
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

PR #2543, opened 2019-09-21
Port Profile node mk3 to 2.80 branch

portnov:
This ports  #2450 and #2463.


nortikin:
Why profile node not supports auto adding nodes (maybe with special defgault-disabled flag)?
profilemk3

PR #2551, opened 2019-09-23
Update jig

zeffii:
this recreates @DolphinDream 's switch input node, from scratch - to better understand the issues in this node.

i try to commit to this when I have time.

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -180,6 +180,8 @@ class SverchCustomTree(NodeTree, SvNodeTreeCommon):
         # print('svtree update', self.timestamp)
         self.has_changed = True
         self.has_link_count_changed
+
+        print('going to do nodetree.process()')
         self.process()
 
     def process_ani(self):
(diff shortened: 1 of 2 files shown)


zeffii:
it occurs to me that doing 
```python
socket.hide_safe = some_bool
```
will disconnect such a socket, if the state of the socket is already not-hidden, and the property is being bassed `False`. This should perhaps be a no-op.

I can't quickly think of a reason why setting the hide_safe would disconnect a socket that way.

zeffii:
so currently
```python
    @hide_safe.setter
    def hide_safe(self, value):
        # handles both input and output.
        if self.is_linked:
            for link in self.links:

        self.hide = value
```
would maybe be
```python
    @hide_safe.setter
    def hide_safe(self, value):
        # handles both input and output.
        if self.is_linked:
            for link in self.links:
        if self.hide == value:
            return
        self.hide = value
```

zeffii:
yikes.
image

zeffii:
see https://github.com/nortikin/sverchok/pull/2554

DolphinDream:
Why "yikes" ? This looks cool :)

PR #2552, opened 2019-09-23
NGon node: add "divisions" input.

portnov:
Add possibility to divide each side of the polygon to several segments.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


portnov:
Screenshot_20190923_202257

Durman:
It would be nice to have separate node which could divide edges of any mesh. 
I thought about implementation of such node but it has low priority in my list of `todo` at least at this moment.

PR #2553, opened 2019-09-23
Optionally register node category panels for the N or T panel.

portnov:
In Blender 2.80, the T panel does not support tabs; so there are no automatically registered panels for node categories under T panel.

This adds the following options to preferences:

* Display node buttons:
  * Under T panel (with popup menu for categories selection)
  * Under N panel (categories are selected with tabs)
  * Do not show (default)
* Display icons only: if set, then for nodes that have icons, only icons will be shown, to save screen space. Checked by default.
* Number of columns to show in "icons only" mode. 4 by default.

Screenshot_20190929_153810

With new icons from #2536:

Screenshot_20190929_152012

Screenshot_20190929_151900

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/menu.py
+++ b/menu.py
@@ -211,6 +211,32 @@ class SverchSeparator(object):
     def poll(cls, context):
         return context.space_data.tree_type == 'SverchCustomTreeType'
 
+class SvPackedLayout(object):
+    def __init__(self, parent, columns=4):
+        self.parent = parent
+        self.columns = columns
+        self._column = 0
+        self._row = None
+
+    def separator(self):
+        self.parent.separator()
+
+    def operator(self, operator_name, **params):
+        if 'icon_value' in params or 'icon' in params:
+            if self._row is None:
+                self._row = self.parent.row(align=True)
+                self._row.scale_x = self._row.scale_y = 1.5
+            params['text'] = ""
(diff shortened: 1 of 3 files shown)


portnov:
Screenshot_20190923_220218
This is what it looks like when enabled.

zeffii:
fine .. but is that really useful / necessary ?    i've never grabbed nodes from that toolbar..

portnov:
I used to use it :)

portnov:
Not sure if the toolbar in the N panel will be as useful as T one, though.

zeffii:
fair enough : )

Durman:
`Presets` panel was definitely useful. I think it should be somewhere. Not sure about the rest of `T` panel.

portnov:
Presets panel is still under T.

portnov:
I managed to get node buttons under the T panel, where they were in 2.79. But T panel still does not support tabs, so there is popup menu of categories instead of tabs.

Screenshot_20190928_191359

matteolegna:
Thanks! I think this is very useful for beginners. My two cents are that it should be on by default... 🙄

PR #2554, opened 2019-09-26
Update jig redux

zeffii:
fixes


zeffii:
it's definitely a headscratcher @DolphinDream .

it seems like there's potentially fewer update calls generated by a single huge socket initialization (+ hide_safe) , in some cases -- but depends on how the node is used.

it's not silly to offer a baseclass node, and have 3 nodes `InputSwitch4*4`, `inputSwitch5*5` .... `+++9*9`

but ofcourse, the baseclass would need to be superslick first.

PR #2555, opened 2019-09-27
Update jig redux2

zeffii:
slight function shuffle


zeffii:
bugs remaining, off top of my head
- selected_set  can exceed num_switches, the question is ... do we 
     - 'cycle/modulo' that value or
     -  hard limit it via prop(set=,  get=)
     -  or hard limit via check in process (remap_indices) function

PR #2556, opened 2019-09-28
B28 throttle update more

zeffii:
add update throttling during `sv_init` , socket creation will no longer trigger process


Durman:
According my exploration here: #2770 creating of new sockets does trigger any update events only handler but handler is not used any more: #3106. I don't see what caused this PR.

zeffii:
> I don't see what caused this PR.

well, if it isn't needed anymore i have no love for it either.

(but it's safe to say that this was added because it _was_ needed at the time )

PR #2557, opened 2019-09-28
implement fast dirty noise displace

zeffii:
makes it easy to do a noise displacement


zeffii:
i'll make a macro instead.

PR #2562, opened 2019-09-30
add draw_gl_polygonoffset property

zeffii:
adds option in buttons.ext , but maybe we'll automatically set it to True...

### DIFF
--- a/nodes/viz/vd_draw_experimental.py
+++ b/nodes/viz/vd_draw_experimental.py
@@ -192,7 +192,9 @@ def draw_fragment(context, args):
 def draw_faces(context, args):
     geom, config = args
 
-    # bgl.glDisable(bgl.GL_POLYGON_OFFSET_FILL)
+    if config.draw_gl_polygonoffset:
+        bgl.glDisable(bgl.GL_POLYGON_OFFSET_FILL)
+    
     if config.display_edges:
         draw_uniform('LINES', geom.verts, geom.edges, config.line4f, config.line_width)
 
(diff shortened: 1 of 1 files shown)


portnov:
Why isn't it enabled by default?..

zeffii:
because it's still tentative.  

i'd be happy to enable by default if someone can show me how to ensure that the opengl state is fully reset after leaving our draw callbacks. currently we have a few parameters that are reset when letting go, but not polygonoffset related stuff.

for some reason there appears to be some interaction with Blender's own viewport drawing.

zeffii:
the condensed version of this new implementation is
```
bgl.glDisable(bgl.GL_POLYGON_OFFSET_FILL)
draw_edges(......)

bgl.glEnable(bgl.GL_POLYGON_OFFSET_FILL)
bgl.glPolygonOffset(1.0, 1.0)
draw_faces()

bgl.glDisable(bgl.GL_POLYGON_OFFSET_FILL)
```

zeffii:
what i would like to guarantee is 
```
get_gl_state()

bgl.glDisable(bgl.GL_POLYGON_OFFSET_FILL)
draw_edges(......)

bgl.glEnable(bgl.GL_POLYGON_OFFSET_FILL)
bgl.glPolygonOffset(1.0, 1.0)
draw_faces(.......)

bgl.glDisable(bgl.GL_POLYGON_OFFSET_FILL)

revert_gl_state()
```

zeffii:
i think the only property that needs to be tracked is the state 
stored by passing `glPolygonOffset(1.0, 1.0)`

zeffii:
so.. maybe..  https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glPolygonOffset.xhtml
```
glIsEnabled with argument GL_POLYGON_OFFSET_FILL, GL_POLYGON_OFFSET_LINE, or GL_POLYGON_OFFSET_POINT.

glGet with argument GL_POLYGON_OFFSET_FACTOR or GL_POLYGON_OFFSET_UNITS.
```

zeffii:
hmmm.. we don't have acces to `glGet`

zeffii:
so.. we use `bgl.glGetFloatV` , maybe

zeffii:
```python
value = new <bgl.Buffer - type GL_FLOAT>
bgl.glGetFloatV(bgl.GL_POLYGON_OFFSET_FACTOR, value)
print(value)
```

PR #2574, opened 2019-10-01
VD attribute node

zeffii:
gives a dedicated node to ViewerDraw for setting internal properties of that node. using a dict.

- [ ] complete 1-1 ability to control current VD EXp attributes
- [ ] add quicklink span operator to attr-socket (to save us from hoving to locate it)

for now the dict would (if entirely specified by the user) look like:
```python
[
dict(
    vector_light="light direction (3f)",
    vert_color="points rgba (4f)",
    edge_color="edge rgba (4f)",
    face_color="face rgba (4f)",
    display_verts="display verts (b)",
    display_edges="display edges (b)",
    display_faces="display faces (b)",
    selected_draw_mode="shade mode (enum)",
    draw_gl_wireframe="wireframe (b)",
    draw_gl_polygonoffset="fix zfighting (b)",
    point_size="point size (i)",
    line_width="line width (i)",
    extended_matrix="extended matrix (b)"
)

]
```

### DIFF
--- a/index.md
+++ b/index.md
@@ -147,6 +147,7 @@
     SvListModifierNode
     SvFixEmptyObjectsNode
     SvDatetimeStrings
+    SvVDAttrsNode
 
 ## List Main
     ListJoinNode
--- a/nodes/list_mutators/vd_attr_node.py
+++ b/nodes/list_mutators/vd_attr_node.py
@@ -13,6 +13,22 @@ import bpy
 from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode
 
+maximum_spec_vd_dict = dict(
+    vector_light="light direction (3f)",
+    vert_color="points rgba (4f)",
+    edge_color="edge rgba (4f)",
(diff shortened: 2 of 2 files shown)


zeffii:
prelim node ui..
image

Durman:
Interesting concept.)) 
It would nice probably if some of the inputs could have default properties. In this case for example this node can be useful for setting one color for multiple vizualization nodes.

zeffii:
i'm going to keep this node super simple with respect to the sockets, but there might be a `layout.ui` in the draw_buttons_ext where a user can set defaults.

vicdoval:
Why is this node in the "List Mutators"  (so in the "List Menu")?

zeffii:
yes!  why :)  i suspect attr nodes will get their own category, plenty of nodes could have an accompanying attr node.  For now i just when with list mutator.

zeffii:
it doesn't strictly need to be shown at all in the menus.. the `quicklink` will spawn an `vd attr node` directly in place.

PR #2576, opened 2019-10-02
Vd attrs node draw group

zeffii:
image

### DIFF
--- a/index.md
+++ b/index.md
@@ -147,6 +147,7 @@
     SvListModifierNode
     SvFixEmptyObjectsNode
     SvDatetimeStrings
+    SvVDAttrsNode
 
 ## List Main
     ListJoinNode
--- a/nodes/list_mutators/vd_attr_node.py
+++ b/nodes/list_mutators/vd_attr_node.py
@@ -11,6 +11,7 @@ from bpy.props import BoolProperty, StringProperty, IntProperty, CollectionPrope
 from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode, enum_item_5
 from sverchok.nodes.viz.vd_draw_experimental import SvVDExperimental
+from sverchok.sockets import socket_colors
 
 sock_str = {
     'enum': "SvStringsSocket",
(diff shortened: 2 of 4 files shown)


zeffii:
vd_attr_node

zeffii:
image

zeffii:
image

zeffii:
image

zeffii:
i'm done for now. this doesn't work in json yet.

Durman:
I think that probably it would be more clearly for users if viewer node would hide the overrided by attr node properties.

zeffii:
yes @Durman, but also some nuance here.  I have thought about this. by removing the properties on the VDExperimental node, i find we loose the easiest way to see the current state of the VD node.  

Perhaps over time a better solution will come, but i'm also for some automatic UI streamlining of VDExperimental when attrs is connected. (maybe display a single row layout showing colours of vert/edge/face, but non-interactive.

a slightly more pressing matter is the IO json storage of attrs node, next on my todo.

Durman:
Ok. Besides, is it possible to visualize a mesh with all face with different random colors with this node?

zeffii:
VD Experimental doesn't support that. I'd like to not make that node more complicated. i'm thinking about alternative VD nodes, maybe a hybrid which can also output `Mesh` or `OpenGL`.

I like fresh starts :)

zeffii:
maybe an `SvMesh` Socket :)

but i also want a `openGL render` function, where we pass a 'virtual camera' object (or use the Viewport Eye) - and experiment with anti-aliasing.

Durman:
It would be nice to have such functionality. As I see it is not too much important but not useless. 

If it were possible I would try to use in showing debug data of half edge data structure. At this moment, I have to use several viewer nodes for showing selected half edge in different color then the rest of them. It is not too much comfortably but nevertheless it works for me.

<details> 
  <summary>img</summary>  



</details>

zeffii:
yeah! a `Dev Viewer` node, with many options to help devs debug algorithms.

maybe you will start using  `Svb28`  then ? :)

Durman:
Well I would like to close all my opened PRs first. I even cut my todo list in `merge mesh` node but work still moving slowly. I hope that in several weeks I'll finish with it and will be able to switch to 2.8 even without `Dev viewer` node.))

PR #2598, opened 2019-10-09
B28 vdexp w dashes

zeffii:
implements dashed line drawing. 

but this needs a refactor - i'm not comfortable with it in this state.   but the draw results are cool and customizable with dash-presets :)

image

image


zeffii:
@portnov  interesting failure in the automated tests.

zeffii:
this (dashed stuff) is super unstable. it might require a rewrite based on everything learnt during implementation..

zeffii:
hmm.. it turns out my machine was unstable. seems OK.  but needs a refactor.

PR #2602, opened 2019-10-09
B28 vdexp w dash reshuffle

zeffii:
this adds dashed lines to vd experimental -- not superduper tested, but it's not ultra easy to break anymore. Seems stable.


zeffii:
the UI for the dashed config will be tweaked eventually, but i wanted to get this in early.

PR #2603, opened 2019-10-09
Add "level" parameter to "select mesh elements by location" node

portnov:

"Select mesh elements by location" node is usually used to process one or several meshes. But sometimes it is required to process a list of lists of meshes, or even more complex structures.

## Solution description

This PR adds a "level" parameter to the node. Minimum and default value of the parameter is 2, which corresponds to old behaivour. 

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


Durman:
Increasing nestedness perhaps is not best solution. 
The more complex data structure the more hard to understand.

PR #2609, opened 2019-10-11
Revert "Litemenu contexthop"

zeffii:
Reverts nortikin/sverchok#2530


zeffii:
instead we should fix the real issue .    node directories.

Issue #2601, opened 2019-10-09
Nodes to forget


vicdoval:
Any reason to keep this nodes out of old_nodes?
CylinderNode
SvProfileNodeMK2
CentersPolsNodeMK2
SvConvexHullNode
Formula2Node
SvInterpolationNode
EvaluateLineNode

zeffii:
nope.

vicdoval:
The EvaluateLineNode has different bl_idname/class_name.
This unables the old_nodes automated system to work.
I suggest to delete it and don't look back.
Any better idea?

zeffii:
so much stuff will just fail - this is not unintentional.   :D

zeffii:
put it this way, no single feature (that i'm aware of) has even been truly 'regressed' out of sverchok, it's always been a matter of improvement / increase convenience.

zeffii:
only regressions are features disappearing/changing from `bpy`, or `bmesh` -- where no one has been interested yet in implementing them afresh. 

like now with the particle nodes being developed by jacques, it feels almost redundant to have any attempt of b28 sverchok-nodes to control particles yet ..

vicdoval:
So you are fine with me deleting it...?

zeffii:
yes. do it.

zeffii:
all particle nodes, will have no use in blender 2.82 when the current particle system becomes end of life.

i suspect the built in particle node system will be sufficient, and to some degree 'remote controllable' via sverchok :)

portnov:
This is done as far as I see....

PR #2614, opened 2019-10-12
Polishing icons, forgetting nodes, matrix nodes to matrix category

vicdoval:
A lot for a PR but I hope you don't mind in this case.
The PR covers this topics
_Polishing some icons and adding some more to the new nodes.
image

_Take action on #2601
_Matrix category filled with Beta Matrix Nodes
image
image

_Added Matrix Normal Docs
- [x] Ready for merge.


portnov:
It seems tests were using old nodes. Could you please update tests to use new nodes?...

vicdoval:
Well..., i was deleting the references to the old nodes... not sure how to deal with some of the tests.. but maybe i should try...

vicdoval:
I did some more cleaning also matching folders and menu and 
Inset Special Docs

CAD menu and folder
image
Transforms Menu and folder
image

I am not very attached to this changes but I think they go in the right direction so if nobody is against I will merge all this changes and they can always be discussed and reverted later

vicdoval:
I am going to merge this PR before I start moving everything (and I leave that to you @zeffii )  :D

PR #2618, opened 2019-10-14
B28 autodpi 3

zeffii:
address DPI/scale issues for nodeview drawing.

### DIFF
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -137,6 +137,11 @@ def sv_post_load(scene):
     Upgrade nodes, apply preferences and do an update.
     """
 
+    # ensure current nodeview view scale / location parameters reflect users' system settings
+    from sverchok import node_tree
+    node_tree.SverchCustomTreeNode.get_and_set_gl_scale_info(None, "sv_post_load")
+
+
     for monad in (ng for ng in bpy.data.node_groups if ng.bl_idname == 'SverchGroupTreeType'):
         if monad.input_node and monad.output_node:
             monad.update_cls()
--- a/docs/404_custom.html
+++ b/docs/404_custom.html
@@ -11,7 +11,7 @@
     <body>
 
         <div id="vh" class="box">
-            Sorry! It appears that <span id="label">Get property</span> doesn't have a document associated with it. 
+            Sorry! It appears that <span id="label">Planar edgenet to polygons</span> doesn't have a document associated with it. 
             <br/><br/>
             Please contact us on the <a href="https://github.com/nortikin/sverchok/issues">GitHub issue tracker</a> or any of the following social sites: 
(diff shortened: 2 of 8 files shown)


zeffii:
i'd appreciate if people would test this, it seems to work here.. but i don't have high res hardware to test on.

zeffii:
in `sv_init` , those nodes that draw to nodeview will get_set the dpi info. This will be fine for a majority of people .  (if the code is doing the right multiplying)

zeffii:
@DolphinDream   before i go ahead and merge this, you should test it.

DolphinDream:
It seems to work fine for me. I reset blender prefs to factory defaults and tested this branch..
The easing node, texture viewer node seem to display output in the right place and with the right size.. The stethoscope seems to display the text in the right place and with a reasonably sized font though it appears to be smaller than the size of the text in the node labels, which to my taste, is a bit smaller so I will probably have to make my sv preferences for the stethoscope size a bit larger than 1.0.

SV-dpi-test

DolphinDream:
After resetting the blender to factory defaults and resetting sv prefs.. (to scale/multipliers = 1.0) the stethoscope, easing and texture nodes all seem to have their render offset and scaled incorrectly.

With the sv preferences set to 2.0 the scale, size and location appears to be correct.

zeffii:
@DolphinDream , currently what happens is the scale configuration is only updated when a new node is created ( the update is called in sv_init ). Therefore when an existing .blend is loaded, and the `sv_init` are not called, the scale settings are not checked/updated.

I can add a call in `post_load` handler. Please try again after you see a commit below with these changes.

zeffii:
please try that again @DolphinDream

DolphinDream:
The Scale/Multiplierxy in sv pref seem to be set to 2.. i try to set the back to 1.0 and save settings but they keep coming back as 2. If these settings are 1.0.. the scale/location error persists. I restarted blen der, new file etc.

zeffii:
the idea is now that you no longer need to set those values manually.

zeffii:
@DolphinDream 

this PR should detect the users system settings on two scenarios
- user adds a node to nodeview (in particular nodes that draw to the nodeview using bgl etc..)
- or user loads a .blend (so if your default .blend has a sverchok nodetree, you'll see some printing in the console that parameters have been set / get )

logically the current code should work as expected, maybe you can investigate a little, i'm losing patience with it.

DolphinDream:
> the idea is now that you no longer need to set those values manually.

of course.. That's what I was expecting.. however these sv prefs setting are still affecting the rendering location/scale.

zeffii:
@DolphinDream   in your console/terminal you should be setting some print outs... and it should be saying `scale  2.0`

DolphinDream:
It seems to be working fine now. I see the preference values are no longer settable by the user but rather displayed as info (2.0 value in my case, which fixes the location/scale).

zeffii:
ok. cool. i'm going to push..anymore issues we can fix later.

PR #2619, opened 2019-10-14
UI for GitHub gist creation API token.

portnov:

This addresses a long-standing issue of Sverchok not being able to export node trees to gist.github.com without manual and tricky creation of some file, which contained your unprotected github password.
The issue was also discussed in #2197.

## Solution description

* Add an instruction for users at https://github.com/nortikin/sverchok/wiki/Set-up-GitHub-account-for-exporting-node-trees-from-Sverchok about how to create a "github access token"
* Add a setting into preferences to store that access token
* Use that token for authentication when creating gists
* Remove old "wad" file support - b28 branch is a point to drop this kind of compatibility
* Show "export to gist" button always; but if user has no github access token configured, display a message and open documentation page in the browser.

Known issue: when trying to open documentation page, a python error may appear. This bug is fixed in Python 3.7.1. 
https://bugs.python.org/issue31014
On my system, the error appears when I press the button first time; but if I press it second time, there is no error.

This PR also takes care of "CERTIFICATE_VERIFY_FAILED" error, which may appear in some blender builds when trying to access HTTPS URL. As far as I understand, the problem is that on some platforms Blender's built-in python does not use system's trusted certificates storage, so we have to use python's own storage, which is accessible via `certifi` package.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -8,8 +8,8 @@ from sverchok import data_structure
 from sverchok.core import handlers
 from sverchok.core import update_system
 from sverchok.utils import sv_panels_tools, logging
+from sverchok.utils.sv_gist_tools import TOKEN_HELP_URL
 from sverchok.ui import color_def
-
 
 def get_params(settings_and_fallbacks):
     """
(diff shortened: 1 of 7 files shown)


portnov:
This has to be tested on all supported platforms:

[x] Linux - tested on Debian testing/unstable by @portnov 
[ ] Windows - could someone please test? Maybe @zeffii ?
[ ] MacOS - could someone please test? Maybe @enzyme69 ?

zeffii:
i'm happy to test win64 v10

zeffii:
yeah works here!

zeffii:
good man @portnov  i didn't know how to do that :)  wasn't too bad after all.

PR #2624, opened 2019-10-18
Build docs with travis (b28)

portnov:
Build documentation upon repository update and publish it in github releases.

This will make Travis CI automatically build HTML documentation and publish it as a github release. This will be done only if:
* This commit is a push towards the master branch
* AND there were changes in the `docs/` directory, or in `.travis.yml`, or in `build_docs.sh` files.

Automatically created releases are marked as draft, so they are not public by default - we can remove them, or rename, or whatever, before publishing.

We may want to update this condition, for example, to make it deploy only if commit message contains "MAKE_RELEASE" string. Or take option which travis suggests: make release only if the commit has any tag.

NB: Travis can not automatically deploy to releases from pull-request branches.

I've set up similar thing for master branch, so here we go, first automatic release with documentation (it is marked as draft, so it is not public yet): https://github.com/nortikin/sverchok/releases/tag/untagged-88b88e6e7be0a07b4cb2

@zeffii  FYI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,6 @@ branches:
     - master
     - b28_prelease_master
 
-addons:
-  apt:
-    packages:
-      - python3-sphinx
-
 install:
   - |
     #set -x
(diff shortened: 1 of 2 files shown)


nortikin:
GREAT. Not always i do updates on docs.

PR #2626, opened 2019-10-18
Test documentation references consistence.

portnov:
This adds test cases to check that:
* all node documentation `rst` files are referenced from their corresponding `_index.rst` files.
* all files mentioned in each `_index.rst` file do exist.
* for each node module, there exists corresponding `rst` file.

I also tried to fix broken links where possible.


portnov:
There is one test failed: there is `docs/nodes/generator/line_mk3.py` instead of `line_mk3.rst` in documentation. And that is really a python file, not just bad file name! :)

@vicdoval ping

portnov:
I took `line_mk3.rst` file from master branch - there it existed for some reason...

portnov:
I decided to tolerate current problems with missing nodes documentation for now, as there are too many problematic nodes to solve this problem right now. I moved that problem into #2627 .
The test for missing documentation will be still useful: it will fail if someone will add a node without documentation.

PR #2628, opened 2019-10-19
Docs ordering b28

portnov:
* Add a test case which checks that each `.rst` file under `docs/nodes` has corresponding `.py` file (i.e. that we do not have obsolete/hanging documentation files)
* Move documentation files around to make that test happy
* Move old nodes documentation into `docs/old/nodes` for now, maybe someone will discover that some of them are the best thing that we have for some undocumented node.


vicdoval:
I have just merged pulga_pysics docs in  #2629

PR #2630, opened 2019-10-19
add example of onboard help

zeffii:
adds console context awareness to node.help invocation

image

something like this is possible, the exact visual representation and whether `n.help` should be added to the history before displaying items is for later implementation

### DIFF
--- a/nodes/viz/vd_draw_experimental.py
+++ b/nodes/viz/vd_draw_experimental.py
@@ -348,13 +348,8 @@ class SvVDExperimental(bpy.types.Node, SverchCustomTreeNode):
     @property
     def help(self):
         if bpy.context.area.type == 'CONSOLE':
-            from console_python import add_scrollback #, get_console
+            from console_python import add_scrollback
             history_append = bpy.ops.console.history_append
-            
-            # get line in console, add to history to show it as 'handled'
-            # m = bpy.context.space_data.history[-1].body
-            # m = m.strip()            
-            # history_append(text=m, remove_duplicates=True)
 
             # print the lines to the console
             add_scrollback(self.console_help, 'INFO')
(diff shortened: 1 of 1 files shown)


zeffii:
but a real drawback about `console` is that it's a dangerous place after the users hits `f8` (scripts.reload)  , any lines previously evaluated will point to a void when you try to execute them, and blender will exit to desktop

```
nodes = bpy.data.node_groups['NodeTree'].nodes
n = nodes['VD Experimental']
n.help
```
if you do `f8` and then
```
n.help
```
blender can vanish, because references are also wiped.

zeffii:
we can print to the console from a node directly, if console is open..

Durman:
Do you know the method for this? I am occasionally wondering if is it possible.))

zeffii:
yeah, you iterate through spaces.. i'll add example code (temporarily)

zeffii:
```python
import bpy
from console_python import get_console

AREA = 'CONSOLE'
for window in bpy.context.window_manager.windows:
    for area in window.screen.areas:
        if not area.type == AREA: continue

        for region in area.regions:
            if region.type == 'WINDOW':
                console, stdout, stderr = get_console(hash(region))
                print(dir(console))
        
        for s in area.spaces:
            if s.type == AREA:
                print(dir(s))
                break

```
i thought I'd figured out how to print to an arbitrary console in the past, but apparently not.  i'll keep digging..

Durman:
Especially it would be nice to creates links to objects for them further inspection via console.

zeffii:
@Durman 

i think you could probably inspect the content of a node at runtime, by returning the values you want from '`process()`

```python

def process(self):
    ...
    ...  # setting / getting socket data

    return [everything you want to query]

>>> nodes = bpy.data.node_groups['NodeTree']
>>> n = nodes[your node]
>>> data = n.process()
>>> print(data)
```

zeffii:
that would be useful during node development / debugging.

if that works it's simpler than implementing `node_dict` to store objects for introspection - so i would prefer not to go there is `returning` data from process is adequate

Durman:
Well, I just create variable in a module and then exports it via console. It works.)

zeffii:
i won't be committing this

Durman:
> Especially it would be nice to creates links to objects for them further inspection via console.

Actually I had found that `breakpoint()` (built in python function) is doing exactly what I need.

PR #2647, opened 2019-11-01
Triangle node

vicdoval:
The node is designed to generate triangles from various combinations of vertices, sides length and angles.
Three ways of getting the same triangle from different inputs t

The node can be used to generate more complex shapes, note that "join last level" is activated the N-Panel
t
 
Advanced example using the node to generate structural ribbons
triangle_node_sverchok_example

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


Durman:
I had wanted to raise one issue but it looks here would be appropriate place. I see that some generator nodes change input data shape. From my point of view it is not logical. If I would like to generate objects I poot several objects in input sockets. But if I'm not wanting? 

Is it really worth to split everything into different objects in this particular example?

vicdoval:
@Durman I was also thinking about this topic lately.., 
That way is how the generators have been working for ages and in my mind makes sense because every "object" is separated in a different list but could not be the best way.
The best way I found to solve it (that is implemented in this node  and i think it could work fr others) is with the property "Join Last Level", on the right-click menu, that also offers a remove doubles cleaning.
I also added a property  "Flat Output" to flat the level 1 of the incoming list so you can have many different data-shapes and take the one that suites the particular case.

Durman:
Actually I faced with problem when I do not need to generate each mesh into separate object. More precisely to say the task was to create some part of mesh in one object and other parts in other objects. 
From my point of view the easiest way is to create input in such shape in which I would like to see output. And as far as I now all nodes (exclude generator) do work in this way. 
Initially I think all generators did not expect any arrays in input. 

Probably it worth to create separate issue anyway.

PR #2648, opened 2019-11-02
Import Quaternion nodes from 2.79 to 2.80

DolphinDream:
- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


portnov:
Maybe @vicdoval could help with icon? :)

vicdoval:
I have them almost ready :) , I will add them once this PR is merged

portnov:
Hmm... it's better have tests passing before merging PR. But as I can see, this PR is from @DolphinDream 's repo, I think you can't push there... So you have to cooperate: @DolphinDream can merge the PR anyway , but that will break tests in b28_prelease_master; then you'll fix them by pushing the icon. Or you can pass icon to DolphinDream directly so that he could add it to this PR before merging it.

DolphinDream:
I could design some icons, too ;) Give me what you have though.

vicdoval:
[quaternion_icons.zip](https://github.com/nortikin/sverchok/files/3801143/quaternion_icons.zip)
there you have them. The quaternion  menu was the hardest, I did my best after some digging.
The others are just variations of the actual icons changing the color 🛩

DolphinDream:
Thanks @vicdoval They look great. Ill add them to this PR

portnov:
:+1:

DolphinDream:
If there are no objections I can integrate this PR.

portnov:
No objections.

DolphinDream:
Btw, should the quaternion_index.rst file be included ? I don’t know how I ended up having this file in the PR. I may have imported it from 2.79. Is it needed? should I leave it in PR? remove it?

portnov:
I think it's needed and should be referenced from `docs/nodes.rst`.

portnov:
@DolphinDream could you please fix tables in documentation. They are not rendering properly https://github.com/nortikin/sverchok/blob/b28_prelease_master/docs/nodes/quaternion/quaternion_math.rst .

DolphinDream:
@portnov I fixed the tables in all 3 quternion node docs (in b28_prelease_master)

DolphinDream:
Btw, while importing from 2.79 I changed the outputs nesteness to have only one level..

[Quaternion((0,0,0,0)), Quaternion((1,0,0,0))]

instead of

[[Quaternion((0,0,0,0)), Quternion((1,0,0,0))]]

This was to match the way the Matrix nodes output.

SV-output-formats

Though I wonder if this change was the way to go.
Compared to other nodes (vector, scalar) it seems that these nodes output is somewhat different.

the scalar nodes have: [[1,2,3]]
the vector nodes have: [[(0,0,0), (1,1,1)]]
the matrix nodes have: [Matrix(...), Matrix(...)]
the quaternion nodes have: [Quaternion(0,0,0,0), Quaternion(1,1,1,1)]

Durman:
What is the benefit of having this nodes in Sverchok? Any examples of usage?

DolphinDream:
@Durman They are cool ;)
The Quaternions are useful for rotations (as you probably know), since they don't suffer from the gimble-lock issue the Euler angles have.. though since the matrix math node dealing with rotation internally (via mathutils?) may already use quaternions to do the rotations, it may be that these quaternion nodes would not be super useful in this case. In general it may be easier to do rotations with quaternions than with matrices (i don't have a proof of that though). The main reason I introduced them was to be able to do 4D math, particularly prompted by the Hypercube node (PR still opened). For such applications it doesn't make sense to provide 4D math nodes if 4D vectors primitives do not exist. Of course, the Quaternion math works on 4D vectors according to the quaternion rules.. but the Quatnerion In/Out can be used both as simple 4D vectors or quaternions. The 4D math node will work on quaternions (as a 4D vector).

PR #2651, opened 2019-11-02
Adaptive polygons mk2 node

portnov:
This provides a new version of Adaptive Polygons node, which is more precise (especially for Tris faces) and has more features. Some features are ported from Tissue addon, I hope Alessandro will not mind :)
Direct translation of Tissue code is not possible, as Tissue is heavily based on `bpy` API.

For the list of new features, just see the documentation https://github.com/nortikin/sverchok/blob/adaptive_polys_mk2/docs/nodes/modifier_make/adaptive_polygons_mk2.rst - always all parameters are new :)

## TODO for later Mk

- Iterative / fractal tessellation - this feature of Tissue gives nice results sometimes, but i'm not sure it is practically useful
- Possibility to provide specific donor object for each recipient face. In simple cases, this can be done by combining several "Adaptive Polygons" nodes with different masks.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


portnov:
Screenshot_20191102_215606
Screenshot_20191102_215945
Screenshot_20191102_220433
Screenshot_20191102_220603
Screenshot_20191102_223247

portnov:
adaptive_polys_2
adaptive_polys_3
adaptive_polys_4
adaptive_polys_5
adaptive_polys_6
adaptive_polys_7

vicdoval:
I was also thinking is re-coding this node 😄 
One feature I would like to have is the possibility to add a "frame" for the donor object like happens with the "barycentric transform" node, that way you could place the donor relative to the frame:
t

portnov:
Hmm. "Adaptive polygons mk2" has two modes for coordinate translation:
* "Bounds": similar to mk1 version; for quad faces, it takes "source area" = bounding rectangle of the donor object; for tris faces, it takes "source area" = bounding triangle of the donor object (i.e. the smallest triangle with equal sides, which covers all donor vertices while having bottom side parallel to X axis). This source area is then mapped to the recipient face.
* "As is": similar to Tissue's "Global" mode. For quads, this takes "source area" = [-1/2; 1/2] x [-1/2; 1/2] square; for tris, "source area" = the triangle with all sides equal to 1 with center at (0,0,0). Then it maps the whole XOY space so that "source area" becomes recipient face.

So, in the later case, you can just manually make donor object smaller than this fixed "source area", or move it from center, or whatever.

portnov:
adaptive_voronoi_1

portnov:
Screenshot_20191103_124417
Screenshot_20191103_124714
Screenshot_20191103_124811

portnov:
adaptive_icosphere_4

portnov:
Screenshot_20191229_174804

PR #2659, opened 2019-11-07
Extrude separate node: possibility to specify transformatrion matrix

portnov:
This adds a parameter and an input to the "extrude separate" node, so that it is now possible to specify transformation of each face being extruded not only by height and scale, but by arbitrary transformation matrix.

This also adds a parameter about what to do with faces that are masked out:

* either do not extrude them at all (the default mode now),
* or extrude them, but do not transform (move/scale) -  this was the default mode before. This way the new vertices are at the same locations as original ones - you have to remove them with **Remove Doubles** node or move with other nodes.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


portnov:
Screenshot_20191107_215731
Screenshot_20191107_220232

PR #2661, opened 2019-11-08
B28 update throttle3

zeffii:
events like `self.inputs.new()` trigger `tree.update()` 

this context manager will enable nodes to have control over whether to end early inside the `tree.update` function, and avoid processing the tree.

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -19,6 +19,7 @@
 
 import sys
 import time
+from contextlib import contextmanager
 
 import bpy
 from bpy.props import StringProperty, BoolProperty, FloatVectorProperty, IntProperty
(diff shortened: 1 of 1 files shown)


zeffii:
this is untested...

portnov:
Maybe it worth wrapping `yield` into try-finally?...

zeffii:
@portnov,  i was under the impression that the context manager `yield` already enters a try/except ?

zeffii:
ok.. a try / finally it is..

zeffii:
https://stackoverflow.com/questions/28157929/python-how-to-safely-handle-an-exception-inside-a-context-manager

zeffii:
not sure the final form really matters...
```python
@contextmanager
def throttle_tree_update(node):
    """ usage
    from sverchok.node_tree import throttle_tree_update

    inside your node, f.ex inside a wrapped_update that creates a socket

    def wrapped_update(self, context):
        with throttle_tree_update(self):
            self.inputs.new(...)
            self.outputs.new(...)

    that's it. 

    """

    try:
        node.id_data.skip_tree_update = True
        yield node
    finally:
        node.id_data.skip_tree_update = False
```

PR #2665, opened 2019-11-09
B28 vtext fix4

zeffii:
resolves the sockets not matching the input types

### DIFF
--- a/data_structure.py
+++ b/data_structure.py
@@ -681,7 +681,7 @@ def changable_sockets(node, inputsocketname, outputsocketname):
     if not inputsocketname in node.inputs:
         # - node not initialized in sv_init yet,
         # - or socketname incorrect
-        info("changable_socket was called on node (%s) with a socket named \"%s\", this socket does not exist" % (node.name, inputsocketname))
+        info(f"changable_socket was called on {node.name} with a socket named {inputsocketname}, this socket does not exist")  
         return
 
     in_socket = node.inputs[inputsocketname]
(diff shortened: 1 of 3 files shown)


zeffii:
it will still create a second  Frame node... todo..

zeffii:
also this message occurs , and is triggered from inside this node's update function.
```
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\nodes\text\viewer_text_mk3.py", line 204, in update
    if not socket.bl_idname == socket.other.bl_idname:
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\core\sockets.py", line 67, in other
    return get_other_socket(self)
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\data_structure.py", line 759, in get_other_socket
    other = socket.links[0].from_socket
IndexError: tuple index out of range

```

zeffii:
i think that error happens when the input socket of viewertext is being disconnected (ie, the noodle is still attached to the mouse cursor..)

zeffii:
wow.   socket.is_linked can be true, and still socket.links can be a zero length tuple

@portnov   when  a noodle is being disconnected from a node, and that node has an update function like viewer text..

portnov:
I think it may also be in case when the socket is connected to "reroute" circle...

zeffii:
there's a remaining bug(s)
- that the node generates two frame nodes..
- sometimes `Data0` has SvNoData (but i think that's part of a larger problem)

zeffii:
still with bug, but better than before.. i think.

PR #2668, opened 2019-11-11
add waveform viewer stub

zeffii:
coming soon.

this node will
- [ ]  display incoming data as a waveform in the nodeview , using bgl + custom shader for lines / grid
- [x]  will have options for num channels / bitrate (generally useful waveform settings )
- [ ] will allow auto normalize
- [ ] will accept both numpy arrays and python lists
- [x] will offer an export to waveform button (initial implementation is non-realtime .wav using `import wave` )

this node is intended to be useful for the Oscillator node by @vicdoval 

prelim UI
image


zeffii:
i'm thinking of adding a feature that also can add those samples to Blender, maybe preview them

zeffii:
bookmark for later.. https://web.archive.org/web/20141226210234/www.sonicspot.com/guide/wavefiles.html

zeffii:
the grid is drawn using a `vertex shader / fragment shader`, but I think there will need to be a `nvBGL` `custom_function_extended` for passing projection view..

PR #2670, opened 2019-11-13
Scalar Math in NumPy

vicdoval:
Scalar Math moved to Numpy with a boost of 2x-30x in  performance

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


zeffii:
that's an epic change,  i would feel comfortable making this a `mk4`, and moving `mk3` to old.  (even if it's for speedtesting / consistency testing.. )

Durman:
When input values are Python lists the node becomes slower I guess.
image
With converting it back to Python numpy implementation becomes two times slower over Python one.
#4487

PR #2672, opened 2019-11-15
B28 waveoutput shader2

zeffii:
https://github.com/nortikin/sverchok/pull/2668 (originally)

addresses a curiosity with respect to 2d nodeview drawing and a custom shader, I don't know if this works yet, but it should allow a fragment shader to control pixels. Useful for grids..

- this builds on the waveoutput PR


zeffii:
one small step ..
image

zeffii:
this is more useful.. 
image

Issue #2456, opened 2019-06-16
Nodes (red /error) don't automatically revert to non-error state/color when copied


BennyGovaerts:
Random copy with list item everything works but there is one node that turns red with blender 2.8 
If i do the same with blender 2.79  then i don't have this problem , is this normally or must i change something ? 

![image](https://user-images.githubusercontent.com/49908008/59557436-7f20a880-8fda-11e9-958c-36192ff051a9.png)

zeffii:
i can reproduce this only in one scenario, namely if the node is red/error state and then you copy it, the copy can not be returned to the default colour.  
![image](https://user-images.githubusercontent.com/619340/59562688-fcc2d380-902f-11e9-8d18-5dd94c6a49d3.png)

there are more nodes that do this,  but from your description of events I think this might not be the same (known...)  bug.

zeffii:
related #2241 #2081

zeffii:
yeah :) i would like this to be fixed.

BennyGovaerts:
yes , you are right if i add a new list item node and remove the red one then it works , thanks for the tip

zeffii:
i don't fully understand the error, and haven't had much quiet time to think about it. no ETA.

it does bug me :)

zeffii:
note to self:

as the node in the error state is copied, the new node is not added to the `ng['error_nodes']` because it has not been told to process yet (and has no input sockets that would force it to do that..)

zeffii:
cool. i think this will be added to the `node base class`

```python
    def copy(self, node):
        """
        if the node being copied currently is listed in the error_nodes dict
        then copy two properties from the dict ( use_custom_color, color ) and apply those
        to the copy.
        """
        ng = self.id_data
        try:
            import ast
            error_dict = ast.literal_eval(ng["error nodes"])
            error_item = error_dict.get(node.name)
            if error_item:
                self.use_custom_color, self.color = error_item
            
        except Exception as err:
            print('cosmetic error during copy', err)
            
```

Durman:
In this case code something like `super().copy(self, node)` should be inserted into overrided method of base class in some nodes.

zeffii:
durman, instead:  ` SverchCustomTreeNode.copy(self,, node) `

zeffii:
altho `super().<function>(args)`  should indeed be the way to go.

for some reason it didn't work https://github.com/nortikin/sverchok/issues/2444#issuecomment-501152982 .

[...]

PR #2675, opened 2019-11-17
Reset node error color on copying

portnov:
refs #2456.

@zeffii  please review and test :)

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/core/update_system.py
+++ b/core/update_system.py
@@ -293,6 +293,22 @@ def update_error_nodes(ng, name, err=Exception):
         node.color = exception_color
     node.use_custom_color=True
 
+def get_original_node_color(ng, name):
+    if "error nodes" in ng:
+        error_nodes = ast.literal_eval(ng["error nodes"])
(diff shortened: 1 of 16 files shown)


zeffii:
thanks for doing this, i do have one discussion point however. It's not a big deal, but i think there's a more consistent way to do this

rather than using `super().copy(node)`, like...
```
class YourNode

    def copy(self):
        super().copy(node)
        ... your other stuff

```
a more consistent (with the likes of `sv_init` ..) would be ... perhaps ...

```
class OurScCommonBassclass:
    def copy(self):
        ...
        if hasattr(self, "sv_copy"):
            self.sv_copy()
        ...

class YourNode

    def sv_copy(self):
        ... your other stuff

```
that would mean mostly just renaming the implemented `copy` functions to `sv_copy`

portnov:
done.

PR #2682, opened 2019-11-18
New Number Range Node

vicdoval:
Written in numpy (10X faster)
image

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


vicdoval:
In the migration from "Range Float" and "Range Int" i am having some troubles with the sockets order, ¿is there a way to set the socket mapping in a dynamic way depending on the parameters of the node?

Durman:
Easiest way that I have found to change order of sockets, is like this:
```Python
def update_sockets_event(node, context):
    with throttle_tree_update(node):
        [node.inputs.remove(sock) for sock in node.inputs]
        [node.inputs.new(ntype, nname) for ntype, nname in sockets_in_new_order.items()]
    updateNode()

**Comment by portnov** [comment]
Maybe reading of `ui/nodes_replacement.py` will help?

**Comment by zeffii** [comment]
@vicdoval 
because there are only a finite set of possible remapping scenarios, and that set is really not gigantic, you might want to set up a remapping table, rather some (potentially) convoluted dynamic thing.

**Comment by vicdoval** [comment]
I added a call to this function that made he trick:
```
    def update_mapping(self):
        if self.mode == 'COUNTRANGE':
            inputs_mapping = dict(Step='Step', Stop='Stop')
        else:
            inputs_mapping = dict(Step='Stop', Stop='Step')
        GenListRangeIntNode.replacement_nodes = [('SvGenNumberRange', inputs_mapping, None)]
```
The point was to change GenListRangeIntNode.replacement_nodes  and not self.replacement_nodes 
image
With the 'Flat Output' activated  (as it is by default) the output will be [[..],[..],...] and it can be plugged to another Number Range Node.
image

zeffii:
this is cool :)

PR #2685, opened 2019-11-19
Svnodata extra error

zeffii:
hopes to locate the origin of an annoying cache reset scenario.


zeffii:
and maybe we add loggin/debug  to the throttle message instead :) ... or  drop it for speed?

zeffii:
i'm comfortable dropping the print / logging  there entirely.

PR #2686, opened 2019-11-20
svnodecommon has a throttle_tree_update wrapper (sv_throttle_tree_upd…

zeffii:
instead of forcing me to import throttle_tree_update from node_tree, we now have an alias, that you can use like 
```python
    def process(self):

           ...do your stuff....
```

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -267,6 +267,9 @@ class SverchCustomTreeNode:
             self.n_id = str(hash(self) ^ hash(time.monotonic()))
         return self.n_id
 
+    def sv_throttle_tree_update(self):
+        return throttle_tree_update(self)
+
     def mark_error(self, err):
         """
         marks the with system error color
--- a/nodes/viz/viewer_bmesh.py
+++ b/nodes/viz/viewer_bmesh.py
@@ -27,7 +27,7 @@ import bpy
 from bpy.props import BoolProperty, StringProperty, BoolVectorProperty
 from mathutils import Matrix, Vector
 
-from sverchok.node_tree import SverchCustomTreeNode, throttle_tree_update
+from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import dataCorrect, fullList, updateNode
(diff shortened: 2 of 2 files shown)


zeffii:
@portnov  @Durman  @nortikin  @vicdoval @DolphinDream   just so you know

portnov:
why the `sv_` prefix is needed?

zeffii:
no, it's not needed but i do like an "alias / hoisting" to have an obviously different name. That's what came out of my fingers.

if there is any massive objection, that should be raised asap.

Durman:
I think it can be real method instead of alias.
And why not just to use name freeze or frozen?

zeffii:
freeze is already in use, i want to avoid messing with that part of the update_system.

i might try to work away the use of freeze entirely, just because the implementatin of freeze/frozen.unfreeze is a little convoluted (i think more convoluted than it needs to be..it's difficult to read)

Durman:
I thought the node does not have freeze method yet. 
As for me `with self.frozen():` looks okay.
But I don't know is that good to use method of an object as context manager.

zeffii:
to be clear here, when we "freeze" or "throttle" the update function, all that is really happening is the following.

- if any code inside a node (inside process, a property `update=updateNode.. or wrapped_updateNode`.. anyhing..) performs some operation that triggers the nodetree.update function, then we return early out of the nodetree update function. We return early because we know it was triggered without the intent to perform a recalcuation of the nodetree.  If we didn't return early we would be likely performing a very tight loop of needless recalculations

- operations that can cause the nodetree.update to trigger are 
    - anything that triggers despgraph update
         - ..it might be useful to list all of these things, but usually it is adding/removing objects/meshes
    - any change UI change in the node editor
         - usually adding/removing a node socket, a link

i think only saying `with self.frozen()`  forgets that better names help understand what the code is doing / supposed to do. This is why i choose `(sv_)throttle_tree_update`

zeffii:
i might even stick throttle update around the `sv_init`, which currently has freeze around it.

PR #2687, opened 2019-11-20
add throttle_tree_update to objinlite

zeffii:
resolve issue with depspgraph in snlite, at the very least.


zeffii:
image

zeffii:
https://docs.blender.org/api/blender2.8/bpy.types.Depsgraph.html#dependency-graph-simple-exporter

zeffii:
i see now objin lite is slightly buggy with get/drop

PR #2690, opened 2019-11-23
Vector math_mk3

vicdoval:
In this case I left both implementations because with huge lists (around 500000 vectors ) the numpy implementation can be slower if you output regular lists.
The numpy implementation will accept list of numpy arrays so you can make all the calculations in Numpy and convert it in the last one improving the performance
image

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/vector/math_mk3.rst
+++ b/docs/nodes/vector/math_mk3.rst
@@ -31,9 +31,9 @@ Modes
 Most operations are self explanatory,
 but in case they aren't then here is a quick overview:
 
-=================== ========= ========= =================================
+=================== ========= ========= =========================================
 Tables              inputs    outputs   description
-=================== ========= ========= =================================
+=================== ========= ========= ========================================
 Scale YZ             v, s     w         scale vector by amount
 Scale XZ             v, s     w         scale vector by amount
 Scale XY             v, s     w         scale vector by amount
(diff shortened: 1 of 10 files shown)


vicdoval:
When I change the operator I get the SvNoDataError(socket).
is this because this branch was made before this was fixed or I have to add something to the node itself?

zeffii:
> Is this because this branch was made before this was fixed or I have to add something to the node itself?

nope, the fix was mostly for bmeshviewer.

without looking at the code, if you switch the operator is it adding/removing a socket? you might want to add a 
```python
    change your sockets here...
    or do anything else that might trigger an update..

#then 
updateNode(self, context)    # or whatever
```

zeffii:
so yes, you should merge b28_prelease_master into your branch, so you can use `self.sv_throttle_tree_update()` from inside a node.

zeffii:
i think what's happening there (and everywhere else we see this) is the following.

1) we update a socket (by removing, replacing, adding... whatever.. maybe even hide...)
2) the event in `1` causes the node tree to fire a call to `nodetree.update`, 
3) inside this update there is a cascade  of process function calls, for the appropriate nodes downstream from the current node
3) the problem part is that the current socket-change has not yet updated `socket_data_cache[s_ng][s_id]`, where `s_id` is the id of the socket that SvGetSocket is called on.

it might be better understand as a classic race-condition violation.

another, more robust solution could be to add a decorator which we can apply to "wrapper_update" functions, that throttles the node tree, while you do whatever you need and then automatically calls updateNode at the end.

I would definitely add that if you find that this resolves the socket data issue..  please test with the long form first

vicdoval:
It solved the issue 😄

zeffii:
cool!  we'll have to make a pass through the codebase and apply a decorator to the `update=...` functions that manipulate sockets - soooon

PR #2691, opened 2019-11-24
Vector In and Vector Out numpy mode

vicdoval:
In *Vector In* there are both implementations, the regular python implementation(default) is faster if you are using regular lists and getting regular lists. The Numpy  implementation accepts lists of flat NumPy arrays and is faster if you want numpy Arrays.
In *Vector out* now lists of NumPy arrays are accepted with shape [n,3] and can output NumPy arrays.

This node-tree shows nodes that accept and create numpy arrays. They can be used with out converting the data to regular lists. Using the this method the tree update was more than 6 times faster than using regular lists
image

- [x] Ready for merge.

### DIFF
--- a/nodes/number/numbers.py
+++ b/nodes/number/numbers.py
@@ -54,7 +54,8 @@ class SvNumberNode(bpy.types.Node, SverchCustomTreeNode):
             prop = kind + '_'
             self.inputs[0].replace_socket('SvStringsSocket', kind.title()).prop_name = prop
             self.outputs[0].replace_socket('SvStringsSocket', kind.title()).custom_draw = 'mode_custom_draw'
-            self.process_node(context)
+
+        self.process_node(context)
 
     int_: IntProperty(
         default=0, name="an int", update=updateNode,
--- a/nodes/number/numbers.py
+++ b/nodes/number/numbers.py
@@ -49,16 +49,17 @@ class SvNumberNode(bpy.types.Node, SverchCustomTreeNode):
     bl_icon = 'DOT'
 
     def wrapped_update(self, context):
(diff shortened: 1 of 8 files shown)


nodes/vector/vector_in.py
```diff
@@ -52,6 +72,19 @@ class GenVectorsNode(bpy.types.Node, SverchCustomTreeNode):
 
     advanced_mode: BoolProperty(name='deep copy', update=updateNode, default=True)
     show_3d_cursor_button: BoolProperty(name='show button', update=updateNode, default=False)
+    implentation_modes = [
```
zeffii: typo

nodes/number/numbers.py
```diff
@@ -49,16 +49,17 @@ class SvNumberNode(bpy.types.Node, SverchCustomTreeNode):
...
-        prop = kind + '_'
-        self.inputs[0].replace_socket('SvStringsSocket', kind.title()).prop_name = prop
-        self.outputs[0].replace_socket('SvStringsSocket', kind.title()).custom_draw = 'mode_custom_draw'
-        self.process_node(context)
+        with self.sv_throttle_tree_update():
+            kind = self.selected_mode
+            prop = kind + '_'
+            self.inputs[0].replace_socket('SvStringsSocket', kind.title()).prop_name = prop
+            self.outputs[0].replace_socket('SvStringsSocket', kind.title()).custom_draw = 'mode_custom_draw'
+            self.process_node(context)
```
zeffii: the process part should be outside of the context manager... so 
```diff
-            self.process_node(context)
+
+         self.process_node(context)
```

PR #2692, opened 2019-11-25
B28 updatefunc decorator

zeffii:
new decorator for updateNode wrappers..

### implementation details

this is only a preliminary implementation, because it might be useful to include a way to end the update function early, and not ultimately `process_node`. 

```python
# use as a decorator
def throttled(func):
    """
    When a node has changed, like a mode-change leading to a socket change (remove, new)
    Blender will trigger nodetree.update. We want to ignore this trigger-event, and we do so by
    - first throttling the update system. 
    - then We execute the code that makes changes to the node/nodetree
    - then we end the throttle-state
    - we are then ready to process

    """
    def wrapper_update(self, context):
            func(self, context)
        self.process_node(context)

    return wrapper_update
```
### usage

```python

from sverchok.node_tree import SverchCustomTreeNode, throttled

class SvYourNode(...):

    ...

    @throttled
    def mode_update_wrapper(self, context):

        # your nodetree update triggering stuff here 
        #             1) changes to sockets
        #             2) changes to links
        # the decorator calls   self.process_node(context)    for you

    mode_direction: bpy.props.EnumProperty(......update=mode_update_wrapper)

```

### Why ?

this means that we can abstract away the task of throttleing, and also assume that the node tree will be udpated at the end.

### issues..

1. ~i'm not sure on the name, but the implementation seems sound.~
2. maybe later after testing the implementation and encountering special cases, we might consider adding the ability to pass an 'exit_function`..


zeffii:
or maybe it should be called 
```python
    @throttled_updateNode
```
and still need to be imported from data_structure.py

or.. import it into sverchok global so it would be
```

    @sverchok.throttle_updateNode
    def your_function(self, context):
        ...
```

nortikin:
what kind of changes need this?

zeffii:
@nortikin  `inputs.new and inputs.remove` and any function that use those.  Some functions call these functions like.. 8 times.

zeffii:
implementing this properly will reduce extraneous calls to process the node tree. this is a good thing.

zeffii:
the reason this has become necessary is the fact that blender devs removed `scene update pre/post handlers` and replace it with a handler (`depsgraph pre/post` and that handler is triggered by more things. than `scene update/post` was.

without a clear way to findout what triggered the depsgraph update, from the event handler.. we must defend against it this way.

zeffii:
> If it is part of an multysocket or typechanger, maybe integrate solution to thouse functions?

yes , good point. should happen too.

zeffii:
i failed to add `throttle` decorator as a classmethod to `SverchCustomTreeNode`, so this is why we must import the decorator manually. I hope for now that's OK. perhaps someone knows how to have the decorator available without import from `from sverchok.node_tree`

zeffii:
essentially, when we used to write

```python
def wrapped_update(self, context):
    with self.throttled_node_tree():
        ...code modifying the node
    self.process_node(context)
```
you now write
```python
@throttled
def wrapped_update(self, context):
    ...code modifying the node
```

PR #2694, opened 2019-11-26
B28 geom 2d package fix2

zeffii:
case sensitivity conflict


zeffii:
@Durman  checks this.. seems OK.

zeffii:
i had to 
```
git config core.ignorecase false

git rm -r --cached nodes/cad
git rm -r --cached docs/nodes/cad

commit ...

git config core.ignorecase true
```

Durman:
Thanks for helping =)

zeffii:
awesome pullrequest @Durman 

this allows us to do some crazy 2d stuff :)

PR #2700, opened 2019-11-27
experiment with context_sensitive_throttle

zeffii:
this attempts to employ a context manager approach to skipping tree updates in our socket ulity functions
- replace_socket
- multi_socket
- changeable_socket 

the reason is that the context manager approach should certainly return the `node_tree.skip_tree_update` to the appropriate state, even if there's an exception mid function.

whether this is more readable, is up for debate.. but it's more pythonic, and it's what we should be doing for the sverchok state-machine anyway..

### DIFF
--- a/data_structure.py
+++ b/data_structure.py
@@ -16,6 +16,7 @@
 #
 # ##### END GPL LICENSE BLOCK #####
 
+from contextlib import contextmanager
 from math import radians, ceil
 import itertools
 import time
@@ -757,6 +758,19 @@ def updateNode(self, context):
 ##############################################################
 ##############################################################
 
+@contextmanager
+def context_sensitive_throttle(node_tree):
+    if node_tree.skip_tree_update:
+        yield node_tree
+
+    else:
+        try:
+            node_tree.skip_tree_update = True
+            yield node_tree
+        finally:
+            node_tree.skip_tree_update = False
+
+
 def changable_sockets(node, inputsocketname, outputsocketname):
     '''
     arguments: node, name of socket to follow, list of socket to change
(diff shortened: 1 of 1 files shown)


zeffii:
@portnov  a sanity check please, what do you think?

data_structure.py
```diff
@@ -757,6 +758,22 @@ def updateNode(self, context):
 ##############################################################
 ##############################################################
 
+@contextmanager
+def context_sensitive_throttle(node_tree):
+    if node_tree.skip_tree_update:
```
portnov: Uh. Sorry, got to this only now.
I do not see the reason for this check. Why do not just `node_tree.skip_tree_update = True`? If it's already True — that's will do nothing, but no problem?

zeffii: if its already `True`, then another function call is already throttling it, and it won't do anything but yield `node_tree`, but it must be sure not to set it to `False` .

portnov: Ah, I see. Then it's probably better to do
```
prev_skip = node_tree.skip_tree_update
node_tree.skip_tree_update = True
...
node_tree.skip_tree_update = prev_skip
```
(plus usual try-catch stuff, off course).

zeffii: all we want there is a `no-op` , because the scenario is that there's a throttle, inside a throttle, (a nested throttle), and we dont want to prematurely 'exit' the throttle state.

exactly what code is used for this is really down to what's most effective. i welcome your suggestion.

portnov:
```
        try:
            prev_skip_state = node_tree.skip_tree_update
            node_tree.skip_tree_update = True
            yield node_tree
        finally:
            node_tree.skip_tree_update = prev_skip_state
```
this approach has the following properties:
1) it guarantees that skip_tree_update will be returned to it's previous state, no matter what it was;
2) it will automatically work with any nesting level, because prev_skip_state will be stored as a local variable in the call stack;
3) the fact that we are just restoring the state that was before is obvious from the code, it does not raise questions like "what if state was True? Or what if it was None for some reason — will it be restored afterwards?"
4) it does not require two copies of try-finally clause — "what if true" and "what if false".

zeffii: ok :) yes!

zeffii:
i'll be using your suggestion here @portnov .  committing this soon, ..and can then gradually remove a few explicit throttles

PR #2703, opened 2019-11-28
B28 waveform output

zeffii:
image

didn't feel like wasting more time on a fragment shader, not this time.


zeffii:
not sure why the test is failing, code be the latest version of Blender handles custom node group (in relation to what we do in Monads) differently..

zeffii:
ok.  it kind of works: 
image

@vicdoval

zeffii:
now with initial tick support

tick every 128 samplers:
image

zeffii:
image


i think minimized/hidden easing and waveoutput node should display their graph x=0 y=-labelwidth , ie.. above the label.

zeffii:
i don't really know how useful this node is, but it's fun to have anyway.

enzyme69:
Zeffii this waveform thing, can it plug into Audio Nodes? https://github.com/nomelif/Audionodes

zeffii:
Audionodes is an entirely different engine, don't they have a waveform viewer?  SvWaveformViewer is Sverchok only.

PR #2217, opened 2018-05-11
Switch mk2

Durman:
## Addressed problem description

#2198

## Solution description

Have a look to the documentation.

## Preflight checklist

- [x] Code changes complete.
- [x] Code documentation complete.
- [x] Documentation for users complete (or not required, if user never sees these changes).
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

PR #2704, opened 2019-11-30
switch node - mk2 - port from 2.79

Durman:
#2217

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


Durman:
I have added supporting of numpy arrays. Output is related with input from socket A and socket B. Output will be numpy array if at least one input sockets (A or B) has numpy array and another socket does not have list with two or more values.

Durman:
Something unexpected can be with none iterable objects like matrix or Blender objects. On the picture below it can be expected that switch should add first matrix and second quaternion:



 but for this states input should have values on first object level not on second data level:

Durman:
Alternative of `list mask out` node

Durman:
More control over sockets:
socket control

PR #2708, opened 2019-12-03
Panel 3d - refactoring

Durman:
List of all nodes which can have properties on 3D panel,
And new location of 3D panel on N panel:

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/number/list_input.rst
+++ b/docs/nodes/number/list_input.rst
@@ -19,15 +19,16 @@ Output
 
 A single *flat* ``list``.
 
+3D panel
+--------
 
+The node can show its properties on 3D panel. 
+For this parameter `to 3d` should be enabled, output should be linked.
+After that you can press `scan for props` button on 3D panel for showing the node properties on 3D panel.
 
 Examples
 --------
 
 Useful when you have no immediate need to generate such lists programmatically.
 
-.. image:: https://cloud.githubusercontent.com/assets/619340/4186017/0f076f00-375d-11e4-92ce-30af77e41307.PNG
-  :alt: ListInputDemo1.PNG
-
-.. image:: https://cloud.githubusercontent.com/assets/619340/4186018/0f61f4b6-375d-11e4-99c8-6f7ef62598b3.PNG
-  :alt: ListInputDemo2.PNG
+.. image:: https://user-images.githubusercontent.com/28003269/70140711-c7c63e00-16ae-11ea-9266-e4f24586e448.png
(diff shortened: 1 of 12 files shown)


Durman:
It looks there is no need always show limits in 3D panel for `A number` node.
n panel a number

Durman:
`Hasattr` is most curious function in Python
```Python
class A:
    def prop(self):
        A.asdfasd
        return True
   
>>> hasattr(A(), 'prop')
False
```

Durman:
After this PR if someone would like to have UI on 3d panel for a node all what should be done is to add one special method and one special property to a node, like this:
```Python
class SvMyNode:
    my_prop: bpy.props.IntigerProperty()
    draw_3dpanel: bpy.props.BoolProperty()

    def draw_buttons_3dpanel(self, layout):
        layout.prop(self, 'my_prop')
```
Property can be dynamic:
```Python
class SvMyNode:
    my_prop: bpy.props.IntigerProperty()

    @property
    def draw_3dpanel(self):
        return True if self.outputs[0].is_linked else False

    def draw_buttons_3dpanel(self, layout):
        layout.prop(self, 'my_prop')
```
@zeffii Can you make a suggestion where this logic could be documented? If there is no objection of such approach.

zeffii:
@Durman 
```
class A():

    def __init__(self):
        ...

    def prop(self):
        return True

a = A()
print(hasattr(a, 'prop'))
>>> True
```

i have no objections to making it easier for nodes to display properties in the other panels, be aware that there is already a property-check in place for when a node is deleted, to clean up the propertygroup used to show the properties.

https://github.com/nortikin/sverchok/blob/a92cea484c3a90dd908bcc1884467e1a6c2b448b/node_tree.py#L533-L544

Durman:
With class A example I only try to point that debugging process can be painful. If you try to get some non existed attribute inside method and try to check whether the method exists you can be surprised that the method does not exist from point of Python.

> be aware that there is already a property-check in place for when a node is deleted, to clean up the propertygroup used to show the properties.

I will check but it looks that this contradicts with 3d panel code:

https://github.com/nortikin/sverchok/blob/a92cea484c3a90dd908bcc1884467e1a6c2b448b/ui/sv_panels.py#L196-L206

Durman:
There are no even one node with `has_3dview_props` attribute. I think it worth to reimplement this part of code. This code is probably outdated or it is just a declaration of functionality without any real job.

Durman:
But probably it worth to discuss how to handle deleting of a node represented in 3D panel. Now it does not removed from 3D panel automatically, but there is warning message about this.

From my point of view it is quite good approach. You can delete a node casually without noticing but then you can see the hint on 3D panel that the node was deleted. Any thoughts?

zeffii:
i don't have a strong opinions about this, i can't help with this at the moment.

PR #2471, opened 2019-06-23
Align mesh by mesh node

Durman:
## Addressed problem description

Sometimes there is need of aligning one object to another. This node represent most basic features. It is something similar to internal Blender function of snapping to objects. More details in examples. 

## Preflight checklist

- [x] Code changes complete.
- [ ] Code documentation complete.
- [ ] Documentation for users complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

PR #2715, opened 2019-12-05
align mesh by mesh node - port from 2.79

Durman:
#2471

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.


Durman:
@zeffii What about `custom_draw` attribute not only for string sockets?

zeffii:
> @zeffii What about custom_draw attribute not only for string sockets?

yes, I'm for it!   do it the way you propose here first, we can later add a method to the Socket common class - if needed.

Issue #1810, opened 2017-10-10
random points on faces (snlite)


zeffii:
no serious error checking implemented here..

```python
"""
in verts_list v
in faces_list s
in num_points s d=20 n=2
out verts_out v
"""

import bpy_extras
from bpy_extras import mesh_utils

for verts, faces in zip(verts_list, faces_list):

    mesh = []
    try:
        mesh = bpy.data.meshes.new(name="xxxxaaaa")
        mesh.from_pydata(verts, [], faces)
        tessfaces = mesh.polygons
        new_verts = mesh_utils.face_random_points(num_points, tessfaces)
        verts_out.append([v[:] for v in new_verts])
        
        if "xxxxaaaa" in bpy.data.meshes:
            bpy.data.meshes.remove(mesh)

    except Exception as err:
        print('nope', err)

```

![image](https://user-images.githubusercontent.com/619340/31381894-2ae5f39a-adb6-11e7-8785-a0d33249a7ef.png)

Durman:
Better than random verts on polygons script.
![2017-10-10_16-23-59](https://user-images.githubusercontent.com/28003269/31386366-9d68f6a8-add7-11e7-9a7f-ac2a768076f8.png)

zeffii:
oh.. it seems that the `face_random_points` expects triangles only?.. or quads.. but not ngons?

zeffii:
turns out `face_random_points` is also written in python

##### scripts/modules/bpy_extras/mesh_utils | face_random_points
```python
def face_random_points(num_points, tessfaces):
    """
    Generates a list of random points over mesh tessfaces.

    :arg num_points: the number of random points to generate on each face.
    :type int:
    :arg tessfaces: list of the faces to generate points on.
    :type tessfaces: :class:`bpy.types.MeshTessFace`, sequence
    :return: list of random points over all faces.
    :rtype: list
    """

    from random import random
    from mathutils.geometry import area_tri

    # Split all quads into 2 tris, tris remain unchanged
    tri_faces = []
    for f in tessfaces:
        tris = []
        verts = f.id_data.vertices
        fv = f.vertices[:]
        tris.append((verts[fv[0]].co,
                     verts[fv[1]].co,
                     verts[fv[2]].co,
                     ))
        if len(fv) == 4:
            tris.append((verts[fv[0]].co,
                         verts[fv[3]].co,
                         verts[fv[2]].co,
                         ))
        tri_faces.append(tris)

[...]

PR #2716, opened 2019-12-06
Random points on mesh

Durman:

#1810

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Ready for merge.


Durman:
Probably it would be interesting to have input weight of faces for correction of points distribution, but i'm not sure. Probably in next time.

Durman:
Distribution of points on sphere surface according distance to floating point:
random point

portnov:
It would be good to be able to generate some sort of faces between those points (non-degenerated faces, I mean).

Durman:
Do you mean something like remesh node? I don't think that this is simple thing to do. 

By the way I have found new function in geometry module for Blender 2.81 only: [delaunay_2d_cdt](https://docs.blender.org/api/current/mathutils.geometry.html#mathutils.geometry.delaunay_2d_cdt)
It would be interesting to have a node based on this function but I'm afraid that in this case Sverchok will lose compatibility with Blender 2.80.

portnov:
Well, worth exploring anyway!

portnov:
>> Vertex coordinates (2d)

:(

Though, that should be faster that our Python implementation anyway...

nortikin:
> > > Vertex coordinates (2d)
> 
> :(
> 
> Though, that should be faster that our Python implementation anyway...

lets use it!!!

Durman:
Voronoi from unevenly distributed points on faces.

PR #2718, opened 2019-12-07
fix output type in a more elaborate way

zeffii:
uses a less defensive update function. is this OK ?

### DIFF
--- a/nodes/list_main/join.py
+++ b/nodes/list_main/join.py
@@ -48,22 +48,24 @@ class ListJoinNode(bpy.types.Node, SverchCustomTreeNode):
         layout.prop(self, "JoinLevel", text="JoinLevel lists")
 
     def update(self):
-        # inputs
-        multi_socket(self, min=1)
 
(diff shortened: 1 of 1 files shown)


Durman:
I don't feel like to use update method for dealing with sockets. I even had thought about creating different method for checking conditions of a node working which are usually located in beginning of process method. I think it would be nice to keep process node only for dealing with sverchok data structure and calling main functions of a node algorithm. But as temporary solutions it is okay I think.

zeffii:
yeah, using the update function is an evil necessity, with the current update mechanism.

i welcome experimentation.. else this stays broken.

zeffii:
important to clarify here, it's the `multi_socket` function that is triggering `nodetree.update` because the `self.inputs` section of multi_socket does no freezing of its own.

maybe def update should be :
```
    def update(self):

        if len(self.inputs) and len(self.outputs):
            multi_socket(self, min=1)
```

Durman:
I have tested this, looks like everything work as expecting.

PR #2719, opened 2019-12-07
B28 join fix

zeffii:
simpler code, but doesn't solve the real problem either.

doing anything inside `self.update`,  will trigger any time the update function of any other node is triggered.

so if you do a print('hahah') in list join's  `def update()` it is triggered by connecting/touching a socket of any other node, even if that node is not connected to anything..

### DIFF
--- a/nodes/list_main/join.py
+++ b/nodes/list_main/join.py
@@ -49,7 +49,7 @@ class ListJoinNode(bpy.types.Node, SverchCustomTreeNode):
 
     def update(self):
 
-        if len(self.outputs) > 0:
+        if 'data' in self.outputs:
             multi_socket(self, min=1)
 
 
--- a/nodes/list_main/join.py
+++ b/nodes/list_main/join.py
@@ -48,22 +48,24 @@ class ListJoinNode(bpy.types.Node, SverchCustomTreeNode):
         layout.prop(self, "JoinLevel", text="JoinLevel lists")
 
     def update(self):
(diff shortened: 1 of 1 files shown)


zeffii:
maybe we should capture the update function also.. in the svnodecommon
```
sv node common:

    update_triggered_by_node: StringProperty(...)

    def update(self):
        update_triggered_by_node = self.name
        try:
            self,sv_update()
        finally:
            update_triggered_by_node = ""

```

i don't know .. this update stuff is starting piss me off and mystify me.

Durman:
Update function is called upon any changes of a node tree. It is [default behaviour.](https://docs.blender.org/api/current/bpy.types.Node.html#bpy.types.Node.update) If new node is created or deleted or new link is created or deleted update function is called of every node in tree.

zeffii:
i didn't realize that every single node which implements the update function, gets that function called by editor changes, it would be more useful ( i think..)  if we could localize what node triggered the update.

Durman:
I have made some investigation in update system of Blender:

If `some_node.outputs.new('StringSocket', 'Name')`:
- nodetree update fuction is called
- the node update fuction is called **(update fuction of other nodes is not called)**

if `node_group.links.new(Socket1, Socket2)`:
- nodetree update function is called
- update function of all nodes in the nodetree is called

zeffii:
yep, and sockets.remove / move also trigger updates.

PR #2727, opened 2019-12-09
add viewer gp code back

zeffii:
super wonky node.


zeffii:
```python
AttributeError: 'GreasePencil' object has no attribute 'palettes'
```
ok..

zeffii:
https://docs.blender.org/api/blender2.8/bpy.types.GreasePencil.html#bpy.types.GreasePencil

zeffii:
maybe i'll make this node take an Integer input for the frame number, rather than passing it `stroke.id_data` from another node.  It would be nice if this was simpler to use, like a BmeshViewer

zeffii:
this is going to take a while..
```
layer = bpy.data.objects['GPencil'].data.layers.new()
frame = layer.frames.new(0)
stroke = frame.strokes.new()
stroke.points.add(20)

bpy.data.objects['GPencil'].data.layers[0].frames[0].strokes[0].points.add(20)
```

zeffii:
image

zeffii:
i'm merging this fully aware that is is not feature complete. it may already be of some use.

PR #2729, opened 2019-12-10
Chess selection node

Durman:
## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/analyzer/analyzer_index.rst
+++ b/docs/nodes/analyzer/analyzer_index.rst
@@ -35,3 +35,4 @@ Analyzers
    volume
    raycaster_lite
    image_components
+   chess_selection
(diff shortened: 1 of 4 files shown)


portnov:
Interesting.
There is also interesting API near bm.edge[i].loops, which theoretically allows to program non-trivial routes through the mesh. I've thought about a node where you can write programs like "step 2 faces to the right, select one face, then go 3 faces to the left of it, then select next 3 faces"...

Durman:
Yes, it like [half edge](https://en.wikipedia.org/wiki/Doubly_connected_edge_list) data structure but a tiny bit trickier.)
https://archive.blender.org/wiki/index.php/Dev:Source/Modeling/BMesh/Design/

PR #2730, opened 2019-12-11
Edge length node

portnov:
I found it odd that we still did not have this node. It outputs length of each object's edge. As an option, it can calculate total sum of all edge lengths.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


portnov:
Screenshot_20191211_214956

vicdoval:
The "path length" node does exactly this. Maybe the name was not right but the functionality is the same

portnov:
Oh. Very un-obvious :)

PR #2732, opened 2019-12-11
Add material sockets to several nodes.

portnov:
This adds "material index" sockets to several nodes:
* Obj In Mk3
* Adaptive Polys
* Remove Doubles
* BMesh Viewer

Such sockets would allow to generate meshes with different materials assigned to different faces.

I think, many sockets may benefit from having such sockets. Also, it seems that bmesh operators are already doing "the right thing" with materials (though I could not find documentation about this behavior, yet). For example, if we assign materials to bmesh faces, then do remove_doubles, then faces of the resulting mesh will have correct materials.

@zeffii @Durman @vicdoval your thoughts?

## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


portnov:
this
adaptive_icosphere_31

is done by this
Screenshot_20191212_001221

zeffii:
i would prefer that materials in bmeshviewer was added below `Matrices` , but am fine with having the socket - it was inevitable to add it some day.

Durman:
According that fact that faces can have different binded information (not only material index) more flexible solution would be to grant face index output. Like here:
https://github.com/nortikin/sverchok/blob/b28_prelease_master/docs/nodes/CAD/merge_mesh_2d_lite.rst

Or another solution could be create input socket which can keep any information binded to faces and after node process to grant this information in output socket according topology changes. 

In general I'm against of creating special socket for material because faces can keep alot of different information and I think that Sverchok should take this in account.

portnov:
Yeah, I thought about adding possibility to add any per-face (and also per-vertex and per-edge, why not?) information. The problem is, how should we implement this. For example, let's consider "remove doubles" node. Let's say we provided some abstract data for each face of the input mesh; then we use vertices, edges and faces inputs to make a `bmesh`; then we do `bmesh.ops.remove_doubles`. If bmesh operation removes some vertices, it can also remove edges or even faces; so indexes of the faces will change after bmesh operation. So, we can't just pass that abstract per-face information to output — it will be messed up. We should somehow ensure that when `bmesh.ops.remove_doubles` removes a face or changes it's index, it will correspondingly move items in that abstract per-face data array.
With material indexes, I put them into `bmesh.face[i].material_index`, and I've found that `bmesh` is already "material-aware", i.e. when it changes the mesh, it updates materials correctly. I could not find this in documentation, so I'm not sure if this is a standard behaviour for all bmesh operations or only for some of them; but if it is for most of them, then we can easily add "materials" input and output to many nodes, and that's good.

For arbitrary data, bmesh module has "custom data layers" mechanism; Again, I can't find any documentation about what bmesh ops are supposed to do with that custom layers. And I didn't manage to check it experimentally, yet.

Durman:
I think it worth keep experimenting. Material is only one possible case among many others. I believe that bmesh features are clever enough and probably they even take in account UV mapping after topological changes of a mesh.

`Inset individual` operator for example has such option:
**Interpolate** - Interpolate mesh data: e.g. UV’s, vertex colors, weights, etc.

This can probably give a little explanation.
https://docs.blender.org/api/current/bmesh.html#customdata-access

Durman:
It looks works.
```Python
import bpy
import bmesh
from bmesh.ops import inset_individual

verts = [[0,0,0], [1,0,0], [2,0,0], [0,1,0], [1,1,0], [2,1,0]]
faces = [[0,1,4,3], [1,2,5,4]]

bm = bmesh.new()
bm.faces.layers.int.new('sv')  # create custom layer
sv = bm.faces.layers.int.active  # pointer to custom layer ???

[bm.verts.new(co) for co in verts]
bm.verts.ensure_lookup_table()
for i, face in enumerate(faces):
    bmf = bm.faces.new([bm.verts[i] for i in face])
    bmf[sv] = i  # assing value to custom layer

inset_individual(bm, faces=list(bm.faces), 
                     thickness=0.1, 
                     depth=0.0, 
                     use_even_offset=True,
                     use_interpolate=True, 
                     use_relative_offset=False)

[print(face[sv]) for face in bm.faces]  # print values from custom layer
```
Integer value can point to a element of array of any type values.

portnov:
I've checked it with remove_doubles. It also works. Splendid! So, we actually can work with (almost) arbitrary per-face data.

The next question is how should we represent such data in Sverchok sockets. Obviously having one socket for "int" layer, one for "float" and one for "float", for each node — is not a way. If we are going this way at all, we have to have some generic data structure for arbitrary data. Maybe a dict would do. On the other hand, in this diversity of "arbitrary data" there are some things with very clear semantics — material index, for example. So, I'd suggest something like

```
class SvItemData(object):
    def __init__(self, *args, **kwargs):
        self.int = dict()
        self.float = dict()
        self.string = dict()
        self.material_index = None
        self.tag = None
        self.select = None

    @staticmethod
    def from_bmesh_faces(bm):
        result = []
        for face in bm.faces:
            data = SvItemData()
            data.material_index = face.material_index
            data.tag = face.tag
            data.select = face.select
            for key, layer in bm.faces.layers.int.keys():
                data.int[key] = face[layer]
            for key, layer in bm.faces.layers.float.keys():
                data.float[key] = face[layer]
            for key, layer in bm.faces.layers.string.keys():
                data.string[key] = face[layer]
            result.append(data)
        return result

    @staticmethod
    def assign_to_bmesh_faces(data_list, bm):
        for data, face in zip(data_list, bm.faces[:]):
            if data.material_index is not None:
                face.material_index = data.material_index
            if data.tag is not None:
                face.tag = data.tag
            if data.select is not None:
                face.select = data.select

            for key, value in data.int.iteritems():
                layer = bm.faces.layers.int.get(key)
                if layer is None:
                    layer = bm.faces.layers.int.new(key)
                face[layer] = value

            for key, value in data.float.iteritems():
                layer = bm.faces.layers.float.get(key)
                if layer is None:
                    layer = bm.faces.layers.float.new(key)
                face[layer] = value

            for key, value in data.string.iteritems():
                layer = bm.faces.layers.string.get(key)
                if layer is None:
                    layer = bm.faces.layers.string.new(key)
                face[layer] = value
```
(plus `__repr__` implementation).
?

portnov:
So,
* Any node that uses bmesh internally can now have "FaceData" input and output. Arbitrary per-face python data list can be fed into, and we should expect some meaningful output.
* Any node that deals with mesh without using bmesh, can also have "FaceData" sockets, but in this case such data has to be handled manually.
* Currently, this is only implemented for "remove doubles" and "adaptive polygons" only. In future, this will be slowly implemented in other nodes.
* "Objects In Mk3" now has "Materials" output.
* "Bmesh Viewer" now has "materials" input.

portnov:
adaptive_icosphere_32

Durman:
As you have seen in the `inset faces` node it is no necessary to keep real data inside bmesh object it is possible just to keep references to data. This approach looks faster and can be provided with any Python data types like user classes or numpy arrays. 

The only reason to put real data inside bmesh object is when an operator involve such data in process for example to interpolat values according position of new mesh elements.

Durman:
Are `Objects in` and `Bmesh viewer` nodes have material data or just indexes of material data in material socket? Probably name of cocket should more explicit if it keeps only indexes.

portnov:
Actually they use material indexes only.

Durman:
In this case I would recommend to rename the sockets to `material_indexes` or something like this because I expect that material objects might occupy the material name one day.

portnov:
"MaterialIndex" is just too long to my taste to be the socket name. So let's call it just "MaterialIdx".

Issue #1910, opened 2017-12-07
Bmesh functions in extrude_separate.py


Durman:
There are linear and cubic complexity of some function. The Analyzing function usually use the cubic complexity. I noticed that `bmesh.ops.scale` and `bmesh.ops.translate` have the cubic complexity without any reason for it. It have became apparently after I tried to extrude many thousands of faces. I don't know why it has that behavior but I think if rewright this to Python, this function will become faster.
Test for 10 faces.
![2017-12-07_14-18-36](https://user-images.githubusercontent.com/28003269/33710368-be058fb6-db59-11e7-87d6-3727e4c7cc53.png)
Test for 100 faces.
![2017-12-07_14-20-45](https://user-images.githubusercontent.com/28003269/33710435-f580c906-db59-11e7-9815-43fd5b6879ef.png)
Test for 1000 faces.
![2017-12-07_14-22-18](https://user-images.githubusercontent.com/28003269/33710454-0d9c4c2c-db5a-11e7-87f9-fffbe55db2ae.png)
Test for 10000 faces.
![2017-12-07_14-23-34](https://user-images.githubusercontent.com/28003269/33710520-410bcef2-db5a-11e7-809f-a847e5b937be.png)
You can see that, that the more faces in object the much time  function take per face. So it is very weird.
And there is no problem if object will be broken into faces.

zeffii:
yep. a very similar node is "Inset Specials" and it is dramatically faster.

zeffii:
see: https://github.com/nortikin/sverchok/issues/1488

zeffii:
I will gladly accept a PR to speed up this node.

portnov:
@Durman Do you mean that blender's internal methods are slow? Or slow is some part of our Python "glue" around them?
It would surprise me if it is possible to rewrite bmesh operations in python faster than they are (implemented in C).

portnov:
Also, if some of bmesh operations are really so slow, it desires a bugreport to blender.

zeffii:
hmm..
https://github.com/dfelinto/blender/blob/1d33843ea6b128242183bb31804474019c606a9d/source/blender/bmesh/operators/bmo_utils.c#L74-L98

and
https://github.com/dfelinto/blender/blob/1d33843ea6b128242183bb31804474019c606a9d/source/blender/bmesh/intern/bmesh_opdefines.c#L758-L795

zeffii:
[...]

PR #2733, opened 2019-12-12
Inset faces node

Durman:

This PR is attempt to utilize `inset face` features ([inset_individual](https://docs.blender.org/api/current/bmesh.ops.html#bmesh.ops.inset_individual), [inset_region](https://docs.blender.org/api/current/bmesh.ops.html#bmesh.ops.inset_region)) of bmesh operators.

There is known issues with bmesh operators. They does not design to deal with multiple values:
#1910

I have made some trick to avoid the problem and it looks like it's working.
Now working time grows linear with linear increasing number of input faces. 

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Ready for merge.


Durman:
@portnov I mean like this approach:

By the way, `inset_individual` operator have managed perfectly with distributing custom data of faces.))

Durman:
10000 custom insertions for 1 second.  Not too bad I think. =)

portnov:
So you are using a "custom data layer" to store "index of the face in the initial mesh" to untangle face indexes after the bmesh op. I wonder if this approach has better or worse performance comparing to storing each field of "per-face data" in separate custom data layer. Will check...

portnov:
It appears that your approach is faster, at least comparing to my implementation. Will leave it here, for future reference

```python
class SvItemData(object):
    def __init__(self, *args, **kwargs):
        self.int = dict()
        self.float = dict()
        self.string = dict()
        self.material_index = None
        self.tag = None
        self.select = None

        if args:
            if len(args) == 1:
                value = args[0]
                if isinstance(value, (int, str, float)):
                    self.value = value
                elif isinstance(value, (list, tuple)):
                    self.set_list("value", value)
                else:
                    raise TypeError("Unsupported type of single argument for SvItemData(): {}: {}".format(value, type(value)))
            else:
                self.set_list("value", args)

        if kwargs:
            for key, value in kwargs.items():
                self[key] = value

    def set_list(self, key_base, values):
        for i, value in enumerate(values):
            key = "{}[{}]".format(key_base, i)
            self[key] = value

    def __getitem__(self, key):
        if key in self.int:
            return self.int[key]
        elif key in self.float:
            return self.float[key]
        elif key in self.string:
            return self.string[key]
        else:
            return None

    def __setitem__(self, key, value):
        if isinstance(value, int):
            self.int[key] = value
        elif isinstance(value, float):
            self.float[key] = value
        elif isinstance(value, str):
            self.string[key] = value
        else:
            raise TypeError("Unsupported type of custom data for key {}: {}: {}".format(key, value, type(value)))

    def __int__(self):
        value = self.value
        if isinstance(value, int):
            return value
        else:
            raise TypeError("This is not an integer value")

    def __float__(self):
        value = self.value
        if isinstance(value, float):
            return value
        else:
            raise TypeError("This is not a float value")

    def __str__(self):
        value = self.value
        if isinstance(value, str):
            return value
        else:
            raise TypeError("This is not a string value")

    @property
    def value(self):
        return self["value"]

    @value.setter
    def value(self, new_value):
        self["value"] = new_value

    @staticmethod
    def from_bmesh_faces(bm):
        result = []
        for face in bm.faces:
            data = SvItemData()
            data.material_index = face.material_index
            data.tag = face.tag
            data.select = face.select
            for key, layer in bm.faces.layers.int.items():
                data.int[key] = face[layer]
            for key, layer in bm.faces.layers.float.items():
                data.float[key] = face[layer]
            for key, layer in bm.faces.layers.string.items():
                data.string[key] = face[layer]
            result.append(data)
        return result

    @staticmethod
    def assign_to_bmesh_faces(data_list, bm):
        data_list_up = []
        for data in data_list:
            if not isinstance(data, SvItemData):
                data = SvItemData(data)
            data_list_up.append(data)
            for key in data.int:
                layer = bm.faces.layers.int.get(key)
                if layer is None:
                    print("new int key:", key)
                    bm.faces.layers.int.new(key)
            for key in data.float:
                layer = bm.faces.layers.float.get(key)
                if layer is None:
                    print("new float key:", key)
                    bm.faces.layers.float.new(key)
            for key in data.string:
                layer = bm.faces.layers.string.get(key)
                if layer is None:
                    print("new string key:", key)
                    bm.faces.layers.string.new(key)

        for data, face in zip(data_list_up, bm.faces):
            if data.material_index is not None:
                face.material_index = data.material_index
            if data.tag is not None:
                face.tag = data.tag
            if data.select is not None:
                face.select = data.select

            for key, value in data.int.items():
                layer = bm.faces.layers.int.get(key)
                #print("L[{}]: {}".format(key, layer))
                #print("F: {}: {}: {}".format(face, layer, value))
                face[layer] = value

            for key, value in data.float.items():
                layer = bm.faces.layers.float.get(key)
                if layer is None:
                    layer = bm.faces.layers.float.new(key)
                face[layer] = value

            for key, value in data.string.items():
                layer = bm.faces.layers.string.get(key)
                if layer is None:
                    layer = bm.faces.layers.string.new(key)
                face[layer] = value

```

Durman:
Face mask selection output socket:
face_mask

Durman:
`Inset region` mode with random face selection of a cube.

portnov:
What is the difference between "extrude separate/region" and "inset separate/region"?

Durman:
I think their are different only in ultimate desirable result. But some extra options make them nearly or maybe entirely equal. At least, extruding does not expect scaling of inserted face. Also they can shoes normals differently for shifting inserted faces.

Durman:
Improving selection mask:
mask generation

PR #2740, opened 2019-12-15
Delaunay from mathutils

Durman:

Utilizing `mathutils.geometry.delaunay_2d_cdt`:

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/modifier_make/delaunay_2d_cdt.rst
+++ b/docs/nodes/modifier_make/delaunay_2d_cdt.rst
@@ -1,0 +1,60 @@
+Random points on mesh
+=====================
+
+.. image:: https://user-images.githubusercontent.com/28003269/70866264-f088e600-1f80-11ea-91b2-f8b785c597a3.png
+
+Functionality
+-------------
+Computes the Constrained Delaunay Triangulation of a set of vertices,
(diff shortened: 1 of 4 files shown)


Durman:
Wow! The function can just find intersection of polygons!

Durman:
I'm absolutely amazed by this new function. Actually capabilities of the function much more larger than its name sound. It looks like that it include all feature of all my `geom2D` package! 

I will be happy to rewrite the nodes which use `geom_2D` package and include this function because it should be much more stable and efficint. But probably it will not be without troubles. 

I already have catched several crushes of Blender until I did not figure out what each mode do. I hope that if input will be correct the function shows itself stable.

Durman:
Well, I have tested it out. The function looks like does not design for large input data. Blender is crashing with 6000 intersections. With 5000 intersections the function shows much less performance then Python `geom_2d`. Nevertheless I think it's great alternative with small input meshes because it should be more sable in this case.

Durman:
200 000 points - time 4 seconds

Durman:
Keeping color initial faces:

Durman:
Combine of a line and random points into Delaunay triangulation:

PR #2744, opened 2019-12-15
Fix lamps handling.

### LINKED ISSUE #2742: lamp node
## Problem statement

Looks like the node should be fixed to new Blender API.

```Python
  File "C:\Users\....\Blender_code\addons\sverchok\nodes\viz\lamp_out.py", line 206
, in make_lamp
    lamps_data = bpy.data.lamps
AttributeError: 'BlendData' object has no attribute 'lamps'
```

Blender 2.81

portnov:
This should fix #2742 .
@Durman please test.

### DIFF
--- a/json_examples/Introduction/BentTorus.json
+++ b/json_examples/Introduction/BentTorus.json
@@ -1,0 +1,352 @@
+{
+  "export_version": "0.079",
+  "framed_nodes": {
+    "A Number": "Frame.001",
+    "Move": "Frame.001",
+    "Note.001": "Frame",
+    "Note.002": "Frame.001",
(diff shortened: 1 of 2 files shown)


Durman:
Is it supposed to work in Eevee? If so then next things does not work:
- strength
- cast shadows (always do)
- hemi type (does not exist any more)

Probably something else but this list is what I think is important.

portnov:
I tested this node with Cycles only. It seems support of Eevee will take more time.
For now I'm merging this as is, please feel free to implement Eevee support :)
Or make another issue, some day someone will implement it.

PR #2750, opened 2019-12-17
Triangle Fill node (WIP)

portnov:
bmesh.ops.triangle_fill.

Screenshot_20191217_230413

## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/index.md
+++ b/index.md
@@ -113,6 +113,7 @@
     SvMeshJoinNode
     ---
     SvFillsHoleNode
+    SvTriangleFillNode
     SvRecalcNormalsNode
     SvFlipNormalsNode
     ---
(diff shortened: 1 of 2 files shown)


portnov:
Screenshot_20191218_230607

Durman:
Strange result:

portnov:
Yeah. This bmesh function is not very clever, it does not even try to detect "holes". It works properly only if you specify the edges of the hole to be filled by EdgeMask input.
We could automatically detect holes by calling bmesh.ops.holes_fill() first; but then it is not clear, why do we need to use triangle_fill() at all — we could triangulate faces returned by holes_fill() instead.
So I'm not currently sure if this node is very useful.

Durman:
I think we should at first to make up a purpose of designed node and then to think how existing tools can help us. Or we should to investigate existing tools and try to think what can they bring to Sverchok.)) Straight implementation of Bmesh operators has less use I think. 

In this particular case I see that this operator can fill closed edge loops with triangles. So the input of this node should be set of closed edge loops I think if it would not use other alorithms.

PR #2756, opened 2019-12-20
track origin of error in objin node after loading .blend

zeffii:
This problem will only be noticed if the object that Objects In is referencing requires you to pick up modifier data too - and that requires despgraph  evaluation.

the new depsgraph acquisition code, has a problem when a .blend is loaded from disk, it seems
```
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\nodes\scene\objects_mk3.py", line 281, in process
    obj = depsgraph.objects[obj.name]
KeyError: 'bpy_prop_collection[key]: key "Sv_0" not found'
```
depsgraph.objects is empty until triggered or sometehing?

no eta.  but maybe the` .blend load ` handler must be scrutinized. (rather than fixing it in the objects in Process function...)

### DIFF
--- a/nodes/scene/objects_mk3.py
+++ b/nodes/scene/objects_mk3.py
@@ -16,6 +16,8 @@
 #
 # ##### END GPL LICENSE BLOCK #####
 
+import traceback
+
 import bpy
 from bpy.props import BoolProperty, StringProperty
 import bmesh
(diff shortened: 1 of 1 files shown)


zeffii:
it seems the problem is only noticed if the Scene object being references is hidden from view. weird.

Issue #2766, opened 2019-12-25
Dictionary socket


Durman:
## Problem statement

Recently nodes have been starting to get face data sockets. Sense of this socket is to include all data which related with given face. After a node processing it rebuild this data according topological changes. The problem is that faces can have a whole bunch of related data. String socket (list socket) can't handle with this problem now.

## Solution

One of the solutions can be adding one more list level. But it looks like that human brain does not design for handling with such data structures.)) My proposal is to add new socket type: dictionary.

Dictionary could have arbitrary number of keys. Values of cays can include any Sverchok data. Dictionary in node can looks in this way:

```Text
|      dictionary in     |
+------------------------+
|                   dict |
| item 1 name (editable) |
| item 2 name ...        |
```

**Vectorization of dictionary in node:**
If to plug list of values like this [[0,1,2,3,4]] it produces one dictionary [dict1]. 
If to plug several lists [[0,1,2,3,4],[5,6,7,8]] then output looks like this [dict1, dict2].

**Nested dictionaries:**
Also dictionaries could include another dictionaries. Probably nested dictionaries could solve problem of handling data structures with arbitrary nestedness or at list to simplify solution of the problem. 

For creating nested dictionary it will be enuff to plug dictionary socket to `dictionary in` node. So it would possible to create complex data structures from tree of nodes. Most tricky thing is to make up how such data structure should me handled. Probably some recursive algorithm should be appropriate.

**Appearance:**
I think the socket should have different color from other type of sockets because it does not have compatibility with them. Also the nodes `dictionary in` and `dictionary out` looks like deserve to have own category right below list category. Also I expect more nodes handling dictionaries can fill this category.

Durman:
![dict in](https://user-images.githubusercontent.com/28003269/71527090-4cddf500-28f3-11ea-9c4c-e39c577ad1c3.gif)

Durman:
Control over construction of dictionaries structure.
![control](https://user-images.githubusercontent.com/28003269/71539920-c2ca7680-295c-11ea-891b-5f17953c2c06.gif)

nortikin:
as additinal socket type yes. levels are great, but maybe levels container could be better

Durman:
What do you mean by levels container?

nortikin:
> What do you mean by levels container?

omg, not remember exactly what i ment.
maybe it was mistake

Issue #2772, opened 2019-12-27
Drawing number of objects on sockets


Durman:
## Problem statement

The function is calling each time when UI is redrawing. UI is redrawing each time when mouse is passed over any interactive element on a screen. The function is calling `other` method of socket which is calling `links` method which is actually quite expensive. Calling `links` method takes O(n) time where `n` is number of links in a tree. 
https://github.com/nortikin/sverchok/blob/61deb2ff87885fae0489c8016e03c2e5fa604dfd/core/socket_data.py#L47-L67
So with growing size of a tree number of calculations is growing exponentially and recalculating quite often. It definitely can be reimplemented. I think that number of objects can be recalculated during tree update event and memorize in sockets somewhere.

zeffii:
perhaps, the socket class `common` can hold a 'object count', which is set only at the end of a successful completion of a node's `process()`  (also means setting to a default 0 prior to calling `process` ).

i've thought about this too in the past when visiting these areas of code, and i'm supportive of reducing redundant cpu cycles, but here i'm apprehensive to implement `object count`.

zeffii:
but whatever :)  we can try it

ly29:
Having a compiled links data structure would be good and would also make the update system easier. 
Removing reroutes and wifi (if that is still is used) is a big win and would make the update system easier.

ly29:
Made a very simple cache implementation for these kind of issues. Untested as a minimal implemenation.
https://github.com/nortikin/sverchok/blob/testNameCache/core/socket_data.py

Durman:
By the way I was thinking about `socket.sv_links` approach and it looks it would be impossible to implement. Because what this method should returns? We can't keep straight references to nodes. Links to nodes can't be created via Pointer property. We can keep their names but search node by their name again returns to O(n**2) performance.

nortikin:
default off
optionally turn it on may be better experience.

Durman:
![lags](https://user-images.githubusercontent.com/28003269/82578548-93333980-9b9d-11ea-85f8-1c05b48d3651.gif)

https://github.com/nortikin/sverchok/blob/09e9788c53c7c92b447c02f60e1851a4aecfedc2/core/sockets.py#L229
And performance without calling the function.
```Python
info_text = t + '. ' 
```

![no lags](https://user-images.githubusercontent.com/28003269/82578703-cd9cd680-9b9d-11ea-937a-86260d28d2dc.gif)

nortikin:
[...]

Issue #454, opened 2014-10-12
Node methods for Sverchok 


ly29:
Note that sverchok isn't added to the python path so all imports from sverchok has to start with ```sverchok```.
Like this.
```python
from sverchok.node_tree import updateNode
```

**WIP**
This is a outcast for documentation for writing a well behaved node for Sverchok 0.5

```python
def update(self):
```
```update``` is called on user interface changes (link/socket creation) in the node group. This method is optional and only for nodes that have dynamic sockets of some kind. Can also be used to change color based on links etc.

```python
def process(self):
```
```process``` is the main processing method for the node.
 
**Note** that it is okay for this method to fail, do not catch errors that are not recoverable, for example a socket that needs data but is empty. For pure ui nodes this isn't needed. No socket creation or linking of nodes should take place in this method.
It is safe to use the ```.is_linked``` property of sockets. This is much quicker than ```.links```.

```python
def sv_init(self, context):
    # create nodes etc
```
Replaces ```def init(self, context)``` for Sverchok, but it will be called from ```def init(self, context)``` in the Sverchok Node baseclass. Optional.

```python
def bake(self):
```
For nodes that can bake themselves this method will be called from Sverchok when the bake all buttons is pressed for the specified node group.

Below are the standard methods for node with some specific notes for using them with Sverchok.

```python
def draw_label(self):
    if self.string_name:
         return string_name
    else:
        return self.bl_label
```
Optional use this for setting dynamic label, if the label is set by the user it won't be called. Return a string to show as label. 

```python
def draw_buttons(self, context, layout):
# draw the ui on the Node
```
```python
def draw_buttons_ext(self, context, layout):
# draw the node UI in the side panel
```
Useful for special properties that require a lot of setup or space. It is good practice to draw the main ui as well, perhaps with different formatting otherwise just use ```self.draw_buttons(context, layout)``` to pass drawing control to main layout.

```python
def copy(self, node):
```
Optional method to be called upon copying of a node, useful for resetting properties. Node is the node self is being copied from.

```python
def free(self):
```
Is called upon deletion of a node, but not closing down a layout, closing a file or similar.

```python
def poll_instance(self, node_tree)
```
To check if this instance of node can be added to the node tree. Not used right now.

**bpy.props**

```python
def process_node(self, context)
```
[...]

PR #2774, opened 2019-12-28
Dictionary in and dictionary out nodes

Durman:

#2766

image

**Todo:**

- [x] Take default name of key from linked socket.
- [x] Keep links of `dictionary out` nodes when a key of a dictionary was renamed. ~Next time or probably it even does not worth complicate too much.~ Simplest solutions was adding custom attribute for dictionaries.
- [x] Robustness of determining type of output sockets. ~Next time.~ Simplest solutions was adding custom attribute for dictionaries.
- [x] Consider handling of several dictionaries in list (ignoring mismatched keys?).
- [x] Сustom draw for all sockets.
- [x] `Is_linked` should have name `was_linked` before start of update event, more correct test is `socket.links` but as I suppose more expensive as well. Looks like in `sv_get` method `is_linked` should be replaced by `links`.
- ~deepcopy = False where it possible~ Deep copy is not implemented for dictionaries. In next PR.
- [x] New dictionary icons :smile: 

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.

### DIFF
--- a/core/socket_data.py
+++ b/core/socket_data.py
@@ -89,7 +89,7 @@ def SvGetSocket(socket, deepcopy=True):
     set to False and increase performance substanstilly
     """
     global socket_data_cache
-    if socket.is_linked:
+    if socket.links:
         other = socket.other
         s_id = other.socket_id
         s_ng = other.id_data.name
--- a/core/socket_data.py
+++ b/core/socket_data.py
@@ -89,7 +89,7 @@ def SvGetSocket(socket, deepcopy=True):
     set to False and increase performance substanstilly
     """
     global socket_data_cache
-    if socket.links:
+    if socket.is_linked or socket.sv_is_linked:
         other = socket.other
         s_id = other.socket_id
         s_ng = other.id_data.name
--- a/core/socket_data.py
+++ b/core/socket_data.py
@@ -89,7 +89,7 @@ def SvGetSocket(socket, deepcopy=True):
     set to False and increase performance substanstilly
     """
     global socket_data_cache
-    if socket.is_linked or socket.sv_is_linked:
+    if socket.is_linked:
         other = socket.other
         s_id = other.socket_id
         s_ng = other.id_data.name
--- a/core/sockets.py
+++ b/core/sockets.py
@@ -62,7 +62,7 @@ class SvSocketCommon:
     expanded: BoolProperty(default=False)
     custom_draw: StringProperty(description="For name of method which will draw socket UI (optionally)")
     prop_name: StringProperty(default='', description="For displaying node property in socket UI")
-    sv_is_linked: BoolProperty(description="Node is linked if (node.is_linked or node.sv_is_linked). "
+    sv_is_linked: BoolProperty(description="Socket is linked if (socket.is_linked or socket.sv_is_linked). "
                                            "`Sv_is_linked` should be set manually. "
                                            "Useful for socket created dynamically during update event.")
 
--- a/core/sockets.py
+++ b/core/sockets.py
@@ -62,9 +62,6 @@ class SvSocketCommon:
     expanded: BoolProperty(default=False)
     custom_draw: StringProperty(description="For name of method which will draw socket UI (optionally)")
     prop_name: StringProperty(default='', description="For displaying node property in socket UI")
-    sv_is_linked: BoolProperty(description="Socket is linked if (socket.is_linked or socket.sv_is_linked). "
-                                           "`Sv_is_linked` should be set manually. "
-                                           "Useful for socket created dynamically during update event.")
 
     quicklink_func_name: StringProperty(default="", name="quicklink_func_name")
 
@@ -285,7 +282,7 @@ class SvObjectSocket(NodeSocket, SvSocketCommon):
             layout.label(text=text)
 
     def sv_get(self, default=sentinel, deepcopy=True, implicit_conversions=None):
-        if (self.is_linked or self.sv_is_linked) and not self.is_output:
+        if self.is_linked and not self.is_output:
             return self.convert_data(SvGetSocket(self, deepcopy), implicit_conversions)
         elif self.object_ref:
             obj_ref = bpy.data.objects.get(self.object_ref)
@@ -313,7 +310,7 @@ class SvTextSocket(NodeSocket, SvSocketCommon):
         return (0.68,  0.85,  0.90, 1)
 
     def sv_get(self, default=sentinel, deepcopy=True, implicit_conversions=None):
-        if (self.is_linked or self.sv_is_linked) and not self.is_output:
+        if self.is_linked and not self.is_output:
             return self.convert_data(SvGetSocket(self, deepcopy), implicit_conversions)
         elif self.text:
             return [self.text]
@@ -344,7 +341,7 @@ class SvMatrixSocket(NodeSocket, SvSocketCommon):
 
     def sv_get(self, default=sentinel, deepcopy=True, implicit_conversions=None):
         self.num_matrices = 0
-        if (self.is_linked or self.sv_is_linked) and not self.is_output:
+        if self.is_linked and not self.is_output:
             source_data = SvGetSocket(self, deepcopy = True if self.needs_data_conversion() else deepcopy)
             return self.convert_data(source_data, implicit_conversions)
 
@@ -372,7 +369,7 @@ class SvVerticesSocket(NodeSocket, SvSocketCommon):
             return {}
 
     def sv_get(self, default=sentinel, deepcopy=True, implicit_conversions=None):
-        if (self.is_linked or self.sv_is_linked) and not self.is_output:
+        if self.is_linked and not self.is_output:
             source_data = SvGetSocket(self, deepcopy = True if self.needs_data_conversion() else deepcopy)
             return self.convert_data(source_data, implicit_conversions)
 
@@ -404,7 +401,7 @@ class SvQuaternionSocket(NodeSocket, SvSocketCommon):
             return {}
 
     def sv_get(self, default=sentinel, deepcopy=True, implicit_conversions=None):
-        if (self.is_linked or self.sv_is_linked) and not self.is_output:
+        if self.is_linked and not self.is_output:
             source_data = SvGetSocket(self, deepcopy = True if self.needs_data_conversion() else deepcopy)
             return self.convert_data(source_data, implicit_conversions)
 
@@ -436,7 +433,7 @@ class SvColorSocket(NodeSocket, SvSocketCommon):
             return {}
 
     def sv_get(self, default=sentinel, deepcopy=True, implicit_conversions=None):
-        if (self.is_linked or self.sv_is_linked) and not self.is_output:
+        if self.is_linked and not self.is_output:
             return self.convert_data(SvGetSocket(self, deepcopy), implicit_conversions)
 
         if self.prop_name:
@@ -461,7 +458,7 @@ class SvDummySocket(NodeSocket, SvSocketCommon):
         return self.other.get_prop_data()
 
     def sv_get(self):
-        if (self.sv_linked or self.sv_is_linked):
+        if self.sv_linked:
             return self.links[0].bl_idname
 
     def sv_type_conversion(self, new_self):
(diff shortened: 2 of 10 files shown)


portnov:
What's the point of the up/down buttons? Python's dicts are unordered anyway. Or are you using collections.OrderedDict instead?...

portnov:
A "dictionary out" node can, I presume, handle a list of dictionaries and convert it to several lists (one per dictionary key)?
What if the list of dictionaries contains dicts with different sets of keys?

Durman:
> What's the point of the up/down buttons? Python's dicts are unordered anyway.

Since 3.7 dictionaries are insertion ordered. `Collection` module is deprecated and should not used anymore.

> A "dictionary out" node can, I presume, handle a list of dictionaries and convert it to several lists (one per dictionary key)?
What if the list of dictionaries contains dicts with different sets of keys?

Yes it handles list of dictionaries but logic can seem a little bit tricky. Output sockets are being created according keys of first input dictionary and will be added with data from other dictionarys with the same keys. But if other dictionaries has other keys data from them will be ignored. But actually such list of dictionary will be impossible to create at least with `dictionary in` node.

Durman:
I found that information kept in dictionary is not sufficient for determining type of output sockets. The node can guess but not always. 
Also there is such case when data related with key is not changed but key does. So such keys should be considered as renaming event and output socket of `dictionary out` nodes should be just renamed. But now the node `thinks` that new data have came and recreates socket and drops links from it. 

So for solving this cases some extra information should be provided into a `dictionary out` node.

Such information can be store in different places. For example it can be stored directly in dictionary data as extra key like `_meta_data` which should be handled separately. I does not like this approach. I thinks users of Sverchok definetly should not look at this data via `stethoscope` node. Also I tried to add this data to socket Blender objects but this also does not look as best way. 

Probably extra method for dictionary sockets like `sv_get_meta` and `sv_set_meta` would be best. This method should give access to hidden from user dictionary which could keep information like from which type of socket data came and what ID the socket had.

For this PR I think I'm not going to make things too complicated. I will leave as it is now. In generally it works. If the cases will be seemed too annoying it is time for thinking about implementation of the solution.

portnov:
As an option, you can generate a special class each time. Maybe a subclass of `dict`. Python is perfectly capable of generating classes on-the-fly...

Durman:
Finally I found simple solution in using dictionary with costume attribute where metadata is keeping. It will let to nodes which deal with dictionary but does not change their shapes don't care about metadata at all.

Durman:
Any objection from adding dictionaries to Sverchok?

@zeffii can you check changes of `spcket_data.py`, `spckets.py`, `data_structure.py` files?

zeffii:
i'll look at those, but really the purple socket is not pleasant to look at, maybe we can use `white` + different `socket shape`.

purple socket color is a hard "no" from me.

core/socket_data.py
```diff
@@ -89,7 +89,7 @@ def SvGetSocket(socket, deepcopy=True):
     set to False and increase performance substanstilly
     """
     global socket_data_cache
-    if socket.is_linked:
+    if socket.links:
```
zeffii: i recall from @ly29 that there was a good reason to favour using ".is_linked" instead of ".links"

https://github.com/nortikin/sverchok/issues/454

`.links` may come with a cpu penality,  the decision to use it was not arbitrary.

zeffii: but hey, we just want stuff to work. so maybe we could track our own `.sv_links` ...eventually.  or propose an improvement to the pynode api.

core/sockets.py
```diff
@@ -44,7 +44,8 @@
     "SvDummySocket": (0.8, 0.8, 0.8, 0.3),
     "SvSeparatorSocket": (0.0, 0.0, 0.0, 0.0),
     "SvObjectSocket": (0.69, 0.74, 0.73, 1.0),
-    "SvTextSocket": (0.68, 0.85, 0.90, 1)
+    "SvTextSocket": (0.68, 0.85, 0.90, 1),
+    "SvDictionarySocket": (0.40, 0.02, 1.00, 1.00)
```
zeffii: NOOOOOOOOOOO :)  i dislike this colour a lot

nodes/dictionary/dictionary_in.py
```diff
@@ -0,0 +1,124 @@
...
+        data = [chain(sock.sv_get(), cycle([None])) for sock in list(self.inputs)[:-1]]
+        keys = [sock.prop_name for sock in list(self.inputs)[:-1]]
+        out = []
+        for i, *props in zip(range(max_len), *data):
+            out_dict = SvDict({getattr(self, key): prop for key, prop in zip(keys, props) if prop is not None})
+            for sock in list(self.inputs)[:-1]:
+                out_dict.inputs[sock.identifier] = {'type': sock.bl_idname,
+                                                    'name': getattr(self, sock.prop_name),
+                                                    'nest': sock.sv_get()[0].inputs
+                                                    if sock.bl_idname == 'SvDictionarySocket' else None}
```
zeffii: I think indentation allows you to write this way, and keep the line short
```python
                out_dict.inputs[sock.identifier] = {
                    'type': sock.bl_idname,
                    'name': getattr(self, sock.prop_name),
                    'nest': sock.sv_get()[0].inputs if sock.bl_idname == 'SvDictionarySocket' else None}
```
it's slightly easier to read

core/sockets.py
ly29: .links was computed on every lookup.

core/socket_data.py
ly29: .links is actually slow enough to be noticed from time to time. The blow is the implemetation.

```
@property
def links(self):
    """List of node links from or to this socket. Warning: takes O(len(nodetree.links)) time."""
    return tuple(
        link for link in self.id_data.links
        if (link.from_socket == self or
            link.to_socket == self))
```

ly29: I am also quite pleased that my old code is still used.

core/sockets.py
Durman: @zeffii @portnov @vicdoval @DolphinDream @nortikin @enzyme69 Please, have a look at examples of color of dictionary sockets, which variant is more pleasant for you?
# Node fotosession
## Lite
**1:** (1,1,1,1)


**2:** (0.55, 0.35, 0.75, 1.0)


**3:** (0.85, 0.55, 0.35, 1.0)


## Dark
**1:** (0.85, 0.15, 0.15, 1.0)


**2:** (0.0, 0.30, 0.80, 1.0)


**3:**  (0.60, 0.00, 0.60, 1.0)

portnov:
I vote for option 1 :)

core/socket_data.py
Durman: according this #2772,solution should not make performance more lousy than it is now.))) 

I agree of course that it is  much more slower. But I think it is better to have more robust solution then faster. I agree that `sv_links` can solve the problem but I think in independent PR. 
`if self.links:` can be temporary solution.

ly29: What is the problem with .is_linked in this context?

Durman: The problem is that when one of a sockets of `dictionary in` node is being plugged it cause update event during which the attribute `is_linked` of the socket returns False.

`Socket: Dict in sock, is_linked: False, links: True`

The result is that during process of the node `sv_get` can't get catched data from the socket.

This is only one case when `is_linked` is not appropriate, in all other cases it works as expected.

Durman: Well, the reason why we did not faced with such problem before is that in `dictionary in` node I replace input sockets by another socket with type according type of `other` socket. 

I don't know probably I should not do this but it looks like that node tree looks more readable with type of inputs socket equal to type of socket to which their are linked.

Durman: If variant below is acceptable then probably we could avoid using `socket.links` test.

ly29:
I understand why you want to do this but It breaks the abstraction of the update system with the current approach since socket are created from process and a number of other issues. The update system certianly isn't flawless and overly complicated but it has some restraints that needs to be respected for it to kind of work.

Replace socket used to work well for managing replace sockets and keeping links, if it doesn't work fix the issue there since it is slightly tricky to keep the updates from working properply and it is used from other places.

core/socket_data.py
Durman: Probably possible variant is to create `sv_is_linked` method for sockets which will be False by default but can be changed manually. And then such checking should work `if (socket.is_linked or socket.sv_is_linked) or something else:`

Durman:
Yes, `dictionary out` node can `redraw` sockets during Sverchok update event indeed. It can't be done during update event of node because view of output sockets is dependent upon input data. So if number of keys or names of keys or even their order is changed the node redraw output sockets. 

Actually we slightly modernized update system recently and handle such cases with throttling update event which can be caused be redrawing sockets in this way. So I don't see what troubles can such approach cause else. 

Probably it is inconsistent to do such job from process. Update system should be rewritten anyway and probably we could include this feature, something like preprocess method.

portnov:
@Durman it appears the node behaves not as I expected...

Screenshot_20200103_180928

I expected it to output
```
[{A: 0, B: 3}, {A: 1, B: 2}, {A: 2, B: 1}]
```

zeffii:
both outputs could be useful, but the one shown in the screenshot is what _i_ would expect.

Durman:
@portnov I does not agree with your expectations.))

I think for simplicity we should forgot about existing of vectorization in Sverchok for a moment. In this case creating list of dictionaries just would not be possible. The purpose of the `dictionary in` node is create dictionary and not create list of dictionaries. So you just should create key and then assign to the key any data you wish of any shape. The `dictionary in` node is absolutely indifferent about input data. 

Now we can recall vectorization. As I understand its purpose is to repeat the same logic with different input data. So the node just takes each list of values [[list of values 1], [list of values 2] ,...] (sverchok layer I would call) and creates dictionaries for each `layer` how I have discribed it above.

Durman:
@zeffii have a look to the last commit. I have ridded of using `socket.links` by adding new `sv_is_linked` parameter which is solving my corner case problem. It is always false and should be changed manually, so it does not impact on logic of all other nodes. What do you think about such solution?

portnov:
Screenshot_20200104_124639
okay...

core/sockets.py
```diff
@@ -62,6 +62,9 @@ class SvSocketCommon:
     expanded: BoolProperty(default=False)
     custom_draw: StringProperty(description="For name of method which will draw socket UI (optionally)")
     prop_name: StringProperty(default='', description="For displaying node property in socket UI")
+    sv_is_linked: BoolProperty(description="Node is linked if (node.is_linked or node.sv_is_linked). "
```
zeffii: do you mean "socket" here?     `description="Socket is linked......"   `  ?

i do not know how I feel about this. I am not able to test extensively now.

Durman: Yes socket. It looks like solution is better than using `if socket.links`. 

I will check one more opportunity later.

zeffii: no.. i mean is the description of `sv_is_linked`  correct?    `"Node is linked if (node.is_linked....."  there is no  `node.is_linked`  ?

Durman: Yes, description is incorrect. Socket is linked if (`socket.is_linked` or socket.sv_is_linked`). I'll fix this later. 

If you don't like the idea with the `sv_is_linked` method I can offer another one more solution. 
The problem is with `dictionary in` node. When socket is being connected it is replaced by socket with type of `other` socket and links is recreated. But in this case `is_linked` attribute of the socket remains False until the end of update event and in process `sv_get` method returnes `NoDataError`. 

It is possible to create one more type of sockets (chameleon) which will include method `set_color` which will dye itself into color of other socket. In this case socket remains unchanged and its `is_linked` attribute returns correct status.

Durman:
@zeffii Have a look to implementation of new socket type (Chameleon). With this type I can stay out of changing logic of testing whether a socket is linked of other sockets.

portnov:
I'm afraid I do not see, why must we change the color of input socket? Currently we do not do this for other sockets (for example, Matrix socket works fine if you plug Vector into it...). And it seems to work...

Durman:
There is no other purpose than make the node more readable. If the node would not have such sockets than for knowing type of input data each link connected to output socket should be tracked. I don't think this will be very comfortable. Keep in mind that this node is only one which can be plugged to any type of sockets in Sverchok.

portnov:
Well, at least FaceData sockets can handle any type of data.

Durman:
It will be list of values in most cases I guess. Yes there is nodes with sockets that technically can handle any type of data but they really does not expect all types. `Dictionary in` node on the contrary expects this. I think this approach would be useful also for such nodes as `list join` or `switch` but `dictionary in` node need this most badly.

DolphinDream:
vote for #1

Durman:
If we are going to create new pre release branch I can merge it to there. Otherwise if there is no objection I can merge this PR.

zeffii:
chameleon socket is probably OK, maybe include an "sv_idname", to indicate what type of socket it's trying to imitate ?

Durman:
Yes, I had the similar thought. I will consider implementation of this property.

nortikin:
1 ok

PR #2776, opened 2019-12-30
Component Analyzer

vicdoval:
Node with miscellaneous functions 
Component Matrix
image

Component Sharpness
image

Number of adjacent edges, face angle, number of neighbor faces...
image

image
image
image

Edge tools
image

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


nodes/analyzer/component_data.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def process(self):
+        if not any(output.is_linked for output in self.outputs):
+            return
+        modes_dict = modes_dicts[self.mode]
+        component_mode = self.actual_mode()
+        output_ops = modes_dict[component_mode][2]
+        local_ops = modes_dict[component_mode][3]
+        func_inputs = modes_dict[component_mode][4]
+        func = modes_dict[component_mode][5]
```
Durman: Just alternative way:
```Python
_, _, output_ops, local_ops, func_inputs, func, *_ = modes_dict[component_mode]
```

portnov: Maybe use a normal class instead of tuple?... Tuples with more than 3-4 parameters tend to be too fragile - it's too easy to confuse the order of parameters.
Or named tuples, as an option...

Durman: Looks like `namedtuple` is also deprecated: https://docs.python.org/3.7/library/collections.html#collections

> Deprecated since version 3.3, will be removed in version 3.9: Moved Collections Abstract Base Classes to the collections.abc module. For backwards compatibility, they continue to be visible in this module through Python 3.8.

Durman: Probably new way is using `dataclasses` module.
https://docs.python.org/3.7/library/dataclasses.html

portnov: @Durman I tend to understand that phrase
> Deprecated since version 3.3, will be removed in version 3.9: Moved Collections Abstract Base Classes to the collections.abc module.

as "Collections ABC are deprecated in collections module, and will be moved to collections.abc; but other classes and methods from collections module are okay".

vicdoval: I would need to create the classes from the original table-looking dicts so it is another step for a dict that is only called 1 time per update...

utils/modules/edge_utils.py
```diff
@@ -81,3 +130,135 @@ def faces_angle(normals, edges, pols):
...
+    vals = [vertices[e[0]] for e in edges]
+    return vals
+
+def edges_end(vertices, edges):
+    vals = [vertices[e[1]] for e in edges]
+    return vals
+
+def edges_inverted(vertices, edges):
+    vals = [[e[1], e[0]] for e in edges]
+    return vals
```
Durman:
```Python
def edges_inverted(vertices, edges):
    return [[vertices[i1], vertices[i2]] for i1, i2 in edges]

File: utils/modules/edge_utils.py [hunk 3]
```diff
@@ -81,3 +130,135 @@ def faces_angle(normals, edges, pols):
...
+
+    return result
+
+
+def edges_vertices(vertices, edges):
+    '''Explode edges'''
+    verts = [[vertices[c] for c in e] for e in edges]
+    eds = [[[0, 1]] for e in edges]
+    vals = [verts, eds]
+    return vals
```
Durman: Redundant naming
```Pyton
return verts, edges
```

utils/modules/edge_utils.py
```diff
@@ -81,3 +130,135 @@ def faces_angle(normals, edges, pols):
...
+    return center
+
+def edges_matrix(vertices, edges, orientation):
+    '''Matrix aligned with edge.'''
+    origin, track, up = orientation
+    normal = edges_direction(vertices, edges, out_numpy=False)
+    normal_v = [Vector(n) for n in normal]
+    center = edge_vertex(vertices, edges, origin)
+    vals = matrix_normal([center, normal_v], track, up)
+    return vals
```
Durman: I think if function are in utils it means that it can be used in another nodes by other developers. I would appreciate if functions have more explicit documentation. What are expecting in input and what will be in output like this:
```Python
"""
vertices: list of tuple(flaot, float, float)
edges: list of tuple(int, int)
orientation: tuple(origin, track, up) where origin is list of tuple(float, float, float) and so on
return: list of Matrixes
"""
```

vicdoval: Ok, I will expand my code documentation skills

utils/modules/edge_utils.py
vicdoval: Actually should be:
```
def edges_inverted(edges):
    '''
    edges: list with edges [[int, int], [int,int]...]
    outputs the same list with inverted index [[0, 1],[1, 2]] ---> [[1, 0], [2, 1]]
    '''
    return [[e[1], e[0]] for e in edges]
```
    `

Durman: Then socket type should be fixed.

utils/modules/edge_utils.py
```diff
@@ -81,3 +130,135 @@ def faces_angle(normals, edges, pols):
...
+    '''Average of normals of the faces that share the edge (only the two first)'''
+    bm = bmesh_from_pydata(vertices, edges, faces, normal_update=True)
+    normals = [tuple(face.normal) for face in bm.faces]
+    bm.free()
+    vals = edges_normal(vertices, normals, edges, faces)
+
+    return vals
+
+def edges_normal(vertices, normals, edges, pols):
+    '''Average of normals of the faces that share the edge (only the two first)'''
```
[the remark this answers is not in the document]
vicdoval: > It won't return a normal if faces are not flat.

It relies on bmesh capabilities to determinate the normals, I'm not sure how it deals with non-planar faces, but i just trusted in bmesh and didn't dig more on it...

>And what will be if an edge does not have adjacent faces?
>
>I think correct way will be get average normal from normals of adjacent vertices.

If there are no faces there are not valid vertex normals (for bmesh) .
I also thought (and coded) first the average or the vertex normals but it gave vertors that were not perpendicular to de edges (for example in the corner edges of a subdivided box) and as I see it the normals should be always perpendicular...

utils/modules/edge_utils.py
```diff
@@ -62,8 +90,29 @@ def adjacent_faces(edges, pols):
...
+                idx = e_sorted.index(e_s)
+                ad_faces[idx] += [pol]
+    return ad_faces
+def faces_angle_full(vertices, edges, faces):
+    '''angle between faces of each edge (only first two faces)'''
+    bm = bmesh_from_pydata(vertices, edges, faces, normal_update=True)
+    normals = [tuple(face.normal) for face in bm.faces]
+    bm.free()
+    vals = faces_angle(normals, edges, faces)
+    return vals
```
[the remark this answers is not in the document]
vicdoval: There could be some cases in which it would be nice to have the `function(bm)` but not for all of them, in this case the function is also using original edges (that get messed in bmesh) and faces so it would need to be `function(bm, edges, faces)`....
The efficiency of the function depends on the use, for the Component Analyzer that just calls one of this function per node it was better to perform the conversion inside the function if it was needed.
I think that if any developer finds more convenient function(bm) they can make a new function taking the example of the previous one

nodes/analyzer/component_analyzer.py
```diff
@@ -0,0 +1,359 @@
...
+        if not 'u' in out_ops:
+            layout.prop(self, 'split')
+        else:
+            layout.prop(self, 'wrap')
+
+    def sv_init(self, context):
+        inew = self.inputs.new
+        inew('SvVerticesSocket', "Vertices")
+        inew('SvStringsSocket', "Edges")
+        inew('SvStringsSocket', "Faces")
```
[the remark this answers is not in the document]
vicdoval: I'll keep it in mind

utils/modules/edge_utils.py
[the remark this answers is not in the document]
Durman: The less Sverchok has functions the better. I would probably like to use some of this functions in other nodes but your approach does not let it. I does not see reasons why all this functions are in `utils` folder if they can be efficiently used only by `Component Analyzer` node. Or probably names of folders should be less ambitious, the name can be like `Component Analyzer node utils` so everyone would know that the fuctions are mostly only for the node usage.

utils/modules/edge_utils.py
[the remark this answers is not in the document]
Durman: >It relies on bmesh capabilities to determinate the normals, I'm not sure how it deals with non-planar faces, but i just trusted in bmesh and didn't dig more on it...

What do you mean? As far as I concern the `bmesh` module does not gives any normals for edges. You are create them according your own logic.

no normals
Your solution also does not gives perpendicular normals as I already mentioned.

utils/modules/edge_utils.py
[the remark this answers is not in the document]
vicdoval: Right now I think is better to have more smaller functions and a bigger one with many conditional cases inside, but i guess there is not a truth for all the cases... this node is a prof of concept and could be expanded or reduced i think that is not a bad approach to a muti-function node work and open the mind of the user
The functions can be used efficiently in many other cases, obviously not all of them, they could be called from a script node (without knowing anything about converting to bmesh) or get inside other algorithms. I started to fill the vertex_utils, edges_utils, pols_utils and matrix_utils to light the node code and in case someone can use them too, if you find it disturbing i can change the filenames and folder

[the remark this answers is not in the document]
vicdoval: Right now I think is better to have more smaller functions and a bigger one with many conditional cases inside, but i guess there is not a truth for all the cases... this node is a prof of concept and could be expanded or reduced i think that is not a bad approach to a muti-function node work and open the mind of the user
The functions can be used efficiently in many other cases, obviously not all of them, they could be called from a script node (without knowing anything about converting to bmesh) or get inside other algorithms. I started to fill the vertex_utils, edges_utils, pols_utils and matrix_utils to light the node code and in case someone can use them too, if you find it disturbing i can change the filenames and folder

utils/modules/edge_utils.py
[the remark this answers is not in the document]
vicdoval: As you point the algorithm was created using my own logic, neither bmesh or blender deals with this concept if any software does (probably because the normal to a line should be a plane) . And it fails without polygons and with no-planar faces but works t to create this 
image
and from it to create this:
image

If you can improve the algorithm in any way covering the missing cases or improving its logic you are very welcome to do it. I was thinking in it for a while and it this is  how far I could arrive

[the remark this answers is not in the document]
Durman: It looks working correctly:
```Python
y = (edge.verts[1].co - edge.verts[0].co).normalized()
_normal = (edge.verts[0].normal + edge.verts[1].normal).normalized()
x = y.cross(_normal)
normal = x.cross(y)
```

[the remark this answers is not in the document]
vicdoval: Except for the fact that bm.edges does not respect the order of the input your function works better and is simpler, I will implement it. 
Thanks!

PR #2782, opened 2020-01-03
Adaptive polygons: recipient face index

portnov:
This adds the following:
* "FaceRecptIdx" output contains, for each output face, the index of recipient object face, which was used to make this face.
* "VertRecptIdx" output contains, for each output vertex, the index of recipient object vertex, which was used to make this vertex.
* "Remove doubles" checkbox and "Threshold" parameter/input, to integrate "remove doubles" / "merge by distance" functionality into this node, as it is required very often.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


portnov:
Screenshot_20200103_210345

portnov:
Use materials from the donor object, but vertex colors from the recipient object:

Screenshot_20200103_211558

PR #2792, opened 2020-01-05
Mask out upgrade

vicdoval:
The Mask Out node had a .pop function that was super slow.
I rewrote it in another way that boost the node from 2 to 20 times
with 100000 verts 
before 0.923sec per call
after  0.098 sec per call
🚀 

- [X] Ready for merge.

### DIFF
--- a/nodes/list_masks/mask.py
+++ b/nodes/list_masks/mask.py
@@ -19,7 +19,7 @@
 from copy import copy
 from itertools import cycle
 import bpy
-from bpy.props import BoolProperty, IntProperty, StringProperty
+from bpy.props import IntProperty
 from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode, changable_sockets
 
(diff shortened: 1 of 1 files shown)


nodes/list_masks/mask.py
```diff
@@ -78,60 +78,61 @@ def process(self):
...
-                    ind_false.append(l2[4])
+                for idx, sub_list in enumerate(list_a):
+                    sub_res = self.put_current_level_list(sub_list, list_b, mask_l, level - 1, idx)
+                    result_t.append(sub_res[0])
+                    result_f.append(sub_res[1])
+                    mask_out.append(sub_res[2])
+                    ind_true.append(sub_res[3])
+                    ind_false.append(sub_res[4])
             else:
                 print('AHTUNG!!!')
```
portnov: wtf? :)

vicdoval: It was already there 😄  You think should delete it?

portnov: If you think this message is somehow usable (gives some information at least to developer), please change print to self.debug(). Otherwise, just delete it.

Durman: Probably more readable would be:
```Python
res_t, res_f, res_m, res_it, res_if = self.put_current_level_list(sub_list, list_b, mask_l, level - 1, idx)
result_t.append(res_t)
result_f.append(res_f)
mask_out.append(res_m)
ind_true.append(res_it)
ind_false.append(res_if)
```

vicdoval: Actually it was never called...  :D I refactor the node again, at this point of the code it was not working properly due a mix of the before and after procedures... Now it works as it should.
I also followed your tip @Durman but finally I wrote it in a compressed two lines  that seem readable :D

PR #2797, opened 2020-01-06
Set custom UV map

Durman:

Continue of experiments from here: #2218

In this time I would like to create very simple node which just takes X Y coordinates of given mesh and creates UV map from them. It will be already useful for 2d meshes but for 3d mesh unwrapping nodes are required (not in this PR).

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/object_nodes/object_nodes_index.rst
+++ b/docs/nodes/object_nodes/object_nodes_index.rst
@@ -14,4 +14,4 @@ Objects
    getsetprop
    assign_materials
    material_index
-
+   set_custom_uv_map
--- a/docs/nodes/object_nodes/set_custom_uv_map.rst
+++ b/docs/nodes/object_nodes/set_custom_uv_map.rst
@@ -1,0 +1,61 @@
+Set custom UV map
+=================
+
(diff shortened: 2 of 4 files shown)


Durman:
Example of result which I'm expecting from working of this node and `Inset faces` node.

Durman:
This works as flat projection:

portnov:
Another possible option will be to start from any face and walk from it in all directions (either depth-first or breadth-first), unwrapping all faces without deformation. Mappings of different faces will obviously overlap, but for some kinds of textures this is okay.

Durman:
Yes, but this should be another node like here:

Durman:
I'm looking for a way of update an object after resetting uv coordinates. It looks like some update event should be caused for this. It works only if change frame number now.

It looks like does not work:
https://blender.stackexchange.com/questions/140789/what-is-the-replacement-for-scene-update

Durman:
https://blender.stackexchange.com/questions/163300/how-to-update-an-object-after-changing-by-script-of-uv-coordinates

Durman:
I have been got the answer =D
`obj.data.update()`

enzyme69:
Cricket with diamond UV https://skfb.ly/6SQxB

PR #2800, opened 2020-01-07
follow active quad

Durman:
## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/modifier_change/follow_active_quads.rst
+++ b/docs/nodes/modifier_change/follow_active_quads.rst
@@ -1,0 +1,75 @@
+Follow active quads
+===================
+
+.. image:: https://user-images.githubusercontent.com/28003269/72198433-af340b00-3446-11ea-95f3-caa61581220a.png
+
+Functionality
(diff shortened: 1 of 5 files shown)


portnov:
I think there should be some input to indicate which quad is active...

Durman:
Yes, I slightly redone interface of the node.)

Durman:
Possible unwrapping of a box, just with Sverchok nodes.

Durman:
Unwrapping relatively to shape of map of active quad:

unwrap active quad

Durman:
Supporting multiple active quads. Useful for unwrapping a mesh with disjointed parts.

Durman:
I have added protection from re unwrapping faces. So it is possible to unwrap all disjoint parts of a mesh even without knowing exactly which indexes of faces related to each disjoint part.

Probably it would be useful to put button unwrap all (or more explicit name?) somewhere.

PR #2809, opened 2020-01-10
B28 oldgroup removal

zeffii:
this code has not been executed in several years, it can be removed.


zeffii:
the Monad experiment rapidly grew into a sufficiently usable feature, but it remains awkward code to reason about because there's so much of it and to be fair it is abstract/dynamic.  It deserves a rewrite building on the lessons learnt.

Issue #2807, opened 2020-01-09
Nodes that use/set Collections


zeffii:
One of the few remaining easy things to fix should be restoring Group support using Collections. I suspect all that is really needed for this to be useful is to always make a new collection for a node, and generate all objects inside that collection.

```python
# setup collection for node
basename = "Alpha" #  or random_available_name()
node_related_collection = bpy.data.collections.new(basename)

parent = ( current scene collection?  )
parent.children.link(node_related_collection)
```
There are two main areas of interest here

1. Object Viewers
    These use SvObjHelper directly and are straightforward upgrades
    
    - [x] BMesh https://github.com/nortikin/sverchok/pull/2810  | SvObjHelper
    - [x] PolyLine https://github.com/nortikin/sverchok/pull/2810  | SvObjHelper
    - [x] Curve https://github.com/nortikin/sverchok/pull/2810 | SvObjHelper
    - [x] Typography https://github.com/nortikin/sverchok/pull/2810 | SvObjHelper
    - [ ] SkinMesher  <-needed?  | SvObjHelper  , low priority
    - [ ] Metaball  <-needed? | SvObjHelper  , low priority

2. "Feature" Viewers
    These can be converted to use SvObjHelper, or implement their own

    - [ ] Lamp 
    - [ ] Empty
    - [x] Instancer https://github.com/nortikin/sverchok/pull/2818  grouping is now standard

Durman:
Probably the node could have property on the N panel with name of collection where objects should be generated with default value `SV Objects` or whatever.
In this case several nodes can generate objects to one collection.

portnov:
As far as I understand the collections are hierarchical? Then we can do always a collection under path "/Sverchok/NodeTreeName/NodeName" for each node.

Durman:
I think a user should have control over creating of collections.

zeffii:
- the collection can use:
   ```python
   collection_name = self.custom_collection_name or self.base_name 
   ```

i also like the nesting,
```
- Collection Sverchok
   |
   +-- Collection NodeTreeName
          | 
          +--  Collection (NodeName or collection_name)
```

zeffii:
i've never used a separate `scene` , that would be an additional nesting.

zeffii:
i don't feel like bumping the Object Nodes for this to a new `MK`,   but it might not be possible to introduce this without undesired effects in existing .blends that use those nodes.

i'll try to get this in before sunday

zeffii:
[...]

PR #2810, opened 2020-01-10
B28 svobj collections

zeffii:
It's been my intention to do this since forever.  `Blender 2.8+` brought collections and removes "groups" support, this PR implements sverchok collections as a groups analog.

following a discussion at  https://github.com/nortikin/sverchok/issues/2807 .


zeffii:
this applies Collection support to the main users. There are 4 more nodes that could be fixed too (but not in this PR)

PR #2816, opened 2020-01-12
adjust collection code.. works now, but will require attention

zeffii:
perform extra administrative work in the viewer object nodes to locate/relocate generated objects into the desired hierarchy (Collection)


zeffii:
this is not the most ideal solution, but it is the solution that takes the least amount of code for now.

PR #2821, opened 2020-01-13
Random points on mesh: ability to disable "proportional" mode.

portnov:
Default behaviour is unchanged ("proportional" mode is enabled by default).
If it is disabled, then the number of points at each face is conrtolled only by the "weight" input.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


enzyme69:
By the way this node is not working right now.

PR #2825, opened 2020-01-14
add external function

zeffii:
since the start the Object Viewers have not supported renaming the `basedata_name` or the resulting group. Now with `collections` being implemented i'm looking into a low impact (ie..doesn't require massive recode) option to allow users to rename collection/basedata_name . collection will still be linked to basedata_name.  (like Henri ford...all colours you want, as long as it's black )


zeffii:
this is not solved.

PR #2831, opened 2020-01-15
get / set property node additions.

zeffii:
getset prop node needs surgery

### DIFF
--- a/nodes/object_nodes/getsetprop.py
+++ b/nodes/object_nodes/getsetprop.py
@@ -161,13 +161,15 @@ class SvGetPropNode(bpy.types.Node, SverchCustomTreeNode):
             self.bad_prop = True
             return
         self.bad_prop = False
-        s_type = types.get(type(obj))
-        outputs = self.outputs
-        if s_type and outputs:
-            outputs[0].replace_socket(s_type)
-        elif s_type:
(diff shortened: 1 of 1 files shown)


Durman:
Portnov insists that all debug message was done via logger.)

zeffii:
in my test branch i will do what is most convenient for _me_.

Durman:
I meant only messages that are intended to be in master.

zeffii:
i will use the logger, maybe even before hitting master :)  This is one of those nodes where a one-off verbose message to the user is supremely useful

zeffii:
fixing every possible conceivably usage of this node is not going to happen today, but here as some thoughts for future me

# Set Mode
- you can use a wrapper (like local_updateNode) to call the node's own process function.
    - why? because this node has no output socket, and is therefore a deadend for "updateNode"
    - example, float_prop and int_prop do nothing when updated on the node, only when they get external socket input.
- what if the property is `type_prop_array` and takes only 3 components, unlike 4 that the default FloatVectorProperty sets... offer a second FloatVectorProperty.. name `Col3` , `Col4` . This can get complicated quick if the color property is expecting some HSV etc.. i don't know.

# Get Mode 
- this probably deserved direct converting to our "SvColorSocket"

PR #2832, opened 2020-01-15
Noisize Node

vicdoval:
image

Seed and scale per vertex can be passed, in this example the seed is chosen by determining the closest point of another mesh and the scale is based on the distance to that point
image

One matrix per point can be passed if the matrix list is wrapped, note that the "Flat Output" checkbox of the matrix in is un-checked
image

The node offers three ways of matching the list lengths "Repeat Last", "Cycle" and "Match Short" in this example "Cycle" is used to alternate the noise matrix
image

In this example the scale output is used to blend with another oscillation texture
image

The "Vector" Mode does not use vertex normals so it can be used just with vertices
image

The "Scale out" input can be used to mask the affected vertices
image

- [x] Code changes complete.
- [x] Manual testing done. 
- [] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/nodes/transforms/noisize.py
+++ b/nodes/transforms/noisize.py
@@ -99,8 +99,8 @@ noise_func = {'NORMAL': v_normal, 'VECTOR': v_noise}
 
 class SvNoisizeNode(bpy.types.Node, SverchCustomTreeNode):
     """
-    Triggers: Add Noise to verts
-    Tooltip: Affect input verts/mesh with a noise values.
+    Triggers: Noise Displacement
+    Tooltip: Move input verts/mesh with a noise values.
 
     """
 
--- a/nodes/transforms/noisize.py
+++ b/nodes/transforms/noisize.py
@@ -1,0 +1,192 @@
+# ##### BEGIN GPL LICENSE BLOCK #####
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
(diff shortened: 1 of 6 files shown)


portnov:
Is this something like "Vector noise" + Move?

vicdoval:
@portnov  talking short: yes
But it also pre-multiplies the vector by the inverted matrix (To use a blender empty to control the noise), allows seed per vertex and multiplies the output by a vector and gives the chance to apply it to the normal. Also with some candy list.machting :)

zeffii:
"Apply Noise"  , good - i had a branch with something similar which can now be ditched :), but i did have a "rescale" option, to scale in the input , then get noise, then add noise to original input

thanks @vicdoval  :)

vicdoval:
@zeffii i was following you attempt at that point :)
I started also with the pre-scale concept, then I added offset, then added rotatation... And finally i found it was more intuitive to invert the matrix and call it Noise Matrix as the Object UVMapping in the Displace Modifier.:)  (Btw i have started a Displace node to work with blender textures)

nodes/transforms/noisize.py
```diff
@@ -0,0 +1,192 @@
...
+
+class SvNoisizeNode(bpy.types.Node, SverchCustomTreeNode):
+    """
+    Triggers: Noise Displacement
+    Tooltip: Move input verts/mesh with a noise values.
+
+    """
+
+    bl_idname = 'SvNoisizeNode'
+    bl_label = 'Noisize'
```
zeffii: this name / label  ,  could be better... "Noise Displacer" even.  Not a big issue, maybe Noisize feels a little gimmicky ?

vicdoval: It was recalling Randomize but it is true Noisize  is a little weird...

zeffii: "Perturb", "Add Noise"

vicdoval: Noise Displacer sounds fine to me, but would you change also the class name and the files names?

vicdoval: Maybe just "Noise Displace" is even cleaner

zeffii: yeah, but do it early.

catzova:
where can I find plane mk2?

Durman:
It's called Plane now.

Issue #2243, opened 2018-07-10
Translate two vectors to quaternion. New node?


Durman:
What about such node? It is possible to create quaternion from two vectors which will present rotation from one vector to another.

![2018-07-10_08-29-55](https://user-images.githubusercontent.com/28003269/42489348-77bfd776-841b-11e8-91fb-79d6a3b9dc9f.png)

<details> 
  <summary>gif</summary>  

![quat rotation](https://user-images.githubusercontent.com/28003269/42489398-c3dd9616-841b-11e8-806e-ad4f772cbb2e.gif)

</details>

--
[my to do list](https://trello.com/b/vvC6ieUZ)

portnov:
Please check if we have something similar in "vector math" node (i'm not sure right now). Maybe it's good idea to extend that mode, or create another node, which gives full solution to this problem.
I know 3 ways of doing something similar:
* Use of Householder matrix
* Use "tracking" logic from blender's TrackTo
* Use "rotation_difference" from blender's mathutils

Rotation itself can be output as quaternion, or as a matrix, or maybe in some another way.

These options give very similar results in most cases, but different results in some corner cases. So user might want to use another algorithm in some cases.
Refer to objects_along_edges.py for example of implementation, and https://github.com/nortikin/sverchok/blob/master/docs/nodes/modifier_change/objects_along_edge.rst .

Durman:
I used rotation_difference indeed.

<details> 
  <summary>code</summary>  

```Python
'''
in v_a v d=[] n=2
in v_b v d=[] n=2
out x s
out y s
out z s
out w s
'''

from mathutils import Vector, Quaternion

quat = Vector(v_a).rotation_difference(Vector(v_b))
w, x, y, z = quat
x, y, z, w = [[x]], [[y]], [[z]], [[w]] 
```

</details>

The same result can be done and with matrix.
![2018-07-12_08-22-29](https://user-images.githubusercontent.com/28003269/42612503-c7866418-85ac-11e8-9e18-5e4fb65b21eb.png)

But it is a little bit tricky and can be less obvious how to do, for common users.
Sverchok have not representation of quaternions at all. May be there is no needs.
But there would be more opportunities if user can rotate objects from one vector to another just plugging two vectors to a node.

zeffii:
> Sverchok have not representation of quaternions at all.

we do, but no nodes / people use it.

[...]

PR #2833, opened 2020-01-15
Rotation difference

Durman:

#2243 Looks like the time has came for this node. It will add some sence to existing quaternion nodes.))

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.


Durman:
First vector of the node is initial direction of monkey head, second one is direction of empty object.

quaternion

Durman:
Also it is possible to make movements of monkey head more natural by applying Z component separately. I sink this is should be more flexible approach than `matrix normal` node or `matrix track to` node.

quaternions

PR #2835, opened 2020-01-15
Cut object by surface

portnov:
This node cuts **edges** of one mesh ("object") by intersecting them with faces
of another mesh ("surface"). It can connect the new vertices by creating new
faces, to produce a cut surface. It can also output the pieces of the object,
that are the result of cutting. This is similar to "Cross Section" and "Bisect"
nodes, but it operates with any surface, not just with plane. In a sense, this
operation can be considered as a very simple analog of Boolean operations, but
it finds intersections of **edges**, not intersections of volumes.

**Implementation restriction**: the node requires that of each face of "object"
mesh, only two edges intersect the "surface" mesh. So, each face can be cut
only in two pieces, not in three or more.

Faces made by this node at places of cut will be usually NGons, so they may be
non-planar. It may be good idea to pass the output of this node through the
**Split faces** node (in **split non-planar faces** mode), or maybe through
**Make faces planar** node.

Note that this node outputs "cut pieces" mesh always as one merged object (not
split into pieces). If you need these pieces in separate, you can pass the
output of this node through the **Separate Loose Parts** node.

An example:

Screenshot_20200114_235009

Here the intersection of eight cubes (green edges) with the icosphere (yellowish edges) is calculated. Red mesh is the "cut surface".

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/modifier_make/cut_object_by_surface.rst
+++ b/docs/nodes/modifier_make/cut_object_by_surface.rst
@@ -76,3 +76,34 @@ This node has the following outputs:
   where the "object" is cut by "surface" (i.e. there will be holes). This
   output is only available if the **Make cut pieces** parameter is checked.
 
+Examples of usage
+-----------------
+
+Sphere cut by the box. Here we intersect edges of the sphere with the box, and
+connect resulting vertices. Note that in this case the number of box's
+subdivisions does not have any meaning, because the places of intersections of
+sphere's edges with the box are always the same.
+
+.. image:: https://user-images.githubusercontent.com/284644/72448547-033d4780-37d9-11ea-9016-79fb80523bcb.png
+
+Box cut by sphere. Here we intersect edges of the box with the sphere, and then
+connect resulting vertices. Note that if we specify number of box's
+subdivisions = 1, we would get nothing at all.
+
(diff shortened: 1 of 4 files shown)


portnov:
The same cut surface passed through the "split faces" node:

Screenshot_20200115_204019

So you can see this node is not at all the boolean operation — it makes one big face where the boolean operation would make a part of sphere.

portnov:
Sphere cut by the box:

Screenshot_20200115_204734

portnov:
Box cut by sphere. Here we intersect edges of the box with the sphere, and then connect resulting vertices. Note that if we specify number of box's subdivisions = 1, we would get nothing at all.

Screenshot_20200115_205017

portnov:
Screenshot_20200115_212550

zeffii:
i don't really understand what i'm looking at, but it's cool : )

portnov:
Similar to the previous picture, but I moved some parts of the resulting object away:

Screenshot_20200115_215948


Cut a number of smaller spheres by surface of the big sphere (red areas are the intersections):

Screenshot_20200115_215330

PR #2836, opened 2020-01-15
Adaptation for extra nodes addons

portnov:
This adds basic support for extension addons, that would be installed together with Sverchok and add new nodes and node categories.

This adds API methods for such addons to register their nodes and node categories in Sverchok's menus and panels.

Currently I use these functions in my "Sverchok-Extra" addon (discussed in #2837). Later this (or similar) API can be used by other addons that extend Sverchok's functionality.

I do not consider this implementation as final, it may be improved and generalized later.

If you do not have any "extra" addons installed, you should not notice any difference in Sverchok's work.

## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/menu.py
+++ b/menu.py
@@ -38,6 +38,9 @@ class SverchNodeCategory(NodeCategory):
     def poll(cls, context):
         return context.space_data.tree_type == 'SverchCustomTreeType'
 
+    def __repr__(self):
+        return f"<SverchNodeCategory {self.identifier}>"
+
 def make_node_cats():
     '''
     this loads the index.md file and converts it to an OrderedDict of node categories.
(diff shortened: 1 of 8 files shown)


zeffii:
i think `menu.py` contains a few things that should be their own `sverchok/utils/..`

zeffii:
is it possible to do this without touching `menu.py` so much? like injecting into `_nodeitems` like
```python
    NC = nodeitems_utils._node_categories
    SV = NC['SVERCHOK']
    SVcatlist, sv_draw, SvCats = SV
    NC['SVERCHOK'] = (augmented_SVcatlist, sv_draw, augmented_SvCats)
    # NC['SVERCHOK'] = (augmented_SVcatlist, bpy.types.NODEVIEW_MT_Dynamic_Menu.draw, SvCats)

```
and have the user preference to autoinjecting if available addon is found 
1. on_load
2. at start_up
3. on demand

it seems like you have most of the code already in pllace, just i have a kneejerk reaction to the location of the new code.

portnov:
One cannot just insert something into lists defined in menu.py, because the order of addons registration is undefined. By the time sverchok-extra is registered, if sverchok is already registered, menus would be already constructed... So I had to do everything dynamic: there is a "registry" of node categories providers, and each time you open the menu it looks through all providers to find all categories. Those providers can be added "on-the-fly".
I moved some of the code away from menu.py though :)

PR #2840, opened 2020-01-16
implement mixin for setget node

zeffii:
makes this possibly easier to read


zeffii:
i'll push this as a beta nodeset for now

PR #2842, opened 2020-01-17
Normal vector node

Durman:

I'm searching a way for setting direction by `direction` widget.
vector normal

It is not obvious how this should be implemented. By default the space has next view.

I gives opportunity to change it. Instead Z axis custom axis can be set in this case Z axis will replace Y axis.
Any suggestions are welcome.

## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Ready for merge.

### DIFF
--- a/nodes/vector/vector_normal_in.py
+++ b/nodes/vector/vector_normal_in.py
@@ -5,7 +5,11 @@
 # SPDX-License-Identifier: GPL3
 # License-Filename: LICENSE
 
+
 from itertools import chain, cycle
+
+from collections import namedtuple
+from math import pi
 
 import bpy
 from mathutils import Quaternion, Vector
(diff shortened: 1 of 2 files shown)


nodes/vector/vector_normal_in.py
```diff
@@ -0,0 +1,104 @@
...
+
+Modes = namedtuple('Modes', ['x', 'y', 'z', 'custom'])
+MODE = Modes('X', 'Y', 'Z', 'Custom')
+MODE_DIR = Modes(Vector((1, 0, 0)), Vector((0, 1, 0)), Vector((0, 0, 1)), None)
+
+QUAT_X90 = Quaternion((1, 0, 0), pi / 2)
+QUAT_Y270 = Quaternion((0, 1, 0), -pi / 2)
+
+
+def get_costume_norm(normal, space_normal, mode=MODE.z):
```
zeffii: "custom"  not "costume"

zeffii:
the 3dview rotation widget would be really cool for this, is that what you mean?

Durman:
Yes 3d view widget would be even more flexible but direction vector is also cool.)

zeffii:
i also don't know @Durman  : [

Durman:
Not sure about this node yet.

PR #2845, opened 2020-01-19
Texture Displace (Displace Modifier)

vicdoval:
image
and his little companion Texture Evaluate
image

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/transforms/texture_displace.rst
+++ b/docs/nodes/transforms/texture_displace.rst
@@ -1,0 +1,91 @@
+Texture Displace
+================
+
+This node displaces a list of Vectors using a texture.
+
+Inputs & Parameters
+-------------------
+
++----------------+-------------------------------------------------------------------------+
+| Parameters     | Description                                                             |
++================+=========================================================================+
+| Direction      | Displacement direction.                                                 |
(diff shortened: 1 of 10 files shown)


portnov:
Uh. In skimage, there is a function to find contour (iso-lines) in images (they can be extracted as blender curve meshes and filled if you wish). Do you think such feature would be useful for textures? If yes, I'll put it on my todo for sverchok-extra :)

vicdoval:
@portnov I think that function can be very useful in many cases, vectorizing an image for example or to create crazy topologies 😄

portnov:
Could you place "color_channels" enum into some module under `sverchok.utils`? Maybe some other functions too, but I'm not so sure about them...

portnov:
@vicdoval Probably I'm mistaken... but as far as I understood, "Texture" data blocks are sort-of deprecated in 2.8x? How are you creating them for this node to use?

vicdoval:
@portnov the texture nodes don't work but brush textures and "displace modifier" textures  seem to work fine
texture_evaluate_sverchok_blender_example_1

Issue #2849, opened 2020-01-20
New normal and center nodes


Durman:
## Problem statement

I'm thinking about combining this nodes into two and spread their logik on all mesh components (vertices, edges, faces).
![2020-01-20_18-57-12](https://user-images.githubusercontent.com/28003269/72736195-b7aee300-3bb6-11ea-8ac5-89089ceba027.png)

```Text
|  Mesh normals      |
+--------------------+
|             normal +
|            tangent +
| modes:             |
| vert | edge | face |
+ verts              |
+ edges              |
+ faces              |
```

```Text
|  Mesh centers      |
+--------------------+
|             center +
|             matrix +
| modes:             |
| vert | edge | face |
+ verts              |
+ edges              |
+ faces              |
```

I would like to try implement this close to blender default behaviour in edit mode.

vicdoval:
How are you planning to create the verts tangent?
Why matrix is associated with center?
I appreciate the propposal, but wouldn't you prefer, following the idea of smaller nodes (smaller than Component Analyzer at least) to do also "Mesh Tangents" and "Mesh Matrixes"?

Durman:
I decided don't touch `normal` nodes at all. Only `center faces` node I'm going to replace by new `origins node`. `Center faces` node already has nearly similar output but the name is not suitable to it. Origins I think will be more correct.
![2020-01-23_16-07-45](https://user-images.githubusercontent.com/28003269/72983385-96d6cf80-3dfa-11ea-900d-fe1448dff5f7.png)
For setting space of a mesh component, center, normal and tangent are all required. I don't think that this output should be splitted because `center face` node already proved usability of such set of output data.

Vertex tangent will be created according Blender logic. I will create PR soon.

Issue #2680, opened 2019-11-17
where can if find the node Center Polygon 2 with blender 2.8


BennyGovaerts:
With version SV 0.6.0.0 (d19fe88) there is just one node with centers polygons how can i do the same  like with the older versions  off sverchok  ? 
![6d65a916451dd29316fa5aa272cc3d60](https://user-images.githubusercontent.com/49908008/69012614-0a321000-0978-11ea-8b4c-f1dfe4bfc226.jpg)

zeffii:
i think it's called " CentersPolsNodeMK3 "  now.

zeffii:
version mk2 was moved to old nodes..

zeffii:
I think the node outputs a different Matrix for centers now..  it does not produce the same transformation as in your screenshot.. 

![image](https://user-images.githubusercontent.com/619340/69013817-9fd39c80-0984-11ea-977b-73e333fc44d8.png)

this reminds me ..we really have to fix the `cache-miss` @portnov (ie.. SvNoDataError )

zeffii:
oh wait.. i think I did this...

```python
source_data = SvGetSocket(self, deepcopy = True if self.needs_data_conversion() else deepcopy)
```
and that's not legal python?..!?!

zeffii:
hmmm. maybe it is legal..but it looks suspect AF.

zeffii:
anyway... offtopic.

BennyGovaerts:
With blender 2.79 if i use center mk2 then i have the same result but not if i use center mk3

Durman:
Mk2 and mk3 nodes use different approach in determining local space coordinates of polygons. It is easy to create own local space coordinates with `matrix normal` node:
![2019-11-18_12-53-47](https://user-images.githubusercontent.com/28003269/69038049-7eb19100-0a02-11ea-9d3f-652afec11877.png)

PR #2850, opened 2020-01-23
Origins node

Durman:

#2849
image

Using of corner case of tangent for verts connected only with one face.

Using matrix output for mesh instancing, looks like work of adaptive polygon node.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.


vicdoval:
Some ideas:
Face tangent modes, bmesh offers different methods maybe would be nice to add them.
Matrix orientation: the Center Polygons aligns the matrix with center and first vertex if you replace the node is that going to change?
Performance:
Wouldn't you like to add some booleans to avoid matrix calculation if matrix output is not plugged?
Name: The origin of a edge is its center or its first point? The same question applies to polygons

vicdoval:
Also the center of the faces can be center bounds, center median and center weighted, wouldn't you like to leave option for the user?

Durman:
> Face tangent modes, bmesh offers different methods maybe would be nice to add them.
Also the center of the faces can be center bounds, center median and center weighted, wouldn't you like to leave option for the user?

We now have `component analyzer` node which can produce any algorithm. This I would like create without any alternatives. It should produce data close to how Blender draws axises in edit (normal) mode.

> Matrix orientation: the Center Polygons aligns the matrix with center and first vertex if you replace the node is that going to change?

Yes. The current approach is not something what is expected from this node. #2680

> Wouldn't you like to add some booleans to avoid matrix calculation if matrix output is not plugged?

I think that no. I don't think that build matrixes essentially slow done the node.

> Name: The origin of a edge is its center or its first point? The same question applies to polygons

It the same how Blender draws axises if component is selected, or at least close to this drawing.

PR #2852, opened 2020-01-24
Post modifiers Objects_In fix

vicdoval:
This fixes the problem of crashing Blender when deleting one random object if you have Objects_in mk3 with post modifiers active using suggestions from #2511

### DIFF
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -131,20 +131,24 @@ def sv_main_handler(scene):
     global sv_depsgraph
     global depsgraph_need
 
-    if not pre_running:
-        pre_running = True
-        if depsgraph_need:
-            sv_depsgraph = bpy.context.evaluated_depsgraph_get()
-        for ng in sverchok_trees():
-            # if P (sv_process is False, we can skip this node tree.
-            if not ng.sv_process:
-                continue
+    # when this handler is called from inside another call to this handler we end early
+    # to avoid stack overflow.
+    if pre_running:
+        return
 
-            if ng.has_changed:
-                print('depsgraph_update_pre called - ng.has_changed -> ')
+    pre_running = True
+    if depsgraph_need:
+        sv_depsgraph = bpy.context.evaluated_depsgraph_get()
+    for ng in sverchok_trees():
+        # if P (sv_process is False, we can skip this node tree.
+        if not ng.sv_process:
+            continue
 
-                ng.process()
-        pre_running = False
+        if ng.has_changed:
+            print('depsgraph_update_pre called - ng.has_changed -> ')
+
+            ng.process()
+    pre_running = False
 
 
 @persistent
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -14,6 +14,22 @@ from sverchok.utils import app_handler_ops
 
 _state = {'frame': None}
 
+pre_running = False
+sv_depsgraph = []
+depsgraph_need = False
+
+def get_sv_depsgraph():
+    global sv_depsgraph
+    global depsgraph_need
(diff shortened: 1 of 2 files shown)


zeffii:
i've reflected on this..  ( untried )
```python
# from A

@persistent
def sv_main_handler(scene):
    """
    Main Sverchok handler for updating node tree upon editor changes
    """
    global pre_running
    global sv_depsgraph
    global depsgraph_need

    if not pre_running:
        pre_running = True
        if depsgraph_need:
            sv_depsgraph = bpy.context.evaluated_depsgraph_get()

        for ng in sverchok_trees():

            if not ng.sv_process:
                continue

            if ng.has_changed:
                print('depsgraph_update_pre called - ng.has_changed -> ')

                ng.process()
        pre_running = False
```
to B, to avoid the indentation
```
@persistent
def sv_main_handler(scene):
    """
    Main Sverchok handler for updating node tree upon editor changes
    """
    global pre_running
    global sv_depsgraph
    global depsgraph_need

    # when this handler is called from inside another call to this handler we end early
    # to avoid stack overflow.
    if pre_running:
        return

    pre_running = True
    if depsgraph_need:
        sv_depsgraph = bpy.context.evaluated_depsgraph_get()

    for ng in sverchok_trees():

        if not ng.sv_process:
            continue

        if ng.has_changed:
            print('depsgraph_update_pre called - ng.has_changed -> ')

            ng.process()
    pre_running = False

```
to.. C  , using a context manager .   i'd be happy with B.
```python

@persistent
def sv_main_handler(scene):
    """
    Main Sverchok handler for updating node tree upon editor changes
    """
    global pre_running
    global sv_depsgraph
    global depsgraph_need

    # when this handler is called from inside another call to this handler we end early
    # to avoid stack overflow.
    if pre_running:
        return

    # inside this context pre_running is True, and once finished becomes False
    with pre_running_true():
        if depsgraph_need:
            sv_depsgraph = bpy.context.evaluated_depsgraph_get()

        for ng in sverchok_trees():
            if not ng.sv_process:
                continue

            if ng.has_changed:
                print('ng.has_changed')
                ng.process()

```

zeffii:
either way.. push the PR,

portnov:
As far as I understand, you're caching the reference to depsgraph? Isn't it being invalidated when we press Ctrl-Z?

zeffii:
maybe it should be invalidated by default, at the end of the last nodetree `<node>.process()`.

zeffii:
also. if there are two nodes that require the depsgraph, and both implement their "free" function , seems then that removing one of those nodes would globally disable the "depsgraph_need"

zeffii:
can't this be implemented without the use of "global" ?

vicdoval:
@zeffii I removed the indentation 😄  To solve it with out the use of global i think i would need to create a different file to host the global variables as stated here: https://www.geeksforgeeks.org/global-keyword-in-python/ do you think it is necessary/better?

> maybe it should be invalidated by default, at the end of the last nodetree <node>.process()
 
you mean adding one end line  after `pre_running = False` like `sv_depsgraph=[]`  ?

@portnov I cant detect any weird behaviour when pressing  Ctrl-Z  :/

vicdoval:
> adding one end line after pre_running = False like sv_depsgraph=[]

this does not work

I'm going to push this so we don't have that critical crash, if someone at sometime finds a better solution I will be happy to use it

zeffii:
i'll find time to test it soon, it might not be super useful to implement this differently.  Code inside the event handler function does get hoisted into a new namespace (iirc) 

if it works it works, conserve energy :)

Issue #2439, opened 2019-06-03
Draft mode for node tree


portnov:
From https://github.com/nortikin/sverchok/issues/1990#issuecomment-354586202

  * Enabled/disabled by a toggle button near "Update tree".
  * Input nodes, such as "A Number", have (and store) separate value for Draft mode (which is usually smaller, but it depends). It is displayed and used instead of main value, when the tree is in Draft mode.
  * Some nodes have (and store) separate values of their settings for Draft mode.
  * All nodes have programmatic access to this mode (smth like `if self.tree.draft:`). In draft mode, complex nodes execute some simplified version of their algorithm.

Durman:
As how I understood your suggestion I didn't catch what is purpose of this mode?

portnov:
@Durman , it is supposed that with smaller values of some parameters (such as number of subdivisions), node tree update will be much faster. So you can switch to "draft mode", play with tree and see result updated in real time, then switch to "final mode" and see final result.

Durman:
Possible variant of implementing of set of properties. Not sure that it is workable variant. But in this case it would not have limitation in number of modes.

```Python
class VectorInProps(bpy.types.PropertyGroup):
    x = bpy.props.FloatProperty(name='X')
    y = bpy.props.FloatProperty(name='Y')
    z = bpy.props.FloatProperty(name='Z')

class SvVectorIn ...

    props = bpy.props.ColectionProperty(name='Vector in props', type=VectorInProps)

    for mode_name in modes:
        props.add().name = mode_name

    x = props[active_mode].x
    y = props[active_mode].y
    z = props[active_mode].z
```

portnov:
done

PR #2861, opened 2020-02-04
Draft mode

portnov:
The initial idea was described in #2439 .

* Introduce draft mode
* Buttons for draft mode in UI
* Tree mode indicator in header
* Hotkeys:
** F5: trigger update of the current tree (the same as "Update   NodeTree" button in the panel)
** Ctrl+F5: the same as previous, but perform update even if "P" is      disabled for the node tree (enable it, process and then disable it      back).
** F6: toggle "P" (processing) of the current node tree.
** F7: toggle "D" (Draft) mode for the current node tree.

This, in fact, creates a new possible approach to working with complex node trees: to disable processing, play around with nodes and settings, and then trigger node tree recalculation only when you really want it (by Ctrl-F5).

For now, "draft" mode is supported in "A number" node, and in "Plane" node. I suggest to add support of this mode to other nodes on-demand.

@zeffii @Durman @vicdoval please review and test.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -31,7 +31,7 @@ from sverchok.data_structure import get_other_socket
 
 from sverchok.core.update_system import (
     build_update_list,
-    process_from_node,
+    process_from_node, process_from_nodes,
     process_tree,
     get_update_lists, update_error_nodes,
     get_original_node_color)
@@ -223,9 +223,30 @@ class SverchCustomTree(NodeTree, SvNodeTreeCommon):
         """
         This is triggered when Draft mode of the tree is toggled.
         """
+        draft_nodes = []
         for node in self.nodes:
             if hasattr(node, 'does_support_draft_mode') and node.does_support_draft_mode():
+                draft_nodes.append(node)
                 node.on_draft_mode_changed(self.sv_draft)
+
+        # From the user perspective, some of node parameters
+        # got new parameter values, so the setup should be recalculated;
+        # but techically, node properties were not changed
+        # (only other properties were shown in UI), so enabling/disabling
+        # of draft mode does not automatically trigger tree update.
+        # Here we trigger it manually.
+
+        if draft_nodes:
+            try:
+                bpy.context.window.cursor_set("WAIT")
+                was_frozen = self.is_frozen()
+                self.unfreeze(hard=True)
+                process_from_nodes(draft_nodes)
+            finally:
+                if was_frozen:
+                    self.freeze(hard=True)
+                bpy.context.window.cursor_set("DEFAULT")
+
 
     sv_draft : BoolProperty(
                 name = "Draft",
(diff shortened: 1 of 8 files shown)


Durman:
- The `A number` node is in draft mode when it is disabled and vice versa.
- It would be nice if changing of draft mode causes update event of node tree from N panel.
- The `Plane` node does not show `draft` label in draft mode.
- Bug: switch off process by F6, press F5 one time, then F7 does not make any effect.

I see that the conception has some downside. All parameters which maintain draft mode should be set twice whether they really should have different value or not. In case big node tree it can be quite painful to adopt whole node tree to draft mode because when you will switch to draft node all parameters are reseted to default.

node_tree.py
```diff
@@ -296,6 +297,14 @@ def node_id(self):
             self.n_id = str(hash(self) ^ hash(time.monotonic()))
         return self.n_id
 
+    def does_support_draft_mode(self):
+        """
+        Nodes that either store separate property values
+        for draft mode, or perform another version of
+        algorithm in draft mode, should return True here.
+        """
+        return False
```
Durman: Probably this could be a dynamic property.

portnov:
> The A number node is in draft mode when it is disabled and vice versa.

It was actually a misplaced `[D]` sign. Fixed.

> The Plane node does not show draft label in draft mode.

I'm not sure if it is required. But maybe it would be helpful.

portnov:
> Bug: switch off process by F6, press F5 one time, then F7 does not make any effect.

I'm not sure what do you mean. "P" and "D" modes are currently toggled independently. If "P" is off, then F5 does nothing (as well as "Update NodeTree" button in the N panel).
If "P" is off, then it does not matter whether "D" is on or off: no processing will be done anyway... But it can still have sense to change "D" mode while "P" is off, just to be sure which mode you will be in when you will enable "P" or when you press Ctrl-F5.

portnov:
> All parameters which maintain draft mode should be set twice whether they really should have different value or not.

It seems to me I've come up with solution. Now all parameters of nodes (ones that support draft mode) are copied from "standard" properties to their "draft" counterparts the first time when the node is switched into draft mode. Please test.

portnov:
> It would be nice if changing of draft mode causes update event of node tree from N panel.

Will try to implement.

portnov:
> It would be nice if changing of draft mode causes update event of node tree from N panel.

Implemented. @Durman  please test :)

I'd also like for @zeffii  for take a look at the code at places where I do some freeze-unfreeze calls. It seems to work for me but still a bit mysterious :)

portnov:
I also have a feeling that it would be good to have a hotkey to enable/disable ViewerDraw (and, maybe, similar nodes, such as BMesh Viewer). But for now I can't figure out what to do if we have several such nodes.

zeffii:
i don't see any obvious issues with Freeze or Throttle @portnov , but i'm not sure why freeze even needs to be used.

vicdoval:
Is this somewhere in the docs?

portnov:
I'm not sure...

PR #2864, opened 2020-02-05
sweep node

zeffii:
maybe we need it, maybe we don't


zeffii:
image

there are several issues to resolve..

zeffii:
image

zeffii:
image

zeffii:
image

enzyme69:
Hi Zeffii, got blend with working sweep node modulator to make that elephant trunk above? Is the node supposed to module different shapes and make interpolation weep?

PR #2871, opened 2020-02-07
List Join Numpy mode and other Numpy candy

vicdoval:
Added a NumPy mode to List join: 
- Pros: 
  Faster numpy array joins, 
  keeps it in Numpy
- Cons: 
  Cant handle different arrays with different axis number
  Just implemented for Level < 4.

On this PR also:
Object baker of Viewer Draw supports numpy
Random Num Gen right click menu
Vector Out  performing only if any output is connected
Data Shape informs about numpy arrays shape

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/data_structure.py
+++ b/data_structure.py
@@ -467,8 +467,8 @@ def describe_data_shape(data):
     """
     def helper(data):
         if not isinstance(data, (list, tuple)):
-            if isinstance(data,(np.ndarray)):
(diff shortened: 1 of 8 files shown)


vicdoval:
Finally the implementation only affects to some cases (when is nicer to concatenate numpy arrays) but in those cases it can be over 10 times faster than regular mode 😃

PR #2881, opened 2020-02-14
Monad add new socket, without noodle linking first.

zeffii:
- [ ] add socket
  
   image

- [ ] allow appropriate property configuration

image

also sneak in a few upgrades
- [ ] allow new empty monad creation


zeffii:
deprecated for `monad fixer 15`

PR #2882, opened 2020-02-15
"Constant list" node

portnov:
This node generates a list by repeating a single constant (integer or
floating-point) value. It is "vectorized", so by putting some lists to it's
inputs it is possible to make lists consisting of different numbers repeated
different count of times. Such lists can be useful as generated masks for some geometry.

Screenshot_20200215_203905
Screenshot_20200215_203957
Screenshot_20200215_204044

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


vicdoval:
Although/Because it seems quite a light node maybe would be interesting if it could accept and produce numpy arrays :smile:

PR #2899, opened 2020-02-24
Move mk3, Scale mk3  and Rotation m3

vicdoval:
Re-written of this nodes y multilevel recursion fashion
image

Move and Scale are in fully numpyzed
Rotate is partially written in numpy except for the case in which of one rotation matrix per vertex is needed and if there are many rotation centers and output numpy is disabled.

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


vicdoval:
I am having troubles replacing the rotation mk2:
The inputs of Euler and Quaternion modes are created after the init function. (on property change)
When the replacement operator runs the inputs don't change so the links are not made.

Why they don't update if the property is changed during the replacement?
Any suggestions to solve this?

zeffii:
i can't yet today.. but briefly looking at sv_init of `rotation_mk2`, i'm not sure it needs a call to `self.mode_change(context)`. because upon initialization the `mode` is the same as `current_mode` anyway, and it will return early + it will trigger another update (because the throttled updatefunction (mode_change) will call updateNode, after execution of the body of the updatefunction.

portnov:
@zeffii maybe in `@throttled` we should call `updateNode` only when there actually was a change of throttle mode from True to False?

zeffii:
@portnov  `@throttled`  could accept functions, 
```python
@throttled(pre=test1, post=test2)
```
and the return value of those could determine:
test1 = if the body of the wrapped function is even run,
test2 = do you update after wrapped function is run

but..  i reversed back a little and thought that was getting too complicated. I think `rotation_mk2` can be adjusted to not do a potential update from sv_init.

zeffii:
i won't have time until later today (at the earliest) stare at this.

portnov:
It is in general a widely used idiom, I think:
```
@throttled
def update_sockets(self, context):
    self.inputs['A'].hide_safe = self.mode != 'A'

mode : EnumProperty(..., default = 'B', update=update_sockets)

def sv_init(self, context):
   self.inputs.new('A')
   self.inputs.new('B')
   self.update_sockets(context)
```
This way you do not have to duplicate the logic of hiding/unhiding sockets, it is always in `update_sockets`, and you can be sure only correct sockets will be shown by default.

What I'm proposing is smth like
```
def throttled(func):
    def wrapper_update(self, context):
              # here skip_tree_update is always False
              func(self, context)
        # here skip_tree_update is restored...
        if not self.skip_tree_update:
            self.process_node(context)

return wrapper_update
```

zeffii:
> It is in general a widely used idiom, I think:

ok, yes it is used more for nodes that have a lot of setup to do. It's not good or bad.

>         if not self.skip_tree_update:
            self.process_node(context)

i don't know. i'm not eager to haste a c0de-change on that.

zeffii:
let's back up a little here

> When the replacement operator runs the inputs don't change so the links are not made.
> Why they don't update if the property is changed during the replacement?

yes. indeed .why?

zeffii:
https://github.com/nortikin/sverchok/blob/master/ui/nodes_replacement.py#L47

possibly we could move the `migrate_from` call to before the inputs are set.

zeffii:
but.. we could also add a call to a `migrate_props_pre_relink` function, which would be called before the link copying and setting.

```diff
        # Copy ID properties
        for prop_name, prop_value in old_node.items():
            new_node[prop_name] = old_node[prop_name]

+       # or add stub to the node base class
+       if hasattr(new_node, "migrate_props_pre_relink"):
+            new_node.migrate_props_pre_relink(old_node)

        # Copy incoming / outgoing links
        old_in_links = [link for link in tree.links if link.to_node == old_node]
        old_out_links = [link for link in tree.links if link.from_node == old_node]


```

vicdoval:
Thanks for the back up :) by the moment i solve it with the hide_safe strategy (with the ķnown downsides of the ctrl+h and the inputs appearing on the n-panel).
I guess is not that bad but probably the pre-link function is a cleaner solution. You think i should implement it?

zeffii:
yeah do it!  at least try it and report back.

vicdoval:
One of the advantages of the pre-link function would make possible the mk2 of nodes as list-join or other future nodes with dynamic sockets..

zeffii:
as a side note:  do we need to make a replacement for `ctrl+h` i'm sure we can come up with something that takes our needs into consideration. Of course we won't easily be able hide the sockets in the node side-panel.

vicdoval:
the pre-link function works great 😄

zeffii:
excellent stuff @vicdoval ! thank you

PR #2902, opened 2020-02-25
Instancer mk2

willneedit:
A modified version of the Mesh Instancer Node that takes an object parameter instead of manually configuring an object as the node's property.

- Allows for integration in a node tree's workflow to create instanced objects of a just-created and configured 'blueprint' object.
- Additional property 'Delete Source', which removes the 'blueprint' object from the final output, leaving the resulting instanced objects in place.
- Generates as many instances as matrices is passed into the node. If more than one object is passed in, instanced objects will loop through the given blueprint objects.


portnov:
Hello.
Thanks for the PR!
Travis complains about unexisting documentation. Could you please add documentation file for the node, and add a reference to it to the corresponding `*_index.rst` file?

willneedit:
Well, here you go :)

zeffii:
seems good to go then!

Durman:
@willneedit I have found that the node does not update instances properly if input object collection is changed.

PR #2906, opened 2020-02-27
Update instancer_MK2.py

zeffii:
adds cycle or randomize with seed


Durman:
Probably creating of a `shuffle` node which could mix any collection would be more wise solution then implement such approach for each node separetly where it seems like useful case.

PR #2914, opened 2020-02-29
Collection picker mk1

zeffii:
adresses some prelim bugs in this new version.

### DIFF
--- a/docs/nodes/scene/collection_picker_mk1.rst
+++ b/docs/nodes/scene/collection_picker_mk1.rst
@@ -1,0 +1,4 @@
+Collection Picker mk1
+=====================
+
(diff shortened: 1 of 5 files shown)


zeffii:
i'm merging these changes, because they are already better than the current state. but more must be done.  

At present, only objects that are of type Mesh will be instanced correctly. Others (Nurbs. Lights, Empties.... anything not MESH) will be instances as an Empty.  

fixes for that coming up.

regarding the Collection Picker Node. I suspect we will want a mode to also pass "nested Collections" found in the collection.

zeffii:
also... what about objects that have Modifiers ? :)

zeffii:
might experiment with object.copy()  which will transfer moifes etc for free :)

PR #2916, opened 2020-02-29
Instancer data agnostic 1

zeffii:
allow fullcopy (with modifiers)


zeffii:
i've updated this node a lot @willneedit @Durman @BennyGovaerts

PR #2917, opened 2020-02-29
NURBS Curve / Surface Output

portnov:
This adds two nodes:

* NURBS Curve Output
* NURBS Surface Output

The nodes create Blender's NURBS objects. Note that NURBS support in Blender is limited; for example, there is no way to specify the knot vector explicitly.

Also there is some problem in Blender (to be reported later, I think), due to which I have to avoid creating an object during scene update events. Updating existing object works fine.

Screenshot_20200229_113920
Screenshot_20200229_214359

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/nodes/viz/viewer_nurbs_curve.py
+++ b/nodes/viz/viewer_nurbs_curve.py
@@ -1,0 +1,128 @@
+# This file is part of project Sverchok. It's copyrighted by the contributors
+# recorded in the version control history of the file, available from
+# its original location https://github.com/nortikin/sverchok/commit/master
+#  
+# SPDX-License-Identifier: GPL3
+# License-Filename: LICENSE
+
+import numpy as np
+from itertools import zip_longest
+import traceback
+
+import bpy
+from mathutils import Matrix, Vector
+from bpy.props import StringProperty, BoolProperty, IntProperty, EnumProperty
+
+from sverchok.node_tree import SverchCustomTreeNode, throttled
+from sverchok.data_structure import Matrix_generate, match_long_repeat, updateNode, get_data_nesting_level, ensure_nesting_level, describe_data_shape, zip_long_repeat, fullList
+from sverchok.utils.sv_obj_helper import SvObjHelper
+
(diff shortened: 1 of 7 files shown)


nodes/viz/viewer_nurbs_curve.py
```diff
@@ -0,0 +1,128 @@
...
+    def sv_init(self, context):
+        self.inputs.new('SvVerticesSocket', 'ControlPoints')
+        self.inputs.new('SvStringsSocket', 'Weights')
+        self.inputs.new('SvStringsSocket', "Degree").prop_name = 'degree'
+        self.outputs.new('SvObjectSocket', "Objects")
+
+    def process(self):
+        if not self.activate:
+            return
+
```
zeffii: stick a  with self.sv_throttle_tree_update()  here, any part of this process function that directly modifies Blender data, will trigger Sverchok updates otherwise.

zeffii:
cool.

 you will no doubt be in a better position to articulate what Sverchok needs to Blender devs in order to make it possible to do this without using `bpy.ops`

portnov:
An example of where Blender's functionality is limited:

Screenshot_20200301_121144

White (selected) line — a curve being interpolated; black line — NURBS Curve object generated by Blender; Blue line - correct interpolating NURBS Curve; Red points - control points being used to build both curves. The difference between blue and black curves is (I believe) caused by lack of knot vector input: correct interpolating curve uses non-uniform knot vector, while Blender's NURBS objects always use uniform knot vectors.

vicdoval:
I was checking the code and behaviour of this nodes  to add them in the #2925 but they both have many unused imports (not very important but useless) some unused variables and if I try the  'Separated lists' in the nurbs surface I get this error
 ```
\nodes\viz\viewer_nurbs_surface.py", line 249, in process
    spline.points.add(n_u - 1)
UnboundLocalError: local variable 'n_u' referenced before assignment
```

portnov:
Fixed. it seems...

PR #2918, opened 2020-02-29
Added object output socket for instancer_MK2

willneedit:
instancer_MK2: Added object output socket providing the newly instanced objects


zeffii:
ok : )

PR #2921, opened 2020-03-01
List split and slice accept numpy

vicdoval:
split and slice

- [x] Ready for merge.


zeffii:
yep good, wicked!

zeffii:
what about `/list_struct/slice_lite.py`

vicdoval:
> what about `/list_struct/slice_lite.py`

If you make the docs I can give it a go

zeffii:
i'll look tomorrow, it might be deprecatable seen as listsplit now does numpy too.

vicdoval:
i have never used it, i see now is a list split with no unwrap possibility... maybe could be added as an option to the list split node

PR #2925, opened 2020-03-03
Viewers accepting numpy

vicdoval:
Viewers accepting numpy
Bmesh Viewer
VDraw
Curve viewer
Polyline viewer
Skin Meshes
Viewer index

Also implicit conversion (verts/matrix) accepts numpy
image

- [x] Ready for merge.

### DIFF
--- a/data_structure.py
+++ b/data_structure.py
@@ -254,13 +254,11 @@ numpy_list_match_modes =  list_match_modes[:3]
 #     ]
 
 def numpy_full_list(array, maxl):
-    # difl = maxl - array.shape[0]
-    try:
-        difl = maxl - array.shape[0]
-    except:
+    if type(array) != np.ndarray:
         array = np.array(array)
-        difl = maxl - array.shape[0]
-    difl = maxl - len(array)
+
+    difl = maxl - array.shape[0]
+
     if difl > 0:
         new_part = np.repeat(array[np.newaxis, -1], difl, axis=0)
         array = np.concatenate((array, new_part))
--- a/data_structure.py
+++ b/data_structure.py
@@ -166,6 +166,22 @@ def fullList(l, count):
         l.extend([l[-1] for a in range(d)])
     return
 
+def fullList_np(l, count):
+    """extends list l so len is at least count if needed with the
+    last element of l"""
+    n = len(l)
+    if n == count:
+        return
+    d = count - n
+    if d > 0:
+        try:
+            l.extend([l[-1] for a in range(d)])
+        except:
+            l = numpy_full_list(l, n)
+    else:
+        l = l[:count]
+    return
+
 def fullList_deep_copy(l, count):
     """the same that full list function but
     it have correct work with objects such as lists."""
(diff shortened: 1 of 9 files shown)


data_structure.py
```diff
@@ -237,6 +253,18 @@ def sv_zip(*iterables):
...
+    try:
+        difl = maxl - array.shape[0]
+    except:
+        array = np.array(array)
+        difl = maxl - array.shape[0]
+    difl = maxl - len(array)
+    if difl > 0:
+        new_part = np.repeat(array[np.newaxis, -1], difl, axis=0)
+        array = np.concatenate((array, new_part))
+    return array[:maxl]
```
Durman: Not too wide exception?
I would consider about possibility of using `np.resize` function here.
Instead of `np.repeat` using of `np.full` looks like more logical.

nodes/viz/viewer_polyline.py
```diff
@@ -71,20 +71,14 @@ def live_curve(obj_index, node, verts, radii, twist):
...
-            polyline.points.foreach_set('tilt', TWIST)
-            
+        if len(RADII) > 0:
+            rad = numpy_full_list(RADII, len(VERTS))
+            polyline.points.foreach_set('radius', rad)
+
+        if  len(TWIST) > 0:
+            twist = numpy_full_list(TWIST, len(VERTS))
+            polyline.points.foreach_set('tilt', twist)
+
```
zeffii: cool

data_structure.py
vicdoval: np.resize has some complications with 2-axis arrays (vector arrays) also np.full would just work with 1d arrays.
About the try/ except y change it to `if type(array) != np.ndarray` that seems more readable :)

Durman: Why don't use `AttributeError`?
It looks work:
```Python
np.full((3, 3), [1,2,3])
array([[1, 2, 3],
       [1, 2, 3],
       [1, 2, 3]])

a = np.array([[1,2,3], [1,2,3]])
a.resize(3,3)
a
array([[1, 2, 3],
       [1, 2, 3],
       [0, 0, 0]])
```

vicdoval:
```
np.full((3, 3), [[1,2,3],[2,3,4]])
ValueError: could not broadcast input array from shape (2,3) into shape (3,3)

```
This alone is awesome for cycle match
```
np.resize([[1,2,3],[2,3,4]], (5,3))
array([[1, 2, 3],
       [2, 3, 4],
       [1, 2, 3],
       [2, 3, 4],
       [1, 2, 3]]
       )
```


So i was testing this two
```
def numpy_full_list(array, maxl):
    if type(array) != np.ndarray:
        array = np.array(array)

    difl = maxl - array.shape[0]

    if difl > 0:
        new_part = np.repeat(array[np.newaxis, -1], difl, axis=0)
        array = np.concatenate((array, new_part))
    return array[:maxl]

def numpy_full_list2(array, maxl, vect=False):
    l= len(array)
    difl = maxl - l
    
    if difl > 0:
        if vect:
            array = np.resize(array, (maxl,3))
        else: 
            array = np.resize(array, maxl)
        array[l:]= array[l-1]
        
    return array[:maxl]
```
for 1d arrays the resize method is 2 or 3 times faster
for 2d arrays the concatenate method is  6 or 7 times faster
So i think i should keep the two of them and use them wisely
Any idea to improve them?

Durman: There is no need in slicing:
```Python
np.resize([1,2,3,4], 2)
array([1, 2])
```
`Vector` argument is not good. Using of something like this should help.
```Python
np.resize(a, [length] + a.shape[1:])
```
The same is fair about `full` function:
```Python
a = np.array([[1,2],[3,4]])
np.full([5] + list(a.shape[1:]), a[-1])

array([[3, 4],
       [3, 4],
       [3, 4],
       [3, 4],
       [3, 4]])

a = np.array([[[1,2]], [[3,4]]])
np.full([5] + list(a.shape[1:]), a[-1])

array([[[3, 4]],
       [[3, 4]],
       [[3, 4]],
       [[3, 4]],
       [[3, 4]]])
```

Also can be used =)
```Python
np.pad(np.arange(5), (0,2), 'edge')
array([0, 1, 2, 3, 4, 4, 4])
```

vicdoval: After some more test the repeat + concatenate method is in most of the cases faster than reshape and more flexible than full, so i think i'm going to leave it like that for the moment.
Anyway thanks for the feedback 👍

zeffii:
i especially like how minimally invasive the code is :)

vicdoval:
i was also gladly surprised :)

PR #2929, opened 2020-03-05
show x location sorting in disjoint sets

zeffii:
addresses race conditions explicitely by sorting the update sequence of disjoint nodes, as a function of the lowest 'x' (or y ) in a given subgraph.


zeffii:
this solution does not take into consideration the `Frame` adjusted location of nodes.. (which complicates stuff a little bit)

core/update_system.py
```diff
@@ -184,7 +184,16 @@ def separate_nodes(ng, links=None):
             nodes.discard(n)
             node_set_list[-1].add(n)
 
-    return [ns for ns in node_set_list if len(ns) > 1]
+    found_node_sets = [ns for ns in node_set_list if len(ns) > 1]
+    sort_component = 'x'  # ["x", "y", "None"]
+    
+    if sort_component == 'x':
+        print(found_node_sets)
+        find_lowest_coordinate = lambda ns: min(ng.nodes[n].location.x for n in ns)
```
zeffii:
```diff
- find_lowest_coordinate = lambda ns: min(ng.nodes[n].location.x for n in ns)
+ find_lowest_coordinate = lambda ns: min(ng.nodes[n].absolute_location[1] for n in ns)
```

zeffii:
@portnov @nortikin @Durman @vicdoval @ly29   yay  / nay ?

zeffii:
i didn't test to see if this still holds true if the graph topology is unchanged,, but the user moves nodes around (to effectively invalidate the sorting order)

portnov:
I don't have any opinion on this... obviously there should be some fixed evaluation order, but imho it does not matter what the ordering algorithm is.

zeffii:
i'll make it a 3 mode thing. `[X, Y, None]`

vicdoval:
When i have needed to ensure the order of disconnected trees i have relayed on the  different nodetrees which the get executed from top to bottom 
image
image

zeffii:
@vicdoval . yep. i think you have very subtle but important point here.

if a nodetree is simple, and has a few disjoint graphs, then some nodetree-subset sorting is useful, but as soon as you get elaborate trees, it's probably a good idea to separate the nodes into individual trees.

zeffii:
image

zeffii:
image

zeffii:
the question marks are dynamic tooltips

zeffii:
i'm happy to push this,  unless objections.

core/update_system.py
```diff
@@ -185,7 +185,15 @@ def separate_nodes(ng, links=None):
...
-    return [ns for ns in node_set_list if len(ns) > 1]
+    found_node_sets = [ns for ns in node_set_list if len(ns) > 1]
+  
+    sorting_type = ng.sv_subtree_evaluation_order  
+    if sorting_type in {'X', 'Y'}:
+        sort_index = 0 if sorting_type == "X" else 1
+        find_lowest = lambda ns: min(ng.nodes[n].absolute_location[sort_index] for n in ns)
+        found_node_sets = sorted(found_node_sets, key=find_lowest)
+
+    return found_node_sets
```
Durman: It would be nice for placing this code in separate function with name like `sorting execution order of nodes` for self-documentation and easier to read.

nortikin:
Снимок экрана от 2020-03-30 17-38-28
Additionally can be used "node arrange addon" to order nodes in area.

PR #2930, opened 2020-03-06
Plane mk3

vicdoval:
New plane mk3 (in numpy) following the changes of the last line_mk4 node, splitting the different behaviours
image
As usual the numpy inplementation is over 2 times faster when outputting regular lists and over 10 times faster when outputting numpy arrays...
- [ ] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/docs/nodes/generator/generator_index.rst
+++ b/docs/nodes/generator/generator_index.rst
@@ -15,7 +15,7 @@ Generators
    image
    line_mk4
    ngon
-   plane_mk2
+   plane_mk3
    random_vector_mk2
    script1_lite
    sphere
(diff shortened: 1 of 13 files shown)


vicdoval:
The actual plane mk2 has this  "separate" property that splits the plane in lines and wraps them.
I have never used this property neither seen any example of its use.
I think this property/functionality could be removed from the node. Any ideas about this?
image

vicdoval:
What about the linking sizes
image
I am also about to delete it, it was fine, before size x and size y had a proper socket...

vicdoval:
Ok, this is ready, 
The performance difference goes from 1.5x of boost  to 100x (ouputting numpy).😃

PR #2931, opened 2020-03-06
Console node pydatatexture

zeffii:
extends console color mode, with texture loading from disk (as np data)


zeffii:
merging to have this in the repo, it may be the backbone of a future indexviewer/debugsatement in-situ/notation system.

zeffii:
wow. i didn't realize that the 256*256  npy ( of a png)  was 2 mb.  I will look for a less expensive alternative.

zeffii:
```python
t = g.repeat(3).reshape(-1, 3)
np.concatenate((t, np.array([1, 1, 1])[:,None]),axis=1)

import bpy
import numpy as np
import bgl

#self.char_image = r"C:\Users\zeffi\Desktop\GITHUB\sverchok\utils\modules\bitmap_font\consolas_0.png"

self = lambda: None
self.char_image = "consolas_0.png"
image = bpy.data.images.get(self.char_image)
pixels = np.array(image.pixels[:])
image_array = np.array(pixels[::4])
print(image_array.size)
# texture = bgl.Buffer(bgl.GL_FLOAT, self.total_size, np.resize(data, self.total_size).tolist())
# destination = r"C:\Users\zeffi\Desktop\consolas_0.npy"
# np.save(destination, image_array)




``

PR #2933, opened 2020-03-07
console node optimizer01

zeffii:
addresses one or more wasteful processing steps that can be cached.
- [x] cache fnt lookup creation, to one time per all instances of the node.


zeffii:
image

PR #2937, opened 2020-03-08
add exception display in the nodeview

zeffii:
this addresses a longstanding pain-in-the-butt issue with distance between the error and the node where the error occurs.

first a simple implementation, with as little overhead as humanly possible.

image

image

image

### DIFF
--- a/core/update_system.py
+++ b/core/update_system.py
@@ -31,7 +31,7 @@ from sverchok.utils.exception_drawing_with_bgl import clear_exception_drawing_wi
 
 import traceback
 import ast
-import sys
+# import sys
 
 graphs = []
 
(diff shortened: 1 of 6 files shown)


zeffii:
@portnov @Durman @vicdoval @nortikin 

maybe we can offer a variety of debug formats here..  despite having used the exception mechanism extensive i'm a little lost as to what is a good message to show us / the users . 

any input is welcome.    remember, the actual rendering of this message will probably not be using blf, but the code i wrote for the console node.

nortikin:
this debug stuff is dark matter for me :(

zeffii:
me too @nortikin .

hope you are doing well :)

portnov:
@zeffii I'd say it's better to output exception message only, without stack trace. Stack trace can be found in the log / console, if someone needs it.

zeffii:
i wish to keep the stack trace, seen as this mode is optional anyway, and I find it useful to see more information from the nodeview when I don't have big screenspace.

but stacktrace could be optional... eventually.

Issue #2898, opened 2020-02-24
SV Animation Remote Control can no longer override keyframe 


enzyme69:
I noticed that since Blender 2.82, SVerchok Bmesh Viewer or Matrix Remote Control can no longer override keyframe animation. In the past, the behaviour is that if I change keyframe, SVerchok will still be calculated last. But not with 2.82. Once keyframed, SVerchok can no longer control the object. --> Animation Nodes can still do it and bake 

SV Object ID Set can actually still override the Matrix of object, but slightly tricky.

<img width="818" alt="Screen Shot 2020-02-24 at 2 28 27 pm" src="https://user-images.githubusercontent.com/3952687/75127208-039efb80-5712-11ea-8248-687649cde2ef.png">

[object_remote_control_001.blend.zip](https://github.com/nortikin/sverchok/files/4243032/object_remote_control_001.blend.zip)

enzyme69:
Video:
https://www.youtube.com/watch?v=tuHeiG8RE9g

In 2.82, when object keyframed, it's pretty much locked and SVerchok no longer transform the object. In the past, if I use arrow key and stepping, I can still see the object moving, hence I can keyframe and bake.

enzyme69:
I have a feeling it's the Frame Info node that's not reacting / locked.

zeffii:
> I have a feeling it's the Frame Info node that's not reacting / locked.

unlikely, all that nodes does is read the current frame.

i think whats happening is that Sverchok is updating the location data before the frame change (using frame_change_pre ) , this is a setting in the preferences ( POST/PRE ) , try POST.  no guarantee.

zeffii:
## What i really think.
I know the following is not your question directly, but i think you're having this problem because you are forced by the current paradigm to approach animation with Sverchok in an arse-backwards way. it's not that I don't want to help you here, Blender dev have not added keyframe support for PyNode Nodetree sliders (last time i checked..).

maybe when add-on's are allowed to set keyframes from their UI (like sliders on nodes in a material...) you'd have more fun doing animation from the UI sliders/graph editor.

## Untill such time...:
i'm not sure it's a good feature tooverwrite keyframes. it feels so unintuitive.  How about using the FCurve In node, and reading evaluated values from that, and using an keyframed-Empty  ?

```
Sverchok =/= Animation Nodes 
```

enzyme69:
It's not Pre and Post thing, I tested it. I should logically this one. Calculate SVerchok last...

[...]

PR #2945, opened 2020-03-10
add throttle and update

zeffii:
issue #2898   may or may not be fixable.


zeffii:
this also doesn't override a keyframe.

PR #2949, opened 2020-03-11
List Mask Out, Mask to index, List Item and Insert Item speak numpy

vicdoval:
The mentioned nodes will output numpy arrays when feeded with numpy arrays.
The list mask out is over 10x faster in that case
I guess it could be refined I little more, but enough for me right now
image

- [x] Ready for merge.

### DIFF
--- a/data_structure.py
+++ b/data_structure.py
@@ -264,6 +264,24 @@ def numpy_full_list(array, maxl):
         array = np.concatenate((array, new_part))
     return array[:maxl]
 
+def numpy_full_list_cycle(array, maxl):
+    '''retuns array with desired length by cycling'''
+
+    difl = maxl - array.shape[0]
+    if difl > 0:
+        if difl < array.shape[0]:
+
+            return np.concatenate((array, array[:difl]))
+
+        new_part = np.repeat(array, ceil(difl / array.shape[0]), axis=0)
+        if len(array.shape) > 1:
+            shape = (ceil(difl / array.shape[0]), 1)
+        else:
+            shape = ceil(difl / array.shape[0])
+        new_part = np.tile(array, shape)
+        return np.concatenate((array, new_part[:difl]))
+
+    return array[:maxl]
 def numpy_match_long_repeat(list_of_arrays):
     '''match numpy arrays length by repeating last one'''
     out = []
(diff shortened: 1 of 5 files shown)


portnov:
Well... next logical step would be to automatically detect that the recipient node can consume numpy arrays and enable numpy mode in that case.
@zeffii any thoughts?

Consumer:
```
def sv_init()
   self.inputs.new('SvVerticesSocket', 'Vertices').can_consume_numpy = True
```

Producer:
```
def process():
    if self.outputs['Vertices'].is_linked and self.outputs['Vertices'].other.can_consume_numpy:
       output_numpy = True
```

zeffii:
> @zeffii any thoughts?

i'm all for the "auto-inter-configuration" of nodes when both can speak numpy, also the `implicit conversion` can be useful in some scenarios, where a node is outputting numpy but the recipient can not ( the node may have multiple recipients ) .

i'll be focusing on monad / snlite fixing for the timebeing.

vicdoval:
As you can do  `for v in np.array ` equally to  `for v in python_list` most of the algorithms would support numpy  but there are some cases in that  fail, the most popular is 
`if any_python_list:  `
this will return true if the python list is other than [] but wont know what to do if it is a numpy array.
I we change it to 
`if len(any_python_list) > 0:`
then it will work equally with both  
The other main fails are `extend` and `append`, this are more tricky and should be studied if turning the numpy array to a regular list or perform the same task in numpy...

By the moment i am using this script to change all the nodes (and change it later on the nodes that didn't work) and on my last piece  with 3.000.000 verts it supposed a 15-20% of speed up  

```
import bpy

state = True

b = bpy.data.node_groups["NodeTree.001"]
for nodes in b.nodes:
    # most of cases
    if hasattr(nodes, 'output_numpy'):
        nodes.output_numpy= state
        
    # line node
    if hasattr(nodes, 'as_numpy'):
        nodes.as_numpy= state
        
    # box node
    if hasattr(nodes, 'out_np'):
        nodes.out_np = state, state, state
```

Durman:
> Well... next logical step would be to automatically detect that the recipient node can consume numpy arrays and enable numpy mode in that case.

My implementation of `switch` node works exactly in this way. By the way it includes functionality of `list mask (out)` node.

vicdoval:
List Mask (Out) rewritten (now up to 10x faster with regular lists and 20x with numpy arrays)

PR #2957, opened 2020-03-11
Less deep copys

vicdoval:
Removed unnecessary deep_copys
deepcopy=False in
Component Analyzer
List Length
Vector math
Vector Polar In
- [x] Ready for merge.


zeffii:
yes!! if none of those nodes modifies incoming data : ) good thinking.

PR #2959, opened 2020-03-12
Snlite revision

zeffii:
this allows user to load blends that crash .blender

### DIFF
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -139,6 +139,8 @@ def sv_main_handler(scene):
     pre_running = True
     if depsgraph_need:
         sv_depsgraph = bpy.context.evaluated_depsgraph_get()
+    
+
     for ng in sverchok_trees():
         # if P (sv_process is False, we can skip this node tree.
         if not ng.sv_process:
@@ -171,7 +173,6 @@ def sv_post_load(scene):
     # ensure current nodeview view scale / location parameters reflect users' system settings
     from sverchok import node_tree
     node_tree.SverchCustomTreeNode.get_and_set_gl_scale_info(None, "sv_post_load")
-
 
     for monad in (ng for ng in bpy.data.node_groups if ng.bl_idname == 'SverchGroupTreeType'):
         monad.freeze(True)
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -174,8 +174,10 @@ def sv_post_load(scene):
 
 
     for monad in (ng for ng in bpy.data.node_groups if ng.bl_idname == 'SverchGroupTreeType'):
+        monad.freeze(True)
         if monad.input_node and monad.output_node:
             monad.update_cls()
+        monad.unfreeze(True)
 
     sv_types = {'SverchCustomTreeType', 'SverchGroupTreeType'}
     sv_trees = list(ng for ng in bpy.data.node_groups if ng.bl_idname in sv_types and ng.nodes)
--- a/core/monad.py
+++ b/core/monad.py
@@ -571,7 +571,7 @@ class SvGroupNodeExp:
         if not self.monad:
             print(">> monad.process didn't find self.monad")
             print(bpy.data.node_groups)
-            print([tree.name for tree in bpy.data.node_groups])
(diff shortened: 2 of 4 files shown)


vicdoval:
Btw is it possible to set deepcopy=False in the Snlite?

zeffii:
>  Btw is it possible to set deepcopy=False in the Snlite?

should be..

zeffii:
found the problem in snlite anyway..

zeffii:
@vicdoval  do you think all input from sockets should be `deepcopy=False` by default? :)

snlite uses sv_get only in `make_new_locals()`
```

       ...
       ...
            if s.is_linked:
                val = s.sv_get(default=[[]])
                if sock_desc[3]:
                    val = {0: val, 1: val[0], 2: val[0][0]}.get(sock_desc[3])
            else:
                val = sock_desc[2]
                if isinstance(val, (int, float)):
                    # extra pussyfooting for the load sequence.
                    t = s.sv_get()
                    if t and t[0] and len(t[0]) > 0:
                        val = t[0][0]
```

zeffii:
i would be happy to make a preprocessor directive like 
```
deepcopy=True
in verts v
out verts_out v
...etc
...etc
```
and otherwise assume user is OK with `deepcopy=False`

zeffii:
or just `deepcopy`

zeffii:
anyway. i need to eat.

Durman:
Is it really need for SNL? I think if you need a performance make a normal node.

zeffii:
i'll add it because it makes sense, it will be the default.

imho it's bad practice to modify incoming data "in-place" in the first place. let these reason be obvious.

Durman:
Am I correctly understand that `deepcopy=False` means that data is get by reference?

vicdoval:
I would go for deepcopy=False as the default, think for most of the cases is better and you can always make a deepcopy inside the node. 
But even better if you add that preprocessor directive :)

@Durman I thought about this testing a SNL vs a node, i think is a good idea to give the chance to make a faster SNL

zeffii:
just to recap, the default will be `deepcopy=False`,  the user  doesn't have to do anything. if the user really wants to deepcopy, then they add the `directive` at the top.

Durman:
I don't think that SNL has such a chance. The performance will be lousy anyway. 

I see the node as a good way for experiments for newcomers. `deepcopy=False` can lead to unexpected results and make SNL more difficult tool. It can lead to new issues and take time of Sverchok developers for explaining of SNL tricky behaviour. We should try to make things more simple and do not overcomplicate them. Than less we have to keep in mind then better. Even standard `sv_get` method has `deepcopy=True` by default. It had took long to understand difference between getting data by reference and as copy for me.

zeffii:
it will take a while to get around to this.

zeffii:
actually, you might be right @Durman  , maybe the default will be `deepcopy=True`, and only if the user wants to not use deepcopy, then they use the directive `no deepcopy`

Durman:
I have made simple test of adding 100000 vectors for comparison with `vector math` node. SNL is faster but I'm not sure that with more complex code it can have the same advantage.

```
2020-03-12 20:50:59,324 [DEBUG] sverchok.core.update_system: Processed  Line in: 0.2084
2020-03-12 20:50:59,642 [DEBUG] sverchok.core.update_system: Processed  Scripted Node Lite in: 0.3160
2020-03-12 20:51:01,641 [DEBUG] sverchok.core.update_system: Processed  Vector Math in: 0.4199
```

```Python
'''
in v1 v d=[] n=1
in v2 v d=[] n=1
out v_out v
'''

v_out = list(map(lambda u, v: (u[0]+v[0], u[1]+v[1], u[2]+v[2]), *[v1, v2]))
```

vicdoval:
@Durman  was you test before #2957?
testing your script with 100000 vectors this is what I get
```
2020-03-12 20:23:36,531 [DEBUG] sverchok.core.update_system: Processed  Line in: 0.1077
2020-03-12 20:23:36,806 [DEBUG] sverchok.core.update_system: Processed  Scripted Node Lite.001 in: 0.2742
2020-03-12 20:23:36,901 [DEBUG] sverchok.core.update_system: Processed  Vector Math in: 0.0878
```

Durman:
Before. The point of test was that the SNL was doing the same code as math node.

PR #2961, opened 2020-03-12
console_node line width limits

zeffii:
## an explanation of the current implementation:

- The console node interprets the incoming text data, and depending on the color mode does full
1. code tokenizing
2. random char colors (why?.. as tests..)
3. no color.
- The incoming text longest line is found, and this is used to set the terminal_width. The longest line is also used to pad all shorter lines with whitespace. The new padded text is stored as a multiline string in `node.terminal_text`.
- Then it makes a grid of `x * y` character cells, each cell is 2 triangles. 
     - So for example 2 lines, of 5 chars is 10 cells, that makes a total of 20 triangles.

The issue is when you load a json of a stored sverchok layout, and the json contains a monad, the `"groups"` key will contain one or more very long lines (because it describes a nodetree but as a flat string ) . This will then push the longest line length to a considerable length, and as a conseqence the total amount of triangles produced by the drawing callback is going to cause considerable slow down in drawing.

-----

the solution in this implementation is to automatically trim the longest line, the problem however is exactly what to do with that line.
   1. simply trim it and break lexing?
   2. trim it by replacing the line with a placeholder to indicate `# this line is trimmed`, and make it parsable as a code comment. (not breaking lexer)
   3. replace it with spaces
   4. completely skip the long line, and display a warning.

### DIFF
--- a/nodes/script/console_node.py
+++ b/nodes/script/console_node.py
@@ -153,16 +153,25 @@ def find_longest_linelength(lines):
     return len(max(lines, key=len))
 
 
-def ensure_line_padding(text, filler=" "):
+def ensure_line_padding(text, filler=" ", max_line_length=200):
     """ expects a single string, with newlines """
     lines = text.split('\n')
     longest_line = find_longest_linelength(lines)
-    
+    longest_line = min(longest_line, max_line_length)
+
     new_lines = []
     new_line = new_lines.append
+
+    # this may produce untokenizable linesm when trimming
     for line in lines:
-        new_line(line.ljust(longest_line, filler))
-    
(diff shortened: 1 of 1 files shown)


zeffii:
rather than trimming ultra long lines, (which causes tokenizer issue, in tokened lexer mode) if will replace the long line with `         ["trimmed"]`  where the indent matches the space indent of the input line.

zeffii:
or allow optional `replacement line`

zeffii:
haha!  OK.  what might work is i'll pretokenize, and if any string is longer than 200 chars it will be replaced with `"<start of string as normal> [...] "`

PR #2973, opened 2020-03-15
Snlite cats

zeffii:
yeah works.

image


vicdoval:
Is not it too long?
Wouldn't be cool to have sub-menus? :)

zeffii:
@vicdoval yes agreed. but this is better than it was.

Issue #2979, opened 2020-03-16
Random Points on Mesh --> Per Face?


enzyme69:
If I am not wrong Random Points On Mesh used to have ability to scatter points on polygon face, but this option is now disappear?

<img width="773" alt="Screen Shot 2020-03-17 at 10 08 18 am" src="https://user-images.githubusercontent.com/3952687/76807081-56626380-6837-11ea-8dcf-f06856369d35.png">

[random_point_mesh_002.blend.zip](https://github.com/nortikin/sverchok/files/4340741/random_point_mesh_002.blend.zip)

Basically I need random points per face and then connect the points to make cable muscle line

enzyme69:
Random Points On Mesh is not vectorized as I tested, but kind of working if I use Monad.

<img width="987" alt="Screen Shot 2020-03-17 at 10 21 36 am" src="https://user-images.githubusercontent.com/3952687/76807627-17351200-6839-11ea-8885-67cd908f3988.png">

enzyme69:
<img width="599" alt="Screen Shot 2020-03-28 at 4 27 25 pm" src="https://user-images.githubusercontent.com/3952687/77815688-0de95680-7111-11ea-9585-6eaa04430fea.png">

Random Points on Mesh is 100% broken :( latest version

PR #2987, opened 2020-03-18
Random point on mesh node - vectorization and refactoring

Durman:

#2979

## Preflight checklist

- [x] Ready for merge.

### DIFF
--- a/core/sockets.py
+++ b/core/sockets.py
@@ -17,6 +17,9 @@
 #
 # ##### END GPL LICENSE BLOCK #####
 
+
+from typing import NamedTuple, Any
+from itertools import count
 import math
 
 import bpy
@@ -34,6 +37,21 @@ from sverchok.data_structure import (
     get_other_socket,
     socket_id,
     replace_socket)
+
+
+socket_config = dict()
+config_id_iterator = count()
+
+
+class InputSocketConfig(NamedTuple):
+    id: int
+    name: str
+    socket_type: str
+    prop_name: str = ''
+    deep_copy: bool = True
+    vectorize: bool = True
+    default: Any = object()
+
 
 socket_colors = {
     "SvStringsSocket": (0.6, 1.0, 0.6, 1.0),
@@ -64,6 +82,13 @@ class SvSocketCommon:
     prop_name: StringProperty(default='', description="For displaying node property in socket UI")
 
     quicklink_func_name: StringProperty(default="", name="quicklink_func_name")
+    config_id: IntProperty(default=-1)
+
+    @property
+    def config(self):
+        if self.config_id in socket_config:
+            return socket_config[self.config_id]
+        raise LookupError(f"Config does not fount")
 
     def get_prop_name(self):
         if self.node and self.node.does_support_draft_mode() and hasattr(self.node.id_data, 'sv_draft') and self.node.id_data.sv_draft:
--- a/core/sockets.py
+++ b/core/sockets.py
@@ -17,9 +17,6 @@
 #
 # ##### END GPL LICENSE BLOCK #####
 
-
-from typing import NamedTuple, Any
-from itertools import count
 import math
 
 import bpy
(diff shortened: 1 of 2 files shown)


zeffii:
it seems to me that this implementation adds a considerably complicated method to initialize sockets, and sv_get.

zeffii:
i like the `get_iterator` tho.

Durman:
This approach is more complicated indeed. But there is an advantages as well. Purpose is automatization the work with sockets which have to be done each time inside a node. 

 In this case all what have to be done is to set all properties for sockets in one place like this:
```Python
    VERTICES = InputProps('Verts', 'SvVerticesSocket', deep_copy=False, vectorize=False)
    FACES = InputProps('Faces', 'SvStringsSocket', deep_copy=False, vectorize=False)
    FACE_WEIGHT = InputProps('Face weight', 'SvStringsSocket', deep_copy=False, default=[[1]])
    ...
```
And after that such automatization as initialization, get data from sockets can be done.
```Python
Inputs.init(node.inputs)

...

input_data = Inputs.get_iterator(node.inputs)
```

One of the problem is where to keep socket properties in such format because classes which are registered in Blender can't keep custom type of data.

I think `node.inputs.get_iterator()` is better then `node.get_iterator()` because this method is working with data which belongs to sockets not to a node.

Durman:
For creating method like `links.get_iterator` which would return `sv` data of all input sockets, input socket should have more properties like: should input data be vectorized or should socket has default value. 

Most appropriate place for assigning such attributes to socket is `node.sv_init` method. But socket can't keep python data such as `[None]` for default value for example. 

This problem could be solved the same way as Sverchok is keeping node results in output sockets. So socket should keep only link to a property which will be kept in Sverchok somewhere. But unlike node result this property can't be recalculated and will be lost after restarting Sverchok. 

So there is no way (I don't see) to assign `advanced` properties to sockets inside `node.sv_init` method.
Probably my way is not worst in such circumstances.

zeffii:
i'm not here to suggest you do not use this technique. i just question the divergence from our conventions. used by the 370+ other nodes.

Durman:
I think we can use new convention for new nodes and old nodes can remain with old convention. Any way this way does not do something new. It just encapsulates part of code inside `Inputs` class. 

I have noticed that vectorization is frequent mistake in my nodes. Probably this approach could guarantee correct work of node vectorization.

zeffii:
I restated the code [here](https://gist.github.com/zeffii/af9ab6f5116fe81af35af09ae75ca5ae) using more conventional code. 

(not tested)

zeffii:
i agree there could be better ways to elegantly state how to vectorize  a node,  but I don't see how adding the data to the socket class is the appropriate solution.  @portnov @nortikin @vicdoval   thoughts?

vicdoval:
I don't have a strong opinion on this, i believe that in a perfect world all inputs should be vectorized and not deepcopied in a multilevel-recursion fashion so i don't se the point on having vectorize property and i would look into making all possible inputs deepcopy=False.
But you want to automate the process i don't get the pros and cons of storing it on the socket or in the node... seems that by the moment it could be stored in the node and if the number of nodes using this strategy rises then it could be moved to the socket class

Durman:
@vicdoval Vectorization method can't be putted to socket class because it's need information about maximum length of data from input socket set. So most appropriate place is `inputs` object but Blender does not let to modify this object so next target is `node` object. 

This method also need extra information about sockets such as should data of a socket be vectorized or should it have default value. Assign such information to socket also is great problem because Blender does not allowed assign arbitrary Python type of data to attributes of Blender ID (?) object. 

@zeffii I like your solution with using just dictionary. But information in it looks random. I think it will be more clear if each socket have ful information about itself including default values. Another thing which I don't like is thet this values does not have any connection with code in `sv_init` method. When next time to change the code of the node will come it will be hard to understand how this information related with sockets or this information can be just ignored and the node will get unexpected behaviour again.

Issue #2443, opened 2019-06-09
List Match property and use of right click menu


vicdoval:
On the last nodes I coded i added a couple of enhancements that could be applied to future coding of nodes.
List Match Global and Local:
I have added a couple of Enumproperties to control the list matching in the node.
https://github.com/nortikin/sverchok/blob/210f48b1d0718820ecc549828f4e3bde133f5fec/nodes/analyzer/distance_point_line.py#L119-L129
In stead of importing match_long_repeat as usual i add 
https://github.com/nortikin/sverchok/blob/210f48b1d0718820ecc549828f4e3bde133f5fec/nodes/analyzer/distance_point_line.py#L23
calling to this:
https://github.com/nortikin/sverchok/blob/210f48b1d0718820ecc549828f4e3bde133f5fec/data_structure.py#L185-L200
and use them as in this function to get inputs data:
https://github.com/nortikin/sverchok/blob/210f48b1d0718820ecc549828f4e3bde133f5fec/nodes/analyzer/distance_point_line.py#L165-L169
or this case:
https://github.com/nortikin/sverchok/blob/210f48b1d0718820ecc549828f4e3bde133f5fec/nodes/analyzer/distance_point_line.py#L184-L188

I add this property to the N-Panel and on the Right-Click Menu

Which brings the next topic: Using the  Right-Click Menu
We already have it in Sverchok and i find a good practice to use it to add advanced parameters of the node
https://github.com/nortikin/sverchok/blob/210f48b1d0718820ecc549828f4e3bde133f5fec/nodes/analyzer/distance_point_line.py#L156-L163

Maybe the List Match Properties could be added as a default node property.. or just use it from now on..
What do you think? Do you find it a good practices? Any idea to improve them?

zeffii:
i have reservations about the usefulness of setting `list_math_global` from one node to apply to all nodes ( which is what 'global' suggests to me.... or do you mean all list_matching for that node ) . For `list_match_local` I 'm supportive.  What if the node has multiple sockets which might benefit from each a different `list_match` routine

I would however not be comfortable retrofitting existing nodes because it does look like that would be extra code for a tonne of nodes, and that's perhaps after a while you realize the feature is only useful in a small portion.

also... that would be more work to then apply those changes to svb28.

maybe i'm missing the point.

vicdoval:
`list_match_global` states for a level 1 data matching of incoming lists, `list_match_local` to a level 2 data.
And they just affect to the node
[...]

PR #2991, opened 2020-03-21
Add new Matrix In node (MK3)

DolphinDream:
Added a new (MK3) version of the MatrixIn node.

This version adds a few modes (similar to the **QuaternionIn** node) essentially to allow to define the rotation part of the 4x4 matrix in a few different ways, not just Axis+Angle as in the previous version. As a bonus, the angle based modes come with the angle units as well (similar to QuaternionIn node) to switch between Degrees, Radians.. and “Unities" (0-1).

The new node has 3 modes (as input for rotation):

- Quaternion 
- Axis+Angle
- Euler Angles

Currently there is a **Matrix Euler** matrix node to create matrices based on Euler angles, though this seems to have a bug at the moment. This new MatrixIn node would replace the Matrix Euler node as well. 

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

Here’s the UI (collapsed and expanded) for each of the 3 modes:

SV-MatrixInNodeMK3-UI1


vicdoval:
Please keep the improvements made at #2735 ( vectorization capabilities and the 'Flat Output' property in r-click menu and n-panel) 
Other idea: I think this node does not modify input data and it can benefit from deepcopy=False in the sv_get

DolphinDream:
Any ideas why a node cannot have a **location** property ? Is this interfering with another attribute defined in base classes? (perhaps in blender nodes?). I’m getting the following error when I instantiate a node that has location as one of its properties:

SV-locationProp-Error1

SV-locationProp-Error2

DolphinDream:
One question.. about inputs.

If i have a Quaternion input socket.. when it’s not connected, when reading the socket via sv_get() I get a list of a tuple.. but when it is connected I get a list of (mathutils) Quaternion(s).

How can I handle the reading so that I always end up with a quaternion? 
no connection:
q =  [(1.0, 0.0, 0.0, 0.0)]

with connection:
q =  <Quaternion (w=0.0000, x=0.0000, y=0.0000, z=0.0000)>

Since I have to perform quaternion operations on q I have to make sure this input is Quaternion type not, a list/tuple.

Durman:
You are getting tuple because you are using `FloatVectorProperty` by default which is connected via `prop_name` attribute.
```Python
            source_data = SvGetSocket(self, deepcopy = True if self.needs_data_conversion() else deepcopy)
            return self.convert_data(source_data, implicit_conversions)

        if self.get_prop_name():
            return [[getattr(self.node, self.get_prop_name())[:]]]
        elif self.use_prop:
            return [[self.prop[:]]]
        elif default is sentinel:
            raise SvNoDataError(self)
        else:
            return default
```
Probably this part can be modificated
```Python
Quaternion(getattr(self.node, self.get_prop_name()))
```


https://docs.blender.org/api/current/bpy.types.Node.html#bpy.types.Node.location

DolphinDream:
Well aside from documentation and relocating the old node I think this PR should be done. Any comments?

Btw, I named the node name and filename and MatrixInNodeMK3 instead of the MatrixGenNodeMK3 (and matrix_in_mk3.py etc).

DolphinDream:
Perhaps some profiling to do a mk2 vs mk3 comparison could be useful.

DolphinDream:
Some profiling... 

Using a simple setup. 1K random inputs to generate 1K matrices.. for 1000 frames, the results for the Angle-Axis (the input format provided by the previous version of the node) gives about 3x speed improvement.

SV-MatrixInNodeMK3-profilingSetup

MK3 profiling:

```
   Ordered by: cumulative time

     1010    2.514    0.002    4.895    0.005 matrix_in_mk3.py:198(process)
     1010    0.862    0.001    1.337    0.001 data_structure.py:92(match_long_repeat)
  1034240    0.539    0.000    0.539    0.000 {method 'to_matrix' of 'Quaternion' objects}
  1035250    0.375    0.000    0.472    0.000 data_structure.py:77(repeat_last)
     1010    0.010    0.000    0.177    0.000 matrix_in_mk3.py:258(<listcomp>)
  1034240    0.165    0.000    0.165    0.000 {method 'to_4x4' of 'Matrix' objects}
     3030    0.014    0.000    0.131    0.000 sockets.py:381(sv_get)
  1038280    0.106    0.000    0.106    0.000 {method 'append' of 'list' objects}
     8080    0.005    0.000    0.106    0.000 sockets.py:78(other)
     8080    0.022    0.000    0.101    0.000 data_structure.py:943(get_other_socket)
  1042320    0.099    0.000    0.099    0.000 {built-in method builtins.len}
     8080    0.020    0.000    0.079    0.000 bpy_types.py:1008(links)
     5050    0.010    0.000    0.079    0.000 sockets.py:245(needs_data_conversion)
     3030    0.011    0.000    0.061    0.000 socket_data.py:85(SvGetSocket)
    16160    0.059    0.000    0.059    0.000 bpy_types.py:1012(<genexpr>)
     1010    0.001    0.000    0.057    0.000 sockets.py:113(sv_set)
     1010    0.048    0.000    0.055    0.000 socket_data.py:70(SvSetSocket)
     3030    0.003    0.000    0.044    0.000 sockets.py:250(convert_data)
     1010    0.002    0.000    0.036    0.000 sockets.py:509(sv_get)
     4040    0.018    0.000    0.020    0.000 sockets.py:86(socket_id)
     2020    0.006    0.000    0.006    0.000 sockets.py:68(get_prop_name)
     1010    0.004    0.000    0.004    0.000 matrix_in_mk2.py:79(process)
     4040    0.002    0.000    0.002    0.000 {built-in method builtins.max}
     4040    0.002    0.000    0.002    0.000 {built-in method builtins.hash}
     1010    0.001    0.000    0.001    0.000 {built-in method builtins.getattr}
     2020    0.000    0.000    0.000    0.000 node_tree.py:396(does_support_draft_mode)
     2020    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}

```

MK2 profiling:

```
Ordered by: cumulative time

     1013    0.057    0.000   13.257    0.013 matrix_in_mk2.py:79(process)
     1013    0.793    0.001    9.139    0.009 matrix_in_mk2.py:27(matrix_in)
     1013    5.140    0.005    8.263    0.008 data_structure.py:729(matrixdef)
     3039    0.016    0.000    2.820    0.001 sockets.py:381(sv_get)
     3039    0.013    0.000    2.749    0.001 socket_data.py:85(SvGetSocket)
2080702/3039    1.462    0.000    2.686    0.001 socket_data.py:37(sv_deep_copy)
5065/3039    0.465    0.000    2.678    0.001 socket_data.py:42(<listcomp>)
     3039    0.004    0.000    1.114    0.000 data_structure.py:704(Vector_generate)
     3039    0.003    0.000    1.111    0.000 data_structure.py:706(<listcomp>)
  4149248    1.031    0.000    1.031    0.000 {built-in method builtins.min}
  4161404    0.759    0.000    0.759    0.000 {built-in method builtins.isinstance}
  1037312    0.595    0.000    0.595    0.000 {built-in method Rotation}
  4164443    0.432    0.000    0.432    0.000 {built-in method builtins.len}
  1037312    0.233    0.000    0.233    0.000 {built-in method Identity}
  2079689    0.209    0.000    0.209    0.000 {method 'append' of 'list' objects}
  1037312    0.192    0.000    0.192    0.000 {built-in method Translation}
  1037312    0.187    0.000    0.187    0.000 {method 'normalized' of 'Vector' objects}
  1037312    0.166    0.000    0.166    0.000 {built-in method math.radians}
  1037312    0.164    0.000    0.164    0.000 {method 'copy' of 'Matrix' objects}
     8104    0.005    0.000    0.107    0.000 sockets.py:78(other)
     8104    0.023    0.000    0.102    0.000 data_structure.py:943(get_other_socket)
    10130    0.027    0.000    0.095    0.000 bpy_types.py:1008(links)
     5065    0.010    0.000    0.082    0.000 sockets.py:245(needs_data_conversion)
    20260    0.068    0.000    0.068    0.000 bpy_types.py:1012(<genexpr>)
     3039    0.003    0.000    0.048    0.000 sockets.py:250(convert_data)
     1013    0.001    0.000    0.047    0.000 sockets.py:113(sv_set)
     1013    0.039    0.000    0.046    0.000 socket_data.py:70(SvSetSocket)
     1013    0.002    0.000    0.042    0.000 sockets.py:509(sv_get)
     4052    0.020    0.000    0.022    0.000 sockets.py:86(socket_id)
     1013    0.014    0.000    0.018    0.000 data_structure.py:92(match_long_repeat)
     2026    0.007    0.000    0.007    0.000 sockets.py:68(get_prop_name)
     1013    0.004    0.000    0.004    0.000 {method 'extend' of 'list' objects}
     6078    0.003    0.000    0.003    0.000 {built-in method builtins.max}
     1013    0.002    0.000    0.002    0.000 matrix_in_mk3.py:210(process)
     4052    0.002    0.000    0.002    0.000 {built-in method builtins.hash}
     1013    0.001    0.000    0.001    0.000 {built-in method builtins.getattr}
     2026    0.000    0.000    0.000    0.000 node_tree.py:396(does_support_draft_mode)
     2026    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

DolphinDream:
Btw, I also checked to see if the MK3 and MK2 generate the same matrices for the same inputs.. and they do.

DolphinDream:
If someone is willing to review the code to check that everything in kosher, that would be super awesome. ~ gracias

vicdoval:
Check this example.
Two objects with two points each. With the "Flat ouput" we get  one list with 4 Matrixes [M,M,M,M] and when it is dissabled we get two lists with 2 Matrix each [[M, M], [M, M]]
image

DolphinDream:
@vicdoval Btw, the line output seems a bit odd..

([v, v, v],)  .. instead of [[v, v, v]] 

and 

([v, v], [v, v, v],) instead of [[v, v], [v, v, v]]

Durman:
I think it does not make any troubles because sverchok does not modify input data without copying it.

DolphinDream:
So if i understand correctly the way MK2 (with flat output off) mixes the input as such:
(say v is for location vector input, a is for the angle input)
```
[[v, v, v]] + [[a, a, a, a]] => [[m, m, m, m]]
```
and 
```
[[v, v], [v, v, v]] + [[a, a, a, a]] => [[m, m, m, m], [m, m, m, m]]
```
essentially, it does a **match long repeat** between each sub-list in the vector input and the list in the angle input:

first sub-list:
```
[v, v] & [a, a, a, a] <= long match => [v, v, v, v] & [a, a, a, a] => [m, m, m, m]
```
and second sublist:
```
[v, v, v] & [a, a, a, a] <= long match => [v, v, v, v] & [a, a, a, a] => [m, m, m, m]
```
Btw, what happens when you have:
```
[[v, v], [v, v, v]] & [[a, a], [a, a, a], [a, a, a, a]] ?
```
Should this case be handled? how would you do the “matching” / vectorization in this case?
Perhaps a long match of sub-lists? and then a long match of elements in list pairs?

```
[[v, v], [v, v, v]] & [[a, a, a], [a, a], [a, a, a, a]]  
<= long match => 
[[v, v], [v, v, v], [v, v, v]] & [[a, a, a], [a, a], [a, a, a, a]] 
(repeating last vector sub-list)
```
and then matching sub-list pairs
```
[[v, v], [v, v, v], [v, v, v]] & [[a, a, a], [a, a], [a, a, a, a]]
<= long match => 
[[v, v, v], [v, v, v], [v, v, v, v]] & [[a, a, a], [a, a, a], [a, a, a, a]]
=>
[[m, m, m], [m, m, m], [m, m, m, m]]
```

vicdoval:
yes, that how I understand it :)
If you would like to add different matching methods you can check  https://github.com/nortikin/sverchok/issues/2443 
The two steps you mention is what i called Global List Match (for object level) and Local List Match (for sub-object level)
I find 'cycle' method many times useful  to create rhythms 
image

DolphinDream:
Voila !

SV-MatrixInMK3-Example1

vicdoval:
When you are ready please remove the MK3 from the bl_label and add https://github.com/nortikin/sverchok/blob/f53457edc8c0861eb852e0664081490dcf78c2aa/old_nodes/bbox.py#L35 to the old node to wake up the replacement operator :)

DolphinDream:
> When you are ready please remove the MK3 from the bl_label and add
> 
> https://github.com/nortikin/sverchok/blob/f53457edc8c0861eb852e0664081490dcf78c2aa/old_nodes/bbox.py#L35
> 
> to the old node to wake up the replacement operator :)

Once I add the replacement_nodes to the old/node/matrix_in_mk2.py .. how does this replacement work? I reloaded blender with a file having an MK2 node and it shows it with a red depripcated frame.. What do i need to do to replace the MK2 with MK3?

DolphinDream:
@vicdoval btw. The output matrix socket link is not preserved. I wonder if it has something to do with the output socket name having changed from Matrix to Matrices.

vicdoval:
Yes if you change the name of the sockets you need to pass the changes
One example
https://github.com/nortikin/sverchok/blob/f53457edc8c0861eb852e0664081490dcf78c2aa/old_nodes/move_mk2.py#L34
You can read the full implementation with other mechanisms here:
https://github.com/nortikin/sverchok/blob/f53457edc8c0861eb852e0664081490dcf78c2aa/node_tree.py#L550

To add it to the right click menu of the Mk2 you need to add a call in the right-click menu of 
` self.node_replacement_menu(context, layout)
`

DolphinDream:
Can something like this be done to preserve output links considering name was changed from “Matrix” to “Matrices”
```
replacement_nodes = [('SvMatrixInNodeMK3', None, {"Matrix":"Matrices"})]
```

It doesn’t work like this but I wonder whether the is a certain parameter i can pass via this list to have output socket be properly relinked.

DolphinDream:
@vicdoval awesome. Adding the `self.node_replacement_menu(context, layout)` to the MK2 matrix node is kewl.. however I wonder if this could be added automatically to any node that has the replacement_nodes defined so that every node would get that via right-click.

vicdoval:
The menu is added automatically if the node does not have a right-click menu 
https://github.com/nortikin/sverchok/blob/f53457edc8c0861eb852e0664081490dcf78c2aa/node_tree.py#L580-L586
I guess this could change somehow now that we are using it more often :)

DolphinDream:
I think this PR is ready to be merged, unless anybody has any other suggestions/comments. If you guys can give it a quick test and/or review the changes before the merge that would be kewl. Otherwise I am going to merge it soon.

DolphinDream:
Btw, I ran the automated tests (run_tests.sh) locally and everything seems to be fine too.

DolphinDream:
Well.. it seems that changing the example json file’s to replace MK2 with MK3 doesn’t quite work. When the example node-trees are created the links/sockets are not recreated correctly. 
How do I go about updating the examples? Can I instantiate them, replace the node than resave them as json?

vicdoval:
Yes. You have to load the example, replace the node and export it as json in the same place

DolphinDream:
> Yes. You have to load the example, replace the node and export it as json in the same place

How do you export a tree as json? 

> Yes. You have to load the example, replace the node and export it as json in the same place

Thanks. I’ll do that.

One note though.. Looking at the json examples, there is this “update_lists” section which seems to be the links between the nodes, but they reference the sockets by index. I wonder if this link list should be done by listing the sockets names instead, so that if the socket indices change (but the names remain the same) the examples could still work after an update of a node.


e.g. for a simple node setup

SV-MatrixIn-Export


```
{
  "export_version": "0.079",
  "framed_nodes": {},
  "groups": {},
  "nodes": {
    "Line": {
      "bl_idname": "SvLineNodeMK4",
      "color": [
        0.0,
        0.5,
        0.5
      ],
      "height": 100.0,
      "hide": false,
      "label": "",
      "location": [
        -136.45831298828125,
        200.99908447265625
      ],
      "params": {},
      "use_custom_color": true,
      "width": 140.0
    },
    "Matrix In": {
      "bl_idname": "SvMatrixInNodeMK3",
      "height": 100.0,
      "hide": false,
      "label": "",
      "location": [
        -143.63677978515625,
        -12.157699584960938
      ],
      "params": {},
      "width": 140.0
    },
    "Viewer Draw Mk3": {
      "bl_idname": "SvVDExperimental",
      "color": [
        1.0,
        0.30000001192092896,
        0.0
      ],
      "height": 100.0,
      "hide": false,
      "label": "",
      "location": [
        63.54167938232422,
        200.99908447265625
      ],
      "params": {},
      "use_custom_color": true,
      "width": 140.0
    }
  },
  "update_lists": [
    [
      "Line",
      0,
      "Viewer Draw Mk3",
      0
    ],
    [
      "Line",
      1,
      "Viewer Draw Mk3",
      1
    ],
    [
      "Matrix In",
      0,
      "Viewer Draw Mk3",
      3
    ]
  ]
}
```

instead, the link list could perhaps be expressed as:

```
"update_lists": [
    [
      “Line”:  “Verts",
      "Viewer Draw Mk3”: “Verts"
    ],
    [
      “Line”: “Edges",
      "Viewer Draw Mk3”: “Edges"
    ],
    [
      "Matrix In”: “Matrices"
      "Viewer Draw Mk3”: “matrix"
    ]
```

where the lists is essentially

```
{
  [
    “FromNode” : “OutputSocketName”,
    “ToNode” : “InputSocketName”
  ]
}
```

DolphinDream:
Btw, I’m getting an error when exporting the Architecture/TensileCover (even in master branch). Can someone confirm this?

to replicate.. run the example to generate the tree.. Select all nodes in the tree.. in the export on the property panel, enable “Selected Only” and Export.

I wonder if there is a bug in the export module.

SV-ExportError

Durman:
Confirm

<details> 
  <summary>error</summary>  

```
  File "C:\Users\User\YandexDisk\3.╨б╨▓╨╛╨╕_╨┐╤А╨╛╨╡╨║╤В╤Л\Blender_code\addons\sverchok\utils\sv_IO_panel_operators.py",
 line 58, in execute
    layout_dict = create_dict_of_tree(ng)
  File "C:\Users\User\YandexDisk\3.╨б╨▓╨╛╨╕_╨┐╤А╨╛╨╡╨║╤В╤Л\Blender_code\addons\sverchok\utils\sv_IO_panel_tools.py", lin
e 331, in create_dict_of_tree
    get_superficial_props(node_dict, node)
  File "C:\Users\User\YandexDisk\3.╨б╨▓╨╛╨╕_╨┐╤А╨╛╨╡╨║╤В╤Л\Blender_code\addons\sverchok\utils\sv_IO_panel_tools.py", lin
e 144, in get_superficial_props
    node_dict['location'] = node.absolute_location
AttributeError: 'NodeReroute' object has no attribute 'absolute_location'
```

</details>

DolphinDream:
For the time being I reverted the MK2 -> MK3 changes in the examples. The user can still use the examples and given the option to replace outdated nodes. Once the Export functionality is fixed I’ll update the examples to use MK3 node.

vicdoval:
Replace this line 
`    node_dict['location'] = node.absolute_location`

at
https://github.com/DolphinDream/sverchok/blob/e864a2d90256c95f21c0170d3ee1ce3db225d5df/utils/sv_IO_panel_tools.py#L144-L145
with this 
```
    if hasattr(node, 'absolute_location'):
        node_dict['location'] = node.absolute_location
    else:
        node_dict['location'] = node.location[:]
```
Probably not the end solution but i think it can be fine by the time being

DolphinDream:
> Replace this line
> ` node_dict['location'] = node.absolute_location`
> 
> at
> https://github.com/DolphinDream/sverchok/blob/e864a2d90256c95f21c0170d3ee1ce3db225d5df/utils/sv_IO_panel_tools.py#L144-L145
> with this
> 
> ```
>     if hasattr(node, 'absolute_location'):
>         node_dict['location'] = node.absolute_location
>     else:
>         node_dict['location'] = node.location[:]
> ```
> 
> Probably not the end solution but i think it can be fine by the time being

Thanks. This seems to fix the problem.

However, as I run this json example, update deprecated nodes, select all nodes and export selected, it seems that a lot of other nodes are changed, not just the matrix In nodes. Is this expected? Actually if i don’t update nodes at all, but simply do the export of the same node tree i get different json file. I suspect that the export mechanism has changed since these examples were originally exported.

Also, in case of Architecture/Tensile Cover.. when I replace node “Elevated Pins matrix” (a Matrix In MK2) with a MK3 version the location setting doesn't seem to be preserved. Neither is the custom property “expanded” to keep the location value expanded. Any ideas where the problem may be?

SV-MatrixInReplacementBug-Demo1

portnov:
By default, replacement operator copies properties with identical names from old node to the new one. If you want to do something more fancy, you have to override `migrate_from()` method in the new node class.

DolphinDream:
> By default, replacement operator copies properties with identical names from old node to the new one. If you want to do something more fancy, you have to override `migrate_from()` method in the new node class (see Sv common node class definition, there is some documentation).

Cool. Btw, this solves the migration problem, however it seem that the migration code still keeps the old attributes around. e.g. MK2 has l_ as the location, while MK3 has location_ for the same attribute. If i have the migrate_from map the location_ to l_ it does successfully copy the property values however the migrated node will also have the old l_ preserved so when exporting the nodes they will have both location_ and l_ attributes.

DolphinDream:
Any suggestions how to proceed with this PR? Should I update all the current json examples to use the new MK3 node (even though the migration/export keeps the old node’s attributes that no longer exist? (See my previous comment)). Or should I just merge this PR without updating the examples?

vicdoval:
Try replacing this:

https://github.com/nortikin/sverchok/blob/6a5b20ea02a404e461bae047f09d7a6cf0b1ec8b/ui/nodes_replacement.py#L103-L104

with this
```
        for prop_name, prop_value in old_node.items():
            if hasattr(new_node, prop_name):
                new_node[prop_name] = old_node[prop_name]
```

DolphinDream:
Thanks @vicdoval That solves the stale attribute problem. 

One more note though.. it seems that the replacement code saves (when exporting to json) the new node's new attributes even though they have the same values as their default values. Is there a quick way to save/exporet these attributes only if they are different than the default values? e.g. the replacement code (or export code?) would save the MK3’s node “axis” attribute (which replaces the old MK2’s r_ attribute) even though the value of the axis is (0,0,1), which is the same as its default value. Not a big deal, but I thought this points out a deficiency in the replacement code (or perhaps in the export code?)

vicdoval:
I think neither the exporter or the replacer check if a value is the default value but  the default value could change with time so maybe is not that bad to keep it anyway

DolphinDream:
Thanks all for your valuable feedback. If all agree I can go ahead and merge/close this PR.

portnov:
Well, since the tests are fixed, I think you can merge it :)

DolphinDream:
@zeffii  @vicdoval @Durman  @portnov  I’d like to add some of the code from this node to a common place (see below). Where would be the most appropriate location to add this code, so that I can re-use it with other nodes that need this same data?

```
euler_order_items = [
    ('XYZ', "XYZ", "", 0),
    ('XZY', 'XZY', "", 1),
    ('YXZ', 'YXZ', "", 2),
    ('YZX', 'YZX', "", 3),
    ('ZXY', 'ZXY', "", 4),
    ('ZYX', 'ZYX', "", 5)
]

angle_unit_items = [
    ("RAD", "Rad", "Radians", "", 0),
    ("DEG", "Deg", "Degrees", "", 1),
    ("UNI", "Uni", "Unities", "", 2)
]

angle_unit_conversion = {
    "RAD": {"RAD": 1, "DEG": 180/pi, "UNI": 1/(2*pi)},
    "DEG": {"RAD": pi/180, "DEG": 1, "UNI": 1/360},
    "UNI": {"RAD": 2*pi, "DEG": 360, "UNI": 1}
}
```

zeffii:
`sverchok/utils/modules/<useful_name>.py`  ,  `common_lookups.py`  i dunno

Durman:
@DolphinDream You can use Python enums for manipulation with data more flexible.

```Python
from enum import Enum

class AngleUnits(Enum):
    RADIANS = 1
    DEGREES = 2
    UNITIES = 3
```
So it can be used in this way:
```Python
>>> AngleUnits.RADIANS
<AngleUnits.RADIANS: 1>
```
It is possible to add methods to enum items:
```Python
    def get_radians_factor(self):
        if self == AngleUnits.RADIANS:
            return 1
        elif self == AngleUnits.DEGREES:
            return 180/pi
        elif self == AngleUnits.UNITIES:
            return 1/(2*pi)
        else:
            raise Error(f"WTF")

    def get_degrees_factor(self):
        ...

>>> AngleUnits.RADIANS.get_radians_factor()
1
```

Class method can be used for converting Python enum to Blender enum
```Python
    @classmethod
    def get_blender_enum(cls):
        if len(cls) == 3:
            return [
                ("RAD", "Rad", "Radians", "", 0),
                ("DEG", "Deg", "Degrees", "", 1),
                ("UNI", "Uni", "Unities", "", 2)
                ]
        else:
            raise Error(f"WTF")

>>> AngleUnits.get_blender_enum()
[('RAD', 'Rad', 'Radians', '', 0), ('DEG', 'Deg', 'Degrees', '', 1), ('UNI', 'Uni', 'Unities', '', 2)]
```

DolphinDream:
Aside from these lists, which could be shared via an angles_utils.py of sort, what about the UI elements that are typically created using these lists? How can I share this code among nodes? Can I put the props definitions and UI elements drawing also in some common place? If so, where would be a good place to put them?

example:

```
    euler_order: EnumProperty(
        name="Euler Order", description="Order of the Euler rotations",
        default="XYZ", items=euler_order_items, update=updateNode)

    angle_units: EnumProperty(
        name="Angle Units", description="Angle units (radians/degrees/unities)",
        default="DEG", items=angle_unit_items, update=update_angle_units)

    last_angle_units: EnumProperty(
        name="Last Angle Units", description="Angle units (radians/degrees/unities)",
        default="DEG", items=angle_unit_items)

and  ...

        layout.prop(self, "mode", expand=False, text="")
        if self.mode == "EULER":
            col = layout.column(align=True)
            col.prop(self, "euler_order", text="")
        if self.mode in {"EULER", "AXISANGLE"}:
            row = layout.row(align=True)
            row.prop(self, "angle_units", expand=True)
```

zeffii:
@DolphinDream  for the viewer nodes  i do this repeating ui / prop definitions by declaring a `mixin` in `sv_obj_helper.py ` and importing it by forexample in bmeshviewer

DolphinDream:
@zeffii Is it fair to say that if i want a node wants to have (share) some “angle features” I could add these features to a separate base class, then have the node inherit from that class ? Such base class would not necessarily be the “object helper” class you mentioned earlier, but a new class, since the object helper class provides a lot more than the “angle feature” capability a node would want.

zeffii:
i was not clear.  yes you make your own mixin class. (a base class) . i'm suggesting to look only at sv_object_helper to see the extent multiple nodes are already sharing mixin code to keep them simpler / less repetitive class code individually.

zeffii:
maybe `sverchok.utils.sv_transforms_helper.py` would be a good place to stick the `items` and `lists` and `mixin` for reuse.  (better than in `modules`, and consistent with ``sv_objects_helper`)

DolphinDream:
To get this straight @zeffii  you’re suggesting the more generic name/file sverchok.utils.sv_transforms_helper.py as a place to put any further transform helper classes (not just my angle stuff)? Perhaps create SvAngleHelper class to handle my angle related stuff.. and add other SvSomethingElseHelper classes for other transform related stuff?

zeffii:
bingo :)

zeffii:
i like `SvTranformHelper` if your doing a mixing with a lot of things, and maybe `SvAngleHelper` mixin if you expect several nodes to only have a subset of what might be present in `SvTransformHelper`..

get something going, it will be obvious quite fast whether the approach is clever or convoluted.

DolphinDream:
Why is it that globally defined lists are initialized but dictionaries are not?

e..g if I have a list and a dict in the helper file:

```
myList [1,2,3]
myDict=  {“1”:2, “2”:2}

def myClass:
  def access_list():
        return myList[0]

  def access_dict():
       return myDict[“1”]
```

a call to the access_list from another node that imports myClass works fine but the call to access_dict doesn’t.

zeffii:
what other programming languages do you write @DolphinDream  ?

DolphinDream:
C++

PR #2815, opened 2020-01-12
Line node vectorizing

Durman:
## Addressed problem description

#2649 #2758

## Preflight checklist

- [x] Code changes complete.
- [x] Code documentation complete.
- [x] Documentation for users complete.
- [x] Manual testing done. 
- [x] Ready for merge.

PR #2995, opened 2020-03-23
Add shift parameter to TorusKnot node

DolphinDream:
Add shift parameter to allow points to be shifted along the curve

Shift values are 0-1, where 0 = no shift, and 1 = a complete shift of the points along the curve.

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


DolphinDream:
Btw, should this update be an MK2 node? (one extra setting parameter / input socket was added)

Durman:
Adding new sockets destroy work of existing layouts with such nodes.

DolphinDream:
> Adding new sockets destroy work of existing layouts with such nodes.

That’s what I thought.. just confirming. So, it’s a MK2 then.
What exactly is the process for making a MK2? Do I need to move the old node to some retired location or should i just add the new node files without moving the old files?

Durman:
File with old version of node should be moved to `old_nodes` directory.
Label of new node should not point to version of the node.
All other things which should be changed will be pointed by passing test of this PR.
You can use this PR as template: #2815 probably I have missed something.

zeffii:
@DolphinDream  you should be trying to merge into "master", not `b28_prelease_master`

DolphinDream:
> @DolphinDream you should be trying to merge into "master", not `b28_prelease_master`

Is **master** now the Blender 2.8x stable branch?

zeffii:
2.83+ even.

DolphinDream:
How do I implement a node replacement feature so that old blends using old torus knot version can easily be replaced and reconnected with the new node version? any examples I can follow?

Durman:
I think you have to rename file with documentation to `torus_knot_mk2.py` as well.

DolphinDream:
> I think you have to rename file with documentation to `torus_knot_mk2.py` as well.

fixed. and btw.. we don’t need to keep the old node documentation, correct?

zeffii:
> we don’t need to keep the old node documentation, correct?

correct: https://github.com/nortikin/sverchok/pull/2998

zeffii:
if someone really wants to read docs of the old node there's always the git history

vicdoval:
Something like this has to be added to the old node to use the replacement operator https://github.com/nortikin/sverchok/blob/ed61384f3024e73ee0930c6518ab6791f88324d1/old_nodes/box.py#L38

DolphinDream:
A live demo of the shift setting.

SV-TorusKnotNodeMK2-Demo1

PR #3004, opened 2020-03-27
Obj boolean internal

zeffii:
hahah.. i don't know..

image

### DIFF
--- a/index.md
+++ b/index.md
@@ -417,6 +417,7 @@
     SvBMOpsNodeMK2
     ---
     SvCSGBooleanNodeMK2
+    SvBlenderBoolean
     SvNumpyArrayNode
     SvParticlesMK2Node
     SvJoinTrianglesNode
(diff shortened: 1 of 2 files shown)


Durman:
Probably it would be possible to create series of `set modifier` nodes and `set boolean modifier` could be one of them.

portnov:
@Durman , the problem is, you will have to write UI and write some code for each modifier. In fact, this is what Sorcar consists of.

Durman:
Well then should we have such nodes? It is possible just adding modifiers via standard Blender UI. 
Or creating such nodes can gain some benefits?

Durman:
`Sorcar` code of `boolean modifier` node:
https://github.com/aachman98/Sorcar/blob/8cd2ed94902a7b16075c5ca3c5ecc447565ef356/nodes/modifiers/ScBooleanMod.py#L1-L39

zeffii:
i'm geared up to add only a few modifier nodes.
- array
- screw
- 
..even if it's just to see how useful that would be.

zeffii:
image

it's a bit wonky.

zeffii:
yeah, i think i'll not do this. People who want this kind of stuff should know that a lot is possible
image

PR #3014, opened 2020-03-30
add listlength to quicknodes

zeffii:
add to quicknodes


zeffii:
list length is possibly one of the most frequently used nodes.

PR #3015, opened 2020-03-30
use import anyway

zeffii:
a bit more defensive fix


zeffii:
https://gist.github.com/96728c4d8dcf41aafcbfa3d78728ec1c

PR #3018, opened 2020-03-30
Multiple documentations 2

zeffii:
many nodes don't have docs.  This PR is an attempt to provide at least a rudimentary documentation for the following nodes.  I'm going for the low hanging fruit first.

many nodes..


zeffii:
FEWER TRIS
image

PR #3021, opened 2020-03-31
Selection grabber lite

zeffii:
Lite version, why so complicated.

### DIFF
--- a/index.md
+++ b/index.md
@@ -364,6 +364,7 @@
     SvDupliInstancesMK4
     SvFCurveInNodeMK1
     SvCollectionPicker
+    SvSelectionGrabberLite
 
 ## Objects
     SvVertexGroupNodeMK2
--- a/nodes/scene/selection_grabber_lite.py
+++ b/nodes/scene/selection_grabber_lite.py
@@ -6,6 +6,8 @@
 # License-Filename: LICENSE
 
 import bpy
+import bmesh
+
 from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode
 
@@ -49,7 +51,7 @@ class SvSelectionGrabberLite(bpy.types.Node, SverchCustomTreeNode):
 
         for obj in objects:
             if obj == bpy.context.edit_object:
-                bm = bmesh.from_edit_mesh(mesh)
+                bm = bmesh.from_edit_mesh(obj.data)
                 if self.include_faces:
                     face_mask.append([f.select for f in bm.faces])
                 if self.include_edges:
@@ -57,6 +59,7 @@ class SvSelectionGrabberLite(bpy.types.Node, SverchCustomTreeNode):
                 if self.include_vertex:
                     vertex_mask.append([e.select for e in bm.verts])
             else:
+                mesh = obj.data
                 if self.include_faces:
                     face_mask.append([f.select for f in mesh.polygons])
                 if self.include_edges:
--- a/nodes/scene/selection_grabber_lite.py
+++ b/nodes/scene/selection_grabber_lite.py
@@ -1,0 +1,73 @@
+# This file is part of project Sverchok. It's copyrighted by the contributors
+# recorded in the version control history of the file, available from
+# its original location https://github.com/nortikin/sverchok/commit/master
+#  
+# SPDX-License-Identifier: GPL3
+# License-Filename: LICENSE
+
+import bpy
+from sverchok.node_tree import SverchCustomTreeNode
+from sverchok.data_structure import updateNode
+
+class SvSelectionGrabberLite(bpy.types.Node, SverchCustomTreeNode):
+    
+    """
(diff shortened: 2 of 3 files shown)


enzyme69:
Zeffii is this how you propose to use this node? Every single time there is a masking needed?

Screen Shot 2020-04-01 at 1 19 26 pm

Durman:
I do not see purpose of having of bool properties in current implementation.

zeffii:
exactly right @enzyme69 . When you want to mix "purely procedural" with some elements of "manual" work, then this is a solution that works right now. 

@Durman , but then an index_to_mask node is needed... if if someone wants the indices, they can use mask_to_index..

zeffii:
@enzyme69   but I recomment swithing the `Live` of the `BMV Alpha` node .  off.

Durman:
I meant `include_vertex`, `include_edges`, `include_faces` properties.

zeffii:
ok.. it could be an `enum`, and that node could have one socket output..  but how much does it really matter..

Durman:
I think the node can be without any properties at all. Now those properties do not have any sense. The simpler UI the better.

zeffii:
@Durman   like this?  https://github.com/nortikin/sverchok/pull/3024

Durman:
Yes, that what I meant.

zeffii:
i might add a button to store the current set in `self.node_dict[hash(self)] = ...`

enzyme69:
After some exploration, @zeffii this might be quite something, because it's somewhat SEMI working with UI and normally it tooks me quite a long time to make this, but using Viewer, and baking "intermediate mesh" (with selection still on) allow me to make this

Screen Shot 2020-04-01 at 10 21 28 pm

enzyme69:
I just realized that if I make changes to 1st selection, then 2nd, 3rd etc will change :) which means some kind of special care needed for this "dynamic masking".

enzyme69:
Screen Shot 2020-04-01 at 11 12 47 pm

I will make video about this, it's going to be a weird one. 

The previous "intermediate mesh" that stores the masking will grow and it keeps history, however, there is no correlation between them however, so if I make changes to any selection, it will suddenly change others.

Especially obvious if I select all of those intermediate, hit Edit, and select all or deselect all ...

zeffii:
you will always find sverchok lacking when it comes to selections based on manual interaction, bmesh face/edge index instability (as a direct result of using bmesh and bmesh.ops from Blender) plays a role in this. You might not always notice "weird indexing" but you will encounter it, and if you look at the code of the node involved you'll see it relies on bmesh for something.

enzyme69:
@zeffii Thanks for that extra info Zeffi. I know this is probably not perfect but I think what you add here can open up ideas. It's a bit like building a giant tower using cards, I supposed. We can't go back but at least this visualize and letting user grab edit the component while figuring out where to go next :)

zeffii:
also..   if you are just picking one index... use indexviewer and then feed it manually.
image

Durman:
> i might add a button to store the current set in `self.node_dict[hash(self)] = ...`

Can you explain what difference will be between this node and this #2465 after that?

Durman:
> you will always find sverchok lacking when it comes to selections based on manual interaction, bmesh face/edge index instability (as a direct result of using bmesh and bmesh.ops from Blender) plays a role in this. You might not always notice "weird indexing" but you will encounter it, and if you look at the code of the node involved you'll see it relies on bmesh for something.

There is no trouble with indexing of bmesh data there is trouble with understanding of such data structure.

zeffii:
@Durman i haven't noticed unstable indices lately either. so maybe.   but I would not count on it..

as howardt put it here:  https://devtalk.blender.org/t/bmesh-adding-new-verts/11108/3

>  I don’t regard this as a bug since there was never any guarantee about what happens to vertex indices as you do operations. 

for us, unfortunately as scripters, stability is super important..

enzyme69:
I wonder how "Sorcar" addon does it... It might be more like Macro of procedures within a Loop.

Let say we start with a grid of 3x3 polygon face.

We select a center face to start with. Extruding it and make a tree somewhat.

Then I suddenly want to have 2 or 3 more face selected from the original 3x3 grid. But the tree does not automatically traverse through for reason that the node dependant on the index.

enzyme69:
Maybe "propagate" is something that I was looking for. So selection can be propagated based on topology... 
https://docs.blender.org/manual/uk/dev/modeling/meshes/editing/vertex/propagate_shapes.html

Durman:
@zeffii Vertices, edges and faces can keep custom attributes. It is possible to create such attribute for indexes and when work with bmesh is finish to rebuild list of indexes of elements. This is the way how `mesh filter` node work for example.

zeffii:
> Can you explain what difference will be between this node and this #2465 after that?

yeah, the Lite version is coded to be as simple as possible.

PR #3034, opened 2020-04-02
Namus robustus

zeffii:
removing confusing implementations

### DIFF
--- a/nodes/scene/instancer.py
+++ b/nodes/scene/instancer.py
@@ -16,11 +16,6 @@ from bpy.props import BoolProperty, FloatVectorProperty, StringProperty, EnumPro
 
 from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import dataCorrect, updateNode
-from sverchok.utils.sv_viewer_utils import greek_alphabet
-
-def get_random_init():
-    return random.choice(greek_alphabet)
-
 
 def make_or_update_instance(node, obj_name, matrix):
     context = bpy.context
--- a/nodes/scene/instancer_MK2.py
+++ b/nodes/scene/instancer_MK2.py
@@ -6,7 +6,6 @@
 # License-Filename: LICENSE
 
 import itertools
-import random
 
 import bpy
 import bmesh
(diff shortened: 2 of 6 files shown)


zeffii:
no

zeffii:
for the time being if you want sorting ability, you need to pick names yourself.

zeffii:
just to be clear here, it is possible to do something different when a node is copied (cloned) than when it is just picked from the new node menu.  But what?

if cloned..  (psuedocode)
```
def up_one(name):
    name = (name += 1)

def new_name(current_name):
    if current_name.endswith(regex="Copy{dd}":
        new_name = up_one(current_name)
    else:
        new_name = current_name + "Copy00"

```

zeffii:
ie..
```python
import re

bpy = lambda: None
bpy.data = lambda: None
bpy.data.objects = ['name', 'nameCopy04', 'nameCopy05', 'AlphaCopy0003']

def make_name_suggestion(current_name):
    m = re.search("Copy(?P<postfix>\d+)$", current_name)
    if m:
        postfix = m.group('postfix')
        padding = len(postfix)
        slice_size = padding
        num = int(postfix)
        new_num = str(num + 1)
        str_postfix = new_num.zfill(padding)
        return current_name[:-slice_size] + str_postfix
    else:
        return current_name + "Copy00"

for name in bpy.data.objects:
    result = make_name_suggestion(name)
    print(name, '->', result)
```
becomes
```
name -> nameCopy00
nameCopy04 -> nameCopy05
nameCopy05 -> nameCopy06
AlphaCopy0003 -> AlphaCopy0004
```

enzyme69:
Yes I think I prefer this convention, thanks Zef~ Like if I start with Blah, I am hoping the next BMeshViewer to be Blah.000., Blah.001, and so on.

zeffii:
not 
```
Blah, Blah.000, Blah.001
```
bmeshviewer and other viewer nodes keep track of their objects, and to avoid confusion the "copy" name should be a bit more obvious for me. It's not the same as simply duplicating a Mesh Object in 3dView. At some point there's just a choice being made between what is easiest for me to get my head around.
```
Blah
BlahCopy00
BlahCopy01
BlahCopy02
BlahCopy99
```
each of those is separate Viewer, and each create `.0001` `.0002`  etc... as its sub objects.

PR #3037, opened 2020-04-04
Factorizing the angle units code into a helper class

DolphinDream:
Factorizing the angle unit / conversion code into a helper class so that nodes that use angles that want to select the angle units (RAD, DEG, UNI) can have a way to create the the angle unit UI elements and conversions performed. 

Note: this PR will focus on the SvAngleHelper class/file but for the sake of illustrating how this is going to be used by various nodes that could adopt it I started to integrate it into some nodes (the changes to these nodes could be added in a separate PR) .

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/utils/sv_transform_helper.py
+++ b/utils/sv_transform_helper.py
@@ -1,0 +1,146 @@
+# ##### BEGIN GPL LICENSE BLOCK #####
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software Foundation,
+#  Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# ##### END GPL LICENSE BLOCK #####
+
+import bpy
+from bpy.props import (EnumProperty, BoolProperty)
+
+from sverchok.data_structure import updateNode
+from sverchok.settings import get_params
+
+from math import pi
+
+euler_order_items = [
+    ('XYZ', 'XYZ', "", 0),
+    ('XZY', 'XZY', "", 1),
+    ('YXZ', 'YXZ', "", 2),
+    ('YZX', 'YZX', "", 3),
+    ('ZXY', 'ZXY', "", 4),
+    ('ZYX', 'ZYX', "", 5)
+]
+
+
+class AngleUnits:
+    RADIANS = "RAD"
+    DEGREES = "DEG"
+    UNITIES = "UNI"
+
+    @classmethod
+    def get_blender_enum(cls):
+        return [
+            (AngleUnits.RADIANS, "Rad", "Radians", "", 0),
+            (AngleUnits.DEGREES, "Deg", "Degrees", "", 1),
+            (AngleUnits.UNITIES, "Uni", "Unities", "", 2)
+        ]
+
+
+angle_remap_options = {
+    # from               to
+    (AngleUnits.RADIANS, AngleUnits.DEGREES): 180/pi,
+    (AngleUnits.RADIANS, AngleUnits.UNITIES): 1/(2*pi),
+    (AngleUnits.DEGREES, AngleUnits.RADIANS): pi/180,
+    (AngleUnits.DEGREES, AngleUnits.UNITIES): 1/360,
+    (AngleUnits.UNITIES, AngleUnits.RADIANS): 2*pi,
+    (AngleUnits.UNITIES, AngleUnits.DEGREES): 360
+}
+
+
+class SvAngleHelper():
+
+    def get_preferences(self):
+        props = get_params({
+            'auto_update_angle_values': False,
+        })
+        return props.auto_update_angle_values
+
+    def angle_conversion_factor(self, from_angle_units, to_angle_units):
+        if from_angle_units == to_angle_units:
+            return 1
+
+        multiplier = angle_remap_options.get((from_angle_units, to_angle_units))
+
+        if not multiplier:
+            raise Error(f"node {self.name} failure to map angle units, {from_angle_units}->{to_angle_units}")
+
+        return multiplier
+
+    def radians_conversion_factor(self):
+        return self.angle_conversion_factor(self.angle_units, AngleUnits.RADIANS)
+
+    def degrees_conversion_factor(self):
+        return self.angle_conversion_factor(self.angle_units, AngleUnits.DEGREES)
+
+    def unities_conversion_factor(self):
+        return self.angle_conversion_factor(self.angle_units, AngleUnits.UNITIES)
+
+    def update_angle(self, context):
+        ''' Wrapper to inhibit angle updates when units are changed '''
+        if self.inhibit_updates:
+            return
+
+        updateNode(self, context)
+
+    def update_angles(self, context, acf):
+        ''' Override this in the derived class to update specific angle values'''
+        # print("SvAngleHelper update_angles called")
+
+    def update_angle_units(self, context):
+        ''' Update all the angles to preserve their values in the new units '''
+        if self.angle_units == self.last_angle_units:
+            return
+
+        auto_update_angle_values = self.get_preferences()
+
+        if auto_update_angle_values:
+            acf = self.angle_conversion_factor(self.last_angle_units, self.angle_units)
+
+            self.inhibit_updates = True  # deactivate updates
+            self.update_angles(context, acf)
+            self.inhibit_updates = False  # reactivate updates
+
+        self.last_angle_units = self.angle_units  # keep track of the last units
+
+        updateNode(self, context)
+
+    def angle_unit_conversion_factor(self, new_angle_units):
+        return self.angle_conversion_factor(self.angle_units, new_angle_units)
+
+    euler_order: EnumProperty(
+        name="Euler Order", description="Order of the Euler rotations",
+        default="XYZ", items=euler_order_items, update=updateNode)
+
+    angle_units: EnumProperty(
+        name="Angle Units", description="Angle units (Radians/Degrees/Unities)",
+        default=AngleUnits.DEGREES, items=AngleUnits.get_blender_enum(),
+        update=update_angle_units)
+
+    last_angle_units: EnumProperty(
+        name="Last Angle Units", description="Angle units (Radians/Degrees/Unities)",
+        default=AngleUnits.DEGREES, items=AngleUnits.get_blender_enum())
+
+    inhibit_updates: BoolProperty(
+        name='Inhibit Update', description='Flag to inhibit update calls', default=False)
+
+    def draw_angle_euler_buttons(self, context, layout):
+        col = layout.column(align=True)
+        col.prop(self, "euler_order", text="")
+
+    def draw_angle_units_buttons(self, context, layout):
+        box = layout.box()
+        box.label(text="Angle Units")
+        row = box.row(align=True)
+        row.prop(self, "angle_units", expand=True)
--- a/settings.py
+++ b/settings.py
@@ -66,7 +66,7 @@ class SverchokPreferences(AddonPreferences):
             color_def.apply_theme()
 
     tab_modes = data_structure.enum_item_4(["General", "Node Defaults", "Theme"])
-    
+
     selected_tab: bpy.props.EnumProperty(
         items=tab_modes,
         description="pick viewing mode",
@@ -195,7 +195,7 @@ class SverchokPreferences(AddonPreferences):
 
     over_sized_buttons: BoolProperty(
         default=False, name="Big buttons", description="Very big buttons")
-    
+
     node_panel_modes = [
             ("X", "Do not show", "Do not show node buttons", 0),
             ("T", "T panel", "Show node buttons under the T panel", 1),
@@ -207,7 +207,7 @@ class SverchokPreferences(AddonPreferences):
         name = "Display node buttons",
         description = "Where to show node insertion buttons. Restart Blender to apply changes.",
         default = "X")
-    
+
     node_panels_icons_only : BoolProperty(
             name = "Display icons only",
             description = "Show node icon only when icon has an icon, otherwise show it's name",
@@ -235,9 +235,13 @@ class SverchokPreferences(AddonPreferences):
 
     stethoscope_view_scale: FloatProperty(
         default=1.0, min=0.01, step=0.01, description='default stethoscope scale')
-    
+
     index_viewer_scale: FloatProperty(
         default=1.0, min=0.01, step=0.01, description='default index viewer scale')
+
+    auto_update_angle_values: BoolProperty(
+        default=True,
+        description="Auto update angle values when angle units are changed to preserve the angle")
 
     def set_nodeview_render_params(self, context):
         # i think these are both the same..
@@ -262,7 +266,7 @@ class SverchokPreferences(AddonPreferences):
     def update_log_level(self, context):
         logging.info("Setting log level to %s", self.log_level)
         logging.setLevel(self.log_level)
-    
+
     log_levels = [
             ("DEBUG", "Debug", "Debug output", 0),
             ("INFO", "Information", "Informational output", 1),
@@ -367,18 +371,23 @@ class SverchokPreferences(AddonPreferences):
             row_sub1 = col.row().split(factor=0.5)
             box_sub1 = row_sub1.box()
             box_sub1_col = box_sub1.column(align=True)
-            
+
             box_sub1_col.label(text='Render Scale & Location')
             # box_sub1_col.prop(self, 'render_location_xy_multiplier', text='xy multiplier')
             # box_sub1_col.prop(self, 'render_scale', text='scale')
             box_sub1_col.label(text=f'xy multiplier: {self.render_location_xy_multiplier}')
             box_sub1_col.label(text=f'render_scale : {self.render_scale}')
-            
+
             box_sub1_col.label(text='Stethoscope')
             box_sub1_col.prop(self, 'stethoscope_view_scale', text='scale')
 
             box_sub1_col.label(text='Index Viewer')
-            box_sub1_col.prop(self, 'index_viewer_scale', text='scale')           
+            box_sub1_col.prop(self, 'index_viewer_scale', text='scale')
+
+            box_sub2 = box_sub1.box()
+            box_sub2_col = box_sub2.column(align=True)
+            box_sub2_col.label(text='Angle Units')
+            box_sub2_col.prop(self, 'auto_update_angle_values', text="Auto Update Angle Values")
 
             col3 = row_sub1.split().column()
             col3.label(text='Location of custom defaults')


DolphinDream:
@zeffii how does this look so far?

DolphinDream:
Demo with some nodes using the angle helper for switching the angle units

SV-AngleHelper-demo1

DolphinDream:
Another demo.. torus knot, matrix in, ellipse, torus node all getting the angle units facelift :)

Evaluate 0-1 drives the full cycle spins in all nodes taking angle as input (in UNI units)

SV-AngleHelper-demo2

Durman:
I think this options should be on N panel only.

utils/sv_transform_helper.py
```diff
@@ -0,0 +1,102 @@
...
+    def update_angle_units(self, context):
+        ''' Update all the angles to preserve their values in the new units '''
+
+        auc = angle_unit_conversion[0][self.last_angle_units][self.angle_units]
+
+        self.last_angle_units = self.angle_units  # keep track of the last units
+
+        self.syncing = True  # deactivate updates
+        self.update_angles(context, auc)
+        self.syncing = False  # reactivate updates
```
Durman, line 72: It looks like here throttling should be used.

utils/sv_transform_helper.py
```diff
@@ -0,0 +1,102 @@
...
+    ("RAD", "Rad", "Radians", "", 0),
+    ("DEG", "Deg", "Degrees", "", 1),
+    ("UNI", "Uni", "Unities", "", 2)
+]
+
+angle_unit_conversion = [{
+    "RAD": {"RAD": 1, "DEG": 180/pi, "UNI": 1/(2*pi)},
+    "DEG": {"RAD": pi/180, "DEG": 1, "UNI": 1/360},
+    "UNI": {"RAD": 2*pi, "DEG": 360, "UNI": 1}
+}]
```
Durman, line 45: Why list?

utils/sv_transform_helper.py
```diff
@@ -0,0 +1,102 @@
...
+        # print("SvAngleHelper update_angle called")
+
+        if self.syncing:
+            return
+
+        updateNode(self, context)
+
+    def update_angles(self, context, auc):
+        ''' Override this in derived class with whatever angles need to be updated'''
+        print("SvAngleHelper update_angles called")
```
Durman, line 61: Raising error will be more safe I think. Or ABC module should be used.

Durman:
@DolphinDream You can use Python enums for manipulation with data more flexible.

```Python
from enum import Enum

class AngleUnits(Enum):
    RADIANS = 1
    DEGREES = 2
    UNITIES = 3
```
So it can be used in this way:
```Python
>>> AngleUnits.RADIANS
<AngleUnits.RADIANS: 1>
```
It is possible to add methods to enum items:
```Python
    def get_radians_factor(self):
        if self == AngleUnits.RADIANS:
            return 1
        elif self == AngleUnits.DEGREES:
            return 180/pi
        elif self == AngleUnits.UNITIES:
            return 1/(2*pi)
        else:
            raise Error(f"WTF")

    def get_degrees_factor(self):
        ...

>>> AngleUnits.RADIANS.get_radians_factor()
1
```

Class method can be used for converting Python enum to Blender enum
```Python
    @classmethod
    def get_blender_enum(cls):
        if len(cls) == 3:
            return [
                ("RAD", "Rad", "Radians", "", 0),
                ("DEG", "Deg", "Degrees", "", 1),
                ("UNI", "Uni", "Unities", "", 2)
                ]
        else:
            raise Error(f"WTF")

>>> AngleUnits.get_blender_enum()
[('RAD', 'Rad', 'Radians', '', 0), ('DEG', 'Deg', 'Degrees', '', 1), ('UNI', 'Uni', 'Unities', '', 2)]
```

Durman:
You should initialize your module in `utils.__init__.py` file.

utils/sv_transform_helper.py
DolphinDream: Raising error instead of what? The print here was just for debugging and it will be removed.

utils/sv_transform_helper.py
DolphinDream: because if i  just define it as dict.. it give the error that the dict is not defined. (see my comment in issue #2991)

utils/sv_transform_helper.py
DolphinDream: I don’t want the change of value of the angle from one unit to another to trigger a bunch of update calls (for each property that is being changed). 

What do you mean by throttling btw?

DolphinDream:
> I think this options should be on N panel only.

I thought about it. Some nodes already have it on the node itself so I was going to be consistent across nodes. Putting it on the ppanel i think it would make it less likely people would find it. I’ve seen noding tutorials where people didn’t even know that some features were on the ppanel and tried to compensate for these features by adding extra nodes. This adds only extra row in the node anyway which may not be as much of a big deal. 

Any reason why you feel this should go on the ppanel?

DolphinDream:
@Durman  I originally implemented the angle conversions like you suggested with if/else statements.. but decided that having a hash/dictionary of these precomputed values without having the 3x3 if/else code stack was going to be not only shorter code, but possibly slightly faster.

Durman:
> Any reason why you feel this should go on the ppanel?

It looks like this property is kind of properties which should be set once when the node is created and then staying without changes.

> the 3x3 if/else code stack was going to be not only shorter code, but possibly slightly faster.

First of all programming is about creating readable code. It is better to make code even longer if it will be more clear what is going on in a code. Performance is on the second place but in your case it even not the way of performance improving.

I think this should be more readable and simpler to reuse.

```Python
angle_unit_conversion[0][from_angle][to_angle]

DEGREES.get_radians_factor()
```

utils/sv_transform_helper.py
Durman: Instead of print. What will happen if the method will be forgotten to override?

utils/sv_transform_helper.py
Durman: It is strange. I had never faced with such behaviour.

utils/sv_transform_helper.py
Durman: Have a look this code:
https://github.com/nortikin/sverchok/blob/b371af4eef3773d0c3132616b6d84591f6ba51ca/node_tree.py#L86

nodes/matrix/matrix_in_mk3.py
```diff
@@ -82,45 +61,17 @@ def update_mode(self, context):
...
-        updateNode(self, context)
-
-    def update_angle(self, context):
-        ''' Wrapper to suppress angle updates when units are changed '''
-        if self.syncing:
-            return
-
-        updateNode(self, context)
 
     mode: EnumProperty(
```
zeffii, line 71: would you not stick this inside the helper class, but use a more descriptive property name other than `mode`..maybe `axis_mode` ... hopefully it won't be too painful

DolphinDream:
@Durman Good points. If no one else has any objections to making the angle units a PP citizen I’m gonna go a head and move this to PP in all nodes. Btw, for euler angle nodes.. would you say this property should also go on the PP ? (XYZ, XZY,...) ? (and expect the user to know that the default is XYZ). Same for the angle, the user should now the default is RAD.

DolphinDream:
@Durman mind giving me some tips.. how to tie the AngleUnits enum class into the blender UI enum property? When user selects the angle unit (the UI enum prop).. how do I get to the AngleUnits conversions? In the node I typically have self.angle_units (which is the current selection of the UI enum property.. and for the current selection i want to get the conversion factor to say radians. Would I have to do an if/else the self.angle_units (UI prop) is say “DEG” .. then set a AngleUnits variable to be AngleUnits.DEGREES  (self.au = AngleUnits.DEGREES), which then I can access the conversions through as self.au.get_radians_factor() ? If so, it would seem that I need to add a if/elif/else to set this self.au (the new enum class) based on the self.angle_units (the UI prop).

DolphinDream:
Btw.. my limited python skills are put to the test here. Can a node have an attribute that is not a blender property? If I want my base class SvAngleHelper to store the proposed AngleUnits (enum) class as an attribute,  then 1) how do i instantiate this in the class.. and 2). will blender remember this attribute ? (i thought all attributes have to be one of the blender properties) to be registered or something.

DolphinDream:
It would appear I would have to do something like this to just find out the conversion from the last to the current? In which case since I have to do this if/else stack why defining another AngleUnits enum class at all when I could just return thee conversion constant directly in this code (below) @Durman Can you see a simpler way ?

```
        # instead of doing this:
        # auc = angle_unit_conversion[0][self.last_angle_units][self.angle_units]

        # would need to do this?
        if self.lastAngleUnits == AngleUnits.RADIANS:
            if self.angleUnits == AngleUnits.RADIANS:
                auc = AngleUnits.RADIANS.get_radians_factor()
            elif self.angleUnits == AngleUnits.DEGREES:
                auc = AngleUnits.DEGREES.get_radians_factor()
            elif self.angleUnits == AngleUnits.UNITIES:
                auc = AngleUnits.UNITIESS.get_radians_factor()

        elif self.lastAngleUnits == AngleUnits.DEGREES:
            if self.angleUnits == AngleUnits.RADIANS:
                auc = AngleUnits.RADIANS.get_degrees_factor()
            elif self.angleUnits == AngleUnits.DEGREES:
                auc = AngleUnits.DEGREES.get_degrees_factor()
            elif self.angleUnits == AngleUnits.UNITIES:
                auc = AngleUnits.UNITIESS.get_degrees_factor()

        elif self.lastAngleUnits == AngleUnits.UNITIES:
            if self.angleUnits == AngleUnits.RADIANS:
                auc = AngleUnits.RADIANS.get_unitiess_factor()
            elif self.angleUnits == AngleUnits.DEGREES:
                auc = AngleUnits.DEGREES.get_unitiess_factor()
            elif self.angleUnits == AngleUnits.UNITIES:
                auc = AngleUnits.UNITIESS.get_unitiess_factor()
```

zeffii:
if you write this
```
        # instead of doing this (like i have it currently):
        # auc = angle_unit_conversion[0][self.last_angle_units][self.angle_units]

        # would need to do this (based on the proposed AngleUnits enum class)?
        if self.lastAngleUnits == AngleUnits.RADIANS:
            if self.angleUnits == AngleUnits.RADIANS:
                auc = AngleUnits.RADIANS.get_radians_factor()
            elif self.angleUnits == AngleUnits.DEGREES:
                auc = AngleUnits.DEGREES.get_radians_factor()
            elif self.angleUnits == AngleUnits.UNITIES:
                auc = AngleUnits.UNITIESS.get_radians_factor()

        elif self.lastAngleUnits == AngleUnits.DEGREES:
            if self.angleUnits == AngleUnits.RADIANS:
                auc = AngleUnits.RADIANS.get_degrees_factor()
            elif self.angleUnits == AngleUnits.DEGREES:
                auc = AngleUnits.DEGREES.get_degrees_factor()
            elif self.angleUnits == AngleUnits.UNITIES:
                auc = AngleUnits.UNITIESS.get_degrees_factor()

        elif self.lastAngleUnits == AngleUnits.UNITIES:
            if self.angleUnits == AngleUnits.RADIANS:
                auc = AngleUnits.RADIANS.get_unitiess_factor()
            elif self.angleUnits == AngleUnits.DEGREES:
                auc = AngleUnits.DEGREES.get_unitiess_factor()
            elif self.angleUnits == AngleUnits.UNITIES:
                auc = AngleUnits.UNITIESS.get_unitiess_factor()
```

i will shoot you.

zeffii:
ofcourse the remap way at the top of that snippet is closer to what you want to end up with. 
btw you are being very kind to the users to auto-convert during the switch :)  nice touch.

btw, i'm not sure what you mean by having a list initialized but a dict not,  in my experience it is the dicts that are more "persistent".

one think you can do in you `sv_transforms_helper`  is a faux container, just as a test.

before you read this.   i know it doesn't make sense.. it's just a syntax trick.
```
angle_enums = lambda: None
angle_enums.some_list = [.... fill this..]
angle_enums.some_ldict = {....fill this...}
```
and in your class you retreive their content
```

         angle_enums.some_list[index]


         angle_enums.some_dict[some_key]
```

i noticed your mac automatically places up and down quotes in your strings. it might be a suggestion to use single quotation marks in the keys.  or use ints as keys

zeffii:
sorry btw, my mother is in hospital (not nCov related.. )   so i'm having a hard time concentrating.

but i still want to hep, i think you're getting close. narrow down you question please.

utils/sv_transform_helper.py
zeffii: yes, take your time with this @DolphinDream   i'm not sure I follow your description of the problem.

DolphinDream:
@zeffii sorry to hear that. Hope she’ll get better soon! 🙏 
Don’t worry about this. I’m glad you are willing to help. I’ll try to be more specific in my questions.

DolphinDream:
Here’s an update.. see the last commit (ignore the debugging code). I’m struggling to understand why the attribute I set in the bases class do can not retain their updated values (they only keep their initial value during the instantiation). I remember this may have something to do with blender doing some tricks with attributes forcing only the predefined props to be persistent.

```
    # in the class I instantiate these as
    angleUnits = AngleUnits(AngleUnits.DEGREES)
    lastAngleUnits = AngleUnits(AngleUnits.DEGREES)

   # and in the update_angle_units method I update these as:
   self.lastAngleUnits = self.angleUnits
 
   # but it seems that the next time I call the update_angle_units method this value is still the same as the initial value (DEGREES)

```

What is the proper way to store node attributes that are persistent?

Durman:
One more way to store attributes in blender classes is using `@properties` but data of such property should be still kept somewhere outside the class. For example in a dictionary with key as node ID and value as value of the property. 
Can you explain why do you use `lastAngleUnits` properties? Are you trying to convert units from one to another? I think such functionality better to be kept in separate node.

DolphinDream:
> One more way to store attributes in blender classes is using `@properties` but data of such property should be still kept somewhere outside the class. For example in a dictionary with key as node ID and value as value of the property.
> Can you explain why do you use `lastAngleUnits` properties? Are you trying to convert units from one to another? I think such functionality better to be kept in separate node.

last angle units is to keep track of the last angle units :) In order to update the values in the node from one unit to another you need to know wha the previous units were. Say you entered the value of an angle as 90 as the units are set to degrees..and then you want to switch to radians.. the value will be updated to be pi/2 .. hence it’s preserving the angle, albeit changing its units. To do this automatic update for the user you need to know the old and new units to do the proper conversion.

DolphinDream:
I tried using the separate AngleUnits class, but it was too much hassle having to match the node’s EnumProperty selected value to the corresponding AngleUnit object (“DEG” -> AngleUnits.DEGREES) just so I can use the AngleUnits class conversion API. Besides, as I said, I couldn’t have these custom properties remember their values once changed, so I decided to use the EnumProperty values instead and added extra API to the SvAngleHelper class itself to do the unit conversions.

Durman:
I see this very confusing. If I'm putting 90 and then switching from degrees to radians do you mean such change will not make any effect and the node just calculate equalent to 90 degrees angle in radians?

DolphinDream:
> I see this very confusing. If I'm putting 90 and then switching from degrees to radians do you mean such change will not make any effect and the node just calculate equalent to 90 degrees angle in radians?

The way I see it if the user enters a 90 DEGREES and the rendering shows things with the angle the user has in mind (a right angle..it would be confusing actually for the rendering to all of the sudden change to 90 RADIANS (which is not a right angle) just because the user changed the angle units. To me it’s more intuitive for the geometric angle (regardless of the units) to be preserved. Changing the units is only so one can alter the angle value albeit in the units that would make more sense to the user at any given time. 

Here’s another example:
Say I have a Torus Knot node.. where The **Revolution Phase** (an angle) is set to 90 DEGREES and the **Spin Phase** (also an angle) is set to 45 DEGREES. Now, as I continue with the noding, I decide to animate the Spin Phase with the Frame Info node’s Evaluate output, which is a 0-1 range value, so I want to plug the Evaluate into the TK's Spin Phase input, but before I do this I would need to change the TK node’s angle units to UNITIES so that the 0-1 input would correspond to a full spin. However, when I change the angle units from DEGREES to UNITIES I don’t want my Revolution Phase angle to change. I don’t want all of the sudden (just because I changed the angle units for the node) for that angle to be now 90 RADIANS instead of simply 0.25 (which is the same right angle expressed as UNITIES). If that value remained 90, albeit now they would be 90 RADIANS (instead of  PI/2 radians), I would have to go and update that value manually to match my originally intended **right angle** value.

Durman:
It is very confusing behaviour I think. For example how can you know does a socket makes any conversions when you open existing node tree?

DolphinDream:
> It is very confusing behaviour I think. For example how can you know does a socket makes any conversions when you open existing node tree?

The existing node trees would not be affected since they will have the old nodes. Only the new versions of the nodes will have angle units feature. The nodes in the old trees will be marked as deprecated (as you load them into blender) and the user can chose to update them to the newer (MK#) version.. and during the migration/update the angle units will be preserved (set) to match the old nodes angle units (radians or degrees whatever happen to be their units) so the overall node tree would not change its original behavior. 

If you guys think this automatic conversion of angle values when changing units is confusing, we can surely add a SV preference option to let the user decide the default behavior they want. I can add a "Angle Units Update” flag in SV prefs and have the angle helper class take this into account and decide whether the auto update of angle values should be done or not (e.g. the auto conversion factor = 1 if this flag is set). I don’t mind having the default behavior be what you suggest it’s more intuitive as I can set it to the behavior I think it’s more intuitive in my own SV settings :) 

Btw.. multiple nodes that had this angle units feature at this time have this auto conversion behavior already so this PR wasn’t introducing/suggesting any new behavior. I was just trying to factorize this code/behavior that current nodes are using into a shared space to avoid replicating the same code all over the place. However, even though this wasn’t my intention with the current PR,  as I am dealing with feature now I don’t mind revisiting this behavior with everyone and come up with a most desirable node behavior moving forward, so your feedback is appreciated as usual. I want to keep everyone happy 👍

DolphinDream:
Here’s a demo using SV settings (currently the auto angle value update is OFF by default). I could also make it ON by default (personally I think ON would be more intuitive as it preserves the angles while switching the units). 

SV-AngleHelper-demo3

utils/sv_transform_helper.py
```diff
@@ -0,0 +1,158 @@
...
+class SvAngleHelper():
+
+    def get_preferences(self):
+        props = get_params({
+            'auto_update_angle_values': False,
+        })
+        return props.auto_update_angle_values
+
+    def angle_conversion_factor(self, from_angle_units, to_angle_units):
+        if from_angle_units == AngleUnits.RADIANS:
```
zeffii, line 60: why not at the top do
```
if from_angle_units == tp_angle_units:
    return 1

```

zeffii: actually, why not
```python
angle_remap_options = {
    # from               to
    (AngleUnits.RADIANS, AngleUnits.DEGREES): 180/pi, 
    (AngleUnits.RADIANS, AngleUnits.UNITIES): 1/(2*pi), 
    (AngleUnits.DEGREES, AngleUnits.RADIANS): pi/180,
    (AngleUnits.DEGREE, AngleUnits.UNITIES): 1/360,
    (AngleUnits.UNITIES, AngleUnits.RADIANS): 2*pi,
    (AngleUnits.UNITIES, AngleUnits.DEGREES): 360 
}

...


    def angle_conversion_factor(self, from_angle_units, to_angle_units):
        if from_angle_units == to_angle_units:
            return 1

        multipler = angle_remap_options.get((from_angle_units, to_angle_units))
        if not multipler:
            raise Error(f"node {self.name} failure to map angle units, {from_angle_units}->{to_angle_units}")
```

zeffii: less verbose, arguably easier to reason about.

DolphinDream:
If we need to migrate the nodes from MKn to MKn+1 for this angle update, can I update the migrate_from of the MKn+1 node and set the default angle units to the native angle units of the previous node? (e.g. in migrate_from() add a call to self.angle_units = whatever_units_MKn_had ?

DolphinDream:
Also, another idea is to have a default angle units for all nodes. currenlty some have degree, some have radians etc. Maybe we can unify this to be the same unit all across .. OR .. perhaps even let he user decide what default angle units their nodes should have. e.g. There can be a SV settings “Default Angle Units” (to be selected from RAD/DEG/UNI) so that any new node created will have the angle units set to the user preference set angle units. Any thoughts?

DolphinDream:
I tried to add “default angle units” to sv preferences but i can’t figure out how to initialize the default of the angle_units properties in the SvAngleHelper class using these sv settings defaults. It seems that the code to get the properties to initialize the EnumProperty default is called only once when loading sverchok. Any ideas how to have these properties default values set to whatever the sv prefs are set to, every time I create a new node that has angle units?

Durman:
You can make class methods for Python Enums to convert Blender Enums to Python Enums.

```Python
current_units = AngleUnits.from_bl_enum(self.angle_units)
last_units = AngleUnits.from_bl_enum(self.last_angle_units)
```
And you can create method for getting factor converting from one units to another.
```Python
last_units.get_convert_factor(current_units)
```

DolphinDream:
> You can make class methods for Python Enums to convert Blender Enums to Python Enums.
> 
> ```python
> current_units = AngleUnits.from_bl_enum(self.angle_units)
> last_units = AngleUnits.from_bl_enum(self.last_angle_units)
> ```
> 
> And you can create method for getting factor converting from one units to another.
> 
> ```python
> last_units.get_convert_factor(current_units)
> ```

sure, that would work. Though, it begs the question, is it worth making the conversion from EnumProp to a custom Enum anyway just to be able to have the sexy angle conversion API. I think in this case it just adds extra code for rather little benefit. Thanks for coding tips though.

zeffii:
>  I think in this case it just adds extra code for rather little benefit. 

: )

```
novice programmer: Flat Array
intermediate programmer: some complex datastrcucture
advanced programmer: Flat array
```

DolphinDream:
> > I think in this case it just adds extra code for rather little benefit.
> 
> : )
> 
> ```
> novice programmer: Flat Array
> intermediate programmer: some complex datastrcucture
> advanced programmer: Flat array
> ```

not sure where I fit in here.. I'm either novice or advanced :)

DolphinDream:
@zeffii Whenever you get a chance.. can you take a look at the current implementation (sv_transforms_helper.py). Also I had a question about how I could have the settings be able to provide the default angle units the new nodes use when first created. Not sure if you think this would be useful or not or even recommended or not for some reason. Or maybe you think i’m trying to do too much in one PR. Let me know. I’m trying to merge this in sooner than later so that I can proceed with the updates of individual nodes (which I intended to promote to MK# in a separate PR).

DolphinDream:
Btw.. with this update, some nodes will have to be promoted to MKn+1 to include the angle units.  I wonder what is the best way to do the migrate_from old nodes in this case. Adding a migrate_from to the new nodes seems to be the solution so far, however I wonder whether the new nodes should be polluted with such migration code.. or whether the migration code should be added somewhere else, perhaps in a separate file. The reason being that after a while this migration code may need to be retired.. hence lead to more updates to the nodes that otherwise would not be necessary. Also, I just added a new MK3 to matrix in node, which has a migration from MK2 already.. but now I need to do another update to Matrix In node to promote it to MK4 (to account for new angle_units and other property renaming) how many generation of old nodes I need to support migrating from? In this case I’d say both MK3 and MK2 since the MK4 follows so shortly after MK3, but in general, in the long run, we may want to support any number of old generations (assuming we can do it in a way that doesn’t pollute and grow the migration_from method in each node.

If we can have a different mechanism / file where these migrations would be added and removed (as needed) the nodes themselves will never need to change to add/remove migration from code. 

Any suggestions?

zeffii:
> Also, do you have an idea how to set the EnumProperty default values defined in this AngleHelper class (angle_units) to the values specified in the SV settings? Basically I want the user to be able to specify via SV settings what the default angle units should be for the newly created nodes.

that..might be possible but ( this is pure conjecture ) you will need to write a "Property Factory". Inside the prop factory, you would return an EnumProperty, where the property attributes ..in particular the default.. would be taken from a lookup.  We have no examples of this in Sverchok yet.
I'm also not sure this is a good idea.  but without trying... i'm not immediately against it, because it's obviously a layer of customization which can be a nicety to the user (including us..hahah)

**sverchok.utils.sv_property_factory**
```python
def enum_prop_factory(named_property, attributes):
    
    import sverchok.some_settings as settings

    # only do this part if the settings can be found at this point
    if hasattr(settings, named_property):
        attributes['default'] = settings.named_property 
    
    return bpy.props.EnumProperty(**attributes)
```
exactly where you store the users default choice wasn't the point of what i'm showing here, but in the node you would import

```
from sverchok.utils.sv_property_factory import enum_prop_factory

# inside the class

class YourNode

      my_property: enum_prop_factory(.....fill the dict here ........)
```

zeffii:
or maybe this..
**sverchok.utils.sv_property_factory**
```python
def enum_prop_factory(property_defaults, named_property, attributes):
    
    # only do this part if the settings can be found at this point
    if hasattr(property_defaults, named_property):
        attributes['default'] = property_defaults.named_property 
    
    return bpy.props.EnumProperty(**attributes)
```
exactly where you store the users default choice wasn't the point of what i'm showing here, but in the node you would import

```python
from sverchok.utils.sv_property_factory import enum_prop_factory
from sverchok .. preferences .. property_defaults
# inside the class

class YourNode

      my_property: enum_prop_factory(property_defaults, items=...,default=..., update=...)
```

zeffii:
i've always found upgrading nodes such an utterly painful practice, i'm going to avoid this question :)

DolphinDream:
> i've always found upgrading nodes such an utterly painful practice, i'm going to avoid this question :)

What if the code that does detect the deprecated nodes and gives the option to migrate to another node (based on the existence of the migration node list) [node_replacement.py?]. can be the one to load another file in which all the migration code specifications are located (outside of all the nodes). There we can then have a (“old node Id”, “new node id”) pair mapped to a migrate_from() code where the migration specifics are specified.

migrate(from_node, to_node):
if from.node.bl_idname == “MatrixInMK2”:
    if to_node.bl_idname == “MatrixInMK3”:
        // do all the node migration magic here 

and of course, have other ifs for other from/to combos.

I’m sure you can do this better than using ifs via some sort of mapping magic. but you get the idea.

When new nodes are added, we’ll only have to add the to_node code from all the from_to code for previously supported old nodes. In time, we may want to retire migration for really old nodes (if needed) in which case we would just have to remove those old_node if cases. 

This way the node code would not need to be polluted with migration code.

Think this would work?

zeffii:
i'm still going to ignore the question, by answering it this way.

if i understand your extended question, what you propose merely changes the location where the complexities are , but don't alleviate the kind of code we have to write to facilitate transfer between A and B.  The upside of adding the code to the node directly, is that you have that code closest to where it makes sense. Contrary to the situation we'd have if we added it to the post_load event handler.

it's tricky, i'd love to periodically ditched the older nodes ..even out of the old nodes list.

DolphinDream:
I guess ignorance is bliss.

zeffii:
before i forget. if you do eventually implement the `enum_factory` , then there's one issue that will need to be handled by iojson.

one detail that might result in interesting issues is that values of node properties, which are not different from their default initialization value, don't get stored in iosjon. This will pose a problem for exchanging iosjon files between users who have different defaults for the angle stuff.

```
user X has a property default set by preferences
user X does not change the value of that property
user X stores iosjon with nodes that initialize their properties from a factory

user Y has a different default set by preferences
user Y loads iosjon with those nodes
user Y will find that nodes that user X has modified defaults, and has not changed them on the node at runtime, those nodes will output not as user X intended.

```
in the case of those nodes that inherent the `angles helper class`, iosjon could ofcourse store those properties anyway, using maybe 1 or 2 lines of code.

zeffii:
i think... but i'll double check to see if i optimized iosjon to store only modified values.

zeffii:
ok looks like I didn't implement "sparse" storage for node properties, (it was implemented for socket properties). so that shouldn't be a problem here.

don't forget I also implemented user node defaults, which can be searched for in the issue tracker..i'm not too sure how that was implemented, it hooks onto `sv_init` anyway.

Durman:
> sure, that would work. Though, it begs the question, is it worth making the conversion from EnumProp to a custom Enum anyway just to be able to have the sexy angle conversion API. I think in this case it just adds extra code for rather little benefit. Thanks for coding tips though.

@DolphinDream My suggestion was just create method of conversion units to enum class. You have did this method. But you add this method to the node class. So my suggestion was not about extra code. 

Actually this method should belong to Enum class because all data which this method need belongs to this class.

utils/sv_transform_helper.py
```diff
@@ -0,0 +1,146 @@
...
+    def angle_conversion_factor(self, from_angle_units, to_angle_units):
+        if from_angle_units == to_angle_units:
+            return 1
+
+        multiplier = angle_remap_options.get((from_angle_units, to_angle_units))
+
+        if not multiplier:
+            raise Error(f"node {self.name} failure to map angle units, {from_angle_units}->{to_angle_units}")
+
+        return multiplier
```
Durman, line 79: I still insist that this method should be in enum class. The fact that self word is not used in method is saying that the method should be moved somewhere.

utils/sv_transform_helper.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def radians_conversion_factor(self):
+        return self.angle_conversion_factor(self.angle_units, AngleUnits.RADIANS)
+
+    def degrees_conversion_factor(self):
+        return self.angle_conversion_factor(self.angle_units, AngleUnits.DEGREES)
+
+    def unities_conversion_factor(self):
+        return self.angle_conversion_factor(self.angle_units, AngleUnits.UNITIES)
```
Durman, line 88: Why this methods?

utils/sv_transform_helper.py
```diff
@@ -0,0 +1,146 @@
...
+}
+
+
+class SvAngleHelper():
+
+    def get_preferences(self):
+        props = get_params({
+            'auto_update_angle_values': False,
+        })
+        return props.auto_update_angle_values
```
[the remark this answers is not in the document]
DolphinDream: I had originally added user default units via SV prefs here, but I couldn't figure out how to read those values from the prefs and updated them in the enum defaults for every node. (they only got updated once when registering the helper class. I removed the default units since but didn’t bother to change the method name as I though I’d add these later at some point in which case I would have had to make this a mixin method anyway (hence i kept the mixin name). But you’re right. I need to change this to perhaps get_auto_update_angle_values or something.

utils/sv_transform_helper.py
DolphinDream: Please feel free to provide the implementation that makes sense to you. I tried my best and spent more time on this already than I originally planned.

Durman: Just take this in account for the future.

Issue #3029, opened 2020-04-01
Unstable Index While Extruding


enzyme69:
Related to this:
https://github.com/nortikin/sverchok/issues/3019

So after "Selection Grabber" node was created by Zeffii I am testing a couple of simple modeling exercise. I quickly noticed some weird things happen, like some Extrude node actually keep changing the index number even without me touching the previous nodes. I think this is quite unstable and could be a bug -- or just an unfortunate limitation that cannot be fixed? 

<img width="1253" alt="Screen Shot 2020-04-02 at 10 31 30 am" src="https://user-images.githubusercontent.com/3952687/78196199-5e740180-74cd-11ea-834d-aeb55a849f25.png">

[table_chair_004.blend.zip](https://github.com/nortikin/sverchok/files/4418096/table_chair_004.blend.zip)

enzyme69:
Video 17 MB.
[Screen Recording 2020-04-02 at 10.37.10 am.mov.zip](https://github.com/nortikin/sverchok/files/4418111/Screen.Recording.2020-04-02.at.10.37.10.am.mov.zip)

Extrude Region node seems to be the culprit...

enzyme69:
If I use select mesh element by locations, it seems to be more stable and sane way as it is not dependant on index.

[table_chair_006.blend.zip](https://github.com/nortikin/sverchok/files/4418143/table_chair_006.blend.zip)

Realized:
Bridge Node is also missing, so maybe "mirror for this kind of thing".

enzyme69:
Inset Faces seems to be most stable. 

<img width="1169" alt="Screen Shot 2020-04-02 at 11 18 54 am" src="https://user-images.githubusercontent.com/3952687/78198268-caf1ff00-74d3-11ea-8383-117ddfb88974.png">

[table_chair_013.blend.zip](https://github.com/nortikin/sverchok/files/4418211/table_chair_013.blend.zip)

zeffii:
>  If I use select mesh element by locations, it seems to be more stable and sane way as it is not dependant on index.

if you reread the responses i've given. You'll appreciate that i've said this over and over and over and over again.

zeffii:
I think , likely , there is instability in face indexing between what you are keeping in the tree, vs what has been baked.

for example, after baking one item.. and rereading from the real geometry seems more stable.
![image](https://user-images.githubusercontent.com/619340/78224979-b7419b00-74c9-11ea-9153-7e72c2701a21.png)

[...]

PR #3040, opened 2020-04-05
Fix face indexing extrude region node

Durman:

#3029

## Solution description

The attempt to solve the problem by using own mesh walker.

## Preflight checklist

- [ ] Create MKNext version
- [ ] Solve multiple disjoint islands problem.
- [ ] Speed test. And new mode if necessary.
- [ ] Ready for merge.

### DIFF
--- a/nodes/modifier_change/extrude_region.py
+++ b/nodes/modifier_change/extrude_region.py
@@ -16,16 +16,18 @@
 #
 # ##### END GPL LICENSE BLOCK #####
 
-from mathutils import Matrix, Vector
-from itertools import chain
+
+from typing import Generator, Set
 
 import bpy
 from bpy.props import FloatProperty, BoolProperty, EnumProperty
 import bmesh.ops
+from mathutils import Matrix, Vector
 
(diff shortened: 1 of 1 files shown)


portnov:
I'd suggest using "bmesh data layers" instead. Not sure how exactly... (not that I understand the problem you are dealing with); something like: remember the original Index of the face in it's data layer. See how it's done in the Bevel node for edges, for example: https://github.com/nortikin/sverchok/blob/master/nodes/modifier_change/bevel.py

Durman:
I understand what do you mean. The problem is the new faces created by the extruded algorithm (side faces) can have different indexes from one execution to another. And it is impossible to set something into their layers directly. So I don't see how layers can help here.

portnov:
Then I'd suggest to add an option to toggle that arranging algorithm on/off. Because in many cases it will just waste time without necessity (most "good" node trees will not depend on this order).

Durman:
I'm going make a speed test. If slowdown will significant I will add the property. I have added this into todo list.

Durman:
I think it will be better to add separate node for mesh sorting.

PR #3042, opened 2020-04-05
Matrix Apply understands NumPy

vicdoval:
Added a Numpy implementation of the node, after coding the performance results are:
As a general rule in this node the new Numpy implementation will be faster if any input is a NumPy array or you want to get NumPy arrays from the outputs. If the surrounding nodes are using python list the performance of both implementations will depend on many factors. With a light geometry but many Matrices the Python implementation will be faster, as heavier gets the input geometry and less the matrices number the NumPy implementation will start being a better choice. Also if the incoming topology of polygons is regular the NumPy implementation will increase its performance while the Python implementation will not be affected by that parameter.

So the default will be the previous Python implementation at this point in time.
 
- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/nodes/matrix/apply_and_join.py
+++ b/nodes/matrix/apply_and_join.py
@@ -40,7 +40,6 @@ def mesh_join_np(verts, edges, pols, out_np_pols):
     accum_vert_lens = np.add.accumulate([len(v) for v in chain([[]], verts)])
 
     v_out = np.concatenate(verts)
-    e_out, p_out = np.array([]), np.array([])
 
     # just checking the first object to detect presence of edges and Polygons
     # this would fail if the first object does not have edges/pols and there
@@ -50,10 +49,11 @@ def mesh_join_np(verts, edges, pols, out_np_pols):
 
     if are_some_edges:
         e_out = np.concatenate([edg + l for edg, l in zip(edges, accum_vert_lens)])
+    else:
+        e_out = np.array([])
 
     if are_some_pols:
         if out_np_pols:
-
             is_array_of_lists = pols[0].dtype == object
             if is_array_of_lists:
                 p_out = [np.array(p) + l  for pol, l in zip(pols, accum_vert_lens) for p in pol]
@@ -62,6 +62,8 @@ def mesh_join_np(verts, edges, pols, out_np_pols):
                 p_out = np.concatenate([pol + l for pol, l in zip(pols, accum_vert_lens)])
         else:
             p_out = [[v + l for v in p]  for pol, l in zip(pols, accum_vert_lens) for p in pol]
+    else:
+        p_out = np.array([])
 
     return v_out, e_out, p_out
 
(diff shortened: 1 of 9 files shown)


nodes/matrix/apply_and_join.py
```diff
@@ -17,27 +17,90 @@
...
+
+        if pols[0].dtype == object:
+            p_out = [np.array(p) + l  for pol, l in zip(pols, lens) for p in pol]
+
+        else:
+            p_out = np.concatenate([pol + l for pol, l in zip(pols, lens)])
+    else:
+        p_out = [[v + l for v in p]  for pol, l in zip(pols, lens) for p in pol]
+
+    return v_out, e_out, p_out
```
Durman, line 54: It would be nice to know which input arguments are numpy array and which are not. Also `out_np` is unclear. According the argument name I have thought that this probably some array with mesh or empty for keeping merging meshes. This looks very magical - `out_np[2]`.

```suggestion
def mesh_join_np(verts, edges, pols, out_np):

    accum_vert_lens = np.add.accumulate([len(v) for v in chain([[]], verts)])

    is_edges_first_object = len(edges[0]) > 0
    if is_edges_first_object:
        e_out = np.concatenate([edg + l for edg, l in zip(edges, accum_vert_lens)])
        #                      is it really edge??
    else:
        e_out = np.array([], np.int32)

    is_pols_first_object = len(pols[0]) > 0
    is_something_else = out_np[2]
    if is_pols_first_object and is_something_else:
        
        is_array_of_lists = pols[0].dtype == object
        if is_array_of_lists:
            p_out = [np.array(p) + l for pol, l in zip(pols, accum_vert_lens) for p in pol]

        else:
            p_out = np.concatenate([pol + l for pol, l in zip(pols, accum_vert_lens)])
    else:
        p_out = [[v + l for v in p] for pol, l in zip(pols, accum_vert_lens) for p in pol]

    v_out = np.concatenate(verts)
    return v_out, e_out, p_out
```

vicdoval: I took part of your suggestions, thanks for the feedback. I hope now everything is more clear

Durman: Why not using `else` condition?
```Python
    if are_some_edges:
         e_out = np.concatenate([edg + l for edg, l in zip(edges, accum_vert_lens)])
    else:
        e_out = np.array([], np.int32)
```
Reading this code it is clear if there are no edges the functions returns empty array. In your case reader should search this logic somewhere else in the code. 
And I think new type of empty arrays should be set explicitly.


The same with `v_out` variable. To know what is assign to this variable I should look through nearly all function. It is logically to move assigning value to variable as close to useing of the variable as possible.
```Python
    v_out = np.concatenate(verts)
    return v_out, e_out, p_out
```

vicdoval: I agree about the else condition.
About the verts line for me feels more natural if is the first input of the node, the first argument of the function, and the first output of the function and the node to be resolved first in the function not in the last line

PR #3043, opened 2020-04-06
Subdivide mk2

vicdoval:
The subdivide node was forgotten... it was too slow....
To divide one time a box 25x25x25:
Before 29.5 secs (!!!)
Mk2 0.071 secs  

The mk2 uses the edge_data layer to match the masking avoiding the previous bottleneck

Anyway I find this node is pretty slow even now so i added a couple of ideas:
**Mute** and **Freeze**.
With 'Mute' active the node is bypassed, is not the fancy blender muting, is a custom one
With 'Freeze' active the node will stop processing the incoming data but will still  be passing the previously computed results. The data is stored in a dictionary and has to be recomputed every time the file is opened but once done the node stops loading the node-tree.
This behaviour could be expanded to other 'heavy' nodes.

What do you think?
image

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


Durman:
I think such modes every node should have.

portnov:
Maybe either
a) work in "mute" mode when the node tree is in draft mode;
b) support special value for "subdivisions" value for the draft mode (see how it is implemented in the Plane node, for example).

vicdoval:
@portnov  I was thinking to about how this should cooperate with the Draft mode.
Maybe muting the node is too aggressive. 
the second sounds better especially combined with the possibility of making 0 subdivisions (and passing the data as it came in which is the same as muting) that way the user can choose between a lower number or total muting on draft mode.
The only question is if the node is 'freezed' and you toggle to Draft mode should it update?

vicdoval:
@Durman The `Mute` for sure would be nice for every node.  But the freeze.. I would prefer the nodes to update only if the upstream has changed and the down stream is connected to a viewer... in other words our dreamed new update system...

nodes/modifier_make/subdivide_mk2.py
```diff
@@ -0,0 +1,399 @@
...
+    def mute_node(self):
+        origin_socket = [0, 1, 2, 3, 0, 1, 2, 0, 1, 2]
+
+        for i, s in zip(origin_socket, self.outputs):
+            if self.inputs[i].is_linked:
+                s.sv_set(self.inputs[i].sv_get(deepcopy=False))
+
+    def freeze_node(self):
+        if 'storage' in self.mem.keys():
+            for m, s in zip(self.mem['storage'], self.outputs):
```
zeffii, line 247: i think you'll find self.mem['storage'] will be used for all instances of this node. so you make a second node and if it also uses `self.mem['storage']`, then both nodes will output the same data.

only when you do
```python
self.mem[hash(self)]['storage'] = ...
```
will each node have unique storage.

zeffii: so
```python
self.mem[hash(self)] = {}
self.mem[hash(self)]['storage'] = ...
```

zeffii: this is why nodes like the viewers and snlite implement `node_dict` in the seemingly convoluted way.

if your immediate thought is : " ok.. then why don't we have a nicer mechanism in svcommon for this " .
then i'm right there with ya.  i don't know.

nodes/modifier_make/subdivide_mk2.py
```diff
@@ -0,0 +1,399 @@
...
+                return True
+        elif self.sv_frezze:
+            self.sv_record = True
+            return False
+
+    def process(self):
+        inputs, outputs = self.inputs, self.outputs
+
+        if self.dont_process():
+            return
```
Durman, line 300: This is no great method. When you are calling like this `if sekf.dont_process()` , I expect that the purpose of this method is return bool value. But actually it is doing some extra job inside. I recommend to split this method on two or more. `dont_process` method should only checking whether it should return `True` or `False` nothing more.

You have lost 2 `else` conditions there.

nodes/modifier_make/subdivide_mk2.py
Durman: Another problem can be that the dictionary will be empty after rebooting Blender file. I don't know whether this was taken in account or not.

Durman: If we would like to have Python attributes for Blender classes we should have function or class which could wright Python data into somewhere in Blender file and restore data from that written data. Also there is need in two triggers - open and save Blender file.

nodes/modifier_make/subdivide_mk2.py
```diff
@@ -0,0 +1,399 @@
...
+    def get_result_pydata(self, geom):
+        new_verts = [v for v in geom if isinstance(v, BMVert)]
+        new_edges = [e for e in geom if isinstance(e, BMEdge)]
+        new_faces = [f for f in geom if isinstance(f, BMFace)]
+
+        new_verts = [tuple(v.co) for v in new_verts]
+        new_edges = [[edge.verts[0].index, edge.verts[1].index] for edge in new_edges]
+        new_faces = [[v.index for v in face.verts] for face in new_faces]
+
+        return new_verts, new_edges, new_faces
```
Durman, line 261: This method looks like should be separate function or should be the method of some other class.

nodes/modifier_make/subdivide_mk2.py
zeffii: when opening a blend `hash(self)` will indeed not be present in the node_dict ( `self.mem` ),  
1) either `self.mem` of all instances will be empty, or
2) self.mem isn't empty but the `hash(self)` of all nodes is different each time the blend is loaded

i'm not sure if 1 is true or 2.  i can't remember and can't test at the moment :).. i think it's `1` .

Durman: It is 1. And we have `n_id` properties for nodes. It is robust `ID` of a node instance.

Durman:
> But the `Freeze`.. I would prefer the nodes to update only if the upstream has changed and the down stream is connected to a viewer... in other words our dreamed new update system...

Freezing can be useful to freeze not only `backward` nodes but and `forward` nodes. For example there is some pattern which is base for some huge algorithm. So every change of this pattern causes of a lot of recalculation. It would be useful to freeze node after pattern so it will be faster to set pattern into appropriate shape.

nodes/modifier_make/subdivide_mk2.py
zeffii: yes `n_id` is preferable, see the implementation @vicdoval .

https://github.com/nortikin/sverchok/blob/7019c9bc5d78dd752debb8cd1319944abe905944/node_tree.py#L367

zeffii: rather than building that into svcommon, it will make more sense to add it to a new mixin, that nodes can inherit from purely if they need a convenient way to use a `node_dict`.

though, i'll argue that this Pullrequest is not the place to implement that mix-in

vicdoval:
Finally I am going to drop the `mute  `and the `freeze`, it is true that this behaviours could be implemented for all the nodes at once, maybe in a improved update system, so it will be for another PR
In this one I will merge only the updated masking and the Draft mode upgrades

nodes/modifier_make/subdivide_mk2.py
vicdoval: you are right is not the most elegant... finally i removed the properties so less hassle.
> You have lost 2 else conditions there.

[No else after return](https://pheanex.github.io/pylint/#R1705)

nodes/modifier_make/subdivide_mk2.py
vicdoval: moved to the sv_bmesh_utils

nodes/modifier_make/subdivide_mk2.py
vicdoval: The idea (for another PR) was it to be reloaded every time you opened the file.
We will see how is the best way to implement it...

zeffii: if a node is set to freeze, it will need to regenerate the data upon blend open and indeed iojson import.
- this can be done automated as a post-load operation, post-import
- can also be built into the node (using the mixin) by providing the function the produce the stored data, and if the node detects that it doesn't contain the data, then it will call that function and store it.

it's not super easy to explain, or indeed to implement in a foolproof way - but I think it is possible (as attested by the fact that SNLite implements it already...everyday :)  )

json_examples/Architecture/Tensile_Cover.json
```diff
@@ -807,214 +831,208 @@
...
     ],
     [
-      "Vector Math.001",
-      0,
-      "UV Connection",
-      1
+      "Reroute",
+      "Output",
+      "List Mask (out).002",
+      "mask"
```
zeffii: you should use indices, not socket name. for all sockets that aren't a Reroute node. Looks like many manual edits to .json files generate a bit of potential problem.

I might have to modify the way iojson stores these links..

vicdoval: I did not edit the json manually...

zeffii: i see, then the function is treating the link differently if either of the nodes is a reroute node.. weird. ..even more pressure to modify that part of the format :)

zeffii: ooops. my bad @vicdoval  

https://github.com/nortikin/sverchok/blob/4260854806b381201e7cd2b89a1a83b8c71f94f9/utils/sv_IO_panel_tools.py#L102

that's working as expected then.

zeffii: i'll have a think about that.

zeffii: forget I said anything :)

PR #3046, opened 2020-04-08
Less global updates

vicdoval:
Reduces the global updates by:

Storing the links and the ouput_sockets connected

If the links don't change the global update is not done.

If the links change `build_update_list `is triggered but the update is done from the affected nodes (not a global update) with the `process_from_nodes` function

It does not affect animation updates.

The socket cache is deleted when you delete the node (not every update) from the `free` function of the node. The nodes that where using this function now use sv_free, called from the common free

With this we would remove just the updates that are called when a new node is created or deleted

The data is stored in dictionaries and has to be refilled every time you open the file.

What do you think?
It could be refined but seems to work nicely with less updates :D


Durman:
https://docs.blender.org/api/current/info_gotcha.html#undo-redo

For this reasons some nodes have `n_id` property.

zeffii:
hat off to you for wanting to tackle this area. it has many tentacles :)

node_tree.py
```diff
@@ -286,11 +291,59 @@ def timestamp(self):
...
+        if self.configuring_new_node:
+            # print('skipping global process during node init')
+            return
+        if self.is_frozen():
+            # print('not processing: because self/tree.is_frozen')
+            return
+
+        if not self.sv_process:
+            return
+        links_has_changed = self.sv_links[hash(self)] != self.links.items()
```
Durman, line 320: This string can crash Blender easily:
https://docs.blender.org/api/current/bpy.types.Node.html#bpy.types.Node.insert_link

If you would like tracking of links creating/deleting you should look toward such events:
https://docs.blender.org/api/current/bpy.types.Node.html#bpy.types.Node.insert_link

First of all you should create list of possible ways of creating/deleting links. Then you should find an appropriate detection event for all ways. It can happen that Blender API does not allow tracking of links creation.

vicdoval: I have changed the hash(self) to a tree_id method similar to the n_id

> This string can crash Blender easily:

Do you have any evidence on that? I have been playing with it and i haven't been able to crash Blender

Durman: You can't compare them anyway. As I have said Blender objects does not have stable hash.

vicdoval: well... it seems to work even after undo/redo

zeffii: instead of using the hash of the noderee, maybe simply use the the name of the nodetree, .. or whatever is currently used for the cache :)

renaming nodetree does pose a problem for that solution tho.

Durman: https://yadi.sk/i/B_ugIFwT3LCZew

vicdoval: @Durman you are right, that check fails after undo, it will return it has changed.
Then the function checks the linked_sockets memory and returns that none of the nodes has been affected so does not trigger the nodetree processing but updates the links_cache and keeps on with normality

vicdoval: or it should.. Im working on it

utils/sv_IO_panel_tools.py
```diff
@@ -433,7 +433,7 @@ def perform_scripted_node_inject(node, node_ref):
         node.load()
 
 
-def perform_profile_node_inject(node, node_ref):
+def perform_profixle_node_inject(node, node_ref):
```
zeffii, line 436: 0_o

node_tree.py
Durman: I have thought about how new links can be detected and came up with next solution. The idea is to generate addresses for each socket in a tree such as `node.n_id + socket.identifier` and create Sverchok links which will keep pare of socket addresses. Then if Sverchok link with current pair of socket addresses is not found it means new links is detected.

```Python
class SvLink(NamedTuple):
    from_socket_address: str
    to_socket_address: str

to_socket_sv_links = {to_socket_address: sv_link}

for bl_link in tree.links:
    to_socket_sv_link = to_socket_sv_links.get(address(bl_link), None)
    if to_socket_sv_link and to_socket_sv_link.is_equal(bl_link):
        continue
    else:
        new_link = bl_link
```

vicdoval:
There is still a way to the end but the error travis outputs  in "Technical Ring with holes"appears only when I try to open that example in the old 2.80, in 2.81 and higher the error does not appear.
Seems the travis is checking everything with 2.80... maybe we should update it?

portnov:
You can do that. You'll have to fix the download link in `.travis.yml` and check that everything is working as expected.
It is also possible to run tests on several versions of blender (google://travis matrix build), I did not try that.

vicdoval:
Despite the test errors that appear on Travis everything seems working pretty good.
Even Undoing, Monads and Reroutes :D
Well... undoing only triggers the update if links have changed or node number has changed, if a node property changes during undo the update in not triggered.
Redoing was not implemented in the actual master and neither in this PR but at least does not crash the system. I dont now if we should do anything about redo... maybe the same as with the undo...
I will be testing it and it would be great if any of you want to test it too.
As @Durman suggested I used a NamedTuple to store the link and a node_dict to store n_id_to_node.name.

node_tree.py
```diff
@@ -131,17 +137,38 @@ def wrapper_update(self, context):
 
     return wrapper_update
 
+# Declaring namedtuple()
+Sv_Link = collections.namedtuple('sv_link',['from_node_id', 'to_node_id', 'from_socket_id', 'to_socket_id'])
```
Durman, line 141:
```suggestion
SvLink = collections.namedtuple('SvLink',['from_node_id', 'to_node_id', 'from_socket_id', 'to_socket_id'])
```

node_tree.py
```diff
@@ -131,17 +137,38 @@ def wrapper_update(self, context):
...
-
+    configuring_new_node: BoolProperty(name="indicate node initialization", default=False)
+    sv_node_dict = {}
+    sv_links = {}
+    sv_linked_output_sockets = {}
+    sv_linked_input_sockets = {}
+    sv_linked_inputted_nodes = {}
+    tree_id: StringProperty(default="")
+    @property
+    def get_tree_id(self):
```
Durman, line 168:
```suggestion
    _tree_id: StringProperty(default="")
    @property
    def tree_id(self):
```

node_tree.py
```diff
@@ -191,7 +218,124 @@ def sv_trees(self):
             if ng.bl_idname in {'SverchCustomTreeType', 'SverchGroupTreeType'}:
                 res.append(ng)
         return res
+    def bl_links_to_sv_links(self):
+
+        sv_links=[]
+        for link in self.links.items():
+            if not link[1].to_socket.node.bl_idname == 'NodeReroute':
```
Durman, line 225:
```suggestion
        for i, link in self.links.items():
            if not link.to_socket.node.bl_idname == 'NodeReroute':
```

node_tree.py
```diff
@@ -191,7 +218,124 @@ def sv_trees(self):
...
+                #recursive function to override reroutes
+                output_socket, output_node = get_output_socket_id(link[1].from_socket)
+                if output_socket:
+                    sv_link = Sv_Link(
+                        from_node_id=output_node,
+                        to_node_id=link[1].to_socket.node.node_id,
+                        from_socket_id=output_socket,
+                        to_socket_id=link[1].to_socket.socket_id )
```
Durman, line 233: It is better to create for this code class method of `SvLink` class.
```Python
class SvLink(NamedTuple):

    @classmethod
    def init_from_link(cls, link) -> SvLink: ...
```

node_tree.py
```diff
@@ -191,7 +218,124 @@ def sv_trees(self):
...
+                    sv_links.append(sv_link)
+                    # sv_links.append((output_socket, link[1].to_socket.socket_id))
+        return sv_links
+
+    def fill_links_memory(self):
+        tree_id = self.get_tree_id
```
Durman, line 239:
```suggestion
```
Just use `self.tree_id` instead.

node_tree.py
```diff
@@ -191,7 +218,124 @@ def sv_trees(self):
...
+        for sv_link in new_sv_links:
+            inputted_nodes.append(sv_link.to_node_id)
+            input_sockets.append(sv_link.to_socket_id)
+            output_sockets.append(sv_link.from_socket_id)
+
+        self.sv_linked_output_sockets[tree_id] = output_sockets
+        self.sv_linked_input_sockets[tree_id] = input_sockets
+        self.sv_linked_inputted_nodes[tree_id] = inputted_nodes
+
+    def use_link_memory(self):
```
Durman, line 256: This method should be definitely splited into 2-3 methods.

Durman:
I will test.

core/sockets.py
```diff
@@ -86,7 +86,17 @@ def set_default(self, value):
...
+        # if not self.socket_id_m:
+            # try:
+            # self.socket_id_m = str(hash(self) ^ hash(time.monotonic()))
+            # except:
+                # return str(hash(self) ^ hash(time.monotonic()))
+            # self.socket_id_m.set(str(hash(self) ^ hash(time.monotonic())))
+            # self.set("socket_id_m", str(hash(self) ^ hash(time.monotonic())))
+        # return self.socket_id_m
+        # return str(hash(self.id_data.name + self.node.name + self.identifier))
+        return str(hash(self.node.node_id + self.identifier))
```
Durman, line 98: Is it really need to change this method?

vicdoval: To support node and tree renaming without losing the track of the data it has to relay on node_id

node_tree.py
```diff
@@ -131,17 +137,38 @@ def wrapper_update(self, context):
...
     has_changed: BoolProperty(default=False)
     limited_init: BoolProperty(default=False)
     skip_tree_update: BoolProperty(default=False)
-
+    configuring_new_node: BoolProperty(name="indicate node initialization", default=False)
+    sv_node_dict = {}
+    sv_links = {}
+    sv_linked_output_sockets = {}
+    sv_linked_input_sockets = {}
+    sv_linked_inputted_nodes = {}
```
Durman, line 165: I would suggest consider of using static class for managing Sverchok links. It will let incapsulate all related code and let read it easier.
```Python
sv_links = SvLinks
```
For example it is possible encapsulate work of getting links for current node tree:
```Python
class SvLinks:
    sv_links = dict()

    @classmethod
    def get(cls, node_tree):
        return cls.sv_links[node_tree]

node_tree.sv_links.get(node_tree)
```
`bl_links_to_sv_links`  method can be moved to this class. It can has methods - `links has changed`, `get affected nodes`.

But updates callings should remain in `SvNodeTreeCommon` class and should be explicit. I don't like how updates are made now. It should be like this:
```Python
class SvNodeTreeCommon:
    def update():
        ...
        if links_has_changed:
            process_from_nodes(get_nodes())
        ...
```

vicdoval: @Durman I separate the code in different files (links.py and node_id_dict.py) to join all the related code and cache dictionaries. is truly cleaner.
From here I'm open to learn the hows and whys of encapsulating it in classes with inheritances and classmethods and so. (To be honest I am pretty new this topics so i have just used them when i didn't find another way)

core/handlers.py
```diff
@@ -65,23 +67,30 @@ def sv_handler_undo_pre(scene):
...
     num_to_test_against = 0
+    links_changed = False
     for ng in sverchok_trees():
         num_to_test_against += len(ng.nodes)
-
+        tree_id = ng.tree_id
+        if tree_id in sv_links_cache:
+            links_changed = sv_links_cache[tree_id] != bl_links_to_sv_links(ng)
+            if links_changed:
+                break
```
Durman, line 84: You are doing this test in many places in the code. You should create method for this test. If you have to change test it will be possible to do in one place.

core/links.py
```diff
@@ -0,0 +1,190 @@
...
+                subtree.has_changed = True
+                node.process()
+                has_been_updated = True
+
+            if has_been_updated:
+                affected_groups.append(node)
+    return affected_groups
+
+
+def use_link_memory(node_tree):
```
Durman, line 139: Does not do updates from this method. We now have `SverchCustomTree` class responsible for updates. We should centralize all updates in one place so it would be easier to deal with update events.
Don't return any `bool` values if purpose of the method is not in doing some test.
Most likely you should create only affected nodes in this method all other code should be moved somewhere.

node_tree.py
Durman: Key to understanding using of classes is that the main their purpose is to combine data with methods. Best method is method without arguments. I would suggest to find good book about OOP.

node_tree.py
```diff
@@ -299,15 +356,19 @@ def update(self):
...
-        self.has_changed = True
-        # self.has_link_count_changed
-        self.process()
+        if link_cache_is_ready(self):
+            use_link_memory(self)
+        else:
+            if self.sv_process:
+                fill_links_memory(self)
+            self.has_changed = True
+            self.process()
```
Durman, line 371: What I'm expecting to see here:
```Python
update_sv_links()  # probably
if links_has_changed():
    process_from_nodes(get_effected_nodes())
else:
    self.has_changed = True
    self.process()
```

core/update_system.py
```diff
@@ -36,7 +36,13 @@
...
 no_data_color = (1, 0.3, 0)
 exception_color = (0.8, 0.0, 0)
-
+sv_first_run = True
+def set_first_run(value):
+    global sv_first_run
+    sv_first_run = value
+def get_first_run():
+    global sv_first_run
+    return sv_first_run
```
Durman, line 45:
```suggestion
def is_first_run():
    global sv_first_run
    return sv_first_run
```

node_tree.py
vicdoval: Modules are also a good place to combine data and methods. There are many node-trees and many nodes but just one place to 'rule' them all so i don't seed the point in making it a class that has to be initiated, it can be perfectly one module that starts when sverchok is loaded

core/handlers.py
vicdoval: this line is only one time in the code, it is similar in another spot but different because i need to store the links first in a variable

node_tree.py
vicdoval: I would also like it to be so simple, but the problem is that we have node-trees inside node-trees so is a recursive search, and if one monad changed it has to be updated and it has to trigger updates in the upper tree and so on.
I started with something like that but solving the monad question made it more complex and returning a boolean, which is not perfect code-wise, works to inform the parent tree

core/update_system.py
vicdoval: ok this one is easy to change

vicdoval:
@portnov  if you could please check the changes i did in the travis.yml obviously i'm missing something because still tries to find 2.8/scripts/... on maybe i have to commit the travis changes to master?

node_tree.py
Durman: Modules looks like can be used as static classes but their are not the same. There are some downsides of using modules in this way:
- You have to create global variables. It is common knowledge that it is better to avoid using them. 
- You have to import each method separately (at least you do this). 
- It is harder to split module into two. 
- Module can't inherit functionality of other modules.
- It is simpler to redo static class to normal class for using its instances rather then redo module in this way.
- And probably most important thing you can't assign functionality of module to other classes fields. It would be more comfortable to have access to this functionality via `node_tree.sv_links.some_method()`

node_tree.py
Durman: Monads have their own `update` method. Why does not put the update code there?

core/handlers.py
Durman: It does not matter is it the line strictly the same or not. If the test of changing of links will change (most likely it can happen) you have to scan all code in catching such places. And this task will much harder after a few months than it looks now. It is much easier to have one common method which can do such operation.

node_tree.py
vicdoval: The Monad is very complex node, is a node and a node-tree, it has two updates, the nodetree update and the node update, my code affects to the nodetree update which is the same as any other node-tree, thats why it is called from the nodetree class and not from the monad class

core/handlers.py
vicdoval: This is one call:
```
if tree_id in sv_links_cache:
   links_changed = sv_links_cache[tree_id] != bl_links_to_sv_links(ng)
   if links_changed:
       break
```
And this is the other one:
```
new_sv_links = bl_links_to_sv_links(node_tree)
before_sv_links = sv_links_cache[tree_id]
links_has_changed = before_sv_links != new_sv_links
if links_has_changed:
   some_var = some_func(new_links)
```

As you can see they are different because in the second case i need to store the links first.
If it changes in the future, the future coder will have also to deal with opening, undoing, updating, reloading and sub-trees. That line would be the less of his problems.

node_tree.py
vicdoval: > You have to create global variables. It is common knowledge that it is better to avoid using them.

the sv_links is a global dictionary as the socket_data_cache affects all sverchok so it makes sense to be a global variable.

> You have to import each method separately (at least you do this).

Everything stays with in the sverchok.core and sverchok.nodetree so is not a big deal to import the at the beginning 

> Module can't inherit functionality of other modules.

with a simple import at the begining you can have whatever functionality you wish. I dont think we are going to need a sv_links_common and many sv_links_classes that need to inherit this 

> It is simpler to redo static class to normal class for using its instances rather then redo module in this way.

There should be no instances of this code, is unique for all nodetrees

> And probably most important thing you can't assign functionality of module to other classes fields. It would be more comfortable to have access to this functionality via node_tree.sv_links.some_method() but instead you have to import this module everywhere where you would like to use it.

you can use sverchok.core.links.some_method(node_tree,...) in the not-so-many places you may need to use it

core/handlers.py
Durman: I see nothing what can prevent to move this code to separate method.
If the goal is only to change the logic link changes detection there is no need in investigation all code of this PR. 
For example some other events can be used for changes detection. It will let avoid expensive comparison between old and new links. New logic can be more tricky and consist from tens of strings. It will be hard to search all places where new code should be pasted. And most likely new code will be separate function any case. And it will be much simpler if new code just replace code of one method.

node_tree.py
Durman: Do you think if you put update code of `monad` tree changes into place where nobody expects to find it is making code simpler?
If you are adding new functionality to monads you should place the code into monads. Hiding functionality just make monads more complex.

node_tree.py
Durman: > the sv_links is a global dictionary as the socket_data_cache affects all sverchok so it makes sense to be a global variable.

`sv_links` does not belong to all Sverchok. They belongs to node trees and nobody else. Socket data catch belongs to sockets. It is quite obvious. Making global in this way is mistake. We already has `update system` based on global variables. Nobody dared to improve it all years.

I have just remembered the words from legend movie:
"Unfortunately no one can be told what the Marix is. You have to see it to yourself."
So you just have to try using classes by yourself.

Durman:
I does not see purpose in further word battle. I estimate the quality of this PR as low. It is effected core of Sverchok so quality do matter. It of course brings new functionality but meanwhile it increases complexity of adding farther improvements. I would suggest to improve knowledge of programming in general before touching this things.

vicdoval:
@Durman Thanks for your feedback, I studied architecture and later on self-taught programming, I am more a creative coder than a back-end programmer, so maybe I don't use the most advanced techniques but I do my best. 
I have been waiting for very long time for a update in the Sverchok update-system and as no-one seemed to have time/courage to do it I tried it my self. When I started I discovered why is hard to change, there are so many areas involved... its from far the hardest programming problem I faced ever.
Finally I made it work in every area. (This is the actual state, on my view it is missing code-documenting and some refactoring)
The updates this PR  avoids for me are precious and i think it is not a simple feature, I don't now if you tried it (apart from reading the code) but seems to work flawlessly. Also i see it as a good point to start reducing other updates. 

I can keep it to myself and keep updating the changes from the master branch until some day when someone faces again this problem and comes up with a higher-quality solution.

But as it does not break anything it could even be dismissed by the next brave coder in the next iterations and take the previous system as starting point. (just removing 8 lines or so)

I would also point that I have used and mimic  the strategies I found in the actual update-system (I don't know if you evaluate it as high or low quality but its been working for all of us for many years)

So what does the crew think @zeffii  @portnov  @nortikin @DolphinDream  @ly29?

portnov:
About Travis: I'd start with clearing caches, but for some reason I can't do it, don't see the link in the menu. Maybe github or travis issue, or maybe permissions.
@zeffii maybe you can do it?
https://docs.travis-ci.com/images/caches-item.png

portnov:
About the topic... Well, I'm afraid I don't know sverchok's update system well enough to have a strong opinion on this matter...

zeffii:
@vicdoval tricky situation here. I've rewritten this comment several times. This is the shortest version.

there are parts of this PR that can definitely be cherry picked and committed anyway
 - the sv_free stuff
 - using the ng.tree_id rather than name
 - whitespace


I do share concerns of increased complexity, though certainly without the strained-tones above. I don't know (_a priori_) if it's worth implementing that kind of bookkeeping detail to hold-off  updates. But you evidently want a certain effect, and i'm sure you don't really care how that is achieved ultimately. You chose brute force. Sverchok is full of brute force :)

it might be an idea to do as you propose: to test these changes more extensively for a while and staying in sync with master, but i suspect that doing so will feel like a sour grape.

ly29:
I think the old update system, which I am mostly still to blame for, is overly complicated and could be much simpler. But having said that, something is to be said for code that keeps solving an actual problemy over the years...

My advice would be to try to makes things simpler before introducing addiontional complexity.  

The code quality (not robustness or functionality) of the existing update system is at best ok.

node_tree.py
vicdoval: @Durman excuse me for resisting your suggestions. I understand we are both here for a greater good (and also to have fun) We both give our best and is good to hear voice of someone trying to help (specially in a topic they know). The ego resists, but later understands.  So thanks again for the feedback and for pushing Sverchok to be better.

This said I don't mind to keep on improving this PR to see if it can achieve all our needs.

I have refactored the code following your suggestions and it would be  very nice if we could keep on the conversation about it

vicdoval:
@zeffii I perfectly understand your concerns because is a lot of 'bookkeeping' and probably with time we will find a way to do it better but check this test.
Now in master
New NodeTree  and load the example 'Adaptative Voronoi panel'
Create a Line node 1.27 secs (+1.27 secs)
Delete this node  2.54 secs(+1.27 secs)
Create Segment Node 3.82 secs (+1.27 secs)
Right click on in "Connect to Viewer Draw"  6.40 secs (2 updates +2.58)

So 6.40 seconds of your life waiting to add a line.

With this branch
New NodeTree  and load the example 'Adaptative Voronoi panel'
Create a Line node 0.0009 secs (+0.0009 secs)
Delete this node  0.0018 secs secs(+0.0009 secs)
Create Segment Node 0.0027 secs (+0.0009 secs)
Right click on in "Connect to Viewer Draw"  0.0147 secs (2 updates +0.012 secs)

So in this case bookkeeping saved 6.38 seconds of life
 
Seconds are precious aren't they?

Anyway I think is a good idea to keep it as clean as possible to change it for a better method whenever it appears

zeffii:
> Seconds are precious aren't they?

yes!  yes they are.  I code on a low end machine, when stuff is slow i'm the first to feel it. :)

btw, any chance you could switch off the auto pep stuff relating to whitespace, this way the PR touches fewer files. it's nice that the files are tidied up, but it's not the goal of the PR. not a hard concern, just wrecks my head a little.

zeffii:
do you think you could make these additions optional, like from a setting in preferences. something like "extended update system" (On, Off) . you'll get more people testing it, and more likely to land it.

zeffii:
because... those are massive time differentials.

node_tree.py
```diff
@@ -709,8 +766,16 @@ def sv_copy(self, original):
         
     def free(self):
```
zeffii, line 764: if a node is removed through the act of an undo, `free` is not called, this is why i added the lame solution to the undo handler. incase you were wondering :)

vicdoval: The Blender undo system should pop a message saying "just undo if is the only choice". :)
If the `free ` function is not called then if there where any socket cache about that node it wont be deleted until the file is closed (as the no-users node-trees :) ) , not a big deal, just a secondary effect...

vicdoval:
Weirdly I place back the white-spaces so the PR keeps only the real changes. :)
Now @zeffii you can tell me if you would like to keep both options with conditional statements in nodetree update, node init, node free, and the  handlers.. it can be done but also adds some noise 

Also would be nice to fix the travis thingy if you know anything about it...

node_tree.py
zeffii: it's a corner case. doesn't bother me, i consider it a bug that `free` is not called by blender, but i got nowhere when asking devs about it.

zeffii: i recall it was needed to remove stale drawing if the last node added was a VD

zeffii:
in reverse

> Also would be nice to fix the travis thingy if you know anything about it...  

i don't , but i'll have a read.

>  Weirdly I place back the white-spaces so the PR keeps only the real changes. :)

thanks, i think you understand :)

> Now @zeffii you can tell me if you would like to keep both options with conditional statements in nodetree update, node init, node free, and the handlers.. it can be done but also adds some noise

it scares me a little, but i'm happy if you merge this as is. But do a release of the current sverchok with its current update system first, if stuff needs meticulous comparing .

zeffii:
it does look like the yaml config should be set from master anyway, feel free to overwrite it.

vicdoval:
This one is cool everything works but Travis says that in the "adaptative voronoi panel" example
```
    x = self.map_bounds(donor.min_x, donor.max_x, v[X])
  File "/home/travis/build/nortikin/sverchok/installation/blender/2.82/scripts/addons/sverchok/nodes/modifier_make/adaptive_polygons_mk2.py", line 497, in map_bounds
    k = 1.0 / (max - min)
ZeroDivisionError: float division by zero
```

When I open the file in 2.82a everything works and no errors are plotted. Any ideas?

vicdoval:
I could change it to `k = 1.0 / math.max((max - min), 1e-6)`  if @portnov  agrees

nodes/viz/viewer_idx28.py
```diff
@@ -157,11 +157,6 @@ def draw_buttons_ext(self, context, layout):
                 colx.scale_x = little_width
                 colx.prop(self, colprop, text="")
 
-    def update(self):
```
zeffii, line 160: compare to vd_experimental..
```python
    @property
    def fully_enabled(self):
        return "attrs" in self.inputs

    def update(self):
        if not self.fully_enabled:
            return

        try:
            socket_one_has_upstream_links = self.inputs[0].other
            socket_two_has_upstream_links = self.inputs[1].other

            if not socket_one_has_upstream_links:
                callback_disable(node_id(self))
        except:
            print('vd basic lines update holdout', self.n_id)
```

it's fully enabled when the last socket is created ( "text" in self.inputs ) and
`socket_one_has_upstream_links = self.inputs[0].other`

-------------------

not sure why this update function was so b0rked.

vicdoval: you mean that could be erased? Or that i should check it?
I did not find any problem with the VD experimental. The Viewer index was disabling itself every update and it was not a good thing if you are not updating all the time

zeffii: i was being too economic with my comment. good catch btw that this update function existed in its current state. i've restated it to something more in line with what the other drawing nodes do.  (this also explains the flickering i noticed the other day.. haha )

https://github.com/nortikin/sverchok/blob/476f26dee0f3faf5ebfea52464d4b241e3a1ae0a/nodes/viz/viewer_idx28.py#L160-L172

vicdoval:
Ok I have created a release with the actual master state in case someone needs to study this changes later on and I will proceed to merge this changes

nodes/viz/viewer_idx28.py
zeffii: note to self. use fstring in self.debug, this probably throws an error at the moment if debug is on.

DolphinDream: I noticed UI flickring too (I‘m not sure it has the same root cause as your observed behavior) (see #3089 for details)

zeffii: no @DolphinDream , but also yes . If you push your most recent changes to your repo i'll have a look at the current code.

PR #3056, opened 2020-04-11
test works

zeffii:
this may be OK for now.


zeffii:
i think there's still a PR regarding an alternative replace_socket function, and context sensitive throttling..

PR #3057, opened 2020-04-11
alternative, preferred

zeffii:
using the update decorator

### DIFF
--- a/nodes/generators_extended/ellipse_mk2.py
+++ b/nodes/generators_extended/ellipse_mk2.py
@@ -19,7 +19,7 @@
 import bpy
 from bpy.props import BoolProperty, IntProperty, FloatProperty, EnumProperty
 
-from sverchok.node_tree import SverchCustomTreeNode
+from sverchok.node_tree import SverchCustomTreeNode, throttled
 from sverchok.data_structure import (match_long_repeat, updateNode, get_edge_loop)
 from sverchok.ui.sv_icons import custom_icon
 from sverchok.utils.sv_transform_helper import AngleUnits, SvAngleHelper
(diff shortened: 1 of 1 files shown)


zeffii:
maybe even
```python
    @throttled
    def update_mode(self, context):
        ''' Update the ellipse parameters of the new mode based on previous mode ones'''

        if self.mode == self.last_mode:
            return

        #              from             to
        switch_state = (self.last_mode, self.mode)
        a = self.major_radius
        e = self.eccentricity
        fl = self.focal_length
        
        if switch_state == ("AE", "AB"):
            self.minor_radius = a * sqrt(1 - e * e)

        elif switch_state == ("AC", "AB"):
            c = min(a, fl)
            self.minor_radius = sqrt(a * a - c * c)

        elif switch_state == ("AB", "AE"):
            b = min(a, self.minor_radius)
            self.eccentricity = sqrt(1 - (b * b) / (a * a))

        elif switch_state == ("AC", "AE"):
            self.eccentricity = fl / a

        elif switch_state == ("AB", "AC"):
            b = min(a, self.minor_radius)
            self.focal_length = sqrt(a * a - b * b)

        elif switch_state == ("AE", "AC"):
            self.focal_length = a * e

        self.last_mode = self.mode
        self.update_sockets()

        # there's an automatic updateNode call from the decorator at the end.
        # people aren't going to be repeatedly pressing the same mode enum 
```

DolphinDream:
So, I take that with @throttle I won’t have to do the trick to inhibit the props update (e.g. self.updating flag can be removed). Correct?

DolphinDream:
Actually it seems that the properties updated in the update_mode (currently only one is updated at a time) will still trigger an update call for those props. (the idea of the updating flag was that changing the mode shouldn’t trigger an update since the ellipse is the same regardless of the mode).

DolphinDream:
I added the @throttle and your code optimization. I kept the updating flag.

zeffii:
hmmm, if you are explicitly setting other properties inside a property update function, then yeah it's not crazy to also do a little bit of local "update prohibiting" like your "is_updating".  The throttle is mostly for the new sockets/replace as they will trigger an update while your node is "mid transition" between states.

PR #3061, opened 2020-04-11
Add Quaternion in & out nodes MK2 with Angle Units support

DolphinDream:
- Add angle units support via the AngleHelper class
>

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

SV-QuaternionInMK2

SV-QuaternionOutMK2


DolphinDream:
@zeffii I know you’re getting tired of this migration topic (but I think migration is important to get right since it can be a bit frustrating for old, obsolete trees to be updated and reused if manual migration & relinking is the only way). So, since you know more about migration, here’s another issue I have:

The quaternion in/out nodes (which i’m promoting to MK2) have various input/output sockets that change based on mode. They are all created initially, but then hidden and shown as needed. When migrating old quaternion nodes it seems that the input/output links are not preserved. I suspect that this has something to do with the way the relinking is established in conjunction with the fact that these nodes hide/show their sockets based on mode (which may mess up socket order?). If you can give me some pointers how to get this migration right I’d really appreciate it.

Here’s the demo showing the link drop:

SV-QuaternionsMK2-demo1

vicdoval:
You can place this function in the node, as you can imagine it was created to make changes in the new node before linking: https://github.com/nortikin/sverchok/blob/f69905427c1434e1f984f3b87e2e2c844c9a0aff/nodes/transforms/rotate_mk3.py#L286-L287

DolphinDream:
Thanks @vicdoval 
Any ideas though why i have to add this to the node to migrate_From to preserve this property?

```
    def migrate_from(self, old_node):
        self.angle_units = old_node.angle_units

```
The old node has the angle_units defined in the node itself.. the new node has it defined in the SvAngleHelper class from which it inherits from. I would assume the property having the same name should have been copied automatically during the migration without having to explicitly state it in migrate_from.  I wonder if there is either a bug in the migration code or whether this is just how python works (e.g. maybe the props are considered different if they are not defined in the same class, but inherited)

zeffii:
> new node has it defined in the SvAngleHelper class from which it inherits

There is only spot where the "non cosmetic" params (height, location etc...) are stored.

https://github.com/nortikin/sverchok/blob/f9618d1f003cd2033e5534b0bc144b289a56d9d8/ui/nodes_replacement.py#L103

if `old_nodes.items()` does not include "angle units" it would explain why you must push it manually.  `old_nodes.items()` is a sparse representation of the parameters on the node that changed.

```
>>> n = nodes['Quaternion In']   # old version
>>> n.items()   # without changing any props
[]

>>> n.items()   # i just changed the mode
[('mode', 1)]

```
it's sparse for a reason, it makes little sense to copy X.A to Y.A  is they are both the same, generally speaking.  But here the defaults must be slightly different, even tho the desired end state in Y.A is the same.

zeffii:
the `migrate_from` is an attempt to offer a relatively simple way handle peculiar migration issues, you can help your future self / others by describing why in particular those parameters requires an extra push.

hopefully i've been able to explain.

zeffii:
i think if we get too complicated with the `node_replacement` function, we end up making a lot more code for relatively few nodes that will ever pass though it.

I don't know if it would help if we added extra checks to compare:

     X.annotations ->  Y annotations 
     X.items()
     testing for all annotations in both if any are different, but do not appear in X.items()

zeffii:
```python
n.__annotations__.keys()
>>> dict_keys(['mode', 'euler_order', 'angle_units', 'component_w', 'component_x', 'component_y', 'component_z', 'scalar', 'vector', 'angle_x', 'angle_y', 'angle_z', 'angle', 'axis', 'normalize'])
```

Issue #2720, opened 2019-12-08
API changes 2.8+


Durman:
https://docs.blender.org/api/current/change_log.html

`bpy.types.NodeSocket`
Added
`bpy.types.NodeSocket.draw_shape`
`links` - List of node links from or to this socket. **Warning: takes O(len(nodetree.links)) time.** (readonly)

`bpy.types.BoolProperty / FloatProperty / IntProperty`
Added
`bpy.types.BoolProperty.array_dimensions`

`bpy.types.UILayout`
a lot of [changes](https://docs.blender.org/api/current/change_log.html#bpy-types-uilayout)

new - `mathutils.geometry.delaunay_2d_cdt`

Durman:
![2019-12-10_12-50-05](https://user-images.githubusercontent.com/28003269/70513740-a1d7e800-1b4b-11ea-9de4-67c0e6fa8b7b.png)

Durman:
Looks like `bmesh.ops.inset_individual` does not work as intended.
<details> 
  <summary>cod</summary>  

```Python
import bpy
import bmesh
from bmesh.ops import inset_individual

verts = [[0,0,0], [1,0,0], [1,1,0], [0,1,0]]
faces = [[0,1,2,3]]

bm = bmesh.new()
[bm.verts.new(co) for co in verts]
bm.verts.ensure_lookup_table()
[bm.faces.new([bm.verts[i] for i in face]) for face in faces]

inset_individual(bm, faces=list(bm.faces), 
                     thickness=0.5, 
                     depth=0.0, 
                     use_even_offset=True,
                     use_interpolate=True, 
                     use_relative_offset=False)

me = bpy.data.meshes.new("Inset face")
bm.to_mesh(me)
bm.free()
obj = bpy.data.objects.new("Inset face", me)
bpy.context.collection.objects.link(obj)
```

</details>

![2019-12-11_09-09-23](https://user-images.githubusercontent.com/28003269/70593355-2aab5e00-1bf6-11ea-92f5-9be6b00e3bad.png)

Durman:
Blender 2.81
![2019-12-16_20-13-15](https://user-images.githubusercontent.com/28003269/70923002-86377a80-2040-11ea-9e58-ffbcf55a2571.png)

Durman:
If I have understood this correctly, now, during update event `is_linked` status of socket can be False and have links and vice versa. 
There are two possible cases: 
- when socket gen new link(s): `is_linked` is False, links are True. 
[...]

PR #3064, opened 2020-04-12
Curve Mapper

vicdoval:
We deserve it.
Using the mythical hack of using a RGB node, in this case inside a node_group.
The node uses node_id to track its related curve and to delete the shader node if the SV node is deleted.
Disclaimer: The changing of the curve does not update the node, you have to click update or any other Sverchok trigger
image
image

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


Durman:
Did you see this?
https://github.com/nortikin/sverchok/issues/2720#issuecomment-572958031

zeffii:
i wonder if there's a way to tag the shadernode, and trigger the sv nodetree associated with that shadernode, so UI interaction on the curve does update live.

vicdoval:
> Did you see this?
> [#2720 (comment)](https://github.com/nortikin/sverchok/issues/2720#issuecomment-572958031)

Yes, but is about bevel profile, different idea that RGB curves

vicdoval:
Editing the curve does not show any info in the Info panel.... maybe adding a handler callback that compares the actual curve with a stored one... in a timer.... looks like pretty complicated and adds processing weight...

zeffii:
if only the despgraph update handler had a property we could query, that would return the thing/object that triggered the update in the first place.

Durman:
According description of new widget it is more suitable for using via Python API. Probably it worth trying?

> New potential use cases for the Profile Widet The profile widget could be used to bring more power to any generated mesh (possibly curves, wireframes..) And it's accessible from Python so any addon can use it.

vicdoval:
> According description of new widget it is more suitable for using via Python API. Probably it worth trying?
> 
> > New potential use cases for the Profile Widet The profile widget could be used to bring more power to any generated mesh (possibly curves, wireframes..) And it's accessible from Python so any addon can use it.

I could not find any example of this widget working with out a hack of somekind:
https://devtalk.blender.org/t/how-do-you-add-custom-profile-prop-to-operator-like-in-bevel-operator/11521/4
https://devtalk.blender.org/t/how-to-define-a-curveprofile-property/12279

zeffii:
from depsgraph handler...
```
id_type_updated(id_type)
'NODETREE', / material / object
```
https://docs.blender.org/api/master/bpy.types.Depsgraph.html#bpy.types.Depsgraph.updates

zeffii:
we could have something in the current main handler, that checks the currently active sverchok nodetree, and currently active node, and if that happens to be CurveMapper node, then it could "consume" the event and trigger process in that nodetree.

zeffii:
with a "live update" toggle in the UI,   if toggled true then updates would be consumed..

vicdoval:
```
id_type_updated(id_type)
'NODETREE', / material / object

```
@zeffii  from this function i can only get a True Value if the changes happened in a node-editor (it does not matter what kind of node-editor) and a False if the update cames from somewhere else... no to much information 😢

vicdoval:
> we could have something in the current main handler, that checks the currently active sverchok nodetree, and currently active node, and if that happens to be CurveMapper node, then it could "consume" the event and trigger process in that nodetree.

The problem is that updating the curve does not turn the node to the active node.... 

We could implement something like the live-in object caller... that compares the curve with the previous curve every 300 ms or something and if detects it has changed then triggers the node update, but I'm not sure it is a good idea

zeffii:
> The problem is that updating the curve does not turn the node to the active node....

yeah, user must make the node active first..  but that would be tolerable :)

> We could implement something like the live-in object caller... that compares the curve with the previous curve every 300 ms or something and if detects it has changed then triggers the node update, but I'm not sure it is a good idea

once you push this node, i will do `mod` and see if it's worth it.  but i definitely don't want a timer operator or rely on animation being active.

vicdoval:
@zeffii  so are you ok with me pushing this node in its actual state?

zeffii:
it will need to implement 
```
storage_set_data
storage_get_data
```
for iojson and things that rely on it... like presets and monad.

zeffii:
i think you understand why i never made this node :)

vicdoval:
"Perfect is the enemy of the good" :)

zeffii:
> rom this function i can only get a True Value if the changes happened in a node-editor (it does not matter what kind of node-editor) and a False if the update cames from somewhere else

yeah, it's kind of useless. it's almost like they don't care :)

portnov:
Found this accidentially: https://blender.stackexchange.com/a/61829

vicdoval:
> Found this accidentially: https://blender.stackexchange.com/a/61829

I read it when I was coding this node and even used the idea of the node-group to hold the RGB Nodes.
Anyway thanks!

vicdoval:
So anyone against merging this non-real-time node?

nortikin:
just do it

zeffii:
@vicdoval  in storage_set_data and storage_get_data you will need to implement these functions, like I do for `snlite - ui curve manual demo`
```python
def get_rgb_curve(material_name, node_name):
    '''
    ShaderNodeRGBCurve support only
    '''
    m = bpy.data.materials.get(material_name)
    node = m.node_tree.nodes.get(node_name)

    out_list = []
    for curve in node.mapping.curves:
        points = curve.points
        out_list.append([(p.handle_type, p.location[:]) for p in points])
    
    x = 'ShaderNodeRGBCurve'
    return dict(mat_name=material_name, node_name=node_name, bl_idname=x, data=out_list)


def set_rgb_curve(data_dict):
    '''
    ShaderNodeRGBCurve support only
    '''

    mat_name = data_dict['mat_name']
    node_name = data_dict['node_name']
    bl_id = data_dict['bl_idname']

    node = get_valid_node(mat_name, node_name, bl_id)

    node.mapping.initialize()
    data = data_dict['data']
    for idx, curve in enumerate(node.mapping.curves):

        # add extra points, if needed
        extra = len(data[idx]) - len(curve.points)
        _ = [curve.points.new(0.5, 0.5) for _ in range(extra)]

        # set points to correspond with stored collection
        for pidx, (handle_type, location) in enumerate(data[idx]):
            curve.points[pidx].handle_type = handle_type
            curve.points[pidx].location = location
    node.mapping.update()

```

zeffii:
otherwise the points/handles won't survive traversing the data black hole that is iojson.

vicdoval:
i don't understand where i need to place those functions

PR #3065, opened 2020-04-13
Add Spiral node MK2 with angle support + new feature + cleanup

DolphinDream:
- Add angle units support via the AngleHelper class
- Add separate feature to separate arms in output
- Various optimization
- Code cleanup
- Variable renaming to match python style
- sanitize radius inputs
- add custom icon
- move node into generator extended category
>

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

SV-SpiralMK2-UI


nodes/generators_extended/spiral_mk2.py
```diff
@@ -0,0 +1,655 @@
...
+    def process(self):
+        outputs = self.outputs
+        # return if no outputs are connected
+        if not any(s.is_linked for s in outputs):
+            return
+
+        # input values lists (single or multi value)
+        inputs = self.inputs
+        input_R = inputs["R"].sv_get()[0]  # list of exterior radii
+        input_r = inputs["r"].sv_get()[0]  # list of interior radii
```
Durman, line 592: Using capitalized letters in names of variables is against Python convention.

nodes/generators_extended/spiral_mk2.py
```diff
@@ -0,0 +1,655 @@
...
+
+        verts_list = []
+        edges_list = []
+        for R, r, e, t, n, s, h, p, a in zip(*parameters):
+            p = p * au
+            arm_verts = []
+            arm_edges = []
+            for i in range(a):  # generate each arm
+                pa = p + 2 * pi / a * i
+                settings = [R, r, e, t, n, s, h, pa, f]  # spiral settings
```
Durman, line 627: I would suggest using named tuple here.

nodes/generators_extended/spiral_mk2.py
```diff
@@ -0,0 +1,655 @@
...
+
+        # integral from l1 to l2
+        u = l1
+        du = (l2 - l1) / M
+        for m in range(M + 1):
+            u = u + du  # u = [l1, l2]
+            phi = u * u * pi / 2
+            x = x + cos(phi) * du
+            y = y + sin(phi) * du
+        l1 = l2
```
Durman, line 306: This quite magical code if someone have to fix something he can have difficulties with code interpretation. More meaningful names of variables are appreciated.

DolphinDream: > This quite magical code if someone have to fix something he can have difficulties with code interpretation. More meaningful names of variables are appreciated.

agree. this spiral is mostly experimental however.  did my best to optimize for speed since this usually involves integrals and left it in a state where it was mostly working. I’ll make another pass to improve it in the future, and will probably try to come up with better notation.

nodes/generators_extended/spiral_mk2.py
DolphinDream: you’re right. I did **most** of the conversion. I only left these R/r ones since I didn’t want to add longer names like input_major_radius & input_minor_radius. it’s not gonna be 100% perfect. Much of the SV code is not 100% compliant at this point. e.g. updateNode() will unlikely change since it’s so pervasive.

nodes/generators_extended/spiral_mk2.py
DolphinDream: Can you give me an example of what you mean?

Durman: Just as example:
```Python
from typing import NamedTuple

class NodeSettings(NamedTuple):
    vertices: list
    polygons: list
    some_property: float = 0.0

node_settings = NodeSettings([v1, v2, v3], [polygon], 1)

extrude(node_settings.vertices, node_settings.polygons, node_settings.some_property)
```
More information here: https://docs.python.org/3/library/typing.html#typing.NamedTuple

PR #3070, opened 2020-04-14
Node update decorator

zeffii:
i don't know if this is needed, or if it's just extra complexity.

it's a decorator for the node's `update` function. For most nodes we don't need to implement the update function, only for nodes which change their ui according to user interaction with sockets.

When we do implement the `update` function, we have a few checks at the top of the function that must be done before any real dynamic UI code can be run.

Usually the code in the update function consists out of testing to see if the last output was generated, and perhaps if certain sockets are connected. The example i'll use is the `combinatorics` node This PR contains the `poll_node` decorator, which wraps the update function, the decorator will call a function called "sv_poll", which will return true is you want to avoid calling update. (in that sense it is in reverse to to normal operator poll)

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -136,7 +136,7 @@ def throttled(func):
     return wrapper_update
 
 
-def throttle_node(func):
+def throttle_node_update(func):
     """
     this function is used to reject calls to node.update
 
@@ -153,7 +153,7 @@ def throttle_node(func):
                 ... usually something like :   
                 ...      if 'last_output' in self.outputs: return True
 
-            @throttle_node
+            @throttle_node_update
             def update(self):
                 ... stuff you do with a highly dynamic ui node
 
--- a/node_tree.py
+++ b/node_tree.py
@@ -136,31 +136,31 @@ def throttled(func):
     return wrapper_update
 
 
-def throttle_node_update(func):
+def poll_node(func):
     """
     this function is used to reject calls to node.update
 
     use as a decorator:
 
-        from sverchok.node_tree import SverchCustomTreeNode, throttle_node
+        from sverchok.node_tree import SverchCustomTreeNode, poll_node
 
         class node:
 
 
-            def hold_check(self):
+            def sv_poll(self):
                 ... test here to return True, if you want to avoid executing the update
                 ... function too early.
                 ... usually something like :   
                 ...      if 'last_output' in self.outputs: return True
 
-            @throttle_node_update
+            @poll_node
             def update(self):
                 ... stuff you do with a highly dynamic ui node
 
 
     """
     def wrapper_node_update(self):
-        if not self.hold_check():
+        if not self.sv_poll():
             func(self)
 
     return wrapper_node_update
--- a/node_tree.py
+++ b/node_tree.py
@@ -82,6 +82,10 @@ class SvLinkNewNodeInput(bpy.types.Operator):
         return {'FINISHED'}
 
 
+
+
+
+
 @contextmanager
 def throttle_tree_update(node):
     """ usage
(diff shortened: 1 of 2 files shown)


zeffii:
@vicdoval  @Durman  @portnov  @nortikin 
opinions?

Durman:
I think what we could create our own update system. Benefits of it will be creating only those events which are needed to Sverchok. For example we can create event of creating links manually and creates methods which should be called only during this event. Also it will let to place all code in one place (I think) instead of creating pieces of code here and there.

Durman:
In this particular case I would prefer to see `rebuild_interface` method or even `rebuild_interface_upon_editor_changes`. This method can be called from `update` method of `Common` class. Using decorator in this case makes everything only more confusing I think.

zeffii:
in the end we'll still need to write a few lines of code, for a select set of nodes, that enables us to prohibit the update function from doing things that will raise errors (ie sockets don't exist yet).

> Using decorator in this case makes everything only more confusing I think.

i don't agree, i want to compartmentalize the "polling" code, out of the `update` function. Just like operators. Operators are only executed if 
1. they don't have a poll
2. their poll returns True  ( i would return False....so that does bug me.. )

i'd be happy to "rename" the update function to 'handle_ui_changes`, because that seems to be the most prominent functionality of the current update function.

Durman:
Why each node should have each own `pool` method?

zeffii:
only nodes that would otherwise implement `update` would optionally implement 'poll`, but most nodes that implement `update` do have a few lines at the top that cause the update function to end early, those nodes would implement the `poll` function, and place the "end early" code inside the poll.

zeffii:
my only issue here is that it's more code...
```python
# from this

def update(self):
    if statement_1:
        return True
    if statement_2:
        return True

    ... handle_ui_changes_here ...


# to this

def poll(self):
    if statement_1:
        return True
    if statement_2:
        return True

def handle_ui_changes(self):
    ... handle_ui_changes_here ...
```

Durman:
I meant whether not all nodes which redraw their UI upon editor changes would have equal `pull` methods? 

I expect that this method should return False if cause of update event is redrawing UI of current node otherwise it should return True. If your suggestion is to create a lot of same methods it is definitely not worth dooing.

zeffii:
>  If your suggestion is to create a lot of same methods

the number of nodes that implement update is not large, but each node that implements update also has a few lines of code at the top of update to determine if the result of the update code needs to be executed.

all i'm considering is to make the poll function explicitly separate from the update function, it's not a big deal

zeffii:
i suppose another solution is a decorator wherein you write the last socket name

```python
@requires_socket("Result")
def update(self):
    ...
```

and `requires_socket` would test self.outputs['Result'], before executing `update`

zeffii:
using it would be entirely optional... but i would gradually update all my own nodes :)

Durman:
It is better to do this because `update` method is either way bad place for redrawing UI.
```Python
def update(self):
    if statement_1:
        return
    if statement_2:
        return

    handle_ui_changes()

def  handle_ui_changes(): ...
```

And in this case your solution should be related with redrawing method. like this:

```Python
def update():
    handle_ui_changes()

@pull
def handle_ui_changes(): ...
```
But either way I think using decorator for `throttling` method is not explicit enough. It feels like use case of decorators lays in different domain. 

And have look at the solution in `switch` node:
https://github.com/nortikin/sverchok/blob/9b3bb084bc6ad3680f13a9888e7e8668db3ddf85/nodes/logic/switch_MK2.py#L164-L166

I see next possible improving here:
```Python
class SverchCustomTreeNode:
    def update():
        if hasattr(self, 'rebuild_ui'):
            if not self.id_data.skip_tree_update:
                self.rebuild_ui()
```

zeffii:
a simple "sverchok" way to address this is:

```python
class SverchCustomTreeNode:
    def update(self):
        if hasattr(self, 'sv_update'):
            if hasattr(self, 'sv_poll'):
                if self.sv_poll():
                    self.sv_update()
            else:
                self.sv_update()
```

Issue #3082, opened 2020-04-17
Origins Node crashing Blender


vicdoval:
Make a Box node of 100x100x100 divisions
create a Origins Node.
Link Vertices
Link Edges
Blender Crashes
Tried in master in 2.82 and 2.83

zeffii:
why is this node not using `bmesh_from_pydata` ?

Durman:
Looks like this line of code is crashing reason.
https://github.com/nortikin/sverchok/blob/e9cfe0c77f07844564a9121ba5821cad96701edf/nodes/analyzer/origins.py#L99

> why is this node not using `bmesh_from_pydata` ?

I was still experimenting with `bmesh` when I was coding the node.

Durman:
It looks like `bm.free()` deletes some points (not all) from memory even if they linked to variable.

https://github.com/nortikin/sverchok/blob/e9cfe0c77f07844564a9121ba5821cad96701edf/nodes/analyzer/origins.py#L60-L62

![2020-04-18_09-50-30](https://user-images.githubusercontent.com/28003269/79629347-1c4bf080-815a-11ea-871a-3ae8ae60459c.png)

PR #3084, opened 2020-04-18
Fix origins node crashing

Durman:

#3082

## Preflight checklist

- [x] Ready for merge.


Durman:
@zeffii Now I'm using `sv_bmesh_utils` module.

Issue #2111, opened 2018-02-05
Enhance Viewer Draw node by adding a “visibility layer" setting


DolphinDream:
Here’s an idea.. I often time want to show/hide multiple viewer draw nodes outputs at the same time and was wondering if adding a **visibility layer** option to the node would be beneficial for that purpose. You could have certain VD nodes set to display only on certain layers, but not others and use blender’s visibility layer feature to simultaneously turn on/off certain VD outputs.

I think this would a be similar to objects behavior (e.g objects have layer visibility).

![sv-viewerdraw-visibilitylayers](https://user-images.githubusercontent.com/2083719/35786866-34375ddc-09f8-11e8-82f6-10c497d341d2.jpg)

zeffii:
I was thinking of using a Collection property per `node_group`. So the VD node only stores the name of the layer it is a child of. The node tree would have a mechanism to add remove layers by name, Not sure how this would work with io/json but something would be possible. Also the problem arrises of how to deal with nodes that are imported into an exist node tree.    these are not nice problems. :)    (whch i suppose would be simlified if there was no named layer system... but i have the anonymous layering.. )

worth some experimenting.. but don't be too jaded if it takes a while to materialize.

DolphinDream:
Importing could be handled via default props (similar to the ones we did for the socket expander). If the node setups (gists) don’t have the “layer visibility” prop set upon import the prop is set to the default “all layers”. Exporting can be sparse to only save this prop if it deviates from default value.

DolphinDream:
Sandboxing the layers in VDMK2. here’s what it looks like. For now I added the layer selection to the property panel only.

![sv-vdmk2-layers](https://user-images.githubusercontent.com/2083719/36063060-07ddce54-0e46-11e8-99b7-dd50d273d5e6.png)

DolphinDream:
And a little demo showing how it would play with blender’s visibility layers.
I used 1-10 layers to display individual components, and 11-20 to combine various components into one layer. 

![sv-viewerdraw-layers-demo2](https://user-images.githubusercontent.com/2083719/36063109-94cc1460-0e46-11e8-89b0-3eaaa8710cd5.gif)

zeffii:
no. please no PR for this yet.

zeffii:
the implementation should be independent from the Blender layer system.

zeffii:
because VDMK2 is such a major node, all changes should be considered even more slowly than normal node changes :)

[...]

Issue #3044, opened 2020-04-07
mixin for node_dict


zeffii:
some nodes implement specific storage, in data structures which are not convenient to store in `bpy.props`, traditionally we've used ` self.node_dict = {} `

as discussed in comments earlier, it's a little convoluted to implement a unique node_dict each time.

I propose a mix-in, and using a `self.node_dict` as the dict name, to be constant with existing nodes , and maybe getting those nodes to use the mix-in.

just thinking out loud here..
```python
class SvDictStorage:

    n_id = bpy.types.StringProperty(name="n_id")
    node_dict = {}

    def node_dict_data_func(self):
        # you must implement this function to ensure that the nodedict will get a valid storage
        # upon open blend
        pass

    def node_dict_set_data(self, data)
        self.node_dict[self.node_id] = {}
        self.node_dict[self.node_id]['data'] = data
        return data

    def node_dict_get_data(self):
        ndict = self.node_dict.get(self.node_id)
        if not ndict or not ndict.get('data'):
            return self.node_dict_set_data(self.node_dict_data_func())
        
        return ndict.get('data')
        
```
it would need to contain some of these..

Durman:
I would suggest to use `@property` decorator for set and get functions. Also `save` function should be implemented for saving data into Blender during saving event. 

Also there should be raising `AttributeError` somewhere in case if requested attribute is not exist. Or probably some default value could be returned.

`Node dict` will be used for keeping only one attribute per node?

zeffii:
> Node dict will be used for keeping only one attribute per node?

no, not necessarily

```python
self.node_dict[n_id]['data'] = ...
self.node_dict[n_id]['other_data'] = ...
```
i want to point out that you can make `data` a dict too, and stick anything inside it.
using `@property` would complicate setting specific multiple keys like `data`/`other_data`

it might be written has
```python
def node_dict_set_data(self, keyname, keyvalue)
    ...
```
-----

of course some errorhandling and recovery will be implemented.

zeffii:
i'd like to do this with as little code as possible, because then there is less to debug.
```python
class SvDictStorage:

    n_id = bpy.types.StringProperty(name="n_id")
    node_dict = {}

    def node_dict_data_func(self, key):
        """
        this function must return the data you wish to store in 
        
            node_dict[self.node_id][key] = ...
        
        to ensure that node_dict will get a valid storage upon open blend
        """
        pass

    def node_dict_set_data(self, key, value)
[...]

Issue #991, opened 2016-11-25
prop_search with filter


zeffii:
In essence it works like this:

This would only list those objects in the scene that have an _id property_ that the user wants..

```python
# ##### BEGIN GPL LICENSE BLOCK #####
#
#  This program is free software; you can redistribute it and/or
#  modify it under the terms of the GNU General Public License
#  as published by the Free Software Foundation; either version 2
#  of the License, or (at your option) any later version.
#
#  This program is distributed in the hope that it will be useful,
#  but WITHOUT ANY WARRANTY; without even the implied warranty of
#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
#  GNU General Public License for more details.
#
#  You should have received a copy of the GNU General Public License
#  along with this program; if not, write to the Free Software Foundation,
#  Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
#
# ##### END GPL LICENSE BLOCK #####

import bpy
from bpy.props import EnumProperty
from bpy.utils import register_class, unregister_class
from sverchok.node_tree import SverchCustomTreeNode
from sverchok.data_structure import (updateNode, replace_socket, enum_item as e)


class SvPropsearchFilterDemo(bpy.types.Node, SverchCustomTreeNode):
    ''' Get Object if it has a certain ID property '''
    bl_idname = 'SvPropsearchFilterDemo'
    bl_label = 'Demo propsearch filter'
    bl_icon = 'OUTLINER_OB_EMPTY'

    def pre_updateNode(self, context):
        ''' must rebuild for each update'''
        self.collection_name.clear()
        print(self.id_prop_name)
        if self.id_prop_name:
            for o in bpy.data.objects:
                if o.get(self.id_prop_name):
                    self.collection_name.add().name = o.name

        updateNode(self, context)

    id_prop_name = bpy.props.StringProperty(update=pre_updateNode)
    obj_name = bpy.props.StringProperty(update=updateNode)
    collection_name = bpy.props.CollectionProperty(type=bpy.types.PropertyGroup)

    def draw_buttons(self, context, layout):
        row = layout.row(align=True)
        layout.prop(self, "id_prop_name", text='id property')
        layout.prop_search(self, 'obj_name', self, 'collection_name', icon='OBJECT_DATA')

    def sv_init(self, context):
        self.outputs.new('SvObjectSocket', "Object")

    def process(self):
        sob = self.outputs['Object']
[...]

PR #3089, opened 2020-04-19
Viewer Layer node

DolphinDream:
Continue the developing of a **viewer layer node** (based on the #2111 conversation) and get a firsts incarnation of such node out there. This is an early stage developing (nothing much as been added since the issue was raised) and will need more brainstorming and playing around to get it to a polished state. 

Considering that since this issue was originally started (2.79) Blender 2.8x has revamped its layering/collection system we might think differently about how we can implement such a viewer layer node in SV. Can (should) we tap into the hierarchical collection system B28 offers? Let’s start the conversation :)

SV-ViewerLayerNode-Example1-loress

<details>
<summary>Live Demo</summary>

SV-ViewerLayersNode-demo1

</details>

ToDo (Feature List):

- [x] Make Layer names editable
- [x] Toggle visibility icon (“show/hide” eye icon)
- [x] Add option to remove viewers within layers
- [x] Add option to remove layers
- [ ] Update layer viewer lists when viewer nodes are removed from the node tree
- [x] Update layer viewer lists when viewer nodes are renamed
- [x] Make layers collapse/expand (add toggle to show/hide the list of viewers in a layer)
- [ ] Add option to add a layer to another layer (hierarchical layers, similar to BL collections)
- [x] Update layer visibility icon (status) based on its children status
- [x] Add option to avoid duplicate viewers in a layer

### DIFF
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -18,18 +18,55 @@ pre_running = False
 sv_depsgraph = []
 depsgraph_need = False
 
+subscription_owner = object()
+
+"""
+The nodes that need to subscribe to object/attribute changes should call this
+convenience function from within their setup_subscription method (typically
+invoked by the sv_setup_subscriptions below, after loading a blender file).
+"""
+def subscribe_to_changes(subscribe_to, message, callback):
+    debug("* handlers: subscribe_to_changes: %s", message)
+    bpy.msgbus.subscribe_rna(
+        key=subscribe_to,
+        owner=subscription_owner,
+        args=(message,),
+        notify=callback,
+        options={"PERSISTENT", }
+    )
+
+
+"""
+This is called automatically after loading a blender file (load_post) to give
+all the nodes the opportunity to (re)subscribe to object/attribute changes.
+
+A node that needs to subscribe to object/attribute changes should implement a
+setup_subscriptions method and within that method use the subscribe_to_changes
+convenience function above to setup all its necesssary subscriptions.
+"""
+def sv_setup_subscriptions():
+    debug("* handlers: sv_setup_subscriptions")
+
+    for node_tree in sverchok_trees():
+        for node in node_tree.nodes:
+            setup_subscriptions = getattr(node, "setup_subscriptions", None)
+            if callable(setup_subscriptions):
+                node.setup_subscriptions()
+
+
 def get_sv_depsgraph():
     global sv_depsgraph
     global depsgraph_need
     if not depsgraph_need:
-
         sv_depsgraph = bpy.context.evaluated_depsgraph_get()
         depsgraph_need = True
     return sv_depsgraph
 
+
 def set_sv_depsgraph_need(val):
     global depsgraph_need
     depsgraph_need = val
+
 
 def sverchok_trees():
     for ng in bpy.data.node_groups:
@@ -53,12 +90,14 @@ def has_frame_changed(scene):
 #  If at any time the undo system is fixed and does call "node.free()" when a node is removed
 #  after ctrl+z. then these two functions and handlers are no longer needed.
 
+
 @persistent
 def sv_handler_undo_pre(scene):
     from sverchok.core import undo_handler_node_count
 
     for ng in sverchok_trees():
         undo_handler_node_count['sv_groups'] += len(ng.nodes)
+
 
 
 @persistent
@@ -150,7 +189,6 @@ def sv_main_handler(scene):
     pre_running = False
 
 
-@persistent
 def sv_clean(scene):
     """
     Cleanup callbacks, clean dicts.
@@ -161,15 +199,14 @@ def sv_clean(scene):
     data_structure.sv_Vars = {}
     data_structure.temp_handle = {}
 
-@persistent
-def sv_pre_load(scene):
 
+@persistent
+def sv_handler_load_pre(scene):
+    """ Various things to do before (re)loading a blend file """
     sv_clean(scene)
-    set_first_run(True)
 
 
-@persistent
-def sv_post_load(scene):
+def sv_ensure_nodetree_data_after_blender_load():
     """
     Upgrade nodes, apply preferences and do an update.
     """
@@ -178,8 +215,7 @@ def sv_post_load(scene):
 
     # ensure current nodeview view scale / location parameters reflect users' system settings
     from sverchok import node_tree
-    node_tree.SverchCustomTreeNode.get_and_set_gl_scale_info(None, "sv_post_load")
-
+    node_tree.SverchCustomTreeNode.get_and_set_gl_scale_info(None, "sv_ensure_nodetree_data_after_blender_load")
 
     for monad in (ng for ng in bpy.data.node_groups if ng.bl_idname == 'SverchGroupTreeType'):
         if monad.input_node and monad.output_node:
@@ -215,6 +251,17 @@ def sv_post_load(scene):
             ng.update()
 
 
+@persistent
+def sv_handler_load_post(scene):
+    """ Various things to do after (re)loading a blend file """
+
+    # update node tree data
+    sv_ensure_nodetree_data_after_blender_load()
+
+    # update setup subscriptions
+    sv_setup_subscriptions()
+
+
 def set_frame_change(mode):
     post = bpy.app.handlers.frame_change_post
     pre = bpy.app.handlers.frame_change_pre
@@ -225,7 +272,7 @@ def set_frame_change(mode):
         post.remove(sv_update_handler)
     if sv_update_handler in pre:
         pre.remove(sv_update_handler)
-    #if sv_scene_handler in scene:
+    # if sv_scene_handler in scene:
     #    scene.remove(sv_scene_handler)
 
     # apply the right one
@@ -238,14 +285,13 @@ def set_frame_change(mode):
 handler_dict = {
     'undo_pre': sv_handler_undo_pre,
     'undo_post': sv_handler_undo_post,
-    'load_pre': sv_pre_load,
-    'load_post': sv_post_load,
-    'depsgraph_update_pre': sv_main_handler
+    'load_pre': sv_handler_load_pre,
+    'load_post': sv_handler_load_post,
+    'depsgraph_update_pre': sv_main_handler,
 }
 
 
 def register():
-
     app_handler_ops(append=handler_dict)
 
     data_structure.setup_init()
--- a/core/update_system.py
+++ b/core/update_system.py
@@ -198,7 +198,7 @@ def separate_nodes(ng, links=None):
             node_set_list[-1].add(n)
 
     found_node_sets = [ns for ns in node_set_list if len(ns) > 1]
-  
+
     if hasattr(ng, "sv_subtree_evaluation_order"):
         sorting_type = ng.sv_subtree_evaluation_order
         if sorting_type in {'X', 'Y'}:
@@ -357,7 +357,7 @@ def do_update_general(node_list, nodes, procesed_nodes=set()):
     timings = []
     graph = []
     gather = graph.append
-    
+
     total_time = 0
     done_nodes = set(procesed_nodes)
 
@@ -387,19 +387,19 @@ def do_update_general(node_list, nodes, procesed_nodes=set()):
             update_error_nodes(ng, node_name, err)
             #traceback.print_tb(err.__traceback__)
             exception("Node %s had exception: %s", node_name, err)
-            
+
             if hasattr(ng, "sv_show_error_in_tree"):
                 # not yet supported in monad trees..
                 if ng.sv_show_error_in_tree:
                     error_text = traceback.format_exc()
                     start_exception_drawing_with_bgl(ng, node_name, error_text, err)
-            
+
             return None
 
     graphs.append(graph)
     if data_structure.DEBUG_MODE:
         debug("Node set updated in: %.4f seconds", total_time)
-    
+
     return timings
 
 
@@ -522,7 +522,7 @@ def process_tree(ng=None):
 def reload_sverchok():
     data_structure.RELOAD_EVENT = False
     from sverchok.core import handlers
-    handlers.sv_post_load([])
+    handlers.sv_handler_load_post([])
 
 def get_update_lists(ng):
     """
--- a/docs/nodes/scene/scene_index.rst
+++ b/docs/nodes/scene/scene_index.rst
@@ -21,3 +21,4 @@ Scene
    particles_MK2
    node_remote
    selection_grabber_lite
+   viewer_layer
--- a/docs/nodes/scene/viewer_layer.rst
+++ b/docs/nodes/scene/viewer_layer.rst
@@ -1,0 +1,77 @@
+Viewer Layer
+============
+
+Functionality
+-------------
+
+This node helps you manage multiple viewer nodes in a centralized way.
+
+It allows you to create multiple layers (groups) to which you can add any of the supported viewer nodes [1] available in the node tree. Once the viewer nodes are added to the ViewerLayer node you can change their attributes (ON/OFF status, vert/edge/face display status and colors etc) directly from within the ViewerLayer node without having to navigate to each viewer node in the node tree to adjust those settings. Additionally, the ViewerLayer node also allows for various operations to be applied to all layers or all viewers in a layer at once (e.g. hide/show all, turn ON/OFF vert/edge/face display for all, collapse viewers in a layer or collapse all layers).
+
+Notes:
+[1] : Currently the supported viewer nodes are the "Viewer Draw" and the "Viewer Index" nodes.
+
+One usefulness of this node is that you can group together various viewer nodes that belong together in rendering some information in the viewport and you can easily turn ON/OFF all those viewers at once with miminal interaction (sometimes with just one click).
+
+
+Node Operations
+---------------
+The node level operations that apply to all layers in the node are:
+* Collapse/Expand all layers (via mix-status toggle button)
+* Turn ON/OFF vert/edge/face display (via corresponding mix-status toggle buttons)
+* Turn ON/OFF visibility of layers/viewers (via corresponding mix-status toggle buttons)
+* Add Layer to node (via "Add Layer" (plus list) button at the bottom of the node)
+
+
+Layer Operations
+----------------
+The layer level operations that apply to all viewers in the layer are:
+* Remove Layer from node (via "Remove Layer" (minus) button following the layer name)
+* Rename Layer (layers are allowed to have same name)
+* Collapse/Expand Layer (via "Expand" button in front of the layer name)
+* Turn ON/OFF (visibility status) of all viewers in Layer
+* Add Viewer to Layer (via "Add Viewer" (plus) button at the bottom of the layer)
+
+Notes:
+- There are no limitations on the number of layers a ViewerLayer node can create.
+- Once a layer is created, a "Add Viewer" (plus) button is shown at the bottom of the layer to allow viewer entries to be created for each layer. Once the number of viewer entries is the same as the number of available viewer nodes in the node tree, the "Add Viewer" button is hidden.
+
+
+Viewer Operations
+-----------------
+The viewer level operations that apply to each viewer in the layer are:
+* Remove Viewer (via "Remove Viewer" (minus) button next to the viewer entry)
+* Select Viewer (from drop down of avaialable viewer nodes)
+* Update ON/OFF status and vert/edge/face display status and colors (via corresponding UI)
+
+Notes:
+- The "Remove Viewer" button is only displayed when the viewer entry is empty.
+- The viewer selection options (viewer name drop down list) of a viewer entry is the list of the names of all the available viewer nodes in the node tree, excluding the viewer nodes already added to the layer. This is to ensure no duplicate viewers are added to a layer, and also to facilitate selection by providing a shorter list with only those viewers that have not yet been selected yet for a layer.
+
+
+ON, OFF and MIX status
+----------------------
+When the layers in the node or the viewers in a layer have an ON status, the corresponding ON/OFF toggle button will show the "Eye Open" icon, indicating the overall status of its descendants. Tapping on the toggle button will turn all its descendants status to OFF.
+
+When the layers in the node or the viewers in a layer have an OFF status, the ON/OFF toggle button will show the "Eye Close" icon indicating the overall status of its descendants. Tapping on the toggle button will turn all its descendants status to ON.
+
+When the layers in the node or the viewers in a layer have a MIX (ON and OFF) status, the ON/OF toggle button  will show a "Dot" icon, indicating the overall status of its descendants (layers or viewers). Tapping on the toggle button will turn all its descendants status to ON.
+
+Note: If a parent level (node or layer) has a MIX status and you want to turn all descendants OFF, you need to tap the ON/OFF toggle button twice: once to turn all ON and second to turn all OFF.
+
+Similar ON/OFF/MIX behavior holds true for the for the vert/edge/face display toggle buttons, except that the icons stays the same (vert/edge/face icon).
+
+
+Renaming viewer nodes externally
+--------------------------------
+When a viewer node is renamed externally the ViewerLayer node will capture the name change and will invalidate all the viewer entries in all layers that reference the older names (those entries will be highlighted in RED, assuming the default Blender color theme). In this case you can either chose to remove that viewer entry from the layers or reselect a new viewer for those entries from the latest list of viewers.
+
+Note: Once SV provides a feature to have a fixed, unique ID assigned to every node created (and saved with the blend file), which could be used as a reference, instead of the viewer nodes name, the ViewerLayer node would be able to capture the external changes to the viewer node names and automatically update the names in the layer entries without needing to invalidate viewer entries. Until then, the manual removal / update of the invalidated viewer entries is necessary. So, keep this in mind when you change the viewer node names externally. For this reason it's best to rename the viewer nodes before adding them to the ViewerLayer nodes, and then resist the temptation to change their names. :)
+
+
+Extra Viewer settings
+---------------------
+Based on the width of the ViewerLayer node additional settings for the viewers are shown/hidden. For narrow width, the viewer name and visibility toggle button are shown. For larger (>300px) width the vert/edge/face colors are also shown. And for even larger width (>400px) the vert/edge/face display status toggle buttons are also shown.
+
+
+
--- a/index.md
+++ b/index.md
@@ -438,6 +438,8 @@
     SvFCurveInNodeMK1
     SvCollectionPicker
     SvSelectionGrabberLite
+    ---
+    SvViewerLayerNode
 
 ## Objects
     SvVertexGroupNodeMK2
@@ -482,6 +484,7 @@
     ---
     SvCombinatoricsNode
 
+
 ## Alpha Nodes
     SvBManalyzinNode
     SvBMObjinputNode
@@ -505,3 +508,4 @@
     ---
     SvGetPropNodeMK2
     SvSetPropNodeMK2
+
--- a/nodes/scene/viewer_layer.py
+++ b/nodes/scene/viewer_layer.py
@@ -1,0 +1,497 @@
+# This file is part of project Sverchok. It's copyrighted by the contributors
+# recorded in the version control history of the file, available from
+# its original location https://github.com/nortikin/sverchok/commit/master
+#
+# SPDX-License-Identifier: GPL3
+# License-Filename: LICENSE
+
+
+import bpy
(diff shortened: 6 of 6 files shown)


portnov:
This thing doesn't seem to have any inputs/outputs... may be this should be a panel instead?

zeffii:
if you give it a 'dict' input socket, you can think about populating / modifying values during an update/process event

DolphinDream:
PP is an interesting idea. And maybe this could be shown per node tree regardless of which node is selected.  Though I like @zeffii  idea of maybe have input sockets that can modify (via dicts) viewer’s properties (colors, visibility etc).

DolphinDream:
FYI .. I added a feature list section to the PR description (first comment). Feel free to add more features to the list.

DolphinDream:
What is the proper way to toggle the eye icon from close to open when the visibility (bool prop) goes from true to false? (visible/hidden) ? Is there a magic attribute in the prop for toggle operations that switches the icon automatically or would i have to set the icon myself based on bool value of the prop ?

DolphinDream:
I would be nice if once I tap the plus button it would automatically select the drop down list to pick a viewer. In current implementation I have to tap plus then tap the new entry to select. Any ideas if such improvement is possible/easily done?

DolphinDream:
Demo of the expand/collapse layer option.

SV-ViewerLayersNode-demo2

DolphinDream:
Another idea.. would it be better to have the layers have both SHOW and HIDE buttons instead of the toggle? I’m thinking since the viewers in a layer could be on or off there isn’t an indication of what the one toggle button would do to the mix viewers visibility .. other than toggling its internal on/off state (which will show/hide its viewers). With the toggle button I often find myself having to click twice to do what I want to do (show or hide).

zeffii:
>  would i have to set the icon myself
yeah, it's a bit verbose so you could make a `mapping dict` outside of the class, and have 
```python

icon_dict = {
   ("prop_name_str", True): "ICON_STR_ON",
   ("prop_name_str", False): "ICON_STR_OFF",
   ...
}

class node
    icon=icon_dict.get((prop_name, state))
```

DolphinDream:
In regards to the updates of the layer’s viewers.. what is the best mechanism to have the Viewer Layer node update its viewer list and names when the viewer nodes are either deleted or renamed externally. Can SV framework keep track of these actions and notify the Viewer Layers node so that it can update itself?

zeffii:
This is the downside to such implementation of a layer system, though I think the object names/renames are a pervasive issue for more complex add-ons. Like if suddenly your local council decides to rename your street without warning, where is your post going.

> what is the best mechanism to have the Viewer Layer node update its viewer list and names when the viewer nodes are either deleted or renamed externally. 

Normally we use StringProperty to store a node name, but as you are aware this poses a problem when these nodes are renamed or removed.  There are more ways to do it. One is (but i do not advice you to instantly do this)
- https://docs.blender.org/api/blender2.8/info_gotcha.html#data-names
( tldr, store references to objects in a main dict... or a node_dict, and only call `node.name` when the ui needs to display it )
- `custom_property = PointerProperty(type=bpy.types.Object)` in conjunction with `prop_search`
- or.. allow renaming directly from the layer manager.  <---- simplest.

> Can SV framework keep track of these actions

if it could then "notify the Viewer Layers node" would be indeed be a logical step, this is where the `msgbus` https://docs.blender.org/api/blender2.8/bpy.msgbus.html?highlight=msgbus#module-bpy.msgbus could potentially be used.

-------------


i propose that you set do have tests in he layer ui to see if a node by that name exists, and turn the nodename stringfield red (using `row.alert = True` ) if not.

zeffii:
alternatively,  display the `label` ... and don't rename the node. ever :)

zeffii:
```python

import bpy
import sverchok
from sverchok.core.update_system import process_from_node

# Any Python object can act as the subscription's owner.
owner = object()

TV_NODE = sverchok.nodes.viz.vd_draw_experimental.SvVDExperimental
subscribe_to = (TV_NODE, "name")

def msgbus_callback(*args):
    nodes = bpy.data.node_groups['NodeTree'].nodes
    for tv in filter(lambda n: isinstance(n, TV_NODE), nodes):
        try:
            print(f'tv triggered, new name {tv.name}')
        except:
            print('failed')

bpy.msgbus.subscribe_rna(
    key=subscribe_to,
    owner=owner,
    args=("Node property has changed",),
    notify=msgbus_callback,
)
print("Done")

```

zeffii:
so if you display the `node.label` isntead, and only rename the labels.. then you save yourself a tonne of work. just don't change the `node.name`.

DolphinDream:
> so if you display the `node.label` instead, and only rename the labels.. then you save yourself a tonne of work. just don't change the `node.name`.

If I understand this correctly.. you are suggesting that the VL (viewer layer) node should list its viewers by label, but still store them internally by name (to be able to reference their props like color, visibility etc)? If so, still, would these labels be updated in the VL when the labels change externally ? The connections may still work (toggling viewers props etc) since the names don’t change, but their labels may be outdated.

zeffii:
>  VL (viewer layer) node should list its viewers by label, but still store them internally by name (to be able to reference their props like color, visibility etc)

yes

>  If so, still, would these labels be updated in the VL when the labels change externally 

because of storing the node.name,  and displaying only node.label  ,  if the label is edited externally, this does not change the node.name, so when the node is found, the node.label is displayed (stays in sync because node.name is constant )

>  Also, I think the label selections for each viewer empty would become invalid if the label changes externally.

initially, if the node is picked, the node.label will be empty so yes, then it would be able to display node.label

hmmmm...

DolphinDream:
Here’s an example with the layer’s 3-state toggle button (off, on, mix). When a layer’s viewers are all ON, the layer toggle button will turn all viewers OFF (icon is eye open to reflect current state of viewers). When the layer’s viewers are all OFF, the layer toggle button will turn all ON (icon is eye closed, to reflect the current state of viewers).. and when the layer’s viewers are a mix (ON & OFF) the layer’s button will turn all ON (the icon is also a dot to indicate the mix content). 

SV-ViewerLayersNode-demo4

vicdoval:
I think the only way of referencing a node reliably is via n_id. in the #3046 (which i hope to merge soon) I needed to add a dictionary to translate from n_id to the actual node. Doing it via name or label will break with renaming... but maybe is not so important in this case

DolphinDream:
And a demo with the “remove viewer” feature.

SV-ViewerLayersNode-demo5

DolphinDream:
One thing I would like to figure out how to do.. is to have the drop down list of each viewer entry only show the non-duplicate viewer options the ones not already added to a layer. I would need to filter the collection somehow. Any clues?

zeffii:
here's a prop_search on a collectionproperty : https://github.com/nortikin/sverchok/issues/991

DolphinDream:
> here's a prop_search on a collectionproperty : #991

yea :) except this is not a simple filtering.  I would need to find a way to propagate a layer’s list of viewer nodes already added to the layer to the pre_updateNode method so it can filter out the nodes already added.

zeffii:
use a `node_dict`

DolphinDream:
Can you have circular class dependency in python ?

SvViewerGroup object has a pointer reference to its (parent) SvViewerLayerGroup object ?

DolphinDream:
One feature I’m trying to accomplish for this node, related to adding/removing of viewers to a layer is to avoid duplicate viewer entries per layer and  also narrow the list of viewers presented to the user each time a new entry is added or updated (e.g. the list would exclude the viewers that were already added to the layer). There may be different ways to accomplish this, and I would need some assistance from the sv/python gurus of what is the best approach.

Essentially, when the user clicks on a viewer entry (empty or not) the drop down list would show only the viewers that have not been yet added to the layer. Also, when the user adds one more viewer to the layer, then taping on a previous viewer entry (e.g. to change the viewer entry selection let’s say) the list of available viewers should reflect the latest list of viewers currently not added to the layer. (adding and removing would both update this list, so the list is dynamic and not necessarily created statically during the creation of a viewer entry).

The way I envisioned this to work was to have the viewer class have a pointer reference to its owner (layer class) object so that when it updates its collection list it can ask its parent layer for the names of all the viewers it currently has. (This however creates a circular dependency between viewer and layer classes that I don’t know yet how to resolve or whether it is even the right thing to do in python. I know how to do this in c++).

Additionally I’d like to be able to trigger this collection/list update, if possible, right before the user clicks on the viewer entry to pick from the available viewers. This would make sure that each viewer’s list gets updated right when it is accessed/needed (rather than once during creation of the viewer entry). This is necessary since the user may choose to not only add new entries, but also remove (tap on the cross button of a given selection) a viewer from the layer list.. this means that all the other viewer entries would need to reflect this change, when the user taps on those viewer entries.

Any suggestions?

zeffii:
experiment with the mixin: https://github.com/nortikin/sverchok/issues/3044

i have not tried this, but it provides the logical building blocks of the `node_dict`, where you can store anything you want locally to the node.  The only drawbacks 
- is that iojson will require you to implement further storage_set / storage_get functions to repopulate the dict from its serialized form. 
- also reloading .blend will introduce new problems to your brain :)

zeffii:
if you want to do any bookkeeping, that's the way we do it for many nodes. just grep `node_dict`

DolphinDream:
Thanks. I’m afraid that goes a bit over my head. I spent too much time trying to figure this out without success so I think i’m going to put this feature on hold for now. 

I tried to add a dict to the SvViewerGroup so that I can pass it a dict[“layer”] = layer (when i create the viewer entry) and have the viewer entry then call the layer (within the pre_updateNode call) to find the layer’s current entries, but it doesn’t seem to work. I don't’ know python/bpy well enough to know what am I doing wrong. 😞

zeffii:
also, i only just realized this, the PointerProperty has a poll function :) 
https://docs.blender.org/api/current/bpy.props.html?highlight=pointerproperty#bpy.props.PointerProperty

zeffii:
not
```python
 dict[“layer”] = layer
```
but
```python
node_dict[self.node_id][“layer”] = layer
```
where layer is preferably a python object or something that can be serialized easily, rather than a reference to a propertygroup element.

DolphinDream:
You mean i could do this:

```
class SvViewerGroup(bpy.types.PropertyGroup):

    collection_name: CollectionProperty(name="List of Viewers", type=bpy.types.PropertyGroup)
    node_name: StringProperty(update=pre_updateNode)
    entry_id: IntProperty(name="Entry ID", description="ID of the viewer's entry in layer list")
   
    node_dict = {}  # btw, node? even if this class is not a node?
```

and then externally do this when creating a new viewer entry:

 ```
  viewers = layer.viewers.add()
  viewer.node_dict[viewer.node_id]["layer"] = layer 
  # btw is the node_id set for the viewer object?
```

where layer is an object of SvViewerLayerGroup class that has viewer as one of its entries.

and then in the pre_updateNode() do:

```
def pre_updateNode(self, context):
    ''' must rebuild for each update'''
    layer = self.node_dict[self.node_id]["layer”] # self is of type SvViewerGroup
    self.collection_name.clear()
    for node in self.id_data.nodes:
        if node.bl_idname == 'SvVDExperimental':
            # check here if already added to layer do not add
            self.collection_name.add().name = node.name

```

**Comment by DolphinDream** [comment]
> not
> 
> ```python
>  dict["layer"] = layer
> ```
> 
> but
> 
> ```python
> node_dict[self.node_id]["layer"] = layer
> ```
> 
> where layer is preferably a python object or something that can be serialized easily, rather than a reference to a propertygroup element.

I’m not sure I understand what you mean here.. and why this is needed. I do want the viewer object to have a reference of its parent layer object so that it can ask its parent what is the latest list of its children (aka viewer’s siblings). I don't want this to be a copy of the layer object since that copy would soon get stale

Sorry, this whole serialization thing is way over my head. I am a bit frustrated/confused because 1) i don’t know enough python, 2) I don’t know enough bpy (and how blender does its python things)  and 3) I don’t know enough how SV uses both to do ITS thing. Any recommendations for must read documentation (hopefully not a 1000 pages manual) to get this things cleared in my head?

**Comment by zeffii** [comment]
i think your reach a level of UI complexity where it is not possible for me to provide easy examples/answers, ...without actually doing the coding for you. ..to see if it's possible.   and i don't feel like doing that, and if i did it would rob you of that journey.

**Comment by DolphinDream** [comment]
I’m considering adding the **Viewer Index** node as a type of viewers the layer node can handle (and possibly add other types). I thought adding the name of another viz node to the search would work out of the box considering that it has same properties (activate, vert/edge/poly colors etc). :) but... actually it doesn’t quite work, because the VD and VID nodes have different attribute names for colors. To fix this of course I will have to add an if/else in the layer and handle the props per viewer based on the type of viewer (VD vs VID). Nevertheless, this gotcha reveals an inconsistency we should probably pay attention to..  I wonder whether we should have an attribute name consistency among nodes that have same type of inputs (like vert/edge/poly colors).

thoughts?

**Comment by Durman** [comment]
I now that polygon is more correct name:
https://en.wikipedia.org/wiki/Polygon_mesh

**Comment by zeffii** [comment]
@DolphinDream 

I appreciate what you say about consistent names in VD and VIDX and in light of what you're doing now, you have a point. instead of making the layer node handle the if / else, i think you might want  to make functions for verts/edges/faces, to set their color; an interface.

```
def set_color_on_node(node, element_kind, color):
    if node.bl_idname == '    ':
        if element_kind == 'vertex':
             ...
    elif node.bl_idname == ' ' : 
       ...
```
this way you can name the elements whatever you want in your node, and have a single place where the hairy inconsistencies are all collected together.

**Comment by zeffii** [comment]
> we should have an attribute name consistency among nodes that have same type of inputs (like vert/edge/poly colors).

i'm all for consistent socket names,, we could do that in one massive PR, but it will break all prior blends :)

**Comment by Durman** [comment]
I think just renaming sockets is not enough. Some system of creating sockets should be implemented which should guarantee that new nodes don't break names convention. Without such system I don't think it is worth to rename sockets of existing nodes.

**Comment by DolphinDream** [comment]
> I think the only way of referencing a node reliably is via n_id. in the #3046 (which i hope to merge soon) I needed to add a dictionary to translate from n_id to the actual node. Doing it via name or label will break with renaming... but maybe is not so important in this case

Yeah.. I think the unique node ID idea is the best. I assume these will be saved with the blender files, right?
I wonder why this feature wasn’t built into blender itself.

**Comment by DolphinDream** [comment]
About the msgbus thing.. where should I put this in the node? so that it gets setup whenever a blend file / node tree is loaded? I originally put this in the node’s sv_init but I think this is only called during the creation of a new node. (see latest PR code)

**Comment by DolphinDream** [comment]
I have some issues with the msgbus callbacks. Any ideas what I'm doing wrong? 
Here’s a screen-capture explaining the problem. Callbacks stop working after reloading saved blend files, or even after simply starting with new blend file. 

Any feedback greatly appreciated.

[](http://www.youtube.com/watch?v=-KwFKuxbnJs "SV Viewer Layer Node - msgbus issue")

**Comment by zeffii** [comment]
the msgbus has a `PERSISTENT` flag , but i think the description suggests that it is not the solution here, the blender devtalk site has a few topics on msgbus and persistence between sessions.

if all fails, it seems like the onload app handler would be a place to re-subscribe to the msgbus.

it's new tech, and i haven't used it with any regularity yet. sorry.

**Comment by DolphinDream** [comment]
Here’s another demo.. slight change in UI layout.. add extra UI dynamically based on node width.

[](http://www.youtube.com/watch?v=V0YE9zvavTY "Viewer Layer Node demo 6")

**Comment by DolphinDream** [comment]
Is there a way (perhaps via msgbus) to monitor when a new (viewer) node is removed from the node tree?

**Comment by Durman** [comment]
`free` (`sv_free`) method of the node.

**Comment by DolphinDream** [comment]
> `free` (`sv_free`) method of the node.

This would work if the node would continuously do this check, which is not cool :) I was hoping that maybe there is a way that my node can be notified once only when a viewer node is deleted. Doesn’t SV framework keep track of when a node is deleted to do some BGL updates or something?

**Comment by zeffii** [comment]
Blender calls `node.free` when it removes a node when you press X, so this function is called only once during the lifetime of a node. The one scenario when Blender does not call `free` is if the node is being removed as part of an `global undo` .  For that explicit reason I added a test in the handlers to see if a update was being triggered-and- the total node count changed. See

read and weep:  https://github.com/nortikin/sverchok/blob/476f26dee0f3faf5ebfea52464d4b241e3a1ae0a/core/handlers.py#L45-L84

**Comment by DolphinDream** [comment]
Talking about flickering (as I noticed the conversation in #3046).. I also noticed some UI flickering in my ViewerLayer node. I don't remember seeing this prior to rebasing my changes to the latest master. I thought maybe my button draw method is doing to much computation.. so I deactivated it, but there are still some calls to get the node from the node tree based on name (I don’t know how expensive these calls are or whether there is a sync issue when doing this). Any ideas why this happens?

SV-UI-flickering

**Comment by DolphinDream** [comment]
@zeffii about the node.free thing. sure, undo is messed up.. but what about the deletion of the node via X? How does my viewer layer node gets notified that some viewer node was freed? (how can the VL node subscribe to a node “free” event?)

**Comment by vicdoval** [comment]
after the #3046 a node_dict is created when you run sverchok
Every node registers on it when created https://github.com/nortikin/sverchok/blob/e2b80f00cdba3c30f858cff466a16464accbc6ef/node_tree.py#L719

and unregisters from it  when deleted (from the free function)
https://github.com/nortikin/sverchok/blob/e2b80f00cdba3c30f858cff466a16464accbc6ef/node_tree.py#L794
 the dict stores {"node_id": Node}
https://github.com/nortikin/sverchok/blob/master/core/node_id_dict.py
You could search if the node_id is present in the dictionary, if it is not then it has been deleted
or even beter you could add a method to the dictionary to get if node_id is present.

**Comment by DolphinDream** [comment]
@zeffii  a question about handlers ..

As I need to add another method to be executed via load_post (aside from the current code in sv_post_load) .. I wonder if these handlers functions should be made generic in handlers.py, e.g. link the app handler for post_load to a generic sv_handler_post_load .. and have this function then call any other method that needs to be executed on load (e.g. factorize the current code in there into a separate function). This way any other new code that needs to be executed on load would require to add a call to that code in the generic function:

```
handler_dict = {
    'undo_pre': sv_handler_undo_pre,
    'undo_post': sv_handler_undo_post,
    'load_pre': sv_clean,
    'load_post': sv_handler_post_load,
    'depsgraph_update_pre': sv_main_handler,
}
...
# the generic post_load handler:
@persistent
def sv_handler_post_load(scene):
       # call the function for the code currently in sv_post_load (perhaps call it some other name if needed)
      sv_post_load()
    
      # call other functions needed to be executed on post load:
      sv_setup_msgbus_subscriptions() # or sv_setup_subscriptions()

```

Alternatively we could make the handler_dict be a dict of lists, where each handler name (load_pre, load_post etc) is mapped to a list of functions to be appended to that app handler:

```
handler_dict = {
    'undo_pre': [ sv_handler_undo_pre ],
    'undo_post': [ sv_handler_undo_post ],
    'load_pre': [ sv_clean ],
    'load_post': [ sv_handler_post_load, sv_setup_msgbus_subscriptions ],
    'depsgraph_update_pre': [ sv_main_handler ],
}
```

Of course, in this case the code that translates this handle_dict into mapping to the bpy.app.handlers will have to change accordingly.

Which of these two implementations do you prefer? Or is there a 3rd way?

**Comment by zeffii** [comment]
why not piggyback on the existing `load_post` handler, it is also persistent.

**Comment by zeffii** [comment]
i'm not confident that the flickering can be entirely eliminated in the `viewerlayers` node, because i've also noticed some glitchy behaviour elsewhere in Blender UI (ie the text editor) randomly hiding sections of code being displayed, if there are multiple Sverchok nodes drawing to 3dview/nodeview.   this may be related, it may be unrelated.

the draw function in the node is doing quite a bit, but i doubt that it's significantly taxing you CPU.

**Comment by DolphinDream** [comment]
> why not piggyback on the existing `load_post` handler, it is also persistent.

Sure i can do that.. but this will mix up other code already in that load_post handler function with more code added (in my case) to setup the msgbus subscriptions. In the future there may be other added to the same handler which will make this handler a mix pot . I was suggestion maybe we can factorize the code to make it easier to read. 

Alternatively i wonder if you can append multiple app handlers for the same event (two callbacks for load_post etc). I don’t know how blender implements this behind the scene.

**Comment by zeffii** [comment]
i don't have issues with making `sv_handler_post_load` a dispatcher as the suggested code showed. I like how it preserves the sequential nature of things that must happen post load.

```
@persistent
def sv_handler_post_load(scene):
       # call the function for the code currently in sv_post_load (perhaps call it some other name if needed)
      sv_post_load(scene)
    
      # call other functions needed to be executed on post load:
      sv_setup_msgbus_subscriptions() # or sv_setup_subscriptions()

```
i agree completely that the `sv_post_load()` function name inadequately describes the code it executes in 2020. it has grown over 6 years.

i'm not a fan of multiple separate handler functions attached, but that's a kneejerk reaction to introducing extra complexity where i prefer to see simplicity. and we might have other opinions about simplicity.

**Comment by zeffii** [comment]
maybe rename `sv_post_load()` to `sv_ensure_nodetree_data_after_load()`

**Comment by DolphinDream** [comment]
Some random questions:

1. What category this node should go into ?
2. Any recommendation for an icon?
3. Is there a preferred quotes to use for single/multiple line docstrings (function documentation) ? (e.g. 3 double quotes  or 3 single quotes). 
4. Double or single quotes for print strings?
5. Double or single quotes for dictionary keys / named tuple etc?

**Comment by DolphinDream** [comment]
And talking about categories :) any chance we can rename the **Viz** category to something more user friendly ? perhaps **Viewers** ?

**Comment by DolphinDream** [comment]
Pushed the latest updates. 
@zeffii  if you can take a look a the handlers.py changes and see if it looks ok or not. I can push the changes to handlers.py as a separate commit  if you prefer.. otherwise i can push it as part of this PR.

**Comment by zeffii** [comment]
changes in handlers seem ok, i will probably do some formatting cleanup there after this is pushed.

**Comment by DolphinDream** [comment]
I’m thinking of a way to drive the Viewer nodes attributes via this ViewerLayer node perhaps. Here’s a setup I came up with. The VLN could have an input socket that takes a dictionary of dictionaries of various attributes of a viewer node. As long as you provide an entry in the format bellow the VLN can process it and update its viewer node attributes accordingly. Should the VLN node be the one handling such operations? What do you guys think? Any suggestions? Alternative ideas? 

SV-ViewerLayerNode-InputSocket

In this example, perhaps a dedicated node that can take all these attributes and convert them to a dictionary would be needed?

SV-ViewerLayerNode-InputSocket

**Comment by zeffii** [comment]
@DolphinDream  you may find someting of interest here: https://github.com/nortikin/sverchok/commit/dcc89b04fdd23008937db190a7307b84d721dceb

it's kind of abandoned

**Comment by DolphinDream** [comment]
> @DolphinDream you may find someting of interest here: [dcc89b0](https://github.com/nortikin/sverchok/commit/dcc89b04fdd23008937db190a7307b84d721dceb)
> 
> it's kind of abandoned

Thanks. I will take a look in the next update of the node. I feel that  tackling this now will make this PR go on for too long.

**Comment by DolphinDream** [comment]
If there are no objections I think I’m done with this PR for now and I’m ready to merge it in.

I also added some documentation.

**Comment by vicdoval** [comment]
> after the #3046 a node_dict is created when you run sverchok
> Every node registers on it when created
> 
> https://github.com/nortikin/sverchok/blob/e2b80f00cdba3c30f858cff466a16464accbc6ef/node_tree.py#L719
> 
> and unregisters from it when deleted (from the free function)
> https://github.com/nortikin/sverchok/blob/e2b80f00cdba3c30f858cff466a16464accbc6ef/node_tree.py#L794
> 
> 
> the dict stores {"node_id": Node}
> https://github.com/nortikin/sverchok/blob/master/core/node_id_dict.py
> You could search if the node_id is present in the dictionary, if it is not then it has been deleted
> or even better you could add a method to the dictionary to get if node_id is present.

With a similar strategy you could handle the external node renaming, just store the node_id and take the node from the dictionary 
```
node_dict = node_tree.nodes_dict.get(node_tree)
if node_id in node_dict:
    node = node_dict[node_id]
    node_name = node.name
    ... ...

```

**Comment by DolphinDream** [comment]
Thanks the tip @vicdoval . To be honest I’m getting kind of sick of this node already :) So I’ll have to tackle this feature in the next (MK2) version, plus whatever other features may come up until then. I’ll let people use it as is for a while and see what they like/hate about it.

**Comment by DolphinDream** [comment]
@vicdoval  btw.. do you know anything about this @persistent decorators in handlers.py. Should i keep these for the app handlers only or for any methods called from within the handlers? Eg. I moved the code from sv_post_load into a separate method and that method i did not make @persistent. The handlers themselves i kept @persistent.

**Comment by zeffii** [comment]
if the handler is decorated with `@persistent`, and that handler calls several functions. Those functions will be considered `as-good-as` persistent without any further work. Though it may be nice to add something like 
```python
def sv_ensure_nodetree_data_after_blender_load():
    """
    This function is called from a @persistent handler

    <existing doc string>
    """
```

if that's what you mean.

zeffii:
looks like a cool node btw.  hope it solves your problem.

zeffii:
maybe it's just my startup .blend , but i think this PR breaks blends that where made without the persistent msgbus calls?

i just get this now:
```
sv: import settings
sv: import all modules
sv: append app.handlers
sv: enable internal debug logging.
2020-04-30 11:18:42,378 [INFO] sverchok.utils.logging: log level, 20
Exception in module register(): \sverchok\__init__.py
  File "\2.83\scripts\modules\addon_utils.py", line 382, in enable
    mod.register()
  File "\sverchok\__init__.py", line 86, in register
    sv_registration_utils.register_all(imported_modules + node_list)
  File "\sverchok\core\__init__.py", line 24, in sv_register_modules
    m.register()
  File "\sverchok\nodes\scene\viewer_layer.py", line 493, in register
    subscribe_to_viewer_nodes_name_changes()
  File "\sverchok\nodes\scene\viewer_layer.py", line 75, in subscribe_to_viewer_nodes_name_changes
    subscribe_to_changes(subscribe_to, "Viewer node name changed", viewer_name_change_callback)
  File "\sverchok\core\handlers.py", line 35, in subscribe_to_changes
    options={"PERSISTENT", }
RuntimeError: subscribe_rna, missing bl_rna attribute from 'RNAMetaPropGroup' instance (may not be registered)
sv: import settings
sv: import all modules
  File "sverchok\core\handlers.py", line 259, in sv_handler_load_post
    sv_ensure_nodetree_data_after_blender_load()
  File "sverchok\core\handlers.py", line 217, in sv_ensure_nodetree_data_after_blender_load
    from sverchok import node_tree
  File "sverchok\__init__.py", line 74, in <module>
    node_list = make_node_list(nodes)
NameError: name 'nodes' is not defined
search for unknown menutype NODE_MT_category_SVERCHOK_BPY_Data
uiItemM: not found NODE_MT_category_SVERCHOK_BPY_Data

sverchok\ui\nodeview_space_menu.py:133
search for unknown menutype NODE_MT_category_SVERCHOK_GROUPS
uiItemM: not found NODE_MT_category_SVERCHOK_GROUPS
```

zeffii:
the menus dont populate

DolphinDream:
@zeffii I’m not getting this error at all.. even when i’m opening older blender files. 
Can you share a blend file example where this error occur ?

zeffii:
i'm using a more recent version of Blender.  2.83.

contrary to my first thought, I don't think this has to do with the name of the handler. rather that the subscribe function is called inside the viewerlayer node's register function. It should be called when all nodes have been registered.

DolphinDream:
The call from within viewer layer register was intended for cases when there are no viewer layer nodes yet in the node tree. The handler will call the setup_subscriptions (on load only) on all the nodes in the tree that have a setup_subscriptions method defined (in this case the VLN). If there is no VLN defined yet (e.g. when starting new blend or loading a blend without VLN nodes). Maybe calling this from within the init of any VLN would work, though in this case every node sets up the subscriptions.

zeffii:
we'll get it working correct . don't sweat it.  but for now understand i just want master playing nicely

DolphinDream:
Btw.. I don't get this error not even with Blender 2.83 Version 2.83 (2.83 2020-04-30)

zeffii:
but i and @vicdoval  do

zeffii:
maybe we're registering objects contained in an unordered structure somewhere, it seems to point at that.

DolphinDream:
@zeffii If you remove the call to subscribe_to_viewer_nodes_name_changes in the register of VLN and put it in the sv_init of the VLN node, do you still get this error?

zeffii:
nope, then Sverchok loads fine.

DolphinDream:
I’m trying to get this node another chance  as it was reverted after merged in due to some conflicts reported by @zeffi when loading old blends. I wonder what the real issue was and why this node couldn’t stay in the main branch. Maybe there is a solution that can make it so it can be reintegrated? Maybe alternative ways to do what the node is trying to do but done in a way that would not break the sv system?

PR #3113, opened 2020-04-25
"Formula interpolate" node

portnov:
This node is designed to perform interpolation between several control points.

The user defines control points by specifying expressions for X and Y
coordinates. Given these control points, the node will take another value of X
and calculate the corresponding Y by interpolating between the control points.

The node supports linear and cubic interpolation.

This node can be considered as a partial replacement of "Curve Mapper" node. A drawback is that the user can not edit the curve visually, he has to provide numbers or formulas for control points. On the other hand, this node allows one to parametrize the curve by use of custom variables.

Screenshot_20200425_133732
Screenshot_20200425_135424
Screenshot_20200425_140022

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/docs/nodes/number/formula_interpolate.rst
+++ b/docs/nodes/number/formula_interpolate.rst
@@ -1,0 +1,112 @@
+Formula Interpolate
+===================
+
+Functionality
+-------------
+
+This node is designed to perform interpolation between several control points.
+
+The user defines control points by specifying expressions for X and Y
+coordinates. Given these control points, the node will take another value of X
+and calculate the corresponding Y by interpolating between the control points.
+
+The node supports linear and cubic interpolation.
+
+Expression syntax
+-----------------
+
+Syntax being used for formulas is standard Python's syntax for expressions. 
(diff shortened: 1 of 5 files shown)


portnov:
@zeffii @vicdoval FYI.
Also, maybe someone has a better name for this node? I'd like to emphasize that this is something like "curve mapper", but without visual representation of the curve, but controlling the control points by formulas instead.

zeffii:
then `formula evaluate` ?

portnov:
I prefer "interpolate", because it does interpolation... "evaluate" is what usual "Formula" node does.

zeffii:
it's like a CurveFitter then.

vicdoval:
super cool node, is great that you can parametrize the points. Thanks :) 

maybe in the tooltip could be changed to  `Triggers: Curve Mapper Pro` so it appears in the search menu when you search for `map `and `curve`

Issue #3091, opened 2020-04-21
nodes with `prop_search` are broken, due to recent Blender API changes.


zeffii:
Currently a vast majority of our nodes use a `bpy.props.StringProperty` to store the name of a Blender datablock (think Material name, Text name, Object name, Image name...etc names), the ui interaction with the search box (called `prop_search` in the Blender API) returns to us the name of the object selected from that box.  like the box below

![image](https://user-images.githubusercontent.com/619340/80480160-7154e700-8950-11ea-97e0-be64528e0b4a.png)

This method of object/material/text/image/font selection now returns more information than just the String name of the object selected, it returns also some prefixes if the object is linked. For more information about why this issue exists, read https://developer.blender.org/T58641.

the `tldr;` prop_search will store object names in `bpy.props.StringProperty` using a 3 character prefix, usually the space character, but depending if an object is linked it will be `L  object_name` amd there may be other conventions.  

-----------------

we shouldn't / can't rely on stringproperty to store object names. it was convenient for a long time, and the solution is relatively easy to implement for all affected nodes. The reason it isn't fixed is because we haven't decided on a nice canonical way and whether its going to be acceptable to break .blends that use the current method.

------

In general use of objects in the .blend the result of a prop_search will be 
```
self.filename = "   visible_filename.txt"
```
internally in the node it will locate that datablock by string name, 
```
file = bpy.data.texts[self.filename]
>>> KeyError: 'bpy_prop_collection[key]: key "   visible_filename.txt" not found'
```
so Profilenode, and whatever other nodes that do an update when `self.filename` is set, will find they they will not work because blender api is inserting chars ...to solve a problem we don't have.

if we trim the StringProperty before storing it in the node, then the prop_search turns red.. which is not a nice UI effect..
![image](https://user-images.githubusercontent.com/619340/79956407-672e7680-8480-11ea-8c8d-42340f768226.png)

zeffii:
profilenodemk3 is working with this PR https://github.com/nortikin/sverchok/pull/3092
but there are many more to do, and i don't like this kind of `kludge`. so...before doing too many of them lets sit back and think

zeffii:
also the viewers that set materials

zeffii:
i suppose we must solve this using a `pointer property` ffsake

zeffii:
the following files contain one or more prop_searches..
```
sverchok\core\sockets.py:
sverchok\nodes\analyzer\image_components.py:
[...]

PR #3114, opened 2020-04-25
multiple prop_search fixes

zeffii:

many nodes use `prop_search` layout function, the value that this stores has been changed over time to include a few leading whitespaces. The leading whitespaces are poorly justified **imo**, but a reality we must solve.

here the list of nodes / utils affected https://github.com/nortikin/sverchok/issues/3091

## Solution description

Here i'm cleaning up the nodes as I see fit..

## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Ready for merge.

### DIFF
--- a/nodes/analyzer/image_components.py
+++ b/nodes/analyzer/image_components.py
@@ -75,13 +75,13 @@ class svImageImporterOp(bpy.types.Operator):
 
     filepath: StringProperty(
         name="File Path",
-        description="Filepath used for importing the font file",
+        description="Filepath used for importing the image file",
         maxlen=1024, default="", subtype='FILE_PATH')
 
     origin: StringProperty("")
 
     def execute(self, context):
-        a = bpy.data.images.load(self.filepath)
+        a = bpy.data.images.load(self.filepath.strip())
         node_tree, node_name = self.origin.split('|><|')
         node = bpy.data.node_groups[node_tree].nodes[node_name]
         node.image_name = a.name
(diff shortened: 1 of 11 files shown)


zeffii:
in profile_mk3 @portnov  you have the top code, i would like to replace it with the second version. objections?
```python
    def write_values(self, text, values):
        """

        """
        texts = bpy.data.texts.items()
        exists = False
        for t in texts:
            if bpy.data.texts[t[0]].name == text:
                exists = True
                break

        if not exists:
            bpy.data.texts.new(text)
        bpy.data.texts[text].clear()
        bpy.data.texts[text].write(values)


    def write_values(self, text_name, values):
        """
        text = some node name, used as a textname
        values = data to write

        """
        
        exists = False
        for text in bpy.data.texts:
            if text.name == text_name:
                exists = True
                break

        if not exists:
            bpy.data.texts.new(text_name)

        bpy.data.texts[text_name].clear()
        bpy.data.texts[text_name].write(values)
```

portnov:
no objections.

zeffii:
in that case i'll go for the more obvious implementation
```
    def write_values(self, text_name, data):
        """
        text_name :  a string, used to name the bpy.data.texts datablock
        data      :  the string to write to the text datablock
        """
        if not (text_name in bpy.data.texts):
            bpy.data.texts.new(text_name)

        bpy.data.texts[text_name].clear()
        bpy.data.texts[text_name].write(data)
```

zeffii:
i think for bpy.data.texts i may write an interface for sv
```python
def get_text_datablock(text_name, make_new=False):
    if text_name in bpy.data.texts:
        return bpy.data.texts[text_name]
    elif text_name.strip() in bpy.data.texts:
        return bpy.data.texts[text_name.strip()]
    else:
        if make_new:
            bpy.data.texts.new(text_name)
            return bpy.data.texts[text_name]

    return None
```

zeffii:
i already know that I don't want to do the rest of these using `.strip()` for each StringProperty that represents a datablock name..

zeffii:
OK. closing this because this is not the way to solve the issue. See new PR here

PR #3122, opened 2020-04-27
add snlite submenus to texteditor template menu

zeffii:
add back snlite templates into the TextEditor

image


zeffii:
i will also make use of these menus in SNLite's `draw_buttons_ext`

PR #3129, opened 2020-04-27
remove print call

vicdoval:
remove print call from links.py


zeffii:
i restarted the travis tests, after removing the json from master. seems ok now?

PR #3130, opened 2020-04-27
Update tree_import_tests.py

zeffii:
dont include this json in tests


nortikin:
When i change draft mode numbers are zeroed. But there are limits inside nodes, so no zeros can be
Снимок экрана от 2020-04-27 22-20-11

zeffii:
sorry @nortikin  i have not used draft mode. @portnov  ? :)

zeffii:
sv init is overwriting these values i think.

vicdoval:
this was not that bad idea, at least for the time being

zeffii:
yes. i we should still resolve the issue.. but that doesn't have to happen in master.

PR #3133, opened 2020-04-28
prop_search for bpy.data.*

zeffii:

in Sverchok nodes we use a `StringProperty` to store the names of objects that we reference in the node tree. (like Materials, Objects, Texts, Fonts... anything `bpy.data.*`). The problem with that approach is that when the user renames a datablock, the StringProperty is not automatically renamed, this results in a 'link being lost'. Because naming is import in Graphics and Modelling scenes, just as it is with programming, it is important to be able name and rename objects to reflect what they are.

At runtime we should be storing object references (object in the broader sense...anything in `bpy.data.*` ) primarily as PointerProperties, and only then store the `object.name` if it's useful, and for iojson serialization.

## Solution description

Those nodes that use prop_search, and there are many, will be revised. The beast is broken.

## Preflight checklist

- [ ] Code changes complete.
    - [x] generator\script1_lite.py
    - [ ] generators_extended\profile_mk3.py
    - [ ] scene\collection_picker_mk1.py
    - [ ] scene\objects_mk3.py
    - [ ] script\multi_exec.py
    - [ ] script\sn_functor_b.py
    - [ ] generators_extended\mesh_eval.py
    - [ ] object_nodes\get_asset_properties.py
    - [ ] scene\node_remote.py
    - [ ] scene\dupli_instances_mk4.py
    - [ ] analyzer\image_components.py
    - [ ] generators_extended\hilbert_image.py
    - [ ] scene\FCurve_in.py
    - [ ] generator\image.py
    - [ ] generators_extended\generative_art.py
    - [ ] text\text_in_mk2.py
    - [ ] text\text_out_mk2.py
    - [ ] viz\viewer_texture_lite.py  ( image file )
    - [ ] viz\viewer_typography.py  ( font file )
    - [ ] field\image_field.py
    - [ ] object_nodes\assign_materials.py
    - [ ] object_nodes\color_uv_texture.py
    - [ ] object_nodes\points_from_uv_to_mesh.py
    - [ ] object_nodes\sample_uv_color.py
    - [ ] transforms\texture_displace.py
    - [ ] vector\texture_evaluate.py
    - [ ] viz\viewer_bmesh.py
    - [ ] old_nodes\profile_mk2.py
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


vicdoval:
If you explain me the strategy you are following maybe i can help

zeffii:
in older blends nodes will store the object names with the true name for most bpy.data.* types , but not in all. Now, since a couple of weeks , new selections using prop_search will include 3 extra characters.

- we can't simply `name.strip()` .. but maybe ` regex   =  ^...(.+)$ `
- but can we write a function that extracts from the `noisyname` the real object name, and in that case still use stringproperty (lazy..but least invasive)
- or do we bite the bullet and implement the PointerProperty, more work.. _breaks shit_, but ultimately more correct.

I have no plan yet, so i welcome the talk.

zeffii:
i'm leaning towards
```
def get_valid_data_name(self, stored_name, bpy_data_kind):
    if isinstance(stored_name, str):
        if stored_name in bpy_data_kind:
            return bpy_data_kind.get(stored_name)
        elif stored_name[3:] in bpy_data_kind:
            return bpy_data_kind.get(stored_name[3:]

    self.error(f"stored_name (string) {stored_name} not found in {bpy_data_kind}")
    return None
```
and expanding on that to ensure older nodes work.

zeffii:
that would mean, that at least the socket SvObjectSocket can be robust first..
```
            return self.convert_data(SvGetSocket(self, deepcopy), implicit_conversions)
        elif self.object_ref:
            obj_ref = bpy.data.objects.get(self.object_ref.strip())
            if not obj_ref:
                raise SvNoDataError(self)
            return [obj_ref]
        elif default is sentinel:
            raise SvNoDataError(self)
        else:
            return default
```
turns into
```python
            return self.convert_data(SvGetSocket(self, deepcopy), implicit_conversions)
        elif self.object_ref:
            obj_ref = self.node.get_valid_data_name(self.object_ref, bpy.data.objects)
            if not obj_ref:
                raise SvNoDataError(self)
            return [obj_ref]
        elif default is sentinel:
            raise SvNoDataError(self)
        else:
            return default
```
I won't be offended if these proposals are thrashed. i'm not happy about these things at all.

vicdoval:
As we have to revisit so many nodes i would try to do it right, maybe trying to hold compatibility but aiming to the future :)

zeffii:
i'm happy to break stuff :)

Issue #3128, opened 2020-04-27
Animatable Nodes


vicdoval:
Right now when the frame changes a global update is performed which is bad idea because many parts of the node tree may not change during the animation.
I propose to add a `is_animatable ` property to the nodes that can be affected by a scene change, (Objects in, F-Curve, Objects Id selector...) so when the frame change is done the system calls `process_from_animatable ` in stead of a global `process`.
This property could be exposed in the N-Panel properties (enabled by default) so the user can disable it in the desired node to make the animation faster

Durman:
I'm totally agree that update whole tree during `frame change` event is inefficient. There is no sense to update some nodes during such events. 

Detect such nodes by properties also looks good. I would suggest to create special class for such nodes (`mixin`) even if this class will have only one properties. It is good because can be documented. 
Also I would suggest to keep all `mixins` one module so it will be simple investigate them during creating new nodes.

zeffii:
@vicdoval  can you have a look at SNlite hot reload (ctrl+enter) 

https://github.com/nortikin/sverchok/blob/766eeedcbe4adc5ab380d9b05094fdb1b3c33e55/utils/text_editor_plugins.py#L60-L65

it isn't triggering a node process, i think since the animatable stuff landed.

zeffii:
hmm.. i think it's something else : )     n.load() isn't even called..

zeffii:
it seems I chopped off part of the file that registers the operator and the keymap.

zeffii:
out of curiosity @vicdoval  have you seen this error before?
```
location: <unknown location>:-1
Traceback (most recent call last):
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\node_tree.py", line 366, in update
    self.sv_update()
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\node_tree.py", line 237, in sv_update
    process_from_nodes(self.get_nodes())
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\core\update_system.py", line 457, in process_from_nodes
    node_names = [node.name for node in nodes]
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\core\update_system.py", line 457, in <listcomp>
    node_names = [node.name for node in nodes]
AttributeError: 'SvFCurveInNodeMK1' object has no attribute 'name'

```
This error appears when i disconnect a node, like 
[...]

PR #3134, opened 2020-04-28
Animatable nodes

vicdoval:
#3128

> Right now when the frame changes a global update is performed which is bad idea because many parts of the node tree may not change during the animation.
> I propose to add a is_animatable property to the nodes that can be affected by a scene change, (Objects in, F-Curve, Objects Id selector...) so when the frame change is done the system calls process_from_animatable in stead of a global process.

Added a mixin SvAnimatableNode with a couple of properties:
`is_animatable`: determine if should be called on frame change
`refresh`:  update the node
and a function to draw the icons on the node:
`self.animatable_buttons(layout, icon_only=True)
`
This are the actual Animatable nodes if you miss any please inform:
image

### DIFF
--- a/utils/nodes_mixins/sv_animatable_nodes.py
+++ b/utils/nodes_mixins/sv_animatable_nodes.py
@@ -1,0 +1,27 @@
+# This file is part of project Sverchok. It's copyrighted by the contributors
+# recorded in the version control history of the file, available from
+# its original location https://github.com/nortikin/sverchok/commit/master
+#  
+# SPDX-License-Identifier: GPL3
+# License-Filename: LICENSE
+
+from bpy.props import BoolProperty
+
+
+# pylint: disable=c0111
+# pylint: disable=c0103
+
+
+class SvAnimatableNode():
+    '''
+    This mixin is used to add is_animatable property to the node.
+    This property is used on frame change to determine which nodes should be updated
+    
+    To allow the user to control it just add layout.prop(self, 'is_animatable')
+    probably in the draw_buttons_ext function
+    '''
+    is_animatable = BoolProperty(
+        name="Animate Node",
+        description="Update Node on frame change",
+        default=True
+    )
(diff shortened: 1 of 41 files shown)


vicdoval:
@zeffii @Durman  Any thoughts?

zeffii:
nothing constructive at this moment @vicdoval

Durman:
I think there is no need include in the list such nodes which get data from text deitor. Do you suppose to type something their during animation? Keywords for including nodes in the animatable list is is it useful to change it input data during animation. I would suggest to reconsider the list from this point of view. For example I does not see reason to change list of materials of an object.

Why `update the node` property?

vicdoval:
> Why update the node property?

In the `objects in` per example if you disable animation of the node and you make any changes on the scene you need to update the node, and better if it is not a global update 
Before we use to relay on updating every-time everything so when you changed the text of a `Mesh Eval` or `Text input `(with auto re-load) or anything in the scene it was re-evaluated in the next global update. Now you can avoid global updates most of the time so we need a way to update the node to trigger `update_from_node`. 


About the materials i admit is quite strange but it could be the first node of a node-tree and gathers information from the scene (at least one object) so it may be needed, it could also be removed if you wish

the `refresh` property is just a property that calls to updateNode it could be an operator but seemed that like that could make the job without the hassle

Durman:
Why only this nodes should have the `refresh` button?

Probably some nodes can initialized with switched off animatable property then.

vicdoval:
Which other nodes could benefit from the refresh button?

The property is_animatable is a boolean, if both properties where merged you would need to click twice on the is_animatable to refresh and still be independent from the animation system

utils/nodes_mixins/sv_animatable_nodes.py
```diff
@@ -0,0 +1,52 @@
...
+
+    refresh = BoolProperty(
+        name="Update Node",
+        description="Update Node",
+        default=False,
+        update=refresh_node
+    )
+    
+    def animatable_buttons(self, layout, icon_only=False):
+        row = layout.row(align=True)
```
Durman, line 50: We have convention of naming such methods with prefix `draw_` I think.

vicdoval:
This is ready and seems to work nicely if anyone has anything against it say it now or _see you in the issues_

PR #3135, opened 2020-04-28
"Curve Mapper": add Curve output.

portnov:
This adds a "Curve" output to the Curve Mapper node.

Screenshot_20200428_224813

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/nodes/number/curve_mapper.py
+++ b/nodes/number/curve_mapper.py
@@ -146,9 +146,9 @@ class SvCurveMapperNode(bpy.types.Node, SverchCustomTreeNode):
         evaluate = get_evaluator(node_group_name, curve_node_name)
         curve_node = get_valid_node(node_group_name, curve_node_name, CURVE_NODE_TYPE)
 
-        curve = SvScalarFunctionCurve(evaluate)
-        curve.u_bounds = (curve_node.mapping.clip_min_x, curve_node.mapping.clip_max_x)
         if 'Curve' in self.outputs:
+            curve = SvScalarFunctionCurve(evaluate)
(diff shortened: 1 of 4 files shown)


nodes/number/curve_mapper.py
```diff
@@ -132,15 +132,25 @@ def sv_copy(self, node):
...
-        n_id = self.node_id
-        evaluate = get_evaluator(node_group_name, 'RGB Curves'+n_id)
+        curve_node_name = self._get_curve_node_name()
+        evaluate = get_evaluator(node_group_name, curve_node_name)
+        curve_node = get_valid_node(node_group_name, curve_node_name, CURVE_NODE_TYPE)
+
+        curve = SvScalarFunctionCurve(evaluate)
+        curve.u_bounds = (curve_node.mapping.clip_min_x, curve_node.mapping.clip_max_x)
+        if 'Curve' in self.outputs:
+            self.outputs['Curve'].sv_set([curve])
```
vicdoval, line 152: `if 'Curve' in self.outputs and self.outputs['Curve'].is_linked` could be before the curve is created

vicdoval:
You were fast :)
I was coding this #3137, maybe you can add the new functions in this PR and i cancel the other.

btw thanks for the cleaning.
Everything seems nice just the little comment I did on the code, maybe is a fast operation but why doing it if is not demanded....

nodes/number/curve_mapper.py
portnov: Ok, I agree. These lines are really fast, but the code will be cleaner.

portnov:
I think different features better be in different PRs.

PR #3137, opened 2020-04-28
Curve mapper iojson

vicdoval:
Curve mapper storage get and set


vicdoval:
any clue why travis was not triggered?

Issue #2427, opened 2019-05-26
svb28


zeffii:
While the pullrequest has been open for a few months, i did most of the actual porting in the first 2 weeks of the PR. From that point it was mostly small, sporadic, bits / bugs and reactions to ongoing API changes.  Essentially hoisting Sverchok from 2.79 to 2.80, with some big changes (ie not fully node tree interoperability between old Blender and new Blender, due to some node class changes)

- [ ] What I would like to see is that the node directories match the node categories, it doesn't make sense to have the arbitrary mismatching of directories and category names.

- [ ]  too many BETA / ALPHA nodes -- I think we should consider an alternative. Many good nodes end up never migrating from there, and consequentially don't get used as much as they deserve.

- [ ]   search_props  usage now sucks for selecting filenames, we must find a replacement approach

- [ ]   several nodes are not yet 100% working (listed in the PR )

- [ ]   the object_helper class for the `Object out` nodes needs a different approach to making new unique names.

- [ ]   resolve long standing (low impact) bugs:
     - [ ] when node is in exception state, and duplicated the duplicate node gets permanent error color
       (overriding def copy in basenodeclass should handle this
     - [x] when a node is removed manually, it might need to do extra handlings if it has properties used by Sv3DProps ..) (DONE)

- [ ]   porting new nodes from master sverchok (support 2.79) to `b28_prelease_master`

Durman:
I especially agree with statement about Alpha and Beta nodes. Probably there is no need in giving to them separate category. We can set label for such nodes with suffix Beta or Alpha, and give them different color, probably slightly reddish for knowing that node should be used carefuly:
![2019-05-30_13-05-48](https://user-images.githubusercontent.com/28003269/58621999-b5bcaa80-82db-11e9-92ef-f9b309028dbf.png)

portnov:
Do I correctly understand, that Alpha/Beta are those for which we are not obligated to maintain backwards compatibility?
What exactly is the difference between Alpha and Beta?

zeffii:
>  Do I correctly understand, that Alpha/Beta are those for which we are not obligated to maintain backwards compatibility?

originally, IIRC, it was intended as a 'sandbox', to get people "not GIT equipped" people to test the node with the implied result of the node being added to the regular catagories. and getting documentation.

> We can set label for such nodes with suffix Beta or Alpha, and give them different color, 

[...]

PR #3148, opened 2020-04-30
prop_search and PointerProperty

zeffii:

Recently blender API changed for the search_prop ui element, users use this ui feature to search through data-blocks (`bpy.data.*` like: materials, texts, fonts, images, objects, collections).

image

Blender devs have changed the return value when search_prop is used in conjunction with a StringProperty by adding 3 extra characters to the start of the object name.

in Sverchok nodes we have used a `StringProperty` to store the names of objects that we reference in the node tree.. The problem with that approach is that when the user renames a datablock, the StringProperty is not automatically renamed, this results in a 'link being lost'. Because naming is important in Graphics and Modelling scenes, just as it is with programming, it is important to be able name and rename objects to reflect what they are.

At runtime we should be storing object references (object in the broader sense...anything in `bpy.data.*` ) primarily as PointerProperties, and only then store the `object.name` if it's useful, and for iojson serialization.

## Solution description

The beast is broken. Nodes that use prop_search, and there are many, will be revised. There are two approaches

1. switch entirely to the PointerProperty for most existing nodes, and all new nodes.
```diff
-    name_texture: StringProperty(
-        name='Texture',
-        description='Texture(s) to evaluate',
-        default='',
-        update=updateNode)
+    texture_pointer: PointerProperty(
+        name="Texture",  
+        poll=lambda self, object: True,  # no filtering, this may be optional.
+        type=bpy.types.Texture, # what kind of objects are we showing 
+        update=updateNode) 
+    # iojson must skip pointers, and implement `storage_get/set_data`
+    properties_to_skip_iojson = ['texture_pointer']
```
2. add a temporary fix that is aware that StringProperty will include extra characters, and perform the lookup accordingly.
```
texture = self.get_bpy_data_from_name(self.name_texture, bpy.data.textures)
```
fixes will need to be tested, most importantly for ![][blend] and ![][json]

### Now
- [ ] Code changes complete.
    - [x] sockets.py
         socket has 3 props now, instead of 1 (`object_ref` )
              - `object_kinds`
              - `object_ref`
              - `object_ref_pointer`
         - `object_kinds` (a string) accepts a single or multiple comma separated object types
                  like: 
                      - `socket.object_kinds = "MESH"`  or 
                      - `socket.object_kinds = "MESH,CURVE"` or
                      - `socket.object_kinds = "ALL"` or `=""` (empty string) to throughput all object types
                      - We could also do `socket.object_kinds = "sverchok.cat.node.filter_func_name"`
         - `object_ref` the socket will use any existing reference to this property until the user picks a new object at which point Sverchok will start using the `object_ref_pointer`.
    - [x] node_tree.py (merged already)
       - implements `get_bpy_data_from_name(self, identifier, bpy_data_kind)` which broadly accepts either a direct bpy.types.Object type or a String that represents the object.name, this function tries to "fuzzily" locate the datablock associated with identifier.
    - [ ] **sv_obj_helper**: (materials ! ) 
         - the mixin will need to set pointerproperty and keep stringproperty for a while.
         - [ ] viz\viewer_bmesh.py  (material prop_search for 3dview.. resolve with obj_helper first)
    - [x] generator\script1_lite.py (**merged already**) ![][blend] + ![][json]
    - [x] generators_extended\profile_mk3.py ![][blend] + ![][json]
    - [x] scene\collection_picker_mk1.py ![][blend] + ![][json]
    - [x] script\multi_exec.py ![][blend] + ![][json]
    - [x] script\sn_functor_b.py ![][blend] + ![][json]
    - [x] generators_extended\mesh_eval.py    ![][blend] + ![][json]
    - [x] object_nodes\get_asset_properties.py   (now mk2)
    - [x] scene\FCurve_in.py
    - [x] generator\image.py
    - [x] text\text_in_mk2.py    ![][blend] + ![][json]
    - [x] text\text_out_mk2.py   ![][blend] + ![][json]
    - [x] viz\viewer_texture_lite.py  ( image file )
    - [x] viz\viewer_typography.py  ( font file )
    - [x] transforms\texture_displace.py (now mk2)
    - [x] vector\texture_evaluate.py (now mk2)
    - [x] viz\viewer_polyline.py ( bevel object, has filter Curves now! )
    - [x] scene\node_remote.py (now mk2)
    - [x] scene\dupli_instances_mk4.py
- [x] Manual testing done. 
- [ ] Ready for merge.

### Later

    - [ ] generators_extended\hilbert_image.py
    - [ ] generators_extended\generative_art.py
    - [ ] field\image_field.py
    - [ ] object_nodes\assign_materials.py
    - [ ] object_nodes\color_uv_texture.py
    - [ ] object_nodes\points_from_uv_to_mesh.py
    - [ ] object_nodes\sample_uv_color.py
    - [ ] analyzer\image_components.py

[blend]: https://user-images.githubusercontent.com/619340/81047553-eed1a780-8eba-11ea-8e3b-dd6b86adeef9.png  "Blend file support"
[json]: https://user-images.githubusercontent.com/619340/81047695-3e17d800-8ebb-11ea-8d6a-8536c64000c1.png  "json file support"


zeffii:
i think the svobjsocket will have to work like this
1. retain existing object_ref  as a string
2. implement `object_ref_pointer` as ppointer
3. implement filter kind ( list of obj.types to aid new filtering in prop_search )

```python
def filter_kinds(self, object):
    if not self.object_kinds:
        return True
    kind = self.object_kinds
    if "," in kind:
        kinds = kind.split(',')
        ...etc.etcetc..

class SvObjectSocket(NodeSocket, SvSocketCommon):
    bl_idname = "SvObjectSocket"
    bl_label = "Object Socket"

    object_kinds: StringProperty()
    object_ref: StringProperty(update=process_from_socket)
    object_ref_pointer: bpy.props.PointerProperty(
        name="Object Reference", 
        poll=filter_kinds(s, o),  # optionally filter items out of the list of Collections presented to user
        type=bpy.types.Object, # what kind of objects are we showing
        update=process_from_socket)
    
    ...

            return self.convert_data(SvGetSocket(self, deepcopy), implicit_conversions)
        elif self.object_ref or self.object_ref_pointer:
            obj_ref = self.node.get_bpy_data_from_name(self.object_ref or self.object_ref_pointer, bpy.data.objects)
            if not obj_ref:
                raise SvNoDataError(self)
            return [obj_ref]
        elif default is sentinel:
            raise SvNoDataError(self)
        else:
            return default
```

zeffii:
i think that will at the very least not "wipe out" the current picked object_ref in existing blends, and when the user repicks an object from that propsearch it will switch over to the `object_ref_pointer`

zeffii:
now... let's see if we can filter CURVE's for the Polyline viewer bevel object.

zeffii:
@vicdoval  @portnov  @nortikin  @Durman 

i require some eyes on this, and some communication .

The easiest and ultimately the best way to proceed (as far as i can tell having stared at this for weeks now....) is to make a clean break from the existing way we store object references ( Text filenames, Image names, Object names.. font names, material names ) and implement a potentially .blend breaking upgrade.

I think the nodes listed above should be considered "Locked until further notice / This or similar PR is merged"

The issue is three fold.
- Adjusting the `SvObjectSocket` to handle PointerProperty solves part of this problem (i think non-intrusively) 
- Reworking the `<datablock>.name` storage (like `self.filename` for SNLite..etc) to add a fuzzy search function for nodes that are most common, i propose using https://github.com/nortikin/sverchok/blob/6f76b309ff4e108fc8ce513a338670addfb738d2/node_tree.py#L845-L878
- Nodes that are used far less commonly would get an overhaul to avoid storing their datablock.names as strings.

I do worry that older blends will fail when this is committed, but I think it will be worth it. It would help me if we stopped adding stuff to Sverchok except critical bug fixed, until this issue is ironed out. It's starting to get in the way of my enjoyment for this project.

portnov:
Hm. Maybe I do not understand something. If we have a PointerProperty, why would we need a fuzzy-search?

zeffii:
if we still want to attempt to do a "seemless" transition between StringProperty and PointerProperty then such a function is needed to try  the conversion automatigically.

maybe i get around to this today..

portnov:
Ah, I see.

Durman:
```Python
object_kinds: StringProperty(default='ALL') 
object_ref: StringProperty(update=process_from_socket) 
```
For what this properties?

zeffii:
@Durman  it is explained above

zeffii:
as discussed yesterday here's a way to store some information about the pointerproperty inside json, which is necessary for various non-export features (monad, presets..) anyway.
```python
def pack_pointer_property_name(datablock, node_ref, key_name):
    """
    store the pointerproperty with a temporary key in the json for exporting
    
    input:
        datacblock : your pointer directly, something like self.texture_pointer
        node_ref   : node_ref you get from inside 'storage_get_data` function.
        key_name   : the temporary name inside the node_ref which the exporter 
                     will use to keep track of datablock.name . This should be 
                     a unique name and not be identical to any attribute or 
                     annotation of the node
    outputs:
        False

    """
    node_ref[key_name] = datablock.name if datablock else ""

def unpack_pointer_property_name(datablock, data_kind, node_ref, key_name):
    """
    attempt to unpack the pointerproperty data found in the json, 
    if the datablock is found inside the .blend being imported into, then the 
    pointer will be set to that datablock

    input:
        datablock  : the attribute on the current node that we want to set
        data_kind  : this can be any bpy.data.*  <images, texts, materials...>
        node_ref   : the reference obtained in `storage_set_data` function
        key_name   : the datablock.name that was stored in json. 

    """
    key = node_ref.get(key_name)
    if key in data_kind:
         datablock = data_kind.get(key)

...

def storage_get_data(self, node_ref):
    pack_pointer_property_name(self.texture_pointer, node_ref, "texture_name")

def storage_set_data(self, node_ref):
    unpack_pointer_property_name(self.texture_pointer, bpy.data.images, node_ref, "texture_name")
```

zeffii:
if `isinstance(datablock, bpy.types.Text)` then pack / unpack can even store/restore it automatically ..from the `datablock_string`.

vicdoval:
in the unpack function `datablock = data_kind.get(key)` does not write in the property
it could be 
```
def unpack_pointer_property_name(data_kind, node_ref, key_name):
    """
    attempt to unpack the pointerproperty data found in the json, 
    if the datablock is found inside the .blend being imported into, then the 
    pointer will be set to that datablock

    input:
        data_kind  : this can be any bpy.data.*  <images, texts, materials...>
        node_ref   : the reference obtained in `storage_set_data` function
        key_name   : the datablock.name that was stored in json. 

    """
    key = node_ref.get(key_name)
    if key in data_kind:
         return data_kind.get(key)
    return  ""
```

and in the node
```
def storage_set_data(self, node_ref):
        self.texture_pointer = unpack_pointer_property_name(bpy.data.textures, node_ref, "texture_name")
```

zeffii:
yes, it's not a settled convention yet.. I hope to get more time later

zeffii:
i'm not sure it's suitable to set `self.some_pointer` to an empty string... maybe `None`

zeffii:
i didn't get the time i'd hoped tonight..maybe a few hours tomorrow.

vicdoval:
i guess this one can stay broken old_nodes\profile_mk2.py

zeffii:
iojson is hairy. OK back on track for real purpose of this PR.

vicdoval:
I guess you know...with the changes in the object socket the information that was holding object_ref is lost, 
(that's why i made the mk2) maybe could be automated? 
```
if not self.object_ref_pointer and self.object_ref:
    self.object_ref_pointer = self.get_bpy_data_from_name(self.name_texture, bpy.data.textures) 
```
?

zeffii:
i think you mean
```python
if not self.object_ref_pointer and self.object_ref:
    self.object_ref_pointer = self.get_bpy_data_from_name(self.object_ref, bpy.data.objects) 
```
i have to grimmace at the thought of moving some of these nodes to an `mk++`, what will the impact be.. these things are broken in sverchok + Blender since maybe 3 weeks now , and only one person has filed bug reports.

by not removing the original stringproperty reference for nodes, and that socket. we can at least open blends made in an old Sverchok, and attempt ( code is maybe not there yet ) to automatically switch over. from StringProp to PointerPropt.  heck.. even if it just failed but didn't cause an exception I would be OK with that,

zeffii:
oh you mean we should add that 
```
if not self.object_ref_pointer and self.object_ref:
    self.object_ref_pointer = self.get_bpy_data_from_name(self.object_ref, bpy.data.objects) 
```
into the socket "handling" code. yeah something very similar to that is a logical step, but maybe with some throttling and instantly setting `self.object_ref = ""` after the switch is completed.

zeffii:
i think the die-hards will be much happier that the prop_search / pointer_property feature treats objects (datablocks i mean) as first class elements of our tree that retain their link if the name changes elsewhere. I really really like that.

zeffii:
we can more meticulously compose the `sv_get` of the object socket,  that was really only my first inclination.

zeffii:
note some of the remaining issues here : https://github.com/nortikin/sverchok/issues/2427

vicdoval:
> note some of the remaining issues here : #2427

Do you mean/mind anything in particular or in general?

zeffii:
well..just look at when it was written : ) and prop_search was mentioned

vicdoval:
Sverchok is so awesome that its still useful in this permanent sub-1.0 state, I guess nobody has used all the nodes available... its a marvellous wild beast, rough and powerful... we learn from it... and when you see what you can do you forget what you can't do or you find another way to do it  😄

zeffii:
@vicdoval  i certainly have not used all nodes yet

zeffii:
at this point I'm not sure why the tests are failing, but on a positive node i'm starting to appreciate that `core/upgrade_nodes` mechanism.

zeffii:
bbl

zeffii:
```python
import sverchok
from sverchok.utils.sv_IO_panel_tools import create_dict_of_tree
tree = bpy.data.node_groups['TestingTree']
f = create_dict_of_tree(tree)
print(f)
```

zeffii:
@portnov i'm a little puzzled why the test for `tree_export_tests.ProfileExportTest)` with `profile_ref.blend.gz` exports a different json to test the local `Profile.json` against, than the json I would produce by manually exporting the [gist](https://gist.github.com/zeffii/53653c653fa2c21d1fc03b7d5771b1a2). 

i should restate that.. 
- i export the following gist manually from `profile_ref.blend.gz` :  [gist](https://gist.github.com/zeffii/53653c653fa2c21d1fc03b7d5771b1a2)  (Seems appropriate)
- that gist compares favourably (with my eyes) to the local `profile.json` stored in references
- if I use the snippet above, also there is nothing wierd.
- the weird stuff happens if I look at what the test is testing against, by uncommenting this line
     # self.store_reference_json("profile.json", export_result)

```
{"nodes": {"Profile Parametric Mk3": {"params": {"filename": "Profile.txt"}, "bl_idname": "SvProfileNodeMK3", "height": 100.0, "width": 140.0, "label": "", "hide": false, "location": [-351.880126953125, 62.399681091308594]}, "VD Experimental": {"params": {}, "bl_idname": "SvVDExperimental", "height": 100.0, "width": 140.0, "label": "", "hide": false, "location": [-150.87026977539062, 68.79962158203125], "color": [1.0, 0.30000001192092896, 0.0], "use_custom_color": true}}, "groups": {}, "framed_nodes": {}, "update_lists": [["Profile Parametric Mk3", 0, "VD Experimental", 0], ["Profile Parametric Mk3", 1, "VD Experimental", 1]], "export_version": "0.079"}
```
notice it's missing the "profile" key.

thoroughly puzzled .

zeffii:
ok cool. saving the blend again seems to do the trick. pitty that's the way it has to be ..but i'll cut my losses.

zeffii:
i see.. it seems that the tests are 
- loading a startup.blend
- appending into that a nodetree 
    - this append/link does not trigger post_load, and no upgrade to nodes
- hence "profile" was not filled automatically, because `self.profile_pointer` also not.

zeffii:
@vicdoval 
```
    return str(hash(self.node.node_id + self.identifier))
AttributeError: 'Node' object has no attribute 'node_id'
```
is this code being called on a ReRoute node?

zeffii:
sockets aren't being created in profilenode.. i must have b0rked the sv_update implementaiton.

zeffii:
at least i'm closing in on the bug.

vicdoval:
> @vicdoval
> 
> ```
>     return str(hash(self.node.node_id + self.identifier))
> AttributeError: 'Node' object has no attribute 'node_id'
> ```
> 
> is this code being called on a ReRoute node?

not in my implementation, the reroutes are out of the link-memory update-system

zeffii:
yeah, i think that's an error coming from an incomplete initialization of profilemk3, not creating sockets... and this the relinking list has socket references/indices that don't exist. I'll find time tomorrow. Maybe a much simpled profilemk3 upgrade.

zeffii:
i reverted all changes to profile mk3.  ready to try again :)  but first something completely different.

zeffii:
well, that was quite a whirlwind. i've been to places i forgot existed.

zeffii:
`mesh_eval` and `get_asset_properties` next.

zeffii:
ok, the state of the upgrade to pointers is still going slow.

i'll document the main issue here, getting in the way of "nirvana".
- `filename` is now intentionally decoupled from the `file_pointer` 
     - the `self.file_pointer` property has an update function which sets the self.filename
     - the `self.filename` does not have an update function, so it does not set `self.file_pointer`

these nodes loaded from a json, will accurately pick up the right file when 
- internally the file is looked for by doing `bpy.data.texts[self.filename]`
- the node doesn't try to programmatically adjust self.file_pointer given an arbitrary self.filename

when a datablock is renamed elsewhere, the `self.<   >_pointer` is indeed changed, but the update function is not called ( in my solutions this then also does not set `self.filename`

-------

bit of a pain, the solution should be simple.

zeffii:
lol..
image
apart from the renaming of the datablock, this works as expected within this minimal approach.

zeffii:
here the red `FILE_REFRESH` icon will show up if the currently loaded file (filename) is different from the current `self.file_pointer.name`

zeffii:
note to self, it seems sane to autosync self.datablock_name with self.datablock_pointer inside the process function. This keeps the changes to the existing code minimal.
```python
        # in the event that the text datablock is renamed elsewhere, this will automatically
        # resync the stored filename. updates to datablock names do not automatically propagate
        # with the calling of the pointer updatefunction. hence this nonsense.
        if self.file_pointer:
            if self.file_pointer.name != self.filename:
                self.filename = self.file_pointer.name
```

zeffii:
hmmmm.. get_asset_properties.py  🍝

zeffii:
get_asset_props will be an mk2

zeffii:
i am almost ready to merge this.

zeffii:
after merging I will `mk++` a few of the more "impacted" nodes, to make their code a bit more elegant again.

zeffii:
lesson learnt. don't bother with "upgrade in place", it gets complicated.  `mk++` allows a clean break.

zeffii:
am merging as is. it passes the tests, and shows how to use PointerProperty. a few of these nodes will go directly to `old_nodes`, but I want these nodes working even in older blends. This is an attempt at that.

Issue #3146, opened 2020-04-30
Select Similar node does not seem to work


enzyme69:
@vicdoval I tried to use this node, supposed to work but I have question about how to use it. Any documentation?

[select_similar_does_not_work_001.blend.zip](https://github.com/nortikin/sverchok/files/4557968/select_similar_does_not_work_001.blend.zip)
<img width="979" alt="Screen Shot 2020-04-30 at 9 34 32 pm" src="https://user-images.githubusercontent.com/3952687/80705813-740b3400-8b2a-11ea-8113-5f8d87e3f8ae.png">

I know the alternative Component Analyzer but I want to see how to use this Select Similar node to work...

zeffii:
https://github.com/nortikin/sverchok/blob/2dbfffccea46730391a65d0837a7a570835f57dd/docs/nodes/analyzer/select_similar.rst

zeffii:
https://github.com/nortikin/sverchok/search?q=select+similar&unscoped_q=select+similar

come on jjimmy :)

zeffii:
https://gist.github.com/56507cdf2d2085b4630494741c4d68dc

![image](https://user-images.githubusercontent.com/619340/80729807-0ef31680-8b09-11ea-960e-7320fa7913f6.png)

vicdoval:
@enzyme69  next time you click to see the online help you will find it if it exists :)

enzyme69:
😆 My SV logic must have been somewhere else last night... Ok now I got it. Thanks again @zeffii

enzyme69:
@vicdoval So many nodes do nothing when I click on help, in fact maybe it does not work anymore... can you make video capture on how to get into the help?

zeffii:
the help system works fine, i mean here exactly that "when you press this button, if Sverchok can find the help file locally, then it can display the file at the same (transformed) path of the online help". 

The problem for any node in the analyzers category is that somehow/someone changed the directory name from `analyzers` to `analyzer`, and that means the help system was looking in a directory that had no docs, and couldn't show the docs it couldn't find. @vicdoval pushed a simple fix to master.

upgrade at your own peril :)   or in future use the Search feature on github like i suggest above, if the node has docs you'll find it there in various forms.  And then if you're feeling helpful tell us one of two things
- " node X has docs, ( link ) but the help system can't find them, HALP
- " node X has no docs ( see my search link .. nothing )

PR #3149, opened 2020-04-30
Docs adventures: analyzer to analyzers 

vicdoval:
#3146 pointed a true bug the `analyzers ` folder was called `analyzer ` so it could never find the docs.

The disparity between categories and folders results in a failure when trying to access the documentation from the `Docs & Help` Panel. This happens with many of the nodes of the CAD category as 'Bevel'
In this PR only the `analyzers` was fixed but at some point we will need to fix the others..


vicdoval:
well, well.... the side problem is that the `analyzers` folder is also called `analyzer` so the easiest solution would be changing the category name to `analyzer` and leave the other folders as they are, because changing the nodes folder name may break scripts and who_knows_what...

vicdoval:
Finally what i did is if it fails to find it in the folder it should be it checks if it is in another folder. Very useful for nodes in Alpha, Beta and CAD that have docs

vicdoval:
Also added some links and logo to the 404 page :)

PR #3154, opened 2020-05-01
sv_first_run to False after reload event

vicdoval:
https://github.com/nortikin/sverchok/issues/3128#issuecomment-622388011
The flag sv_first_run had to be turned to False after reloading.

Also i notice something very strange, sometimes, very few times, when asking for node_tree.nodes after reloading it was returning a random panel data!! (or any other kind) really alien stuff but it happens so weirdly that it could not called or reported as a bug, I did as suggested here https://github.com/nortikin/sverchok/issues/3128#issuecomment-622402343  making a extra check for those cases so it does not report an error in stead reports the rejected data

### DIFF
--- a/core/update_system.py
+++ b/core/update_system.py
@@ -454,7 +454,13 @@ def process_from_nodes(nodes):
     if not nodes:
         return
 
-    node_names = [node.name for node in nodes]
+    node_names = []
+    for node in nodes:
+        if hasattr(node, "name"):
+            node_names.append(node.name)
+        else:
+            print("Something not very important happend in Blender memory", node, type(node))
+
     ng = nodes[0].id_data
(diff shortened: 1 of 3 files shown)


core/update_system.py
```diff
@@ -454,7 +454,13 @@ def process_from_nodes(nodes):
...
         return
 
-    node_names = [node.name for node in nodes]
+    node_names = []
+    for node in nodes:
+        if hasattr(node, "name"):
+            node_names.append(node.name)
+        else:
+            print("Something not very important happend in Blender memory", node, type(node))
+
```
Durman, line 463: Several more such commits and crashing Scerchok will be regular thing.

PR #3155, opened 2020-05-01
texture_displace prop_search

vicdoval:
@zeffi Is it like this?

Like this a file saved in 2.82 with a texture selected when opened in 2.83 the texture appears in red and in order to work you have to select it again.
that is how it is supposed to work?

Or i need to add a pointer property?


zeffii:
If a blend is saved in an older Blender, then the texture name will be "clean", but more recent versions of Blender, the prop_search result will be padded with 3 chars (not necessarily spaces ) at the beginning. 

All `self.get_bpy_data_from_name(self.name_texture, bpy.data.textures)` does . is try to find the currently stored value in `self.name_texure` , with or without the 3 chars padding.  in `bpy.data.textures`

it makes no attempt to "trim" the content of `self.name_texture`, because when you do that then the prop_search box becomes red.  So if Blender finds `self.name_texture` without the 3 chars it considers it as "not part of the list of available choices" and marks it red. (even if that is the proper name,, and that's the way Blender used to pass the result to a StringProperty)

Ultimately yeah we want to use PointerProperty,, i've already tasted the future.

vicdoval:
The question that appears to my head is: 
What is the point of storing a reference to a object that most likely wont be there when I import in other file the same json?

zeffii:
in `json` storing references "somehow" has little to no use. Except in a scenario where the node is able to recreate the object from information stored in the json. (SNLite does this.. by storing `script_str` and `script_name` , and writes script_str to bpy.data.texts[script_name] upon import.

The pointerproperty is really for layouts/nodes/blends , picking any bpy.data.* that you might end up renaming at runtime. Adding `get_bpy_data_from_name` is the poorman's first step to migrating to a pointer property.

for nodes that haven't been in circulation long we need to implement a pointerproperty.

for this node that would look like  ( untested.. you try it )
```diff
-    name_texture: StringProperty(
-        name='Texture',
-        description='Texture(s) to evaluate',
-        default='',
-        update=updateNode)
+    texture_pointer: PointerProperty(
+        name="Texture",  
+        poll=lambda self, object: True,  # no filtering, this may be optional.
+        type=bpy.types.Image, # what kind of objects are we showing 
+        update=updateNode) 
...
- params[2] = [[bpy.data.textures[self.name_texture]]]
+ params[2] = [[self.texture_pointer]]]
# or if it's more stable.
+ params[2] = [[self.texture_pointer or EmptyTexture()]]]

```

vicdoval:
Thoughts:
With the pointer property if you append the nodetree to another file everything that is linked will be imported too.... 
maybe we could also provide examples strored as blendfiles with a importer that only imports the sv nodetrees and the linked data

zeffii:
we could definitely have more .blend examples that show objects (datablocks like material, image, Object..etc ) linked and used.

I don't see why not.   json support for pointerproperty will need to be implemented soon.

vicdoval:
When i move to the pointer property if i try to export i get this report
`TypeError: only strings are allowed as keys of ID properties`

zeffii:
you mean exporting as json ?

zeffii:
that means when iojson is populating the json, it's trying to do
```
node_dict["params"]["texture_pointer"] = <the object>
```
in the exporting step we might just do 
```
node_dict["params"]["texture_pointer"] = <the object>,name
```
but we might have to add another property or method to the node to list all properties in the node that are PointerProperties, and use that during the import

i haven't worked that out yet.

zeffii:
it maybe simpler for the first few nodes that implement pointerproperty, to implement storage_get/set_data and add the pointerproperty attribute to the skip attributes.

zeffii:
like in collection picker
```
properties_to_skip_iojson = ['text_pointer']
```

zeffii:
in case of texture displace node.. maybe you would
```python
def storage_get_data(self, node_ref):
    node_dict["texture_name"] = self.texture_pointer.name if self.texture_pointer else ""

def storage_set_data(self, node_ref):
     texture_name = node_ref.get("texture_name")
     if texture_name in bpy.data.images:
          self.texture_pointer = bpy.data.images.get(texture_name)
```

i don't know.

zeffii:
handling the serialization of such stuff is going to be really similar for all `bpy.data.*`, so a convention would quickly emerge after the first few.. and perhaps a Mixin.

zeffii:
with code comments..
```python
def pack_pointer_property_name(datablock, node_ref, key_name):
    """
    this function stores the pointerproperty with a temporary key in the json for exporting
    
    input:
        datacblock : your pointer directly, something like self.texture_pointer
        node_ref   : the references to node_ref you get from inside 'storage_get_data` function.
        key_name   : the temporary name inside the node_ref which the exporter will use to keep track of
                     datablock.name . This should be a unique name and not be identical to any attribute / annotation of the node
    outputs:
        False

    """
    node_ref[key_name] = datablock.name if datablock else ""

def unpack_pointer_property_name(datablock_name, data_kind, node_ref, key_name):
    """
    this function attempts to unpack the pointerproperty data found in the json, if the datablock is found
    inside the .blend being impoted into, then the pointer will be set to that datablock

    input:
        datablock_name : the attribute name on the current node that we want to set the pointer for
        data_kind      : this can be any bpy.data.*  <images, objects, texts, materials...>
        node_ref       : the reference obtained in `storage_set_data` function
        key_name       : the name that the datablock was temporarily stored as in json. 

    """
    key = node_ref.get(key_name)
    if key in data_kind:
         setattr(self, datablock_name, data_kind.get(key))

...

def storage_get_data(self, node_ref):
    pack_pointer_property_name(self.texture_pointer, node_ref, "texture_name")

def storage_set_data(self, node_ref):
    unpack_pointer_property_name("texture_pointer", bpy.data.images, node_ref, "texture_name")

```

zeffii:
that should in theory fail gracefully for json.  but ofcourse.. completely untested.

zeffii:
or.. if the above worked.. this so should this
```python
def pack_pointer_property_name(datablock, node_ref, key_name):
    """
    store the pointerproperty with a temporary key in the json for exporting
    
    input:
        datacblock : your pointer directly, something like self.texture_pointer
        node_ref   : node_ref you get from inside 'storage_get_data` function.
        key_name   : the temporary name inside the node_ref which the exporter 
                     will use to keep track of datablock.name . This should be 
                     a unique name and not be identical to any attribute or 
                     annotation of the node
    outputs:
        False

    """
    node_ref[key_name] = datablock.name if datablock else ""

def unpack_pointer_property_name(datablock, data_kind, node_ref, key_name):
    """
    attempt to unpack the pointerproperty data found in the json, 
    if the datablock is found inside the .blend being impoted into, then the 
    pointer will be set to that datablock

    input:
        datablock  : the attribute on the current node that we want to set
        data_kind  : this can be any bpy.data.*  <images, texts, materials...>
        node_ref   : the reference obtained in `storage_set_data` function
        key_name   : the datablock.name that was stored in json. 

    """
    key = node_ref.get(key_name)
    if key in data_kind:
         datablock = data_kind.get(key)

...

def storage_get_data(self, node_ref):
    pack_pointer_property_name(self.texture_pointer, node_ref, "texture_name")

def storage_set_data(self, node_ref):
    unpack_pointer_property_name(self.texture_pointer, bpy.data.images, node_ref, "texture_name")
```

zeffii:
i'd be tempted to generated the `key_name` automatically ..

zeffii:
maybe 
```
- pack_pointer_property_name(self.texture_pointer, node_ref, "texture_name")
+ pack_pointer_property_name("texture_pointer", node_ref,)  # inferred key name

- unpack_pointer_property_name(self.texture_pointer, bpy.data.images, node_ref, "texture_name")
+ unpack_pointer_property_name("texture_pointer", bpy.data.images, node_ref)
```
only slight modification of the unpack/pack functions would let us abstract out the picking of temporary keys for json.

zeffii:
i've added more detailed recap of this to the PR  https://github.com/nortikin/sverchok/pull/3148

vicdoval:
this node is done, i think i will do in the same branch the "Texture Evaluate" node before merging to `nortikin:prop_search_v4` if you are okay with it @zeffii

zeffii:
yes please @vicdoval

zeffii:
we can gradually migrate nodes out of `nortikin:prop_search_v4` once it appears that we are doing the correct approach. first ones will be slow... and rapidly speed up.

vicdoval:
Ok this one was fast. :D 

Maybe i should merge this branch with yours at this point so we have the same pack/unpack functions?

zeffii:
yeah, merge unpack/pack to master if you want

zeffii:
you mean merge your pack/unpack into this PR too? yeah, is OK with me. best we use the same functions

zeffii:
i'd like to hold off merging that utility function into master, until we can look at all effected nodes and see if the pattern was wise.

zeffii:
yeah do it.

there may be more specific unpack/pack functions to emerge.

PR #3158, opened 2020-05-02
Add new Timer node

DolphinDream:
Add timer node to manage elapsed/remaining time via start, pause, stop, reset operations. (useful for triggering and handling all kinds of animations*)

Note: in this case  **animation** does not pertain to solely altering translations, rotations, colors properties etc.. it can also apply to handling the generation of new geometry, triggering a sequence of processing of other nodes (e.g. the node can be used as decision switch in the node tree during playback).

The node operations can be set externally via input node.

SV-TimerNode-UI1

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/nodes/scene/timer.py
+++ b/nodes/scene/timer.py
@@ -1,0 +1,576 @@
+# ##### BEGIN GPL LICENSE BLOCK #####
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software Foundation,
+#  Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# ##### END GPL LICENSE BLOCK #####
+
+import bpy
+from bpy.props import IntProperty, FloatProperty, BoolProperty, StringProperty, EnumProperty, CollectionProperty
+
+from sverchok.node_tree import SverchCustomTreeNode
+from sverchok.data_structure import updateNode, match_long_repeat
+from sverchok.utils.nodes_mixins.sv_animatable_nodes import SvAnimatableNode
+from sverchok.utils.modules.statistics_functions import *
+from sverchok.utils.logging import debug
+
+TIMER_STATUS_STOPPED = "STOPPED"
+TIMER_STATUS_STARTED = "STARTED"
+TIMER_STATUS_PAUSED = "PAUSED"
+TIMER_STATUS_EXPIRED = "EXPIRED"
+
+timer_status_items = [(TIMER_STATUS_STOPPED, "Stopped", "", 1),
+                      (TIMER_STATUS_STARTED, "Started", "", 2),
+                      (TIMER_STATUS_PAUSED, "Paused", "", 3),
+                      (TIMER_STATUS_EXPIRED, "Expired", "", 4)]
+
+TIMER_OPERATION_STOP = 0
+TIMER_OPERATION_START = 1
+TIMER_OPERATION_PAUSE = 2
+TIMER_OPERATION_RESET = 3
+TIMER_OPERATION_EXPIRE = 4
+TIMER_OPERATION_LOOP_BACKWARD = 5
+TIMER_OPERATION_LOOP_FORWARDS = 6
+
+EPSILON = 1e-10  # used to avoid division by zero
+
+
+class SvTimerPropertyGroup(bpy.types.PropertyGroup):
+    ''' Timer class '''
+    timer_id: IntProperty(
+        name="Timer ID",
+        description="ID of the timer entry",
+        default=0)
+
+    last_timer_status: EnumProperty(
+        name="Last Timer Status", description="Timer status before it expired",
+        items=timer_status_items,
+        default=TIMER_STATUS_STOPPED)
+
+    timer_status: EnumProperty(
+        name="Timer Status", description="Timer Status",
+        items=timer_status_items,
+        default=TIMER_STATUS_STOPPED)
+
+    timer_time: FloatProperty(
+        name="Time", description="Elapsed time",
+        default=0.0)
+
+    timer_duration: FloatProperty(
+        name="Duration", description="Time after which the timer loops or expires",
+        default=10.0, min=0.0)
+
+    timer_speed: FloatProperty(
+        name="Speed", description="Time speed as a multiplier of playback time",
+        default=1.0)
+
+    timer_loops: IntProperty(
+        name="Loops", description="Number of loops before expiring (0 = no loop)",
+        default=0, min=0)
+
+    timer_loop_count: IntProperty(
+        name="Loop Count", description="Number of realized loops",
+        default=0, min=0)
+
+    timer_last_operation: IntProperty(
+        name="Last Operation", description="Last external control operation executed",
+        default=0)
+
+    timer_start_frame: IntProperty(
+        name="Start Frame", description="Frame at which timer was started",
+        default=0)
+
+    timer_last_frame: IntProperty(
+        name="Last Frame", description="The last frame the timer was updated",
+        default=0)
+
+    def id(self):
+        return self.timer_id
+
+    def status(self):
+        return self.timer_status
+
+    def elapsed_time(self, normalized=False):
+        if normalized:
+            return self.timer_time / self.timer_duration
+        else:
+            return self.timer_time
+
+    def remaining_time(self, normalized=False):
+        remaining_time = max(0, self.timer_duration - self.timer_time)
+        if normalized:
+            return remaining_time / self.timer_duration
+        else:
+            return remaining_time
+
+    def expired(self):
+        return self.timer_status == TIMER_STATUS_EXPIRED
+
+    def loop_count(self):
+        return self.timer_loop_count
+
+    def stop(self):
+        debug("* Timer {0}: STOP".format(self.timer_id))
+        self.timer_time = 0
+        self.timer_loop_count = 0
+        self.timer_status = TIMER_STATUS_STOPPED
+        self.last_timer_status = TIMER_STATUS_STOPPED
+
+    def start(self):
+        debug("* Timer {0}: START".format(self.timer_id))
+        if self.timer_status in [TIMER_STATUS_STOPPED, TIMER_STATUS_EXPIRED]:
+            debug("starting from zero")
+            self.timer_time = 0
+            self.timer_start_frame = bpy.context.scene.frame_current
+            self.timer_loop_count = 0
+        self.timer_status = TIMER_STATUS_STARTED
+        self.last_timer_status = TIMER_STATUS_STARTED
+
+    def pause(self):
+        debug("* Timer {0}: PAUSE".format(self.timer_id))
+        if self.timer_status == TIMER_STATUS_STARTED:
+            self.timer_status = TIMER_STATUS_PAUSED
+
+    def reset(self):
+        debug("* Timer {0}: RESET".format(self.timer_id))
+        self.timer_status = TIMER_STATUS_STOPPED
+        self.timer_time = 0
+        self.timer_loop_count = 0
+
+    def expire(self):
+        debug("* Timer {0}: EXPIRE".format(self.timer_id))
+        self.timer_status = TIMER_STATUS_EXPIRED
+        self.timer_time = self.timer_duration
+        self.timer_loop_count = self.timer_loops
+
+    def loop_backward(self):
+        debug("* Timer {0}: LOOP BACKWARD".format(self.timer_id))
+        if self.timer_loop_count == 0:  # first loop ?
+            self.last_timer_status = self.timer_status
+            self.timer_status = TIMER_STATUS_STOPPED
+            self.timer_time = 0
+        else:  # not the first loop
+            if self.timer_status == TIMER_STATUS_EXPIRED:
+                self.timer_status = TIMER_STATUS_PAUSED
+
+            if self.timer_time > 0:  # has some time in this loop? => reset to 0
+                self.timer_time = 0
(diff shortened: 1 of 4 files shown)


DolphinDream:
SV-TimerNode-Setup1

DolphinDream:
Currently I had to add this “Driver” input socket in order to drive the node updates when the playback is on. Any suggestions how to make the update be triggered automatically (when playback is on) just like the Frame Info node, without having this “Driver” input?

Durman:
Probably this can help:
#3134

DolphinDream:
Latest update... 
- updated UI (add expire button)
- add time scrub slider
- remove external driver (node is animatable)

SV-TimerNode-UI2

SV-TimerNode-Setup2

DolphinDream:
Btw, I don’t know what changed lately but when i reload the scripts (F8) after the rendering gets cleared, I used to have to change the frame in order to have the rendering from the node tree come back .. it seems that this no longer works, so I have to actually change values in the node in order for the rendering to update. On this node, I always thought that it was a bit odd that i had to change the frame after reloading the scripts in order to update the rendering. I think this update should be automatic after reloading the scripts.

vicdoval:
It would be cool but i could not achieve to update the nodetree after reloading scripts automatically.
The best way I found is to press F8 (reload scripts) and after F5 (node-tree process).

Anyway if the node is animatable it should update on frame_change... 🌵

DolphinDream:
I’m updating the Ring node.. this doesn’t really need to be animatable (or perhaps i miss the idea of what this animatable feature is all about). So, for this node if i reload the scripts, and move the frame the rendering doesn’t come back. (unless i change a value in the ring node).

vicdoval:
I that case F8 + F5 is the way to go

(or you could link a animatable node to any input to propagate the changes on frame_change)

vicdoval:
Does this node referees to 'real time' or to 'frame time'? 
I you stop playback it will keep running?

DolphinDream:
> Does this node referees to 'real time' or to 'frame time'?
> I you stop playback it will keep running?

It is frame time .. tied to blender's fps. It only plays when the playback is on. 
Not sure if it would be a good idea to have the node use a real time and play when the playback is not on. What do you think?

vicdoval:
I don't see a clear use for the real time,  specially thinking on how smooth could run the animation with a elaborated nodetree.. :)  
I said it because in you node everything seemed like real seconds and maybe you had something in mind related to realtime liveshow or any  other un-expected idea 😃 (triggering events to a outside program via udp client... 👽

DolphinDream:
And a setup/demo showing external control of the timer node.. and a chained timer setup to showing how a timer can trigger another timer to start an animation :)  

SV-TimerNode-setup3

SV-TimerNode-demo1

DolphinDream:
And one more demo.. showing off the latest vectorization feature. Sorry, I couldn’t resit the temptation to vectorize it. :) 

The demo illustrates a vectorized timer with different durations / speeds hooked up to different animations and other timers (e.g. note the different translation velocities along with their ring progress indicators). The demo also shows timers triggering other timers start/stop (e.g. as the “racers" get to the finish line their timer trigger the torus knot animations/timers to stop). 

[](http://www.youtube.com/watch?v=9Ge759tEBFM "SV TimerNode Demo2")

Durman:
Can you merge master into this PR for testing the test?

DolphinDream:
> Can you merge master into this PR for testing the test?

Can you be more specific? I rebased this branch on top of latest master. “testing the test” = ?

nortikin:
not se how you use timer. is it threading or sleep function or what do you use?

DolphinDream:
> not se how you use timer. is it threading or sleep function or what do you use?

It doesn’t use external timing :) it simply relies on the blender playback time, so they become active once as the frame changes (scrub, select, playback frames).

nortikin:
thanks, got it.

DolphinDream:
Perhaps you guys already expect this, given the updates to the update system, but I experienced a bunch of issues with different nodes not updating properly. E.g. When I enter a value into an input socket (not connected) I expect its update method to be called and sometimes it’s called sometimes it’s not.  ☹️

vicdoval:
@DolphinDream the changes haven't affected that event (changing a node which triggers the updateNode function) That works as it was working before. The changes only are used when linking/un-linking, creating/deleting nodes and animation updates.
Anyway when it happens check the console, maybe you can find something there to start with

nodes/scene/timer.py
```diff
@@ -0,0 +1,391 @@
...
+            elif operation == 2:  # PAUSE timer
+                self.pause()
+
+            elif operation == 3:  # RESET timer
+                self.reset()
+
+            elif operation == 4:  # EXPIRE timer
+                self.expire()
+
+    def update(self, speed, duration, operation, lapse, loop):
```
zeffii, line 123: to avoid confusion please avoid naming a method just "update", maybe `update_config` ? or something else more to your liking.

DolphinDream:
Latest update..

- UI updates
- add input loopss socket (number of loops after the timer expires)
- add output loop socket (the loop number timer is currently in)
- various other code updates

SV-TimerNode-UI3

DolphinDream:
A **video** demo showing the uses of the Timer Node to re-create the Matrix movie's famous "Bullet Time" cinematic effect.The animation starts off with the "real time” flow, then the time flow slows down (to almost freezing the objects animation), then camera rotates one time around the “frozen in time” objects, then the time flow ramps up and continues at its original flow.  Four TimerNodes were used to emulate these transactions.

[](http://www.youtube.com/watch?v=dJMVICRp2tU "SV TimerNode - Bullet Time scene - Demo1")

Durman:
It's seems that this node would look fine in `animation nodes`.

nodes/scene/timer.py
```diff
@@ -0,0 +1,510 @@
...
+        self.outputs.new('SvStringsSocket', "Loop")
+
+    def process(self):
+        # print("processing timer node")
+
+        # print("frame: {0} , last frame: {1}, frame diff: {2}, time diff: {3}".format(frame, self.last_frame, frame_diff, time_diff))
+
+        # return if no outputs are connected
+        if not any(s.is_linked for s in self.outputs):
+            return
```
Durman, line 436: Update system itself does not call not connected nodes.

DolphinDream:
What do you mean @Durman? Are you suggesting Sverchok Addon doesn’t need a timer node? (This node can animate geometry not just object properties)

nodes/scene/timer.py
vicdoval: @Durman  the update system will call a node that was linked when you un-link it. Because it calls the affected nodes by the link change. this is good (for example) for the viewers to disable the bgl call

zeffii:
i'm happy that Sverchok finally has a dev that is interested enough in Animation that such nodes need to emerge :) .  i'm all for this kind of stuff @DolphinDream

nodes/scene/timer.py
DolphinDream: Btw, I copied the code from other nodes I created in the past and use as a template for new node and always thought this check was a recommended way to do things. If this has changed or never have been the case then let me know and I’ll remove it. I’m not knowledgeable enough the updating processes underneath.

zeffii: it appears that with recent changes to the update system, some nodes that implement `sv_update` may be simplified a little, but if you do remove the `sv_update` then the `process` function may be where you do the extra defensive code.

don't worry about it.

Durman:
I mean that you could probably solve your problem easier with tool which is dedicated for animation.

nodes/scene/timer.py
Durman: Process of nodes looks like does not call if it does not connect with any other nodes. I'm not going find proves of it in code because with it current readability it will probably will take a whole day. But where I'm wrong is that if a node has input connection it will be called by update system and in this case to check linking of output has sense.

enzyme69:
I will test this node once it's LIVE, is this a kind of node that allows user to simply have something animating and playing back without timeline running?

DolphinDream:
> I will test this node once it's LIVE, is this a kind of node that allows user to simply have something animating and playing back without timeline running?

@enzyme69 Currently the node animates/measures time only when the playback is on (when the frame changes). I don’t know what use would have to animate outside of the blender’s animation timeline. Of course you can use the time slider within the node itself to animate without changing the frame but the speed of animation is dependent on the speed of scrubbing the slider. 

I can look into animating this without changing the frame and see if it works and if it make sense.

DolphinDream:
@Durman Gotcha. You’re right, if that Bullet Time is the thing I was going after, then the AN would be perhaps the right way to do it. I only picked the bullet time concept as an illustration of what the SV timer node could do (perhaps not the best demo subject idea).

nortikin:
would you merge it?

DolphinDream:
> would you merge it?

I’ll try to wrap it up over the weekend and push it.

DolphinDream:
Various updates.. getting close to wrapping this up for an initial release.

- layout & ui updates
- looping fixes
- remove old lapse/loop settings
- add new loop back/forth operations
- variable renaming, comments updates, various cleanup
- replace prints with debugs
- add sticky mode (PP setting)
- update documentation

SV-TimerNode-UI4

DolphinDream:
If anyone feels like taking a quick look at the PR and let me know if you have any suggestions before I merge this into master, I’d appreciate it. If not, I’ll merge this in shortly.

DolphinDream:
Historical progression of the UI :) I hope #5 looks cleaner / more intuitive.

SV-TimerNode-UI-Progression

nodes/scene/timer.py
```diff
@@ -0,0 +1,574 @@
...
+class SvTimerNode(bpy.types.Node, SverchCustomTreeNode, SvAnimatableNode):
+    """
+    Triggers: Timer, Elapse
+    Tooltip: Manage elapsed time via Start, Stop, Pause, Reset, Expire operations.
+    """
+    bl_idname = 'SvTimerNode'
+    bl_label = 'Timer'
+    bl_icon = 'PREVIEW_RANGE'
+
+    timers: CollectionProperty(name="Timers", type=SvTimerPropertyGroup)
```
zeffii, line 335: you'll want to have a way to rebuild this `timers` data from the serialized state, required by 
- gists...and
- presets..and 
- monads to some degree.

This will tell the export code to ignore that property (it's too complex to decompose automatically... at least for the current export code)
```diff
+  properties_to_skip_iojson = ["timers"]
```
see: https://github.com/nortikin/sverchok/search?q=properties_to_skip_iojson&unscoped_q=properties_to_skip_iojson

and implement `storage_get_data`, `storage_set_data`
https://github.com/nortikin/sverchok/blob/e3fc7a5dc900935d607bd8f3167ace1956f09c46/nodes/script/multi_exec.py#L251-L266

zeffii: look at a few nodes' implementation of `storage_get_data`, `storage_set_data`

DolphinDream:
Thanks @zeffii . If you dont’ mind for the time being I was going add the skip iojson to the node and plan to add the set/get storage feature for export in the next update. Some of the timer attributes are updated every update cycle from the node itself, but things like current state (started, paused etc) are not so export/import SV json trees will not restore these. 

Blender seems to preserve these settings (timer collection group settings) when saving/loading blend files. I wonder if SV import/export could somehow tap into blender’s mechanism to export these collection properties.

nodes/scene/timer.py
DolphinDream: Thanks @zeffii . If you don’t mind for the time being I was going add the skip iojson to the node and plan to add the set/get storage feature for export in the next update. Some of the timer attributes are updated every update cycle from the node itself, but things like current state (started, paused etc) are not so export/import SV json trees will not restore these. 

Blender seems to preserve these settings (timer collection group settings) when saving/loading blend files. I wonder if SV import/export could somehow tap into blender’s mechanism to export these collection properties.

zeffii: it's easier to implement the storage_get/set stuff while the datastructure is still fresh in your mind.

i'll have a look.

zeffii:
it's something like to store them
```python
node_dict['serialized_timer_data'] = {}
timer_storage = node_dict['serialized_timer_data']

for timer in self.timers:
    tiner_storage[timer_id] = {}
    for a in timers.__annotations__:
        # print(a, getattr(timer, a))
        tiner_storage[timer_id][a] = getattr(timer, a) 

```

zeffii:
but i'll implement that if you want after pushing.  just make a few demo .blends and i'll get the .blend - .json 1-1

nodes/scene/timer.py
DolphinDream: I’m sure it would be best to do it now, though i’m just afraid it may be a rabbit hole for me and I don’t have much time right now to tackle this, and I wanted to merge the PR for others to start playing with the node. 

Of all the timer properties.. 

# this is currently used for debugging
timer_id

# these are updated by the node with every update cycle (so i dont’ think they need to be stored)
timer_duration
timer_speed
timer_loops

# these would need to be stored
last_timer_status
timer_status
timer_time

timer_loop_count
timer_last_operation

timer_start_frame
timer_last_frame

DolphinDream:
> but i'll implement that if you want after pushing. just make a few demo .blends and i'll get the .blend - .json 1-1

Where should I put the demo blend files?

DolphinDream:
Here’s a demo Timer node setup blend file.

SV-TimerNode-Setup

[timerNode-dev1.blend.zip](https://github.com/nortikin/sverchok/files/4640980/timerNode-dev1.blend.zip)

DolphinDream:
Here’s another demo Timer node setup blend file. (bullet time demo)

SV-TimerNode-Setup2

[timerNode-dev2.blend.zip](https://github.com/nortikin/sverchok/files/4641212/timerNode-dev2.blend.zip)

DolphinDream:
And another demo setup/blend file. (spinning arcs with different speeds and loop counts)

SV-TimerNode-Setup3


[timerNode-dev3.blend.zip](https://github.com/nortikin/sverchok/files/4641220/timerNode-dev3.blend.zip)

zeffii:
hey just an aside  @DolphinDream  you can export blends directly from sverchok which will add date information..
image
on windows that will automatically spawn an explorer window for me, and highlight the Blend. Unfortunately I can't test on osx, but i could try adding that for *nix systens too.
Sverchok will print the location of the saved file to console / or in `info` even, as:
```
Info: saved current .blend as archive at C:\Users\zeffi\Desktop\fcurve_show_2020_05_18_08_07.zip
```

Issue #3159, opened 2020-05-02
issues need resolving.


zeffii:
i'm going to say a few unpopular things here. 

The update system in the current master is broken, the bug may be easy to fix, but it's there and i will not be the only person experiencing it. I think it was ill-advised to full on merge various update system augmentations directly into master without first having them merged into a dedicated `nortikin/some_branch` test for more convenient testing. The allure of a snappier update system was too powerful : )

as @ly29 suggested a while back, the existing update system has worked reasonably well for a long time now with only minor modifications at any given moment.

so.. what to do..?
- fix that bug ( those bugs? ...)
- how long to keep master in this state ?
- is it a waste of time pushing more commits onto master in this state
- do we revert to https://github.com/nortikin/sverchok/commit/476f26dee0f3faf5ebfea52464d4b241e3a1ae0a last known good configuration , or https://github.com/nortikin/sverchok/pull/3046

zeffii:
how persistent are 
```
class SvNodeTreeCommon(object):
    ...
    sv_links = SvLinks()
    nodes_dict = SvNodesDict()
```
between reloads?

Durman:
I think reverting of next additions from master can bring it to stable condition especially last one.
#3134
#3089
#3046

zeffii:
yeah #3046

vicdoval:
I would not say the update system is broken, is the reload event what is broken.
Have you experience any issue during a regular sverchok session? 
I understand this is specially annoying for us, becacuse the regular user does no reload scripts ever. Just opens and closes files.
After #3046 when I have been using Sverchok as user i have been delighted with its performace.
The problem of the branches is the we don't have testers, so none of this issues appear when it was in a branch.
Let me study it today and if i don't find a solution I will turn the last updates into a "experimental feature" that can be activated through the preferences

ly29:
On reload, reset everything basically is the only sensible strategy.

zeffii:
yeah, the system seems to behave well until i refresh python. Makes it painful to debug my own branches which may themselves not be entirely stable

portnov:
For me, sometimes the node tree does not get recalculated when I link / unlink nodes. Did anyone notice something like that?

zeffii:
yes @portnov ,  come to think of it that has happened too. sometimes i can retrigger an update with a framechange (left right) or updating a node upstream from the "stale" node

vicdoval:
I think now it is fixed please test and report 😃

[...]

PR #3162, opened 2020-05-02
clearing cache when un-registering and on pre-loading

vicdoval:
Following this adivse https://github.com/nortikin/sverchok/issues/3159#issuecomment-622935604 
I found not clearing the memory was breaking things during the reload.

This fixes (i believe) #3159 (i have been reloading hundreds of times with no issues) if you find anything please report 😃


zeffii:
yes!!! looks good so far.

PR #3166, opened 2020-05-03
"Adaptive tessellate" node

portnov:

Given a Surface object, it is easy to convert it to mesh by evaluating it on a cartesian grid, by use of "Evaluate Surface" node. However, in some cases a cartesian grid is not suitable. When the surface has some "almost flat" places and others, that are much more "bent" or even "sharp", a naive cartesian grid approach:

* generates too many points on flat areas
* and generates too few points in the curvy areas.

## Solution description

* Start with a cartesian grid.
* Then add more points into "most interesting" grid cells. "Interesting" cells may be defined as:
  * Having larger area (more precisely, area stretching factor)
  * Having bigger curvature value. Curvature may be defined as Gaussian curvature, mean curvature or as bigger one of the principal curvatures.
* Make a Delaunay triangulation of all these points
* And then map this triangluated thing onto the surface.

This approach can not automatically handle cases where the surface should have sharp edges. However, if we just happen to know where these sharp edges are, we can manually add points on these edges before building a Delaunay triangluation.

On the left: a surface evaluated with cartesian grid; on the right: the same surface with adaptive tessellation. Each of these meshes has (nearly) the same count of triangles - 41K.
Screenshot_20200503_224620
The same with wireframe enabled.
Screenshot_20200503_224647

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/surface/adaptive_tessellate.rst
+++ b/docs/nodes/surface/adaptive_tessellate.rst
@@ -46,6 +46,36 @@ the surface's U/V coordinates frame.
 Since the node uses Delaunay triangulation, it is enough to just apply "Dual
 Mesh" node after it to have a Voronoi subdivision.
 
+**Assumptions and Limitations**:
+
+This node uses a relatively simple algorithm without any "AI" or too complex
+computations involved. Thus, the node can not handle all cases. Especially,
+
+* The node can not generate very good subdivision if the surface has places
+  where it is much more stretched along one direction compared to another. For
+  example, if at some place a `[0; 1] x [0; 1]` square of U/V space is mapped
+  onto a rectangular part of the surface with side lengths of 10 and 100 - the
+  generated faces will appear stretched in one direction. However,
+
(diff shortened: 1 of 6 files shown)


portnov:
Since the node uses Delaunay triangulation, it is enough to just apply Dual Mesh node to have a Voronoi subdivision:

Screenshot_20200503_000638

portnov:
Screenshot_20200504_131747

portnov:
Screenshot_20200504_200952
Screenshot_20200504_201725
Screenshot_20200504_201756
Screenshot_20200504_201858
Screenshot_20200504_202116
Screenshot_20200504_202734
Screenshot_20200504_202840

vicdoval:
Wouldn't you like to add some separators to the menu of the new categories? Would make them easier to read/understand or even a any sub-menu

portnov:
Yeah, probably it's the time.

Issue #3171, opened 2020-05-05
Viewers color


vicdoval:
![image](https://user-images.githubusercontent.com/10011941/81107819-09863980-8f18-11ea-99d2-4853c30f6891.png)
The default viewers color is almost identical to one of error colours, 
and  the default red on the top of the node is too similar and different generating un-confortable feelings to the eye... (imho)

I propose to change it reducing the saturation to 0.5
![image](https://user-images.githubusercontent.com/10011941/81109204-07bd7580-8f1a-11ea-8aaa-be6adf167892.png)
![image](https://user-images.githubusercontent.com/10011941/81109450-64209500-8f1a-11ea-9388-797d0bddfc8c.png)
![image](https://user-images.githubusercontent.com/10011941/81111475-6fc18b00-8f1d-11ea-8c82-c6bfb195f820.png)

What do you think?

nortikin:
hope we can extend color schemes after nippel blossom and so on to please all needs

PR #3172, opened 2020-05-06
Viz color change

vicdoval:
Attends #3171 

image


nortikin:
we can add some color schemes also

Issue #3073, opened 2020-04-14
2D gl Viewer


zeffii:
## the what

- input geometry is verts + edges + faces (not matrices)
- flatten input verts to 2d (dropping Z, user must pre-transform to keep complexity out of this node)
- shows input geometry, with a grid (optionally, to indicate the boundaries of the geometry) 
- in case of multiple input geometry sets, the display will tile vertically / horizontally + include object index (optional)
- the viewer will display in nodeview
- color will be available for verts / edges / face   , all faces are single colour

## the why

- i'm looking for easier way to display patterns / motiefs / shapes / profiles inside the nodeview

## the when

no ETA.

PR #3177, opened 2020-05-07
Viewer  2D Node

vicdoval:
A visual reference in the node-tree
Number, Path, Curve and Mesh modes
Colors per vertex, per polygon and per edge, or per object
image
image
image
image
image
image

image

image

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


nodes/viz/viewer2d.py
```diff
@@ -0,0 +1,744 @@
...
+    ]
+    plane = [
+        ('XY', 'XY', 'Input UV coordinates to evaluate texture', '', 1),
+        ('XZ', 'XZ', 'Matrix to apply to verts before evaluating texture', '', 2),
+        ('YZ', 'YZ', 'Matrix of texture (External Object matrix)', '', 3),
+
+
+    ]
+    n_id: StringProperty(default='')
+
```
Durman, line 424: `SverchCustomTreeNode` already has this property I guess.

zeffii:
cool !  another todo less https://github.com/nortikin/sverchok/issues/3073

nodes/viz/viewer2d.py
```diff
@@ -0,0 +1,744 @@
...
+        nvBGL.callback_disable(node_id(self))
+
+    def sv_copy(self, node):
+        # reset n_id on copy
+        self.n_id = ''
+
+    def sv_update(self):
+        # handle disconnecting sockets, also disconnect drawing to view?
+        if not ("Float" in self.inputs):
+            return
```
zeffii, line 735: because this is quite a dynamic node, with many sockets (hidden), and none of them are called "Float", you could instead do a length test.  Say if this node has a total of 7 sockets (in various states of hide/shown), then the first line in sv_update could be 
```python
if not len(self.inputs) == total_expected_socket_count:
    return

```

zeffii: because this function is called several times, i think each time a new socket is added to the ui (like sv_init). Best to return early when we know the node isn't entirely initialized.

zeffii: in these nodeview viewer nodes, the purpose of this code in the try/except is to make sure the node stops drawing when there are no longer any incoming links

vicdoval: That part of the code was a vestige of the base node i started from (easing 0..1). But is it really necessary? the node stops drawing when i un-link the sockets... (de bgl call is disabled at the beginning of the process function)

zeffii: there are two things here:
- when the node is entirely disconnected, the sv_update is called one more time. that where the try/except gets a change to disable drawing.
- the initial "early" return is probably not that important anymore.. if the sv_update function itself is only doing very minimal processing (ie.. none in this case) , but...  if you don't check that all sockets are created you'll get some Exceptions during the sv_init (at least.. that was always the case in the old update system..i'm not sure that's changed)

vicdoval:
Random vertex colours added
image

zeffii:
if you can remove `sv_update` , then I assume either
- process is called a last time on that node, after the last connected socket is disconnected?
- the update system is automatically disabling callbacks for those nodes that are no longer linked in the tree? 
- magic

vicdoval:
> * process is called a last time on that node, after the last connected socket is disconnected

😄 it is called a last time, just disables the callback and returns

zeffii:
oh cool, then the "hold out" checks can be removed from the other viewers too.  but not a priority.

zeffii:
really neat node btw

vicdoval:
thanks! @zeffii

portnov:
Interesting...
Screenshot_20200508_221044

Issue #3077, opened 2020-04-15
Overriding Blender update events.


Durman:
## Problem statement

- During coding of nodes you should take in account update system of Blender. 
- Blender `update` events does not suit well to the Sverchok required `update` events. 
- It is cheaper to spend some calculation to make update smarter and save time on nodes recalculations.

## Possible solution

First of all such system should get all signals from Blender. It is possible to do by placing function in update methods. For example:
```Python
class SverchCustomTree:
    def update(self):
        tree_update(self)
```
After getting signals the system can call appropriate methods.
```Python
def tree_update(node_tree):
    for node in node_tree:
        if hasattr(node, 'sv_update'):
            node.sv_update()
```
It is simpler to make update smarter.
```Python
def tree_update(node_tree):
    type_update = get_type_update()
    if type_update == 'new_links_creation`:
        new_links = get_new_links()
        fro link in new_links:
            update_node_ui(link.from_node)
            update_node_ui(link.to_node)
```

Durman:
Probably for the start I will try to implement class which could just print type of detected events (node tree update, property update and so on). For this I think enumeration of possible blender events would be nice to create and class which can contain current event(s). Also this class should detect nested events. 

Printing should looks like this: `update of tree="NodeTree.001" was detected`. 

If first stage will be succeeded next step will be to utilize this functionality for generation Sverchok own events. 

List of method for listening:

- [NodeTree.update](https://docs.blender.org/api/current/bpy.types.NodeTree.html#bpy.types.NodeTree.update)
- [NodeTree.interface_update](https://docs.blender.org/api/current/bpy.types.NodeTree.html#bpy.types.NodeTree.interface_update) - not sure about this

- [Node.update](https://docs.blender.org/api/current/bpy.types.Node.html#bpy.types.Node.update)
- [Node.insert_link](https://docs.blender.org/api/current/bpy.types.Node.html#bpy.types.Node.insert_link)
- [Node.init](https://docs.blender.org/api/current/bpy.types.Node.html#bpy.types.Node.init)
- [Node.copy](https://docs.blender.org/api/current/bpy.types.Node.html#bpy.types.Node.copy)
- [Node.free](https://docs.blender.org/api/current/bpy.types.Node.html#bpy.types.Node.free)

- Property update
- Frame change update
- Undo update ?

If someone know other update methods please let me know.

Durman:
The process should consist with the next steps:
1. Listening update events. 
[...]

Issue #3058, opened 2020-04-11
Update system requirements 


Durman:
I'm going to create list of desirable features which new update system should include. What can be included in this list is not obvious and can be discussed.

- Performance issue.:

  - Don't recalculate what already was calculated:
    - Nodes should keep results of their calculations.
    - All changes (in parameters, inputs, data from Blender scene ...) should be tracked.
    - Manually prohibition of recalculation (freezing mode)

  - Don't calculate if result is need for nobody:
    - Create output nodes which cause work of other nodes (viewers).
    - Mode for output nodes switch on/off of showing of calculation results

- User should understand what's going on:

  - Use annotations for showing error and other messages.
  - Color system for showing error nodes and different modes or whatever.
  - Showing performance information.

- Performance of workflow:

  - supporting of node subtrees
  - pass data without changes throu node (bypass mode. It is not possible for all nodes) 

related: #2770 #2380 #1934 #1990

Durman:
- User should understand what's going on:

![node statistics](https://user-images.githubusercontent.com/28003269/81104852-2de81300-8f24-11ea-8ece-e382e92db0d4.gif)

vicdoval:
Really nice! 
No sure the recalculations number is that useful, I guess after 3 minutes tweaking the node-tree it must be a very high number...

Durman:
I have found that `Viewer Draw` nodes effect on each other performance. For example first node update time is 1s. Then another node will update faster if the first node will be disabled. I think there should be possibilities for performance improving. 

`BeshViewer` has similar performance but they does not effect on each other. In this way they are faster.

enzyme69:
If only we can ESC when Blender or SV is frozen calculating ....
-- 


------------------------------------------------------------
- Jimmy Gunawan
http://blendersushi.blogspot.com.au
http://mayaspiral.blogspot.com.au
http://houdoodles.blogspot.com.au
http://puppetar.blogspot.com.au

Durman:
I'm thinking about workflow where error nodes returns their previous result so the rest of a tree still could do something.
![ignore error](https://user-images.githubusercontent.com/28003269/81272920-385af780-905f-11ea-8f20-cc0e6f627f81.gif)

Durman:
Tool which can help to find nodes with slow performance. All nodes with update time more than given value will be dyed into a color.
[...]

PR #3179, opened 2020-05-07
New update core

Durman:

#3077 #3058

Previous attempts: 
#2380 #2393

After a year from previous attempt I'm trying to handle this task again. I have made some code and it looks like the approach is correct. Actual result is a node tree is never recalculate everything (only if base node for all other was not changed). Also there are some extra logic which prevents from useless updates.

In general process can described in next steps:
- get signals from Blender
- analyze them with using tree reconstruction
- translate them into Sverchok events
- correct tree reconstruction according this events
- get outdated nodes from tree reconstruction
- try update given nodes
- colorize nodes after all

I splited all code into several modules:
- tree_recnstruction.py
> This data structure is coping data structure of existing Sverchok node trees
It helps to analyze changes which happens in a node tree
Also it is used for walking on tree for searching nodes which should be recalculated
It is always consistent to Blender trees
So it can be used for searching connections between nodes
The only problem is it has not any information about sockets for simplicity
They can be added in future
- hashed_tree_data.py
> This module keeps node and links instances of bpy.types.NodeTree objects
They are available by their IDs
It used for converting nodes of Reconstruction tree to Blender nodes
But also can be used for any other purposes when collection of node_tree.links and node_tree.nodes are unchanged
It is known that keeping such links can lead to Blender crash
So all this data is refreshing if collection of nodes or links was changed
This module can be adopted to getting links of a socket for constant time
- events.py
> It used for getting signals of Blender, analyze them, keeping tree reconstruction, updating and colorizing nodes.

If someone would like to test keep in mind next things:
- Many things are broken now
- `node statics` and `colorize` panels can help to understand what is going on
- also extra info can be get from console in debug mode
- not all node properties works correctly
- keep in mind what does not still work according list below

Panels:

## Preflight checklist

- [ ] Code changes complete.
  - [x] handle reroutes
  - [x] handle frame nodes
  - [x] handle show in viewport node tree property
  - [ ] handle freeze whole tree node tree property
  - [ ] handle update current tree operator
  - [ ] handle update all trees operator
  - [ ] handle macros efficiently
  - [ ] handle post load signal
  - [ ] support monads
  - [ ] support monads loops
  - [ ] support color presets
  - [ ] handle animation event
  - [ ] handle animate node tree property
  - [ ] handle undo event
  - [ ] handle property update events with costume update function
  - [ ] handle `nodes_dict`
  - [ ] open examples
  - [ ] all nodes should calculate the data for all output nodes
- [ ] Manual testing done. 
  - [ ] open existing layouts
  - [ ] open existing layouts with deprecated nodes
  - [ ] export/import json
- [ ] performance tests
- [ ] Unit-tests implemented.
- [ ] cleaning outdated code
- [ ] Ready for merge.


vicdoval:
Does this override all the last changes to the update system?

Durman:
Yes and even more. It overrides existing color system, update core of monads, refreshing IDs, recreating node trees, handle property updates and I think it is not the full list. I have tried to impact as less as possible but in such matters it is not as simple.

In this way this PR is quite revolutionary. It undo many efforts of other developers.

zeffii:
i welcome any effort to introduce less updates of the node tree. When you feel it's ready, then it's ready for testing by as many people as possible. but i think it's unwise to merge it into master without extensive testing (not just "it passes the automated tests i wrote" , tests are only as good as the mortals that write them). Users will find ways to break even the most hermetic code.

You will get no resistance from me if there's a way to have the current update system and this implementation life side-by-side , hot swap-able, letting users go into "experimental update system" mode. That way you get many more people testing it harder, yet we can still switch to the current update system if we need to.

big changes need big testing, and expect big skepticism.

Durman:
> You will get no resistance from me if there's a way to have the current update system and this implementation live side-by-side, hot swap-able, letting users go into "experimental update system" mode. 

 Than more I'm working on this than less possible it seems to me to maintain two update system simultaneously.  Anyway it is far from ready to merge.

Durman:
show in viewport

PR #3181, opened 2020-05-07
Draw haxk prop search

zeffii:
is this acceptable behaviour if a datablock pointer changes, there's no simple way to propagate that signal to the node. but we can use the draw_button cycle to inform the user that the currently loaded path/txt / datablock  has been renamed,  and the node must be refreshed to reflect the change.

image

### DIFF
--- a/nodes/generators_extended/profile_mk3.py
+++ b/nodes/generators_extended/profile_mk3.py
@@ -392,6 +392,20 @@ class SvProfileImportOperator(bpy.types.Operator):
         updateNode(context.node, context)
         return {'FINISHED'}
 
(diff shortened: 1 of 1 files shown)


zeffii:
unless we have some stupid handler / msgbug that only listens to the renames of pointerproperties on nodes in the current blend. .. that's just compicated.

Durman:
It looks more like it can't be refreshed than name is outdated for me.

zeffii:
if the user is renaming a textblock, it's probably happening from TextEditor, and the user can hit Ctrl+Enter from there and that will refresh the PFMK3 node, in all ways.

it's not idea, but better than broken.

Issue #127, opened 2014-05-02
Time to study numpy


zeffii:
![bonus](https://cloud.githubusercontent.com/assets/619340/2860041/8feabb82-d1bd-11e3-86a3-6cc1efda2207.PNG)
http://docs.scipy.org/doc/numpy/reference/

PR #3183, opened 2020-05-08
"Scalar field curvature" node

portnov:
This node calculates surface curvatures for surfaces defined by scalar field iso-surfaces - a.k.a. implicit surfaces, `F(x,y,z) = C`. The most clearly useful this will be in combination with "marching cubes" from sverchok-extra, but may give interesting effects by itself.

If we have a scalar field defined by `V = F(x,y,z)`, then at each point in space `(x,y,z)` it has a value of `V`, then through each point in space goes an iso-surface defined by `F(x,y,z) = V`. We can calculate curvature of that surface at that point. So, it appears that given one scalar field, we can define another one, defined by `K(x,y,z) = Curvature(F(x,y,z) =V at (x,y,z))`. We can simply evaluate that new scalar field at any point, for example at points of the surface `F(x,y,z)=V` itself; or we can do other strange things with this new scalar field...

Screenshot_20200508_234500
Screenshot_20200508_234800

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/nodes/field/scalar_field_curvature.py
+++ b/nodes/field/scalar_field_curvature.py
@@ -8,7 +8,7 @@ from sverchok.node_tree import SverchCustomTreeNode, throttled
 from sverchok.data_structure import updateNode, zip_long_repeat, match_long_repeat, ensure_nesting_level
 from sverchok.utils.logging import info, exception
 
-from sverchok.utils.field.scalar import SvScalarField, SvScalarFieldGaussCurvature
+from sverchok.utils.field.scalar import SvScalarField, SvScalarFieldGaussCurvature, SvScalarFieldMeanCurvature, SvScalarFieldPrincipalCurvature, ScalarFieldCurvatureCalculator
 
 class SvScalarFieldCurvatureNode(bpy.types.Node, SverchCustomTreeNode):
     """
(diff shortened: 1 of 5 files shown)


portnov:
We can even draw iso-surfaces of curvature field of another scalar field...

Screenshot_20200508_235728

portnov:
Screenshot_20200509_004303

zeffii:
goursat? :)   https://github.com/nortikin/sverchok/issues/127#issuecomment-42117489

portnov:
Huh.
Screenshot_20200509_102532

zeffii:
that _is_ ridiculous @portnov .

nodes/field/scalar_field_curvature.py
```diff
@@ -0,0 +1,74 @@
...
+            default = 0.001,
+            precision = 4,
+            update = updateNode)
+
+    def draw_buttons(self, context, layout):
+        layout.prop(self, 'step')
+
+    def process(self):
+        if not any(socket.is_linked for socket in self.outputs):
+            return
```
Durman, line 41: Let to the update system to decide, whether it is worth call update or not.

portnov: Currently most of nodes do similar check. I agree this is better for update system to decide, but as far as I understand in it's current state it is not clever enough to do not recalculate nodes that do not have their outputs connected. Moreover, there are nodes that should do some work even when their outputs are not connected - all kinds of Viewers.
When you'll merge more clever update system, I think we'll all be happy to remove all these checks from all nodes :)

portnov:
Screenshot_20200509_162946

nodes/field/scalar_field_curvature.py
Durman: You understand it wrong. I'm saying about how exactly current update system works. It is redundant code relatively to current update system.

vicdoval: Also some nodes calculate different things based on the outputs connected. For example the Box node does not compute the polygon data if the polygons output is not connected, so I'm not sure how that could be handled from the update system...

Durman: In my current attempt I have sacrificed such optimisations. Because for solving this issue workflow of all nodes should be reconsidered.

nortikin:
Снимок экрана от 2020-06-24 21-41-22

PR #3186, opened 2020-05-09
Snlite update

portnov:
* Support Curve (`C`), Surface (`S`), ScalarField (`SF`), VectorField (`VF`) sockets in the SNlite node.
* Add some initial documentation for the node.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/generator/script1_lite.rst
+++ b/docs/nodes/generator/script1_lite.rst
@@ -1,6 +1,205 @@
-SNLite Docs
------------
+Script Node Lite
+================
 
-There's no rst for this at present, please read the preliminary (but comprehensive) docs on :
+Functionality
+-------------
 
-https://github.com/nortikin/sverchok/issues/942
+This node allows the user to write arbitrary node logic in Python. It is useful
+when there is no standard node in Sverchok for your particular purpose; you may
+want to use this if you want some complex piece of logic to be used in a couple
+of your own projects, or if you are prototyping a new standard node for
+Sverchok itself.
+
+The script node implementation is intended for quick, non serious, experimentation.
+
+Features
+--------
+
+- easy template importing directly to node, or to text block (the latter is useful for making changes to a template before loading... as is needed with the `custom_draw_complex` example )
+
+.. image:: https://cloud.githubusercontent.com/assets/619340/25421401/ff29eed2-2a5c-11e7-9594-f85295178f57.png
+
+- doesn't force an output socket, but an input socket is expected (else how does it update?!)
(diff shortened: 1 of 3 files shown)


zeffii:
cool.

portnov:
```
"""
in   h_in s
out  field_out SF
"""

# This script shows an example of generating a custom Scalar Field object.
# This may be useful when the formula defining the field is too complex
# to be typed in the "Scalar Field Formula" node.

import math
import numpy as np

from sverchok.utils.field.scalar import SvScalarField

# You can implement some auxiliary funcitons
# to be used in your scalar field definition.
def erdos(z, n):
    return abs(z**n - 1)**2 - 1

def lemn3(x, y):
    return erdos(x + y*1j, 3)

# To make your own scalar field, you have to
# subclass a SvScalarField class.
#
# You have to define two methods in it: evaluate and evaluate_grid.
#
class MyField(SvScalarField):
    def __init__(self, h):
        # the field can have any parameters you want
        self.h = h
        
    def evaluate(self, x, y, z):
        # This method must calculate one real value for given x,y and z coordinates.
        # If you've already implemented evalaute_grid, you may simply write:
        #
        # def evaluate(self, x, y, z):
        #     return self.evaluate_grid(np.array([x]), np.array([y]), np.array([z]))[0]
        #
        # but employing all the numpy's vectorization magic for one single point
        # may have some performance overhead.
        w = x + y*1j
        h = self.h
        return lemn3(x,y)**2 + (16*abs(w)**4 + 1)*(z*z - h*h)
    
    def evaluate_grid(self, xs, ys, zs):
        # This method must calculate a numpy array of shape (n,)
        # given three numpy arrays xs, ys and zs of the same shape.
        # If you've already implemented evaluate() method, and do not
        # want to dig into numpy magic, or you do not care about
        # performance, or your algorithm is just impossible to vectorize,
        # you can write simply:
        #
        # def evaluate_grid(self, xs, ys, zs):
        #    return np.vectorize(self.evaluate, signature='(),(),()->()`)(xs, ys, zs)
        #
        ws = xs + ys*1j
        h = self.h
        return lemn3(xs,ys)**2 + (16*abs(ws)**4 + 1)*(zs*zs - h*h)

# And now you return an instance of your class for each set of the input parameters.

field_out = []
if h_in:
    for hs in h_in:
        for h in hs:
            field = MyField(h)
            field_out.append(field)
```

Screenshot_20200509_162306

PR #3189, opened 2020-05-09
Presets functionality upgrade

portnov:
This adds:

* Support for "Standard" presets, which are supposed to be distributed with Sverchok (some kind of example settings). 
* Support for "Per-node presets", i.e. presets which store settings of one and only one node. Such presets are displayed in the node's N panel, so when the node is selected it is possible to switch it between several pre-saved settings. Some node classes already had similar functionality (I remember Spiral and Regular Solid).

"Per-node" presets are saved in special preset categories, which are distinguished with "/Node/" title prefix. For example, category "/Node/ Box" can only store presets for the "Box" node. Such categories are created automatically when you first create a preset for the certain type of node from the node's N panel. It is possible to use the usual Presets panel under the T panel to manage such presets.

Screenshot_20200510_104238

Also the list of all presets defined for the node is available in the node tree editor by *Shift-P* shortcut.

Screenshot_20200510_104511

## Difference between standard presets and Examples

Historically, our Examples menu contained examples of more or less complete setups, consisting of several nodes. Some of those examples are relatively complex. So, Examples is for examples of what can one do with Sverchok.

Per-node Presets, which are to be distributed with Sverchok, are just some combinations of settings of specific complex node, which developer thinks will be frequently used.

Standard Presets that are not per-node may be some small parts of logic that developer thinks may be frequently used. Probably it worth wrapping such pieces into Monads before packing as presets. As an example of such preset, I included Lloyd 2D monad.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/utils/sv_IO_panel_tools.py
+++ b/utils/sv_IO_panel_tools.py
@@ -255,7 +255,7 @@ def handle_enum_property(node, k, v, node_items, node_enums):
         node_items[k] = v
 
 
-def create_dict_of_tree(ng, skip_set={}, selected=False, identified_node=None):
+def create_dict_of_tree(ng, skip_set={}, selected=False, identified_node=None, save_defaults = False):
     nodes = ng.nodes
     layout_dict = {}
     nodes_dict = {}
@@ -283,7 +283,11 @@ def create_dict_of_tree(ng, skip_set={}, selected=False, identified_node=None):
 
         IsMonadInstanceNode = (node.bl_idname.startswith('SvGroupNodeMonad'))
 
-        for k, v in node.items():
+        if save_defaults:
+            node_props = ((k, getattr(node, k)) for k in node.bl_rna.__annotations__.keys())
+        else:
+            node_props = node.items()
+        for k, v in node_props:
 
             display_introspection_info(node, k, v)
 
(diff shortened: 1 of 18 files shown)


utils/sv_IO_panel_tools.py
```diff
@@ -283,7 +283,11 @@ def create_dict_of_tree(ng, skip_set={}, selected=False, identified_node=None):
 
         IsMonadInstanceNode = (node.bl_idname.startswith('SvGroupNodeMonad'))
 
-        for k, v in node.items():
+        if save_defaults:
+            node_props = ((k, getattr(node, k)) for k in node.bl_rna.__annotations__.keys())
```
zeffii, line 287: can i ask you to turn this into a function, which skips over pointers.
```python
node_props = get_node_annotations(node, pointers=False)
```
if this is to be used by nodes in general

portnov: Ok

vicdoval:
Great feature!! Wouldn't make more sense to place it in a different box layout just under the label? It seems a little strange to see it between "Edit Source" and "recreate node"

portnov:
@vicdoval you mean move Presets block higher, above Help block?

vicdoval:
Yes

portnov:
No problem :)
Screenshot_20200510_011132

portnov:
Well, I think I'll better replace this list of buttons with a menu box. Otherwise the panel will grow too big when there will be a lot of presets...
Screenshot_20200510_104238

Or maybe make it an option in settings?

Also it is possible to create a menu with list of presets for active node, which pops up by a hotkey (Shift-P, for example).

Screenshot_20200510_104511

What do you think?

zeffii:
i think a popup is more preferable per node, than any additions of lists/menus in the NodeITem N panel menu.

PR #3195, opened 2020-05-10
List sort mk3

vicdoval:
The actual List Sort node have some problems:
Can't sort level 1 data, as objects or curves.
The level is not consistent with the other List nodes. (Goes in oposite direction)

This iteration solves this problems.
I also changed the class name from `ListSortNodeMK3 `to `SvListSortNode`

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


vicdoval:
Coding this node I discovered a blender glitch that I find interesting to share:

This node uses the  `changable_sockets` method. When it is called the first time that you check if output is linked it returns False even if it is connected. this results in the node not being updated.

This behaviour is invisible most of the times because if you change the input type usually you have also to change the output link. Anyway I added a boolean check to override the glitch

PR #3200, opened 2020-05-10
"Rounded rectangle" node

portnov:
This node generates a Curve object, representing a rectangle with rounded corners.

Screenshot_20200510_235429

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


vicdoval:
I'm not sure if it should be an option in "evaluate curve" or it could be automatic depending on the curve class, you may know, but wouldn't be nice that in the case of cyclic curves the "evaluate curve" node could remove the repeated vertex?

zeffii:
@portnov  is the samples setting something like intrange- in interpolation node?
image

portnov:
@vicdoval yeah, probably you're right, each curve could know if it is closed, and bring this information to Evaluate Curve node. A point to think about.

@zeffii the resulting curve is composed out of 8 segments: 4 arcs and 4 line segments. Arc segments have domain of `[0; pi/2]`, line segments have domain of `[0; L]`, where L is the length of that segment. So the whole domain of the curve is something like `[0; pi/2] + [pi/2; pi/2 + L1] + [pi/2 + L1; pi/2 + L1 + pi/2] + ...`. What "Evaluate Curve" does is simple: it takes the whole domain and splits it into N equal segments, that's all. Since domains of different segments have different length, we see different density of points on the curve.
There is an "Even Domains" switch on this particular node, which rescales domains of all segments into `[0;1]` - then each arc and each straight segment will have equal number of points after "Evaluate Curve", but that does not mean they will be distributed evenly in space - distribution of points in space will still depend on sizes of rectangle and radius of arcs.
To make points actually evenly spaced, you'll have to either use "Curve Length Parameter" node instead of "Evaluate Curve", or "Naturally Paramertized Curve". These nodes do non-trivial calculation to figure out how the curve domain is to be split (into non-even parts) so that corresponding segments of the curve would have even lengths.

zeffii:
@portnov understood.  I'm looking forward to taking this stuff for a ride :)  i'd like to stay blissfully ignorant of the code for as long as possible :)

PR #3212, opened 2020-05-12
poll fix

zeffii:
this suppresses a poll exception when `F3` is triggered in 3dview, and the Temporal viewer operator is encountered during the search input.


portnov:
@zeffii probably it's better to add `bl_options = {'REGISTER', 'UNDO', 'INTERNAL'}` for this operator. INTERNAL to make it not appear in menus.

zeffii:
yeah, do that.

PR #3213, opened 2020-05-12
overloads G function nicely

zeffii:
adds a `bpy.types.Macro` to Sverchok, currently no keymap 

- have a node outputting into Stethoscope.
- select the stethoscope node.
- `f3` -> "Overload "   (shows Overload G function)
- hit Enter, move the mouse around
- hit enter to let go.   booooom.  automatic repositioning of callback / nodeview drawing.


zeffii:
the `execute` function could be a bit smarter about active node tree, and testing if a bgl node is present in the list of selected nodes -- incase there are multiple nodes being moved about.

zeffii:
i think this operator/code could be orphaned now, seen as the callback itself is dynamically populating data location at runtime.

PR #3216, opened 2020-05-12
Add a couple of examples.

portnov:
Technical PR to run Travis on it.

### DIFF
--- a/json_examples/Fields/Mesh_Attractor.json
+++ b/json_examples/Fields/Mesh_Attractor.json
@@ -4,7 +4,7 @@
   "groups": {},
   "nodes": {
     "Apply Field to Surface": {
-      "bl_idname": "SvExApplyFieldToSurfaceNode",
+      "bl_idname": "SvApplyFieldToSurfaceNode",
       "height": 100.0,
       "hide": false,
       "label": "",
@@ -18,7 +18,7 @@
       "width": 140.0
     },
     "Attractor Field": {
-      "bl_idname": "SvExAttractorFieldNode",
+      "bl_idname": "SvAttractorFieldNode",
       "custom_socket_props": {
         "0": {
           "prop": [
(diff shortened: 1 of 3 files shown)


portnov:
Wtf. What does it need with Donut example...

zeffii:
looks like the .blend being used to try all these .jsons , isn't being object wiped between imports.
so if there was a Curve Viewer node or something creating an object called Alpha, and then next json import is a BMeshViewer ,creating a mesh object also called Alpha..   I think this might happen

zeffii:
we can add something in sv_obj_helper to 
 - tell the user that the node is picking up some other node's data

if this is the problem i'll think about it for a bit.

zeffii:
i think we should also add a "whipe object and bpy.data.mesh/curve/... prior to eacc json import tests

zeffii:
something like this:
```python
def blend_cleanup_remove_obj_and_data(context):

    collection = bpy.context.scene.collection
    objects = bpy.data.objects
    
    meshes = bpy.data.meshes
    curves = bpy.data.curves
    metas = bpy.data.metaballs

    # first remove all the objects
    for obj in objects:
        collection.objects.unlink(obj)
        objects.remove(obj)

    # then remove their data
    for data_kind in [meshes, curves, metas]:
        for data in data_kind:
            data_kind.remove(data)
```

enzyme69:
🤟One suggestion: with EXAMPLES from Help can we make it as zipped blend with node tree or JSON?

zeffii:
relevant. https://blender.stackexchange.com/questions/46990/how-to-completely-remove-all-loaded-data-from-blender?noredirect=1&lq=1

zeffii:
i renamed the `basedata_name` to `RV_Alpha` , but there's a few nodes with `SvEx+` bl_idnames.

PR #3218, opened 2020-05-12
Node xform translate test

zeffii:
keep going.


zeffii:
i don't know how simple it is to overwrite leftclick + mouse move to trigger this Macro-Operator..

PR #3222, opened 2020-05-13
Exception mod 1

zeffii:
explore alternative repr


zeffii:
the issue here, and it will be the same for implementing backgrounds in idxviewer too, is to first obtain the `(h, w)` of each text element prior to drawing it,  then drawing the rect, and then drawing the text. Seen as the drawing context is 2d, that's the order it should happen in.

PR #3224, opened 2020-05-14
Zeffii patch 3

zeffii:
should test OK.

### DIFF
--- a/json_examples/Fields/Random_Voronoi.json
+++ b/json_examples/Fields/Random_Voronoi.json
@@ -152,7 +152,7 @@
         194.48263549804688
       ],
       "params": {
-        "basedata_name": "Alpha"
+        "basedata_name": "Foxtrot"
       },
       "use_custom_color": true,
       "width": 140.0


zeffii:
ideally we should start using more unique names for these Object basenames within our example .jsons , to avoid object/datablock "polution".

zeffii:
@portnov right?

zeffii:
the other side is ofcourse that these json tests are going to be repeatedly and sequentially loaded into a .blend which is unlike the normal user experience. But the .blend file in a state where there are many Objects from viewers in previous json imports,  but we lack a cleanup step.

portnov:
So, you've just renamed objects in those jsons to make tests pass?...

zeffii:
just a different basename , i think we can call this issue "name space polution".  

until we have a "wipe current blend datablocks" function prior to each json test,   this renaming is a sane solution.

zeffii:
because what would be the behaviour in real usecase scenario , where the is an active BMeshViewer generating Alpha objects,  and you then import a json that creates PolyLines with basedata_name Alpha.

not very satisfying :) but at least it's clear why the test is/was failing.

zeffii:
maybe SvObjectHelper should include a test functing, called at the very top of `process`, to end early and raise an Exception if it is in the same .blend as a Node generating objects with the same Basename.

i've tried to avoid thinking about that logic.

PR #3237, opened 2020-05-16
Nested Matrices for Apply matrices node

Durman:

Supporting nested list of matrix as input for `matrix apply` node. Use case is copy mesh inside mesh.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/matrix/apply_and_join.rst
+++ b/docs/nodes/matrix/apply_and_join.rst
@@ -45,9 +45,23 @@ This node has the following outputs:
 - **Edges**. Input edges list, repeated the number of incoming matrices. Empty if corresponding input is empty.
 - **Faces**. Input faces list, repeated the number of incoming matrices. Empty if corresponding input is empty.
 
+Usage of nested lists of matrices
+---------------------------------
+
+The node can handle with list of lists of matrices.
+It means that each object mesh has it's own list of matrices to apply.
+Applying list of matrices to a mesh creates its copies inside an object and transform its.
+Usage case is when there are bunch of seperate meshes which should be copied inside their objects.
+In this mode and with numpy implementation mode you will get numpy output for vertices and edges anyway.
+
+
 Examples
 --------
 
 .. image:: https://cloud.githubusercontent.com/assets/284644/6096652/ac13659e-afbf-11e4-83c9-e13b75c0e346.png
 
 .. image:: https://cloud.githubusercontent.com/assets/284644/6096654/b300fbfa-afbf-11e4-901b-1361a44238c2.png
+
+**Example of usage of nested list of matrices (list of lists of matrices):**
+
+.. image:: https://user-images.githubusercontent.com/28003269/82114656-2bc95400-976f-11ea-8b6a-b539060b03e6.png
--- a/nodes/matrix/apply_and_join.py
+++ b/nodes/matrix/apply_and_join.py
@@ -162,7 +162,7 @@ def apply_nested_matrices_py(
 
     if do_join:
         mesh = mesh_join(*list(zip(*meshes)))
-        return mesh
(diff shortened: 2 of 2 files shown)


Durman:
@vicdoval I did not catch of the idea of `output numpy: face`?

Durman:
I think the node should be simplified (not in the PR). 
- It should  not have any options other than `join mesh`. 

- If inputs vertices are numpy array numpy implementation should be used. 
- If numpy implementations is used output also should be numpy array.

In this PR I'm not going change existing logic so new version `mk++` of the node does not need.

vicdoval:
> @vicdoval I did not catch of the idea of output numpy: face?

The node only changes the faces if "join" is applied other ways outputs a copy of input faces "as is"



> I think the node should be simplified (not in the PR).
> 
> * It should  not have any options other than `join mesh`.
>   
> * If inputs vertices are numpy array numpy implementation should be used.
> * If numpy implementations is used output also should be numpy array.
> 
> In this PR I'm not going change existing logic so new version `mk++` of the node does not need.

I don't agree bout this topic because although is a very "expert stuff" I think it does not harm and gives more control to the user about  where to start and to stop the numpy chain for performance fine tuning.
Maybe a "Auto" mode could be added that makes some assumptions and hides the other buttons but i would definitely keep that possibility to choose somewhere

Durman:
> The node only changes the faces if "join" is applied other ways outputs a copy of input faces "as is"

If you set property `output is numpy` and get python list instead you think it is Okay? Is this documented behaviour?

Having of a node tree where switch between python and numpy should be done in every node manually has too obvious downsides to explain them I think.

vicdoval:
> If you set property output is numpy and get python list instead you think it is Okay? Is this documented behaviour?

That is a bit weird and could be improved 😄 

The thing of transforming manually is a tricky thing because 
Numpy Node to Numpy Node (+1 +1) (boost the first and the last)
Numpy node to python node (+1, -1)  (boost the first and slows the last )
Numpy node to two python nodes( +1 -2)
and so on...
So it is really dependant of the nodetree when is the best to cut the chain... it could be super cool to have a optimizer operator that makes some test in our nodetree and chooses when is better to switch....

Durman:
If a node does not have numpy implementation it won't output numpy arrays. So you will never get such chain: numpy - python - numpy - python. If second node does not support numpy then all next nodes will be deal with python data and use python implementation automatically.

I would suggest you to make concludes about performance of a node tree on a concrete examples. I would expect that in many layouts difference between python and numpy implementation of the node is nothing by comparison performance of such nodes as `remove doubles` or `adaptive polygons` for example.

vicdoval:
Is true it could be very small difference defending on the case, what I mean is:
_ `.tolist()` is slow (in some times is the 90% of the node processing time) 
_ `np.array(another_np_array)` is faster than `np.array(python_list)`
_ `for v in np_array` is slower than `for v in python_list`
_ in some cases  `np.array(python_list) + numpy implementation + .tolist()` is faster than using a python implementation.

With this rules i see that depending on the nodetree structure de performance can be improved if you know where to output numpy and when not. Anyway probably just a few users can use this wisely...

Durman:
> Anyway probably just a few users can use this wisely...

Exactly.

My rules are simpler:
- Numpy is faster than Python. 
- If a node get numpy data numpy implementation is used otherwise Python. 
- If you want boost performance of a node will try to grant numpy data to it. 

With this rules everyone can use Sverchok wisely.

vicdoval:
> If you want boost performance of a node will try to grant numpy data to it.

If you feed a viewer draw or a bmesh.ops node with numpy data it will make it slower...

Durman:
I already have said about performance and how it should be measured.

PR #3245, opened 2020-05-17
Component analyzer v is interior

vicdoval:
A enum index was repeated making inaccesible Vertex is_Interior function
- [x] Ready for merge.


vicdoval:
i don't get why all this commits appear again...  if my branch is up to date and i only did one commit....

vicdoval:
maybe i need to re-start  my whole repo to get rid of them...

Issue #3268, opened 2020-05-22
Join(123) macros does not place nodes properly


Durman:
Join(123) macros does not place nodes properly if they are in frame nodes.

zeffii:
yep
https://github.com/nortikin/sverchok/blob/c6661e810504cf5e4c00f07b8134b586f156cc90/utils/macros/join_macros.py#L47
and
https://github.com/nortikin/sverchok/blob/c6661e810504cf5e4c00f07b8134b586f156cc90/utils/macros/join_macros.py#L59

no ETA. soon.

zeffii:
the Join macro expects two selected Nodes. (let's call them A and B )

what to do if A but not B is framed, or in the same frame. 
what to do if B but not A is framed, or in the same frame.

i'm tempted to using absolute location,  but not Frame the two List Join nodes, let the user frame the List Join nodes if either of A and B aren't framed, (or in the same frame).

zeffii:
after some more thinking i think it makes sense to add the _ListJoin_ nodes using the `absolute_location`, and let the user frame the created nodes.

it's better than nothing, and it also doesn't make assumptions about things that the user might not have considered.

zeffii:
if i get some people agreeing, then i'll do it.

zeffii:
watching : https://www.youtube.com/watch?time_continue=61&v=7DqeJJTCbhc&feature=emb_logo

i think it should go like this:

- if both A and B nodes are in the same Frame,  then also add the ListJoins to that frame
- if Not (either they are both in different frames, or both unframed), then add ListJoins to the absolute (visual) location.

PR #3271, opened 2020-05-23
testing different join approach

zeffii:
this addresses the issue #3268


zeffii:
doesn't yet solve the "both in same frame" issue.

zeffii:
in particular the viewerdraw.

zeffii:
fixes #3268

PR #3284, opened 2020-05-27
New Evolver node (Genetic Algorithm)

vicdoval:
New Evolver node implementing a Genetics Algorithm system  inspired in Galapagos (the GH plugin) and the 9th chapter of "The Nature of Code" by Daniel Shiffman https://natureofcode.com/book/chapter-9-the-evolution-of-code/

For the genotype "A Number" nodes are used with its min and max values. 
To restrict which `A Number` nodes to use a Frame Node can be used.
image

Example of finding the box with less area (sum of all faces) where Suzzane can fit.
image
image

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/docs/nodes/scene/evolver.rst
+++ b/docs/nodes/scene/evolver.rst
@@ -1,0 +1,58 @@
+Evolver
+=======
+
+This node implements a Genetics Algorithm system inspired in Galapagos (the Grasshopper plugin) and the 9th chapter of "The Nature of Code" by Daniel Shiffman https://natureofcode.com/book/chapter-9-the-evolution-of-code/
+
+The system creates a starting population, evaluates the fitness of each "agent" and creates a new population bases on the crossover of the fitter agents and a mutation chance. The new population is evaluated and mixed every iteration rising the global fitness of the population by generating fitter agents.
+
+Every agent is composed by a set of genes that can oscillate among determined values, the fitter agents have more chances to spread their genetic information.
+
+In Sverchok every gen is defined by a "A Number" node. This node type has a minimum and a maximum that will control the variation possibilities of this gen. The node can use all the "A number" nodes of the node-tree or only some of them if they are inside a "Frame Node" that can be selected from the "Genotype" dropdown menu.
+
+The fitness of every member of the population will be evaluated by running the node-tree with the genes of that member and recording the value that is inputted in the "Fitness" socket.
+
+Parameters
+----------
+
+**Genotype**: dropdown menu to chose to use as genes either all 'A Number' nodes or only the ones inside a 'Frame' Node
+
+**Mode**:  "Maximum" and "Minimum" with the maximum selected a higher fitness value is considered better with the minimum a lower fitness will be considered as more fit.
+
+**Population amount**: Number of members of the population.
+
+**Iterations**: Iterations of the system
+
+**Random Seed**: Random Seed used in the system, this affects the initial population and the crossover and mutation process.
+
+**Fitness Boost**: This value affects how the ancestors are chosen. Taking two members, one with a fitness of 2 and another with a fitness of 1: With a value of 1 the first member will have the double chances to get selected than the second one. With a value of 2 the first one will have 4 times the chances of the second, with a boost value of 3 the first will have 8 times the chances of the second....
+
+**Mutation**: 0 the gene will never mutate, 1 will always mutate. The mutation has a 50% of being totally new and other 50% of being inherited from its ancestors and changed by a random amount (small mutation)
+
+**Max Seconds**: Maximum time to run the system, when achieved the system will stop providing the last valid generation of members
+
+Operators
+---------
+
+**Run**: Triggers the system
+
+**Set Fittest**: Will place the best evaluated values on the genes nodes
+
+Options
+-------
+
+**Output all generations**: When enabled the node will output all the members of all the generations. When disabled it will only return the last generation of members
+
+Inputs
+------
+
+**Fitness**: Input to get the fitness of each member (has to be a single number)
+
+
+Outputs
+-------
+
+**Genes**: Outputs the genes of the system
+
+**Population**: Outputs the population. It can be only the last generation or all generations.
+
+**Fitness**: Outputs the fitness of the population. It can be only the last generation or all generations.
--- a/nodes/logic/evolver.py
+++ b/nodes/logic/evolver.py
@@ -31,33 +31,91 @@ from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode
 from sverchok.core.update_system import make_tree_from_nodes, do_update
 
-Gene = namedtuple('Gene', 'name g_type, min_n, max_n, range, init_val')
+Gene = namedtuple('Gene', 'name, g_type, min_n, max_n, range, init_val')
+GeneList = namedtuple('GeneList', 'name, g_type, num_length, init_val')
 
 evolver_mem = {}
 
 def is_valid_node(node, genotype_frame):
-    if genotype_frame == 'All' and node.bl_idname == "SvNumberNode":
+    if genotype_frame == 'All' and node.bl_idname in ["SvNumberNode", "SvListInputNode"]:
         return True
-    if node.parent and node.parent.name == genotype_frame and node.bl_idname == "SvNumberNode":
+    if node.parent and node.parent.name == genotype_frame and node.bl_idname in ["SvNumberNode", "SvListInputNode"]:
         return True
     return False
+
+def number_gene(node):
+    num_type = node.selected_mode
+    if node.selected_mode == "float":
+        min_n = node.float_min
+        max_n = node.float_max
+        initial_value = node.float_
+    else:
+        min_n = node.int_min
+        max_n = node.int_max
+        initial_value = node.int_
+    gene = Gene(name=node.name, g_type=num_type, min_n=min_n, max_n=max_n, range=max_n-min_n, init_val=initial_value)
+    return gene
+
+def list_gene(node):
+    num_type = node.mode
+    if num_type == 'int_list':
+        num_length = node.int_
+        init_val = [node.int_list[i] for i in range(num_length)]
+    elif num_type == 'float_list':
+        num_length = node.int_
+        init_val = [node.float_list[i] for i in range(num_length)]
+    else:
+        num_length = node.v_int
+        mem_list = node.vector_list
+        init_val = [[mem_list[3*i], mem_list[3*i + 1], mem_list[3*i + 2]] for i in range(num_length)]
+    gene = GeneList(name=node.name, g_type=num_type, num_length=num_length, init_val=init_val)
+    return gene
 
 def get_genes(target_tree, genotype_frame):
     genes = []
     for node in target_tree.nodes:
         if is_valid_node(node, genotype_frame):
-            num_type = node.selected_mode
-            if node.selected_mode == "float":
-                min_n = node.float_min
-                max_n = node.float_max
-                initial_value = node.float_
+            if node.bl_idname == "SvNumberNode":
+                gene = number_gene(node)
             else:
-                min_n = node.int_min
-                max_n = node.int_max
-                initial_value = node.int_
-            gene = Gene(name=node.name, g_type=num_type, min_n=min_n, max_n=max_n, range=max_n-min_n, init_val=initial_value)
+                gene = list_gene(node)
+
             genes.append(gene)
     return genes
+
+def list_cross(ancestor1_gene, ancestor2_gene, o_gene):
+    mixing_factor = int(random()* o_gene.num_length)
+    new_gene = [ancestor1_gene[i] for i in range(mixing_factor)]
+    for g in ancestor2_gene:
+        if not g in new_gene:
+            new_gene.append(g)
+    return new_gene
+
+def set_list_node(gen_data, agent_gene, tree):
+    if gen_data.g_type == 'int_list':
+        for i in range(gen_data.num_length):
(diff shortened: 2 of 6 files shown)


nodes/scene/evolver.py
```diff
@@ -0,0 +1,357 @@
...
+    fitness = []
+    for agent in population:
+        tree.sv_process = False
+        for g, s in zip(genes, agent):
+            if g.g_type == 'float':
+                tree.nodes[g.name].float_= s
+            else:
+                tree.nodes[g.name].int_= s
+
+        tree.sv_process = True
```
Durman, line 88: If you get an error here the `sv_process` attribute will remain unexpectedly false.

vicdoval: What would you suggest? A try/except block or repeating 'sv_process=True' outside of the function?

portnov: try/finally.

Durman: Also you can use with statement. An example:

https://github.com/nortikin/sverchok/blob/9fb8060b96e733059324780a42b6bfb673b49694/core/events.py#L510-L519

Durman:
It will be quite unusual node for Sverchok. Probably it would be better to use group node (`monad`) as base for current node. The idea is that the node evaluate only nested tree and nested tree only used for genetic algorithm and nothing more. But implementation of monad is so complicated and fragile that this task should be impossible.

vicdoval:
Also Galapagos plugin is a unusual node in Grasshopper, it even opens a new window because this simulations sometimes are done along many hours. 
My first approach was using a different nodetree, but as the simulation is finally done in a operator it works fine in the same nodetree and can have all kind of nodes (even monads) or calls to the blender scene inside it.

vicdoval:
I am not sure about which category this node should fall into. Maybe it could go with `Pulga Physics` in a new `Simulation` category
Any ideas?

vicdoval:
Or it could go into the `Logic` category

Durman:
> The system creates a starting population, evaluates the fitness of each "agent" and creates a new population bases on the crossover of the fitter agents and a mutation chance. The new population is evaluated and mixed every iteration rising the global fitness of the population by generating fitter agents.

Would not you like to try to describe a population as a class? It looks it would perfectly fit to OOP paradigm.

I'm sure that OOP is popular now for reasons and Sverchok should learn to use it. OOP is definitely harder the procedural programming. But there are benefits as well. In particular our case it would let to reuse code. It is much easier to reuse class then a bunch of functions with a lot of input parameters. And finally it would let to understand code easier.

nortikin:
https://github.community/t/checkout-a-branch-from-a-fork/276/2
Howto checkout fork wants to merge commits
now, have you json?

nortikin:
> It will be quite unusual node for Sverchok. Probably it would be better to use group node (`monad`) as base for current node. The idea is that the node evaluate only nested tree and nested tree only used for genetic algorithm and nothing more. But implementation of monad is so complicated and fragile that this task should be impossible.

i'm trying now to monadify GA

nortikin:
I just can say, it works!
And it fits me.
Maybe monad can wait.

nodes/scene/evolver.py
```diff
@@ -0,0 +1,365 @@
...
+    return genes
+
+def initial_population(genes, init_pop_num):
+    '''Fills first population with random values ranged between min and max of each gene'''
+    population = []
+    for i in range(init_pop_num):
+        agent = []
+        for gene in genes:
+            agent_gene = gene.min_n + random() * gene.range
+            if gene.g_type == 'int':
```
nortikin, line 67: here can be start from number initial values

docs/nodes/scene/evolver.rst
```diff
@@ -0,0 +1,58 @@
...
+In Sverchok every gen is defined by a "A Number" node. This node type has a minimum and a maximum that will control the variation possibilities of this gen. The node can use all the "A number" nodes of the node-tree or only some of them if they are inside a "Frame Node" that can be selected from the "Genotype" dropdown menu.
+
+The fitness of every member of the population will be evaluated by running the node-tree with the genes of that member and recording the value that is inputted in the "Fitness" socket.
+
+Parameters
+----------
+
+**Genotype**: dropdown menu to chose to use as genes either all 'A Number' nodes or only the ones inside a 'Frame' Node
+
+**Mode**:  "Maximum" and "Minimum" with the maximum selected a higher fitness value is considered better with the minimum a lower fitness will be considered as more fit.
```
nortikin, line 19: also, why not to establish threshold for stop with optimal value and with optimal fitness value in enum with minimum and maximum. That will give us possibility to break before last iteration and will save time

nodes/scene/evolver.py
vicdoval: Good idea! Now the first agent of the first iteration will be the initial values. That way you can "accumulate" analysis by running it once,  clicking on `Set Fittest` and running it again.

docs/nodes/scene/evolver.rst
vicdoval: Done!

nodes/logic/evolver.py
```diff
@@ -0,0 +1,452 @@
...
+        if random_val:
+            for gene in genes_def:
+                agent_gene = gene.min_n + random() * gene.range
+                if gene.g_type == 'int':
+                    agent_gene = int(agent_gene)
+                self.genes.append(agent_gene)
+        else:
+            for gene in genes_def:
+                agent_gene = gene.init_val
+                self.genes.append(agent_gene)
```
Durman, line 81: It is not very good to do some complex stuff in init method. I would suggest to move this logic in separate method. The name of the method would say what are you trying to do here by itself.

nodes/logic/evolver.py
```diff
@@ -0,0 +1,452 @@
...
+        self.fitness = agent_fitness
+        # return agent_fitness
+    def cross_over(self, other_ancestor, mutation):
+        # new_agent = []
+        new_agent = DNA(self.genes_def, empty=True)
+        for g, g1, o_gene in zip(self.genes, other_ancestor.genes, self.genes_def):
+            if random() < mutation:
+                if random() < 0.5:
+                    #total gene reset
+                    new_gene = o_gene.min_n + random() * o_gene.range
```
Durman, line 105: It probably could be written in more human way. I don't know exactly how it should but something in a way like this:
```suggestion
            next_mutation_success= random()
            if next_mutation_success < mutation:
                success_barrier = 0.5
                if next_mutation_success < success_barrier:
                    #total gene reset
                    new_gene = o_gene.min_n + random() * o_gene.range  # I believe this string also can be improved
```

nodes/logic/evolver.py
```diff
@@ -0,0 +1,452 @@
...
+        self.time_start = time.time()
+        self.max_time = node.max_time
+        self.genes = genes_def
+        self.use_fitness_goal = node.use_fitness_goal
+        self.fitness_goal = node.fitness_goal
+        self.fitness_v = []
+        self.population_g = []
+        self.population_g.append(DNA(genes_def, random_val=False))
+        for i in range(node.population_n-1):
+            self.population_g.append(DNA(genes_def))
```
Durman, line 142: I would not suggest to copy node attributes to the class attributes. It would be better to use this properties via a node like this: `self.node.node_id`.

nodes/logic/evolver.py
```diff
@@ -0,0 +1,452 @@
...
+
+    def evaluate_fitness(self, tree, update_list, node):
+        tree.sv_process = False
+        for gen_data, agent_gene in zip(self.genes_def, self.genes):
+            if gen_data.g_type == 'float':
+                tree.nodes[gen_data.name].float_ = agent_gene
+            else:
+                tree.nodes[gen_data.name].int_ = agent_gene
+
+        tree.sv_process = True
```
Durman, line 91: as discussed

nodes/logic/evolver.py
```diff
@@ -0,0 +1,452 @@
...
+def is_valid_node(node, genotype_frame):
+    if genotype_frame == 'All' and node.bl_idname == "SvNumberNode":
+        return True
+    if node.parent and node.parent.name == genotype_frame and node.bl_idname == "SvNumberNode":
+        return True
+    return False
+
+def get_genes(target_tree, genotype_frame):
+    genes = []
+    for node in target_tree.nodes:
```
Durman, line 47: I see this function as class method of the Population class

nodes/logic/evolver.py
```diff
@@ -0,0 +1,452 @@
...
+
+        genotype_frame = node.genotype
+        evolver_mem[node.node_id] = {}
+
+        seed_set(node.r_seed)
+        np.random.seed(node.r_seed)
+
+        genes = get_genes(tree, genotype_frame)
+        update_list = make_tree_from_nodes([g.name for g in genes], tree)
+        population = Population(genes, node, tree)
```
Durman, line 253: I see this like this:
```suggestion
        population = Population.get_genes(node, tree, genotype_frame)
```

vicdoval: as there is no other call to the constructor I added the get_genes to the init function so it is 
`population = Population(node, tree, genotype_frame)`

nortikin:
manual testing done

vicdoval:
I added another gene type:

`List Input` node can be used a gene, the node will swap the elements order to generate the different variations.
It can be used to solve (for example)
Which is the shortest path that cycles through all the points?
image

nortikin:
Love this node!!!
Now we should prepare video around that.

vicdoval:
@nortikin do you have any supercool nodetree for a even cooler video? I am still pretty new exploring the GA possibilities

nortikin:
> @nortikin do you have any supercool nodetree for a even cooler video? I am still pretty new exploring the GA possibilities

just now had prepared.
What about missed mode with exact fitness?
I see only minimum and maximum.

Here i am:
https://gist.github.com/31bda4a5253fbd3cca63b8f5f034d5bd

nortikin:
OMG, my object missed matrices when recreated...

nortikin:
https://gist.github.com/f5698b868b1160a36df0e02d5a28fc31
here i applied locations

nortikin:
By adding new six nodes you can add new building and condition. We can group that stuff to make it work simpler.
Снимок экрана от 2020-06-06 01-12-56

vicdoval:
> What about missed mode with exact fitness?

If you activate `Stop on Goal` toggle you can define the fitness you are looking for and the system will stop is it is achieved or improved (in `Minimum` mode improved means under this value and in `Maximum ` improved means over this value

nortikin:
> > What about missed mode with exact fitness?
> 
> If you activate `Stop on Goal` toggle you can define the fitness you are looking for and the system will stop is it is achieved or improved (in `Minimum` mode improved means under this value and in `Maximum ` improved means over this value

ok, i expected additional socket to take values from frame and compare.
to be able adjust from 3d panel. But maybe it will work as node to 3d panel to have run/set/goal values in one row

nortikin:
checked your example, it doing length 5.66 and not less. But when i reorder manunally by radians, it gives 4.8. And as you can see, line is gong back and not connect shortest.
Снимок экрана от 2020-06-06 17-04-30

nortikin:
no, it works

nortikin:
Ok, made enhancements - listinput node now support till 128 items for float and integers.
But for sure, there should be some mode for genesis node to support huge genes
wrote SNL to parse kml files from maps:
```python
"""
in verts v .=[[]] n=2
enum = золотое.kml серебрянное.kml
inject
include <золотое>
out  verts  v
out  names  s
"""


    layout.prop(self, 'custom_enum',expand=True)
    cb_str = 'node.scriptlite_custom_callback'
    a = layout.operator(cb_str, text='.kml>names+verts')
    a.cb_name='my_operator'

#ND = self.node_dict.get(hash(self))               #
#if ND:                                            #
#    callbacks = ND['sockets']['callbacks']        #    this will be hidden in the node eventually 
#    if not 'my_operator' in callbacks:            #
#        callbacks['my_operator'] = my_operator    #

def my_operator(self, context):
    ''' converting kml to sverchok data '''

    import bpy
    import re
    import ast
    from sverchok.utils.listutils import listinput_set_rangeI

    t = bpy.data.texts['золотое']
    tt = t.lines[0].body
    nc = re.findall(r'<name>([\d\w\-\"\.\,\s]+)</name>\S+<coordinates>(\S+)</coordinates>',tt)
    names,verts = [],[]
    for n,c in nc:
        names.append(n)
        verts.append([i for i in ast.literal_eval(c)])
    self.id_data.nodes['Scripted Node Lite'].outputs['names'].sv_set([names])
    self.id_data.nodes['Scripted Node Lite'].outputs['verts'].sv_set([verts])
    print('Парсинг городов закончился: ',names,verts)

    # all the shit with list input node
    LN = self.id_data.nodes.new('SvListInputNode')
    listinput_set_rangeI(LN,verts)
    return {'FINISHED'}

self.make_operator('my_operator')

```
you can handle some real world track with till 128 points

nortikin:
Golden ring of an Architecture in Russia:
Снимок экрана от 2020-06-07 05-06-33
Not really run evolver, but prepared data.

portnov:
@vicdoval many not-so-complex optimization problems can be much easier (and faster, computationally) solved with help of `scipy.optimize` module. That can be done in a scripted node or as a full-fledged node in sverchok-extra.
https://docs.scipy.org/doc/scipy/reference/optimize.html
"Evolver" may be good for some really complex problems (though I'm not quite sure which exactly) :)

nortikin:
https://scenecitydoc.cgchan.com/home
i like the way it enlarged with icon (execute node of this addon)

PR #3312, opened 2020-06-03
WIP: Socket menus

portnov:

Lists manipulation, such as "wrap list", "make nesting level deeper", "concatenate lists" and so on, is a very important part of Sverchok. And in complex setups, there can be a dozen of nodes that do just that.

On the other hand, we have restrictions of nesting level of input data in many nodes — for example, a node can usually consume list of lists of vertices, but not list of lists of lists of vertices; and such restrictions are not only because it would be harder to code support for more nesting levels, but also because if you support more nesting level on the input you will most probably have more nesting levels on the output, so the nesting level of the data can grow to infinity while passing through a series of nodes. The user will be just unable to handle such deep data nesting in mind, so there will appear more and more nodes in each tree that just "simplify" data shape.

## Solution description

Let's consider the following approach (see also Grasshopper's implementation). Each socket has a drop-down menu with postprocessing options:

1. Stage 1. Optionally, one of:
* Flatten - convert any nesting level to level 1, e.g. `[[[1], [2], [3,4]]]` => `[1, 2, 3, 4]`.
* Simplify - similar to Flatten, but leave some small nesting. This is actual, for example, for Vertices sockets - you do not want to concatenate coordinates of different vertices into one big list. E.g., `[[[1,2,3], [4,5,6]], [[7,8,9], [10,11,12]]]` => `[[1,2,3], [4,5,6], [7,8,9], [10, 11, 12]]`.
2. Optional stage 2: Graft (the word is taken from GH, I do not find it too intuitive, suggestions are welcome). If enabled, this takes data from the previous stage, with whatever nesting level it had, and adds another nesting level by wrapping each "atomic item" of data into a separate list. For example, for numbers this should do `[[1, 2, 3]]` -> `[[[1], [2], [3]]]`. For vertices, it should do `[[1,2,3], [4,5,6]]` -> `[[[1,2,3]], [[4,5,6]]]`. So this is similar to List Split node with split length = 1.
3. Optional stage 3: Wrap. If enabled, take data from previous stage and add another pair of `[]` around. `[[1,2,3]]` -> `[[[1,2,3]]]`.
4. Optional. Socket class-specific postprocessing options. For example, for Curve and Surface sockets, there will be "Reparametrize" option, to change the domain of each curve/surface from whatever it was to `[0; 1]`.

Not all options have sense for all types of nodes, and in some cases some of options will just do nothing (for example, if the node already outputs list of level 1, then Flatten will do nothing). Set of available options is defined for each socket class, and sometimes socket class refers to node class to decide if the option should be available.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.


core/sockets.py
```diff
@@ -72,6 +92,34 @@ class SvSocketCommon:
 
     quicklink_func_name: StringProperty(default="", name="quicklink_func_name")
 
+    allow_flattern : BoolProperty(default = False)
+    allow_simplify : BoolProperty(default = False)
+
+    use_graft : BoolProperty(
```
Durman, line 98: Would you like to move all this new fields and methods in separate mixin class?

Durman:
I think there is problem of recognizing vertices, edges and faces in a list. How one can be confident that `[[[1,2,3]]]` is a vector and not a list of figures?

core/sockets.py
portnov: I think all Sverchok socket classes should have such possibilities, no?

portnov:
Just by socket type. If this is SvVerticesSocket, then `[1,2,3]` is a single vertex, otherwise this is a list of numbers.

core/sockets.py
Durman: I meant something like this

```Python
class NewClass: ...

class SvSocketCommon(NewClass): ...
```
It will let separate new features and in future it will be easier to handle them.

Durman:
What about faces and edges? They does not have their own socket type.

portnov:
Yeah, they present a problem. For StringsSocket, I tend to just support both options: "flatten", "simplify to level 2", "graft at level 1 (each number into a list)", "graft at level 2 (each edge to a list)". Though the last option probably does not have sense...

Durman:
I have thought a little how edges and faces could be specified and have come to thought that the most correct way should be to create their own socket type. Actually it is has sense because from my experience connecting edges to faces sockets and vice versa is quite common mistake. 

They even probably can have one type for both. Something like `TopologySocket`. It will be easy to specify them from each other by number of items in sequences. Or at least the sockets can have a same color.

portnov:
@zeffii  could you take a look at this (probably better to checkout and test to better understand how is it working). Any comments / ideas?

zeffii:
i've only superficially looked at it
image

it's less visually disturbing that I thought from the description.  if you need this, i'm totally down with that.  it would certain help declutter the nodes that are only added to unwrap / wrap...etc.

zeffii:
how it plays with iojson, i don't know.

Durman:
Should not icon says something which mode is used?

portnov:
I had this idea too, the only problem I see is that several modes can be enabled simultaneously.

zeffii:
the pynodes API is starting to feel limiting in these areas.

in retrospect i would also be interested in seeing an implementation of this that doesn't add that menu to the socket, but instead is done in a single node, which  can be minimized beside the socket and the drawlabel could better indicate what modes are enabled.  The extreme upside of doing it in a node is that the socket classes don't grow.

but it's easy for me to just armchair theorize about that :)

Durman:
I'm not sure that such functionality should be inherited by Sockets. I think it is possible to use aggregation instead. It will let to avoid creating huge object with a lot of properties and methods and split it into several more simple objects which will easier to understand and manage.

```diff
- class SvSocketProcessing(object):
+ class SvSocketProcessing(bpy.types.PropertyGroup):

- class SvSocketCommon(SvSocketProcessing):
+ class SvSocketCommon:
+    processing: bpy.psop.PointerProperty(type=SvSocketProcessing)
```

portnov:
What advantages would that have? I'm not sure that several objects are somehow easier to manage than one.

Durman:
Two objects is simple to manage when their has good interface. One object shouldn't know about what is going on inside another object. Another object should not know what one object expects to get. In this way you have to think only about one object simultaneously during its coding. Another words it is calling encapsulation.

```
import this

The Zen of Python, by Tim Peters
Beautiful is better than ugly.
Explicit is better than implicit.
Simple is better than complex.
```
The less an object has methods and the less one has attributes the simpler master the object. The more methods and attributes the more bugs. 

Another less reason is that `SvSocketProcessing` is not a socket. It should not be ancestor of `SocketCommon` class. Inheritance usually is used when one object is more specific type of another object. For example `VectorSocket` is descendants of `SokcetCommon` because it is more specific version of it. Or inheritance can be used for adding some qualities like iterable, animatable.

portnov:
> Another less reason is that SvSocketProcessing is not a socket. It should not be ancestor of SocketCommon class. 

This reasoning I understand. However, initially I put all code of SvSocketProcessing into SocketCommon. In fact, all properties and methods of SvSocketProcessing are properties and methods of a socket (there is no separate "socket processing" entity in GUI or wherever else). The only reason there are two classes instead of one is separation of code that implements different pieces of socket functionality; so this is not really inheritance, this is more like "mixin" or even "aspect".

Durman:
I would not be so sure that those methods and attributes certainly belong to sockets. Sockets have got input and output data but should they be able to manipulate them? Data can be an object by itself and know by itself how to manipulate with its.

> Not all options have sense for all types of nodes, and in some cases some of options will just do nothing (for example, if the node already outputs list of level 1, then Flatten will do nothing). Set of available options is defined for each socket class, and sometimes socket class refers to node class to decide if the option should be available.

This can be sign of that it is a mistake to use inheritance. Have a look at the text. It looks like it describes exactly the current situation.

<details> 
  <summary>text</summary>  



</details>

And just imagine how flexible it could be with using aggregation. If it necessary each socket Instance could have got its own processing object.

portnov:
Well, this way we will end up with a separate class for each option in the dropdown menu.
Who will be then responsible for logic like "these two options can not be enabled simultaneously"?

Durman:
> Well, this way we will end up with a separate class for each option in the dropdown menu.
Who will be then responsible for logic like "these two options can not be enabled simultaneously"?

Probably I don't understand the question but it is possible to put a handler on option from where you can enabled any option you like in Blender. And it is possible to create object for each option in the menu but they should be something really complex to worth dooing such a thing.

Let's have a look at the code.

```Python
class SvSocketProcessing(object):
    # These properties are to be set explicitly by node classes
    # for input sockets, if the node knows it can handle simplified data.
    # For outputs, these properties are not used.
    allow_flatten : BoolProperty(default = False)
    allow_simplify : BoolProperty(default = False)
    allow_graft : BoolProperty(default = False)
    allow_wrap : BoolProperty(default = False)
```

And just imagine that for one type of socket we would like to have allow something else. Does it means that base class should be changed like that? And all other sockets should share this option when it's known that they will never support such allow?

```Python
class SvSocketProcessing(object):
    # These properties are to be set explicitly by node classes
    # for input sockets, if the node knows it can handle simplified data.
    # For outputs, these properties are not used.
    allow_flatten : BoolProperty(default = False)
    allow_simplify : BoolProperty(default = False)
    allow_graft : BoolProperty(default = False)
    allow_wrap : BoolProperty(default = False)

    allow_something_else: BoolProperty(default = False)
```

portnov:
Currently, base class contains properties for features that are possible for most of sockets. Some of socket classes have their specific allow_somethings.

So, your proposal is to have

```python
class SvFlattener(PropertyGroup):
  allow_flatten : BoolProperty(default = False)
  def do_flatten(...): ...

class SvSimplifier(PropertyGroup):
  allow_simplify: BoolProperty(default = False)
  def do_simplify(...): ...

...

class SvVerticesSocket():
  flattener : PointerProperty(type=SvFlattener)
  simplifier : PointerProperty(type=SvSimplifier)

  def draw_menu_items(layout):
    layout.prop(self.flattener, 'allow_flatten')
    layout.prop(self.simplifier, 'allow_simplfiier')
...
```

correct?

Now, "flatten" and "simplify" do not have any sense to be enabled at the same time. Which class must contain an update handlers for allow_simplify and allow_flatten, which would disable one of them when another is enabled ?

Durman:
No, I mean something like this:

```python
class SvSocketCommon:
    def sv_set(self, data):
        if hasattr(self, postprocessor):
            data = self.postprocessor.process_data(data)
        SvScoketSet(self, data)

    def draw(self, layout):
        if hasattr(self, postprocessor):
            self.postprocessor.draw(layout)


class ListPostprocessor(PropertyGroup):

    def update_mode(self, context, changed_mode): ...

    mode_1: BoolProperty()
    mode_2: BoolProperty()
    mode_3: BoolProperty()

    def draw(self, layout) ...

    def process_data(self, data): ...


class SvVertiecesSokcet(SvSocketCommon):
    postprocessor: PointerProperty(type=ListPostprocessor)


class SvStringsSocket(SvSocketCommon):
    postprocessor: PointerProperty(type=ListPostprocessor)


class CurvePostprocessor(PropertyGroup):

    def update_mode(self, context, changed_mode): ...

    mode_1: BoolProperty()
    mode_2: BoolProperty()
    mode_3: BoolProperty()

    def draw(self, layout) ...

    def process_data(self, data): ...


class SvCurveSocket(SvSocketCommon):
    postprocessor: PointerProperty(type=CurvePostprocessor)

```
Also probably it would not be bad idea to create abstract base class for `postprocessors`.
https://docs.python.org/3/library/abc.html#abc.abstractmethod

nortikin:
@Durman 
would it slow sverchok?

Durman:
Using aggregation? Why it should slow down code?

nortikin:
@Durman just asking

Durman:
> Should not icon says something which mode is used?
> 

Probably modes could be indicated in this way:

portnov:
Well, I'm fine with this. People?

vicdoval:
I dont love it but i guess is the best option

portnov:
@Durman  I will have to have two tightly coupled, very similar but different, hierarchies of classes. 
```
SvSocketProcessor:
  | properties:
  |   allow_flatten, allow_simplify, allow_graft, allow_wrap # To allow specific nodes to disable some options in the menu
  |   use_flatten, use_simplify, use_graft, use_wrap # Options to be enabled by user
  | methods:
  |   process_data, which consists of:
  |     flatten, simplify, graft, wrap
  |
  |> SvStringSocketProcessor:
  |  properties:
  |    allow_graft2, use_graft2
  |  methods:
  |    process_data: call super and then optionally graft2()
  |
  |> SvCurveSocketProcessor:
  |  properties:
  |    use_reparametrize
  |  methods:
  |    process_data
  |    override flatten, simplify, graft, wrap, because they should work slightly different
  |
  |> SvSurfaceSocketProcessor:
  |  properties:
  |    use_reparametrize
  |  methods:
  |    process_data
  |    override flatten, simplify, graft, wrap, because they should work slightly different
  |
  |> SvSvgSocketProcessor:
  |  methods:
  |    override flatten, simplify, graft, wrap, because they should work slightly different
```

```
SvSocketCommon
  | processor : SvSocketProcessor
  |
  |> SvStringsSocket
  |  processor : SvStringSocketProcessor
  |
  |> SvCurveSocket
  |  processor : SvCurveSocketProcessor
  |
  |> SvSurfaceSocket
  |  processor : SvSurfaceSocketProcessor
  |
  |> SvSvgSocket
  |  processor : SvSvgSocketProcessor
  |
  |> several other socket types do not override `processor` property
```
I'm not sure if this will be any simpler. Two hierarchies of classes will be in fact tightly coupled, because it does not have any sense to hide knowledge about SvCurveSocketProcessor from SvCurveSocket...

Durman:
What is the advantage of inheriting from SvSocketProcessor? I think if class hase something what another class does not need last should not inherit from it.

portnov:
But all socket classes need "processing" features.

Durman:
Why this features can't be set individually for each class?

portnov:
Why should the code be duplicated?

portnov:
NB: `allow_*` properties are not to disable the feature in *socket class*, they are to disable the feature in *socket instance* in particular node class. Though I do not have examples atm when some node does not want to have some of these options, I think there should be such possibility.

portnov:
Let's try this way...

| Option  | Which sockets need it |
|---------|-----------------------|
| Flatten | All, probably except VerticesSocket |
| Simplify | All |
| Graft | All |
| Graft2 | StringsSocket |
| Reparametrize | SvCurveSocket and SvSurfaceSocket, with a bit different implementation - reparametrizing surface is not the same as reparametrizing curve |

Durman:
I don't like using `allow_` attributes they look redundant to me. If some mode is not allowed then a class should not have `use_some_mode` attribute.

> Though I do not have examples atm when some node does not want to have some of these options, I think there should be such possibility.

Probably it would be better don't implement this attributes now if there is no particular usage for them and only implement them if they really will be needed.

In general I look positive on such API. We have separate classes for preprocessing data (`ImplicitConversionPolicy`) it looks logical step to add seperate classes for postprocessing in seperate module.

portnov:
> . If some mode is not allowed then a class should not have use_some_mode attribute.

Again, the mode can be disabled at *socket instance in the specific* **node** class, not in *socket class*.

> Probably it would be better don't implement this attributes now if there is no particular usage for them and only implement them if they really will be needed.

Well, may be.

>  We have separate classes for preprocessing data (ImplicitConversionPolicy) it looks logical step to add seperate classes for postprocessing in seperate module.

The difference is that hierarchy of ImplicitCoversionPolicy classes is not related to hierarchy of socket classes. While with SocketProcessors, there will be two almost identical hierarchies, tightly coupled. To me, this looks like one class torn apart at the middle.

portnov:
I'm thinking about adding a checkbox to node tree to hide / show these menu buttons. In some cases you do not need them, so they just waste space. On the other hand, if you hide buttons, enabled options will be still effective.
?

portnov:
I've added the checkbox into N panel, and it is disabled by default. So users who do not want these features will not be surprised.

core/sockets.py
```diff
@@ -17,31 +17,196 @@
...
+from sverchok.utils.logging import warning
 
 def process_from_socket(self, context):
     """Update function of exposed properties in Sockets"""
     self.node.process_node(context)
 
+class SV_MT_SocketOptionsMenu(bpy.types.Menu):
+    bl_label = "Socket Options"
+
+    test : BoolProperty(name="Test")
```
Durman, line 52: What is it for?

core/sockets.py
```diff
@@ -531,6 +744,67 @@ def draw_property(self, layout, prop_origin=None, prop_name=None):
...
+
+    def do_graft(self, data):
+        return graft_data(data, item_level=0, data_types = SIMPLE_DATA_TYPES + (SvCurve, SvSurface))
+
+    def do_simplify(self, data):
+        return flatten_data(data, 2)
+
+    def do_graft(self, data):
+        return graft_data(data, item_level=0, data_types = SIMPLE_DATA_TYPES + (SvCurve, SvSurface))
+
```
Durman, line 770: Looks like your IDE does not highlight redundancies

portnov:
Fixed.

PR #3320, opened 2020-06-05
viewer 2d ui love

vicdoval:
trying to update the viewer 2d

### DIFF
--- a/nodes/viz/viewer_2d.py
+++ b/nodes/viz/viewer_2d.py
@@ -91,7 +91,7 @@ def fill_points_colors(vectors_color, data, color_per_point, random_colors):
     else:
         for nums, col in zip(data, cycle(vectors_color[0])):
             if random_colors:
-                r_color = [random(),random(),random(),1]
+                r_color = [random(), random(), random(),1]
                 for n in nums:
                     points_color.append(r_color)
             else:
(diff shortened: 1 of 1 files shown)


vicdoval:
@zeffii  I have manage to update the background, (`2D_UNIFORM_COLOR`) but when i try to follow your intructions with the 2D_smooth_color all the vertices appear in one point... have you tested `get_2d_smooth_color_shader` i could not find any example among the changed files

zeffii:
waveform viewer
```
config.background_shader = get_2d_smooth_color_shader()

```

zeffii:
and easings' line_vertex_shader/line_fragment_shader 
```
    shader2 = gpu.types.GPUShader(line_vertex_shader, line_fragment_shader)
    batch2 = batch_for_shader(shader2, 'LINES', {"pos": geom.vertices, "color": geom.vertex_colors}, indices=geom.indices)
```

zeffii:
a recap
- you pass a function instead of x, y:
    `"loc":  somefunction `
- somefunction takes `node` as a parameter, it returns the correct `x, y`
    - through some combination of `node.absolute_location` + `node.width` + `some offset` 
    - and the resulting x, y is multiplied each by `self.location_theta`
- all geometry data is calculated as if `loc = (0, 0)`,  indifferent about where it is being drawn.
- you add x_offset and y_offset somehow into your shader binding.

zeffii:
i think you're missing a few `<shader>.binds()`

vicdoval:
Now it works.
The `batch_for_shader` had to be called from the `view_2d_geom`
Also it was working without the shader.bind 😆  Now I added the missing calls.
@zeffii  take a look if you want and if you like it we can perform the merge

zeffii:
> Also it was working without the shader.bind

sure, remove as much as you can :)  i'm not entirely sure what `.bind` does.

zeffii:
commit when i'm done with Texture nodes

PR #3351, opened 2020-06-16
Evolver upgrade: Genes Holder and Re-use population

vicdoval:
Some improvements for the Genetics Algorithm work-flow:

Genes Holder: A node that stores a list of data (int, float or vectors) that can be filled with another node data and can be manipulated by the Evolver Node. Offers Order manipulation (switching the order of the elements) and Range manipulation (changing the data within a  defined range).
genes_holder_genetics_algorithm_sverchok_blender_example_01

Re-use population:  Option for the Evolver node that allows using last iteration of previous analysis as first generation of a new analysis.

Evolver node and Genes Holder node now store the data in a String property so it is not lost when the file is closed (btw it is a lot faster than storing it in a text-block)

Also some refactoring was done in the Evolver node, now genes are classes too and everything is much cleaner.

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


vicdoval:
Also bigger RUN button
image

Durman:
It's strange that `Gens holder` node did not get into the example in the documentation.

vicdoval:
> It's strange that `Gens holder` node did not get into the example in the documentation.

You are right, I have just added it

PR #3359, opened 2020-06-21
Searh Extra nodes

### LINKED ISSUE #3358: Extra Nodes in Search Menu
Right now the sverchok-extra nodes  don't appear on the search menu.

I have been digging into it and if I add 

```
fx.append((str(idx),'Location','',idx))
loop_reverse['Location'] = 'SvExLadyBugLocationNode'
idx +=1
```
into the gather_items function
https://github.com/nortikin/sverchok/blob/74f2e7b46ea4c8a10faf5a2c555d7699f129a9e8/utils/sv_extra_search.py#L100

Then the Location node appears in the search menu and loads succesfully

So it is only needed a function that returns bl_idname and class name from the valid extra nodes to make it work.
But I could not find such a function. `the get_extra_categories` function seems return the categories with a items field but  could not find the way to access them (requires a context?)  
@portnov  am I missing anything?

I think is quite important the nodes to appear in the search menu. Could you please throw some light into this task?

vicdoval:
Gather extra nodes to appear in the search menu
Fixes #3358
- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/utils/sv_extra_search.py
+++ b/utils/sv_extra_search.py
@@ -28,6 +28,7 @@ from sverchok.utils.docstring import SvDocstring
 from sverchok.ui.sv_icons import custom_icon
 from sverchok.utils.sv_default_macros import macros, DefaultMacros
 from nodeitems_utils import _node_categories
+from sverchok.utils.extra_categories import get_extra_categories
 
 # pylint: disable=c0326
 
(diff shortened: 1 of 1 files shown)


Moult:
Really great PR!

Just to give an idea of the numbers: the Ladybug nodes currently total about 60 nodes. But Ladybug is only about a quarter of the suite, so I expect the full Ladybug Tools (+ Honeybee, Dragonfly...) to be approximately 200 nodes.

I am also planning to build yet another extension to allow crafting IFC BIM data in Sverchok. This would be another approximately 30 nodes (ballpark figure).

PR #3377, opened 2020-06-28
Solid Nodes and Extra-Nodes

vicdoval:
image
This PR adds two features:

**Extra nodes**:  The ability to add nodes with external dependencies.
A new tab has been added to the sverchok preferences to handle the dependencies
image

In case of missing dependencies the nodes wont be loaded and they wont appear on the menu.
If a file is loaded with nodes that require dependencies not found a Dummy node will be created on the Fly
image
The dependent nodes have to be added to the dummy_nodes_dict in utils/dummy_nodes
I followed the strategy of the old_nodes loading but generating the dummy_nodes when needed
Also if a json is loaded with  nodes that require dependencies not found they will be replaced by a Dummy node

**Solid Nodes**:  A set of nodes dependent on FreeCad to perform operations with solids (BRep)
Inspired in https://blenderartists.org/t/freecad-sverchok-subd-to-nurbs/1218702
Using FreeCAD as Python module a set of nodes to deal with solids

Installation
A python 3.7 FreeCad is needed. a windows version can be found here https://github.com/sgrogan/FreeCAD/releases/tag/PY3.7-win
Then the path to the FreeCad bin folder has to be placed in the Extra-Nodes ->FreeCad-> TextField and click on Set Path

image
Solids are heavier than meshes but perform better boolean operations (even with co-planar faces) 
The module also offers variable radius Fillet
image

Chamfer with different angles:
image
image
 
And more to came (import/export to STL and BRep, center of mass, solid to NURBS...)

- [ ] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


zeffii:
awesome!  does freecad have a `manifold verts+edges` to `Solid` function?

zeffii:
"Solid From Shell "  :)  cool the code is out there :)  https://www.xsim.info/articles/FreeCAD/en-US/HowTo/Create-a-solid-from-surface-meshes.html

zeffii:
oh lol. i see you have Mesh to Solid already.  i'm blind.

portnov:
Well. We have to decide something... We discussed this in https://github.com/portnov/sverchok-extra/issues/15
If there will be "Extra nodes" subsystem in Sverchok, do we need Sverchok-Extra at all?

vicdoval:
@portnov  I took part some of the code from sverchok-extra to make this PR work, Sverchok-extra is a extra step, the easiest,  but another step. I think moving sv-extra to extra-nodes will encourage the users to install the dependencies and to use the nodes ...

vicdoval:
Solid Edges node outputs Curve objects:
image

portnov:
1. Do I correctly understand that the path that I have to set up in preferences is the path to `FreeCAD.so` / `FreeCAD.dll` ? This is not too clear. I suggest to write it in the tooltip or in special label nearby. Also I suggest to add examples of typical paths on different platforms there.
2. Why separate `Set` button is required?
3. When I put `/usr/lib/freecad-python3/lib/` there, I have 

```
FreeCAD 0.18.4, Libs: 0.18.4R
No modules found in /usr/lib/freecad-python3/Mod
During initialization the error No module named 'freecad' occurred
```

However the nodes seem to work. What is going on? :)

4. If I try to do "Mesh to Solid", and then do something with that Solid (Fillet, for example, or Champfer), I have

```
  File "/home/portnov/.config/blender/2.82/scripts/addons_contrib/sverchok/core/update_system.py", line 383, in do_update_general
    node.process()
  File "/home/portnov/.config/blender/2.82/scripts/addons_contrib/sverchok/nodes/solid/fillet_solid.py", line 62, in process
    solid_o = solid.makeFillet(r_s, r_e, selected_edges)
Part.OCCError: BRep_API: command not done
```

The same node works if I pass primitive like "Box (Solid)" into it.

vicdoval:
> Do I correctly understand that the path that I have to set up in preferences is the path to FreeCAD.so / FreeCAD.dll ? This is not too clear. I suggest to write it in the tooltip or in special label nearby. Also I suggest to add examples of typical paths on different platforms there.

The path has to be to the folder where FreeCadApp.dll (and the FreeCadBase.exe..) is placed... In my case is something like `E:\programas\conda-0.18.3\bin`  (it is a portable Windows version of FreeCad downloaded from the provided link) 

> Why separate Set button is required?

The Set button saves a file in python\lib\site-packages called freecad_path.pth with the supplied path so it stays permanently on the system... maybe it could be set every time you open blender.. I thought this solution was cleaner but I'm open to suggestions

> However the nodes seem to work. What is going on? :)

Probably you already had a  path in your system to the right version of FreeCAD

About the 4th point: The FreeCAD  api report is a bit obscure..., the Fillet and Chamfer nodes will fail if the Radius/Distance is to high to perform the operation without errors, reducing the value should make it work. maybe I should add a try/except to raise a more clear explanation

zeffii:
looks like the vdmk3 might need an alternative, for solids to avoid that chevron shading artifact in the fillet.

vicdoval:
Solid Faces node outputs surface objects and curves with the outer curves of each face:
image

vicdoval:
@zeffi i was thinking about this... in grashopper some nodes have a draw method that works like a inner viewer draw... now that we have Surfaces, Curves and Solids maybe could try something similar... 
As you see in this image there could be a cleaner way to represent the Solid objects image
The faces are tesselated but only the "real" edges are represented....

zeffii:
"Solid Viewer" as a separate node maybe, an opportunity to go beyond my silly experiments.

zeffii:
coooool.

image

zeffii:
should the Solids get their own separate menu?  even if it's only a really minor menu system.

```python
- input
- operators
- convert
- output
```

vicdoval:
@zeffi you mean taking it out of the general menu or keep it there and also making another smaller menu?

zeffii:
yes, to do this justice i think a separate menu has merit. Especially for the new users, who will be able to see what is available from the FreeCAD module in a nice menu/submenu form

whats happening here?
image

vicdoval:
> whats happening here?

It was me messing with curves... now is fixed 
image

I understand you mean keeping in main + new menu?

The  question that pops up is: A freeCad module menu or a Solids Menu? Once FreeCad is inside Sv we could import more than Solids... (BIM, Fem analysis, Surface, Mesh tools [there is a mesh boolean operator])... how do you imagine it?

zeffii:
> how do you imagine it?

well, in the overarching case that you paint, maybe a freeCAD menu would indeed make sense. 

In that case I would vote to not mix those nodes in the Sverchok menu at all.  but let's see.

portnov:
@vicdoval 
> The Set button saves a file in python\lib\site-packages called freecad_path.pth with the supplied path so it stays permanently on the system... maybe it could be set every time you open blender.. I thought this solution was cleaner but I'm open to suggestions

Does this mean I have to press this button each time I run blender?

portnov:
Screenshot_20200630_210958
```
2020-06-30 21:09:10,122 [ERROR] sverchok.core.update_system: Node Chamfer Solid had exception: BRep_API: command not done
  File "/home/portnov/.config/blender/2.82/scripts/addons_contrib/sverchok/core/update_system.py", line 383, in do_update_general
    node.process()
  File "/home/portnov/.config/blender/2.82/scripts/addons_contrib/sverchok/nodes/solid/chamfer_solid.py", line 63, in process
    solid_o = solid.makeChamfer(d_a, d_b, selected_edges)
Part.OCCError: BRep_API: command not done
```

portnov:
5. "Solid faces" node. Is it possible to have additional output similar to "Outer Wire", but with curves in UV coordinates of each surface? I just don't know if Freecad provides such.

portnov:
6. Dummy nodes. Do we really need separate classes for them? Maybe that "dummy" functionality could be incorporated in the nodes themselves?
```
from sverchok.dependencies import FreeCAD

class SvSomeNode:
    def is_dummy(self):
        return (FreeCAD is None)
    
    def draw(...):
        if self.is_dummy(): layout.label(text="This is a dummy node")
        else: draw actual UI

    def process(self):
        if self.is_dummy(): raise Exception("This is a dummy node")

```

?

portnov:
https://github.com/microelly2/NodeEditor huh.

vicdoval:
Ok by the moment I will add add a extra Solid menu... (that also would be nice for the Surface and Curve categories btw)

vicdoval:
@portnov check the last version 
image

> Does this mean I have to press this button each time I run blender?

No just once, but its true it could be set on every runtime, i will try :)

> "Solid faces" node. Is it possible to have additional output similar to "Outer Wire", but with curves in UV coordinates of each surface? I just don't know if Freecad provides such.

I think it does not provide a curve object like that but maybe there is some workaround.. i will dig into it.

> Dummy nodes. Do we really need separate classes for them? Maybe that "dummy" functionality could be incorporated in the nodes themselves?

It was the first approach to avoid them to load in menu and only registering the ones that are demanded, but probably it would be easier for the coders to keep it all in the node...

> https://github.com/microelly2/NodeEditor huh.

Looks promising but what can be better than having FreeCad inside the mighty Sverchok  :D

vicdoval:
> "Solid faces" node. Is it possible to have additional output similar to "Outer Wire", but with curves in UV coordinates of each surface? I just don't know if Freecad provides such.

@portnov  there is a function in the FreeCad Solid Faces that given a point in space returns the nearest uv coordinates of the surface, that should work no?  give a U of the curve, get the point in space and use that point to get the UV of the surface.
It could be:
- A method of the OuterWire curves 
- A method of the Solid Faces Surface (that could work with the outerwire or with any other curve/point) 
- A different curve called UV wires that generates a 2D curve (and when exposed to Evaluate Curve adds a 0 to create the regular ·3D vector)
- The three of them

What use do you imagine for this curves?

portnov:
@vicdoval As far as I understand, freecad is in fact a fancy GUI for OpenCASCADE (or am I mistaken?). And, there are python bindings for the OpenCascade (https://github.com/trelau/pyOCCT). So, may be better to integrate directly with OpenCascade?

vicdoval:
@portnov  I was thinking about that but i could not find an easy way to install it, seems you have to first install conda and then install pyOCCt  by typing on the conda prompt. 
Maybe to much for a regular user and definitely harder than dowloading the freeCad binarys....

vicdoval:
@portnov 

> Dummy nodes. Do we really need separate classes for them? Maybe that "dummy" functionality could be incorporated in the nodes themselves?

This approach 
-keeps everything in the node 
-avoids registering unusable nodes creating the dummys if necessary
-keeps the node code clean from dummy state

```
from sverchok.dependencies import FreeCAD
from sverchok.utils.dummy_nodes import add_dummy

if FreeCAD is None:
    add_dummy('SvSolidEdgesNode', 'Solid Edges', 'FreeCAD')
else:
   regular clean node
```

portnov:
> here is a function in the FreeCad Solid Faces that given a point in space returns the nearest uv coordinates of the surface, 

Interesting. The thing is, there is a node in Sverchok-Extra, which finds the closest point on a surface, by solving minimization problem numerically.  So, it works for any Surface object. So, if I understand correctly that now our intention is to integrate Sverchok-Extra functionality into Sverchok (@zeffii just in case, I ask you again if you have something against it), then let's do as follows:

1. Move most of Sverchok-Extra into Sverchok, except for most questionable or most hard-to-install things
2.  Then add a method to SvSurface, that finds the nearest point on the surface to the given one; default implementation will solve minimization problem numerically, as currently existing node does;
3. Off course, port the "nearest point" node to use of that method;
4. Then write a SvCurve subclass, that takes a Curve and construct another curve by something like `surface.find_nearest_point(point_on_curve)`;
5. Then it will be possible to use that class in a new node, probably it would be useful by itself;
6. And then we can use the same class in "solid faces" node.

The only problem I see, can we understand which edges should we project onto which surface?

zeffii:
>  (@zeffii just in case, I ask you again if you have something against it), 

i can't think of a robust reason to keep these things separated indefinitely, especially if dependency-failures are handled well.

zeffii:
it will be cool to have a Mesh viewer that takes a Solid socket, and does all this "Patch Boundary" (Outer wires) stuff internally. It could inherit from sv_obj_helper, 
- and there could be two implementations
1. using MESH type.  Setting "Creases" on those edges that are determined to be Patch Boundaries, and add a Edge Split modifier.
2. using the route @portnov went with the NURBS/Surface Viewer node.

vicdoval:
Solid Viewer working
image

portnov:
"Bake" button would be appropriate :)

vicdoval:
@portnov  Looks like we have a roadmap :D 

> The only problem I see, can we understand which edges should we project onto which surface?

Right now the Solid faces OuterWire outputs for each faces a set of curves, which means curves are repeated but is easy to determine to which surface they are attached

> "Bake" button would be appropriate :) 

I would be appropriate  :D, the only thing is that the I have to add the options off the "solid to Mesh node" to determine which algorithm  should be used for the baking... right now uses the basic mode, seems the fastest to visualize but gives worst topology... and brings the topic @zeffii  mentions...creases? nurbs?

At this point I think I should start with the docs so this PR can be merged and you can start with the sv-extra to extra-nodes movement

zeffii:
also time to look at how to render edges with some antialiasing : )

nortikin:
> also time to look at how to render edges with some antialiasing : )

maybe we need no edges?

zeffii:
technically these lines aren't edges, but "patch boundaries" / "patch perimeters" , it's useful to see them, and the information needed to generate those lines would be necessary to have anyway for creasing/edgesplit.

vicdoval:
@portnov 

> Wouldn't it be enough to just save the path in sverchok settings and initialize sys.path accordingly at registration?

The tricky part is that settings.py is loaded after dependencies.py so I can't read a property from the settings at the point dependencies are loaded. Any suggestion solve this better that how it is done now?

portnov:
Well, one possible approach is to initialize everything lazily... idea is to store a flag "is_path_initialized" in `dependencies.py` module, and initialize path first time when "from dependencies import FreeCAD" is performed, and set the flag. But this may be tricky. So probably the way you have is better. You decide :)

luzpaz:
CC @microelly2

vicdoval:
Solids Menu (Shift +S)
image
image
image

image

portnov:
> "Solid faces" node. Is it possible to have additional output similar to "Outer Wire", but with curves in UV coordinates of each surface? I just don't know if Freecad provides such.

There is such a function:

```
 |  curveOnSurface(...)
 |      curveonSurface(Edge) -> None or tuple
 |      Returns the curve  associated to the  edge in  the
 |      parametric  space of  the  face.  Returns None if this
 |      curve  does not exist. If this curve exists then a tuple
 |      of curve and and parameter range is returned.
```

portnov:
@vicdoval if preparing documentation is going to take more time, maybe it will be simpler to split this PR into two? 1) infrastructure and 2) new nodes with documentation.

vicdoval:
> There is such a function:

Would you like me to include that output in the solid faces node

> @vicdoval if preparing documentation is going to take more time, maybe it will be simpler to split this PR into two? 1) infrastructure and 2) new nodes with documentation.

I understand your enthusiasm but splitting the PR will also increase the developing time. I'll try  to only daft the documentation by the moment so i can merge this PR asap

portnov:
> Would you like me to include that output in the solid faces node
not in this PR, I think :) Maybe I later will do that.

portnov:
:+1:

zeffii:
🎉 

now to make some coool stuff with it.

nortikin:
https://vk.com/wall-35076122_11309
short installation instructions on linux and windows:
```
Linux:
sudo apt install libfreecad-python3-0.18
/usr/lib/freecad-python3/lib/
Windows:
https://github.com/sgrogan/FreeCAD/releases/tag/PY3.7-win
E:\programs\conda-0.18.3\bin
```

rastart:
Hello everyone!
I'm the author of the blender artist post: https://blenderartists.org/t/freecad-sverchok-subd-to-nurbs/1218702
I'm starting to study this implementation and I rewrited my original sv project node.
Here node features:
- it takes a freecad project file and output the solids from it
- it writes a mesh from blender overwriting the given name

I see that is possible to slice the solid list and make operation on it, but It would be nice to add a node to select parts by name, I have to try…

freecad_project_to_sverchok

If could be of some use here the node code:
[SvFCproject.zip](https://github.com/nortikin/sverchok/files/5214561/SvFCproject.zip)

portnov:
I would suggest
* Split reading and writing into different nodes;
* For reading node, add "Read List" button, which would read list of objects from FreeCAD's file and populate a dropdown list (or a list with checkboxes, so that the user can select several objects)

rastart:
> I would suggest
> 
> * Split reading and writing into different nodes;
> * For reading node, add "Read List" button, which would read list of objects from FreeCAD's file and populate a dropdown list (or a list with checkboxes, so that the user can select several objects)
> * Probably, support "SvFilePath" socket for file path.

FCStdNodes

Hello! working in progress... Slowly learning Sverchok framework... 
FCStd node splitted:

"read node"
It can import from multiple files but I didn't find a way to populate a dropdown list of the parts to select, so I added a field to filter by name, just writing parts name separated by comma and no spaces ( split(',')... )

"write node" ( mesh/solid modes)
I still have to code multiple file writing in write node.
I tried to write a solid input from the "mesh to solid" node, it works! 
The parts writed from blender are overwrited by name in the FCStd file.
The solid write mode support multiple solids and it will add an index on the base name, on changes it will delete all parts with base name before write them again. This system could be quite dangerous...  It's just a way to experiment with blender/CAD workflow.

[FCStd_nodes.zip](https://github.com/nortikin/sverchok/files/5251545/FCStd_nodes.zip)

nortikin:
> > I would suggest
> > 
> > * Split reading and writing into different nodes;
> > * For reading node, add "Read List" button, which would read list of objects from FreeCAD's file and populate a dropdown list (or a list with checkboxes, so that the user can select several objects)
> > * Probably, support "SvFilePath" socket for file path.
> 
> FCStdNodes
> 
> Hello! working in progress... Slowly learning Sverchok framework...
> FCStd node splitted:
> 
> "read node"
> It can import from multiple files but I didn't find a way to populate a dropdown list of the parts to select, so I added a field to filter by name, just writing parts name separated by comma and no spaces ( split(',')... )
> 
> "write node" ( mesh/solid modes)
> I still have to code multiple file writing in write node.
> I tried to write a solid input from the "mesh to solid" node, it works!
> The parts writed from blender are overwrited by name in the FCStd file.
> The solid write mode support multiple solids and it will add an index on the base name, on changes it will delete all parts with base name before write them again. This system could be quite dangerous... It's just a way to experiment with blender/CAD workflow.
> 
> [FCStd_nodes.zip](https://github.com/nortikin/sverchok/files/5251545/FCStd_nodes.zip)

image
would it be here or where?

rastart:
Yes, I think Exchange would be perfect

rastart:
hello! I fixed some issues related to file array and filter by name( now use .label prop in freecad instead of .name... I noticed that if you rename something in freecad it changes only the label
[FCStd_nodes_25_10_2020.zip](https://github.com/nortikin/sverchok/files/5434775/FCStd_nodes_25_10_2020.zip)
)

vicdoval:
@rastart wouldn't you like to make a pull request? (https://www.youtube.com/playlist?list=PLRqwX-V7Uu6ZF9C0YMKuns9sLDzK6zoiV)

nortikin:
@rastart @vicdoval 
agree, git / github is must-have skill.

rastart:
Ok, I should have pull the request!

rastart:
Hello, I'm evaluating a convenient method to pass filter labels to read node.
I added a string input and it works fine ( I scripted a node to pass a simple string )
There are already some nodes in Sverchok to pass simple text line? maybe also in list format? (like 'list input' in number cat)
I could also try to make a node with an enum function to generate a dropdown list to select like Portnov pointed:

> * For reading node, add "Read List" button, which would read list of objects from FreeCAD's file and populate a dropdown list (or a list with checkboxes, so that the user can select several objects)

but in huge assembly would be a mess. Have you some feedbaks about it?

filter_input

portnov:
```python
class SvShowFcstdNamesOp(bpy.types.Operator):
    bl_idname = "node.sv_show_fcstd_names"
    bl_label = "Show contents"
    bl_options = {'INTERNAL', 'REGISTER'}
    bl_property = "option"

    def get_items(self, context): ...

    option : EnumProperty(items = get_items)
    tree_name : StringProperty()
    node_name : StringProperty()


    def execute(self, context):
        tree = bpy.data.node_groups[self.tree_name]
        node = tree.nodes[self.node_name]
        node.name_filter = self.option
        return {'FINISHED'}

    def invoke(self, context, event):
        context.space_data.cursor_location_from_region(event.mouse_region_x, event.mouse_region_y)
        wm = context.window_manager
        wm.invoke_search_popup(self)
        return {'FINISHED'}
```
...
```python
def draw_buttons():
    wrapper_tracked_ui_draw_op(layout, SvShowFcstdNamesOp.bl_idname, icon=...)
```

rastart:
Meanwhile I made up this... works nice!
part_list
Thanx for the dropdown code, I'll try to implement it in the next days

rastart:
> This will show drop-down menu similar to one that appears by Alt-Space.
Working!

DuncanLithgow:
I've referenced this discussion at https://wiki.osarch.org/index.php?title=Sverchok if anyone wants to expand the text that would be great. Or at least check that what I've written makes sense.

rastart:
added read sketch node!
The sketch geometry is converted in polylines by freecad 'discretize' function and passed to verts/edge data.
Added a resample function with a parameter 'max_points' that set the max number of point in relation to the max length in the sketch geometry.

Until now I just tested the nodes with random operations and I didnt experiment critical issues, mainly **write node** need some attention to avoid recursive overloading if the same freecad file is involved in read/write operation.
From here I'm going to make some little design project to experiment in freecad/sverchok using what we have done until now, 
I'm thinking about some simple design involving mechanical, precedural and some direct blender modeling features all togheter.

fcnodes_sketch

portnov:
Can you output curves as `sverchok.utils.curve.core.SvCurve` instances? you probably will have to write special class for that, or maybe one of classes from `sverchok.utils.curve.freecad` will suffice.

rastart:
> Can you output curves as `sverchok.utils.curve.core.SvCurve` instances? you probably will have to write special class for that, or maybe one of classes from `sverchok.utils.curve.freecad` will suffice.

Fc_sketch_to_curve


OK! sverchok.utils.curve.core.SvCurve instances worked for translate freecad line, circle and arc.
Still to translate conics:  maybe this won't be so easy...

rastart:
I just updated the freecad exchange nodes.
Mainly:

READ NODE -> add some sort logic (ex. to read just the body final shape without import all features shape)
WRITE NODE -> no changes
SKETCH NODE -> it apply to the sketch geometry the transform matrix of the sketch placement and the parent body placement
SPREADSHEET NODE -> no changes

Screenshot (200)

I don't think I'm going to add more features so If you have some reviews from now I'll work just on refactoring!
Let me know!

Meanwhile I'll keep making design experiments with these nodes...

rastart:
> Can you output curves as `sverchok.utils.curve.core.SvCurve` instances? you probably will have to write special class for that, or maybe one of classes from `sverchok.utils.curve.freecad` will suffice.
for other kind of curves other than segment and circles I manged like that:
freecad_sketch_curve.toNurbs() -> SvNurbsCurve.build( SvNurbsMaths.FREECAD, freecad_sketch_curve.Degree ecc...

It seems to work fine:
fc_sketch_node

58911031:
freecad link disabled  could you give me the link again please

nortikin:
@58911031 https://disk.yandex.ru/d/NhCrTKi3nIgTXA

PR #3379, opened 2020-06-28
New sorting modes for Vector Sort node.

portnov:
This adds several sorting modes:

* **Auto XYZ**. Automatically detect a plane where all vertices are lying (or a
  plane most similar to that). Then sort vertices in the reference frame of
  that plane: first along X (which is some arbitrary direction along the
  plane), then, if there are some points with the same X - sort them along Y
  (which is some another direction in the plane, but perpendicular to X), and
  then if there are still points with same X and Y - sort them along Z (i.e.
  along the normal of the plane).
* **Auto Direction**. Automatically detect a straight line along which all
  points are lying (or a line most similar to that). Then sort all vertices
  along that line.
* **Auto Phi / Z**. Automatically detect a plane where all the vertices are
  lying (or a plane most similar to that); also calculate a central point in
  that plane (barycenter of all vertices). Then express all vertices in terms
  of cylindrical coordinates in reference frame of such a plane: Z axis is
  pointing along plane's normal, Phi is going counterclockwise around plane's
  normal. Then sort the vertices: first according to Phi value; then, if there
  are points with the same Phi, sort them according to Z value; and, if there
  are points with the same Phi and Z, sort them according to Rho value
  (distance from central normal).

This also adds **reverse** toggles where it was not so hard to add:
* **Reverse** for **Auto Direction** and **Auto Phi / Z** modes;
* **Reverse X** / Y / Z for XYZ and **Auto XYZ** modes.

I've also tried to update documentation accordingly. However, I still don't quite understand what exactly **Axis** mode does, so I had do leave it undocumented.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/vector/vertices_sort.rst
+++ b/docs/nodes/vector/vertices_sort.rst
@@ -15,19 +15,61 @@ and optional inputs(Vector, matrix and user data)
 Parameters
 ----------
 
-+----------------+---------------+-------------+----------------------------------------------------+
-| Param          | Type          | Default     | Description                                        |
-+================+===============+=============+====================================================+
-| **Vertices**   | Vector        |             | vertices from nodes generators or lists (in, out)  |
-+----------------+---------------+-------------+----------------------------------------------------+
-| **PolyEdge**   | Int           |             | index of polgons or edges     (in, out)            |
-+----------------+---------------+-------------+----------------------------------------------------+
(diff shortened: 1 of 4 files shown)


nortikin:
cylindrical coordinates long musthave feature

PR #3381, opened 2020-06-29
re-reset n_id of SNL

### LINKED ISSUE #3380: SNL node
Weird bug. I guess that it is related with last update system changes because I did not manage to reproduce the bug with my experimental update system. 

In my case it does not reproduce each time. I have to open the file several time to get it. Probably such instability it is related with Sverchok loading order. Also it is related with a similar socket names of SNL nodes I think.

[test1.zip](https://github.com/nortikin/sverchok/files/4844876/test1.zip)

vicdoval:
Although it seems redundant fixes #3380
Just one line added at 306
- [x] Ready for merge.

### DIFF
--- a/nodes/generator/script1_lite.py
+++ b/nodes/generator/script1_lite.py
@@ -60,7 +60,7 @@ class SV_MT_ScriptNodeLitePyMenu(bpy.types.Menu):
             if (node.selected_mode == 'To_TextBlok'):
                 args = dict(operator='text.open', props_default={'internal': True})
             else:
-                args = dict(operator='node.scriptlite_import') 
+                args = dict(operator='node.scriptlite_import')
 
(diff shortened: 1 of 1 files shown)


zeffii:
the `less global updates` stuff added `n_id` by default to all nodes in the nodetree.py 

https://github.com/nortikin/sverchok/pull/3046/files#diff-497b36879012785883fb099d45451e32R415

so yeah, it makes sense that this will be needed to keep the node fresh.

Durman:
Does not this code already fixes the node id property? Why it should be done internally in the node?

https://github.com/nortikin/sverchok/blob/08d6fcd59efebb117b01db5403aeafbc9444aabf/node_tree.py#L776-L787

PR #3388, opened 2020-07-05
prep for bg drawing

zeffii:
maybe i get around to finally doing this.

old node . https://gist.github.com/zeffii/9451340

image

almost :)


zeffii:
ok cool.

image

vicdoval:
It would be also cool to have a way to see 'Object index'

zeffii:
yeah, object index, possibly a relatively coherent way would be to do it like 
```
"f{obj_index}: {text_to_display}"
```
so if object is 4 and index = 7 then `4: 7`

- positive:, 
       - each index is labelled with the object index.
       - even if multiple disjoint meshes are displayed close to eachother / intermingled.. you'll still be able to distinguish.
- negative: maybe it's verbose  (but that's acceptable i think if this option is just a toggle on/off)

PR #3393, opened 2020-07-07
add normals viewer shader to solid viewer

zeffii:
- [x] add normal to rgb
- [x] add vector light incidence

image

looks cool. could be a faster implementation


nortikin:
??

zeffii:
the custom shader code was broken anyway. and the vector light was needed

PR #3406, opened 2020-07-18
WIP add numba dep handler

zeffii:
adds numba to dependencies and an initial recode of inset_specials.

- almost works , with the exception of where the user passes a random list of ignore masks.
- also, currently not speed boosted by `numba`.

once it works with random ignore masks, then it's appropriate to adjust the code to try to apply numba techniques to it. I've done all i can for now. back to reading the `numba` docs.

### DIFF
--- a/dependencies.py
+++ b/dependencies.py
@@ -138,6 +138,16 @@ except ImportError:
     info(freecad_d.message)
     FreeCAD = None
 
+numba_d = sv_dependencies["numba"] = SvDependency("numba", "https://numba.pydata.org/")
+try:
+    import numba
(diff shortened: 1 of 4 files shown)


zeffii:
the get_normal_from_polygon function will be multi-part and initially deal with convex only, (it will handle concave as-if convex...)

zeffii:
the `do` functions require some rewrites to accept flat arrays and loops, instead of nested polygons. the `do` functions might even become one function, with a few if / elif / else statements.

vicdoval:
What happen to this investigation?  Now that we have extra nodes numba is also pip-instalable

Issue #780, opened 2016-05-20
beta SVG writer/viewer


zeffii:
The idea is to write a node that takes in 3d verts/edges/faces and generates appropriate 2D SVG path data. Perhaps initial implementation will be only a Path generator, because its easy to take 3D data and strip off the 3rd dimension.

inputs, I think 
- linewidth (vectorized)
- stroke (color, vectorized)
- fill (color, vectorized)
- verts, edges, faces

![image](https://cloud.githubusercontent.com/assets/619340/15429269/7acaa802-1ea0-11e6-8865-468aa5936ff6.png)

zeffii:
![image](https://cloud.githubusercontent.com/assets/619340/15437271/86c803fe-1ec6-11e6-92d2-d4a92b90f71e.png)

zeffii:
![image](https://cloud.githubusercontent.com/assets/619340/15448476/10754d90-1f63-11e6-8ca1-da04adcfa44b.png)

nortikin:
ok. how will you use this data? 3D for site? or how?

zeffii:
imagine you want to quickly draw some parametric geometry as an svg. No this node will decompose 3d-xyz to 2d-xy

zeffii:
I have not decided how it will interpret incoming socket data..  like..
- if verts only input, then create a group with ellipses at each coordinate
- if verts+edges then generate paths vert-edge-vert,  ...vert-edge-vert
- if verts+faces then generate closed polygons

nortikin:
nice

zeffii:
don't know if I'll find time to work on this again..
![image](https://cloud.githubusercontent.com/assets/619340/15801386/361f8160-2a93-11e6-8c27-646d979077c7.png)

i'm trying to think of another way to compose svg with nodes

PR #3407, opened 2020-07-18
points inside solid

vicdoval:
Points inside solid node
Also solid viewer accepts alpha values :D 
points_inside_solid_blender_sverchok_example

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


zeffii:
slightly offtopic, but maybe it's time to do a not-so-realtime renderer, with beauty-passes for AA with oversampling and render to disk.

vicdoval:
@zeffii i could be cool, but I'm not sure it would be better than eevee 😄 other idea would be a  SVG exporter which would be awesome to generate blueprints from sverchok  (i have a couple of scripts for the SNL for that task but nothing too elaborated...)

zeffii:
an svg viewer, totally.  i gave up https://github.com/nortikin/sverchok/issues/780

vicdoval:
> slightly offtopic, but maybe it's time to do a not-so-realtime renderer, with beauty-passes for AA with oversampling and render to disk.

I was thinking about this lately... it could pretty be cool to render the viewer draw from camera view to a image file so it can be mixed with the render in the compositor to saved to file.... eevee is cool but does not render edges and also with the viewer draw another kind of effects could be achieved... did you ever try something like that @zeffii ?

PR #3420, opened 2020-07-24
Update sample_uv_color.py

Kosvor2:
Now with Obj. socket (only first object is used). And works on Blender 2.91 alpha.

### DIFF
--- a/nodes/object_nodes/color_uv_texture.py
+++ b/nodes/object_nodes/color_uv_texture.py
@@ -28,35 +28,33 @@ from sverchok.data_structure import (updateNode, second_as_first_cycle as safc)
 
 
 class SvMeshUVColorNode(bpy.types.Node, SverchCustomTreeNode, SvAnimatableNode):
-    ''' Find pixel on UV texture from surface'''
+    ''' Find pixel on UV texture from mesh object and change its color'''
     bl_idname = 'SvMeshUVColorNode'
     bl_label = 'Set UV Color'
     bl_icon = 'UV_DATA'
 
-    mode1: BoolProperty(name='normal_update', default=True, update=updateNode)
     image: StringProperty(default='', update=updateNode)
-    object_ref: StringProperty(default='', update=updateNode)
-
     unit_color: FloatVectorProperty(name='', default=(1.0, 1.0, 1.0, 1.0),
         size=4, min=0.0, max=1.0, subtype='COLOR', update=updateNode)
 
     def draw_buttons(self, context,   layout):
         self.draw_animatable_buttons(layout, icon_only=True)
-        layout.prop_search(self, 'object_ref', bpy.data, 'objects')
-        ob = bpy.data.objects.get(self.object_ref)
+        ob = self.inputs[0].sv_get()[0]
         if ob and ob.type == 'MESH':
             layout.prop_search(self, 'image', bpy.data, "images", text="")
 
     def sv_init(self, context):
         si = self.inputs.new
+        si('SvObjectSocket', 'Mesh Object')
         si('SvVerticesSocket', 'Point on mesh')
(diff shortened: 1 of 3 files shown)


Kosvor2:
Fixed two other UV nodes.

zeffii:
@Kosvor2  please look at how existing nodes get the dependency graph.

https://github.com/nortikin/sverchok/blob/master/nodes/scene/objects_mk3.py  is a useful example

( i realize that `obj_in_lite` directly accesses the depsgraph, it should not.. it should also be implemented like objects_mk3 )

enzyme69:
Thanks Kosvor, 2.91 alpha, Sample UV Color works for a single vertex only.

Screen Shot 2020-07-24 at 9 22 45 pm

enzyme69:
@Kosvor2 https://github.com/enzyme69/blendersushi/issues/507 It used to work like this

Kosvor2:
@enzyme69  List Item node outputs only one vertex here? Is sample color node trows error in console when you feeding to it several verts?

enzyme69:
Yes one vertex works, but if I plug all the verts: 

```
2020-07-25 09:08:45,809 - sverchok.core.update_system - ERROR - Node Sample UV Color had exception: index 1025 is out of bounds for axis 1 with size 1024
  File "/Users/jimmygunawan/Library/Application Support/Blender/2.91/scripts/addons/sverchok-master/core/update_system.py", line 383, in do_update_general
    node.process()
  File "/Users/jimmygunawan/Library/Application Support/Blender/2.91/scripts/addons/sverchok-master/nodes/object_nodes/sample_uv_color.py", line 71, in process
    outc.append(pixels[int(V.x*(width-1)), int(V.y*(height-1))].tolist())
IndexError: index 1025 is out of bounds for axis 1 with size 1024

```

enzyme69:
[color_UV_011colorPerPoint.blend.zip](https://github.com/nortikin/sverchok/files/4974554/color_UV_011colorPerPoint.blend.zip)

Screen Shot 2020-07-25 at 9 12 40 am

Sometimes, it seems to work, but sometimes throwing error and I tried points on surface, but look a bit different to color on UV. And I wondering if it is because of the "sampling" 1x1 of the pixel. Maybe we could sample 3x3 or 5x5 average pixel?

enzyme69:
Sometimes the error is like this:
```
  File "/Users/jimmygunawan/Library/Application Support/Blender/2.91/scripts/addons/sverchok-master/core/update_system.py", line 383, in do_update_general
    node.process()
  File "/Users/jimmygunawan/Library/Application Support/Blender/2.91/scripts/addons/sverchok-master/nodes/object_nodes/sample_uv_color.py", line 65, in process
    found_poly = obj.data.polygons[ind]
IndexError: bpy_prop_collection[index]: index 104 out of range, size 1

```

enzyme69:
Screen Shot 2020-07-25 at 9 28 18 am

Kosvor2:
@enzyme69 well, i pushed some fixes in master, try again please.

enzyme69:
It might be getting there but there is still something not quite right with the color it is sampling

Screen Shot 2020-07-26 at 11 55 13 am


Screen Shot 2020-07-26 at 11 55 20 am


[plane_sampling_001.blend.zip](https://github.com/nortikin/sverchok/files/4977020/plane_sampling_001.blend.zip)

Kosvor2:
Progress so far:
sample_uv_node_1231235

points on mesh must be in local object space so i centered the "Plane node". Then i resized image to be 24x24 pixels resolution like a "plane" from "plane node" (but i cant understand yet why it helped) and then image was rotated 90 degree, which i now fixed in master.

Kosvor2:
Vertical and horizontal resolution of the image must be exactly the same for this node to work properly. I will add this to some places.

PR #3431, opened 2020-08-01
allow dynamic config of recursion limit

zeffii:
use this to find the limit of recursion where python dies on the exception heap :)


zeffii:
this is a stupid idea. the code must be replaced instead.

PR #3446, opened 2020-08-08
Solid Distance Node

vicdoval:
The node returns the distance between a Solid, a solid Face or a solid Edge to another Solid (or solid face, solid edge or vertex).

It will also return which is the closest point between the selected objects and information about where this closest point is.
solid distance

solid distance
solid distance

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


nodes/solid/solid_distance.py
```diff
@@ -0,0 +1,164 @@
+
+from sverchok.dependencies import FreeCAD
+from sverchok.utils.dummy_nodes import add_dummy
+
+if FreeCAD is None:
+    add_dummy('SvSolidDistanceNode', 'Solid Distance', 'FreeCAD')
+else:
+    import bpy
```
Durman, line 8: What is the reason of coding all node in the else condition? Extra nestedness is not good.

vicdoval: Is not my favourite either but is the convention I took from sverchok-extra and that now is in all the nodes with dependencies

portnov: @Durman the original problem is that we do not want use "FreeCAD" symbol (in this case) to be used in case it is None. It would be good if we could write something like
```python
if FreeCAD is None:
    add_dummy(...)
    return_from_module

class SvSomeNode...
```
but python does not have such construct afaik...

Durman: @portnov Did not understand. If the class will be declared outside else statement what it effects?

PR #3458, opened 2020-08-13
SVG nodes (WIP)

vicdoval:
Scalable  Vector Graphics Nodes:

image

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


nodes/svg/text.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def draw(self, height, scale):
+        svg = '<text '
+
+        svg += 'font-size="{}px" '.format(self.size)
+        svg += 'font-family="{}" '.format(self.font_family)
+        x = self.location[0] * scale
+        y = height-self.location[1] * scale
+        if self.angle != 0:
+            svg += 'transform="translate({},{})rotate({})" '.format(x,y,self.angle)
```
zeffii, line 48: you don't like `f string` formatting?
```
svg += 'transform="translate({},{})rotate({})" '.format(x,y,self.angle)
svg += f'transform="translate({x},{y}) rotate({self.angle})" '
```

zeffii: either way. super cool stuff.

portnov:
Eeh :)
What do these nodes do? Generate an SVG file? or generate some geometry in Blender / Sverchok?

zeffii:
would be neat to spawn a localhost, pointing at that .svg for a live response.

vicdoval:
@portnov  there was a node that fade in from a previous failed experiment (solid from loft) but this PR is about generating SVG files from Sverchok

nodes/svg/text.py
vicdoval: @zeffii Thanks! i didn't know the f'string formatting but looks super neat and i will move forward to using it

portnov:
Hmm. @vicdoval @zeffii I'm not sure it's a good idea to have such nodes in sverchok... May be it's better to put them into separate addon? We do not put audio or texturing nodes into sverchok, after all...

vicdoval:
> would be neat to spawn a localhost, pointing at that .svg for a live response.

That would be ultra-cool but i don't know how to implement it (and how responsive it would be). Any help would be appreciated 😄

vicdoval:
@portnov the basic idea is to be able to produce blueprints, technical drawings or files for CNC production from Sverchok. I needed that feature in the past and I did it with some SNL scripts. At the beginning I thought of a single node to do it but I found more convenient to split it in some smaller nodes easier to maintain and combine.
Of course it opens many other possibilities that lead to the idea of a svechok-extra addon but that has another troubles that would need to be fixed (addition to examples, searching from menu, tests, addon-github-update)

image

docs/nodes/svg/svg_index.rst
```diff
@@ -0,0 +1,11 @@
+*****
+Solid
```
zeffii, line 2: svg

vicdoval:
Seems that Travis is on holidays (at least from the last 5 days) 😄

zeffii:
i've got a nice csv that'll test this on :)

vicdoval:
I would appreciate an update in the docs or even better to establish a automatic update system of the documentation

nortikin:
> I would appreciate an update in the docs or even better to establish a automatic update system of the documentation

yes, it will be better. Maybe to publicate on github.io automatically?

Issue #3123, opened 2020-04-27
node.create_sockets


Durman:
It looks like it also is not used anywhere.
https://github.com/nortikin/sverchok/blob/93e9d4d4bdaa849ffddf5f9da2e32bea260f2c1d/node_tree.py#L571-L581

zeffii:
make a branch, and see what breaks :) i'd only be worried about affecting Monads.

zeffii:
then also the call in `sv_init`  to `create_sockets` is also a no op.

i think once upon a time @portnov made a wrapper for `node.process()`, (grep "input_descriptors") which we retain in data_structure.py, but it was never used in a node. I think the wrapper can be removed too
https://github.com/nortikin/sverchok/blob/b74b9f0f8076b823c4bd8bc1842b77d39a8b6422/data_structure.py#L1077

PR #3469, opened 2020-08-19
delete node.create_sockets method

Durman:

#3123

- [x] Code changes complete.


portnov (approved):
Agreed. It will remain in git history in case someone needs it.

Issue #3371, opened 2020-06-26
Obj instanser node


Durman:
`BV Alpha` node with `obj instancer` node together change source mesh of input object of `obj instancer` node. Renaming `basedata_name` property to anything else can't bring back work of the node in expected way.

![2020-06-26_13-03-42](https://user-images.githubusercontent.com/28003269/85840509-a2fdf900-b7ad-11ea-955e-4f4c738d28ff.png)
 
@willneedit

zeffii:
maybe the `object base name` property should do a check first to see if there are any viewer nodes that use that basename, in any tree in the blend. and refuse + show warning to change desired object basename to something that isn't used.

Durman:
It of course would be good. But it would be okay if they just override each other as `viewers` nodes do now. Probably the `instancer` node could have the same policy about name field as `viewers` nodes have. 
And if a user names `instanser` a same name as `viewer` node accidently he will face with trouble anyway.

In the particular case something worse is happening. Together they change mesh of another object which they should not at all. It quite unpleasant because user can lost his mesh without possibility restoring it.

zeffii:
painful, but not a catastrophy.

- unlive the viewer node that generates Alpha objects,
- rename the basename
- live that same viewer node again.
- remember in Instancer node not to use the basename  that is being used by any viewer nodes

zeffii:
object instancer   must  only allow unique basenames.

Durman:
I think this should not happen at all without any instructions. Both nodes in the example should change only those objects which start with prefix `Alpha`. But in the example `Cube` object was changed without any reason for this.

zeffii:
feel free to fix the node @Durman

zeffii:
there's two bugs here i think
1. we allow basename to be non-unique,  this is the main problem
2. because of 1. other functionality is "undefined".

Durman:
I consider bug with naming (1) as good because it helped to reveal the another worse bug. I'm not in a hurry to fix the node right now. Potentially I would like to gain from advantages of the node but not in current stage.

Durman:
I understood the problem. When `instancer` creates its objects it names them Alpha.0000, Alpha.0001 and so on. When `Viewer` create its objects it checks whether the objects already was created and determines them by names. So it sees object with name Alpha.0000 and rewrite mesh in its mesh data block. According that the data block is shared with the source object, the source object also loose its original mesh.

[...]

PR #3471, opened 2020-08-19
Using pointer property - object instanser node

Durman:

#3371

I already nearly recreated all previous functionality and I'm quite fond with code appearance. Now I'm thinking about extra functionality. It won't be too hard instasing any type of objects like metaballs or cameras. The question is usefulness of it.

Also I have added mixin class for keeping collection property. I expect that this code can be reused for implementing using of `pointer property` for other `viewer` nodes.

## todo:

- [ ] Renaming base name parameter
- [ ] Remove objects when matrix socket is disconnected
- [ ] Hide objects button - new feature
_
- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


Durman:
Advantages of using pointer property: two nodes have the same basename but their object list does not intersect with each other. Blender itself is responsible for generating objects without name intersections.

Durman:
I have added buttons of standard object properties:

Durman:
Now it is possible to instancing all types of objects exclude metaballs, images, force fields.

Durman:
handling collections

Durman:
Looks like the node more fitting into `viz` category. More likely nodes in `Scene` category should get some data from file. Main purpose of current node is to put data in scene.

PR #3473, opened 2020-08-20
"Deconstruct curve" node

portnov:
This node deconstructs any NURBS or NURBS-like curve into it's main components:
control points, weights and so on.

At the moment, this node can effectively work with the following types of curves:

* NURBS curves
* Bezier curves
* Cubic splines
* Polylines
* Circular arcs
* Line segments

Some nodes, that output complex curves composed from simple curves (for
example, "Rounded rectangle"), have **NURBS output** parameter; when it is
checked, such nodes output NURBS curves, so "Deconstruct curve" can work with
them.

Other types of curves are considered as not having NURBS components (control
points, for example), so outputs will be empty.

Screenshot_20200820_223627
Screenshot_20200820_223802

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


nodes/curve/deconstruct_curve.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def sv_init(self, context):
+        self.inputs.new('SvCurveSocket', "Curve")
+        self.outputs.new('SvStringsSocket', "Degree")
+        self.outputs.new('SvStringsSocket', "KnotVector")
+        self.outputs.new('SvVerticesSocket', "ControlPoints")
+        self.outputs.new('SvStringsSocket', "Edges")
+        self.outputs.new('SvStringsSocket', "Weights")
+
+    def deconstruct(self, curve):
```
Durman, line 30: This method does not look like belong to the node class. It even does not use self argument. Most likely the proper place would be: `curve.deconstruct()`

Issue #3390, opened 2020-07-06
Renaming viewer nodes


Durman:
## Problem statement

Renaming objects names in viewer nodes creates copy of an object with new name. I would expect current name of an object would be changed.

![2020-07-06_21-44-22](https://user-images.githubusercontent.com/28003269/86623291-62368900-bfd2-11ea-9f81-d2b7132d1bed.png)

zeffii:
this is tricky, there is a scenario where i have used the feature (changing basename) to have a trail effect of old objects.

My thoughts on this are that the more liberal implementation of renaming the basename + autorenaming of generated objects, is done via an Operator. 
 - select the node
 - trigger the (unwritten) operator to change the basename + update children.

zeffii:
https://github.com/nortikin/sverchok/pull/2825

zeffii:
@vicdoval @Durman   check out the rename operator :)  only implemented in bmview for testing.

Durman:
It looks it is working. I will agree with UI solution if the text field won't be editable. Because it too gives opportunity to rename object but it will create copy of an object with new name instead.

![2020-07-13_22-50-32](https://user-images.githubusercontent.com/28003269/87342054-a211e900-c55b-11ea-8731-e9d273bdfa05.png)

Durman:
I think with setter and getter the UI can be done more elegant. I mean it can remain as it was without new button and window.

https://docs.blender.org/api/current/bpy.props.html#getter-setter-example

vicdoval:
I like it, it gives a new feature that was not possible before...
Other approach that could also work would be:
- if the text-field is edited -> rename 
- Button to make a "unconnected" copies

vicdoval:
And some improvements for the actual implementation:
"Cancel" Button although it can be cancelled with Escape key it would make it clearer
Cancel operator if pop-up text-field is empty

zeffii:
I'm partial to using the operator _only_ method to set the string in `basedata_name` , and making the string non-editable on the node UI, possibly removing the updatefunc for that property completely.

but i'm distracted with other worries right now, so if you collectively can settle on an approach that treats the user with more respect than I do, then certainly move towards implementing that.

Issue #2767, opened 2019-12-25
Mesh data structure


Durman:
## Problem statement

![2019-12-25_13-53-30](https://user-images.githubusercontent.com/28003269/71442112-f53a6080-271d-11ea-999a-b479efe242ed.png)
As you can see on the picture process of plugging sockets to each other is becoming too boring. And the node does not have `vertices data` and `edges data` sockets yet. 

**So the goal is to make something like alias for group of sockets.**

## Solution

Such issues already was in the past. #184
My proposal is much more simple. Actually it developing of the idea of this #2766

I think for such nodes as `bevel`, `inset face` and similar others it would be nice to have mesh input socket instead of (verts, edges, faces, verts data, edges data, faces data). For creating data for such sockets Sverchok should have Mesh in and out nodes (or SvMesh in / out ?)

```Text
|  Mesh in   |
+------------+
|       mesh |
| verts      |
| edges      |
| faces      |
| vert data  |
| edge data  |
| face data  |
```

Mesh output socket will have dictionary type and have certain keys according input data. So algorithm of working with input mesh in such format will be next: a node search certain keys like `verts` or `face data`, if such keys like `verts` and `faces` which are mandatory for inset faces algorithm are not found then NoDataError is raised, if other keys are not found the node continues working without this data.

**Nestedness:**
Actually it lays out of the boundary of the current problem but with the solution it looks like it can be quite easy to create arbitrary number of nested meshes. More precisely to say meshes can be joined to categories with arbitrary level of nestedness.
Some examples:

```Text
Category A:
    mesh 1
    mesh 2
```

Here is dictionary with two keys: `mesh 1` and `mesh 2`. `Mesh 1` and `mesh 2` are also dictionaries in which has at least key `verts`. The category dictionary can't include any values exclude dictionary with mesh data and other dictionary which include dictionaries with mesh data. So algorithm for searching mesh dictionaries should be next:

```Python
stak_dicts = [input_dict]
while stak_dict:
    current_dict = stack_dicts.pop()
    if 'verts' in current_dict:
         # this is mesh dict
         result.append(main_function(**{current_dict[key] for key in required_keys if key in current_dict})
    else:
        # most likly this is category dict
        stack_dicts.extand([val for val in current_dict.values() if isinstance(val, dict)])
```

Durman:
Comparison of using mesh socket with separately setting elements of mesh:
[...]

PR #3475, opened 2020-08-22
Mesh viewer refactoring - using pointer property

Durman:

#3371
#3390

The node should work much faster. More readable code. Collection selector will be available. No name intersection problems.

## TODO
- [x] Selection objects and create material operators
- [x] Picking names refactoring ?
- [x] Apply matrixes (two variants to object/mesh ?)
- [x] Merge mesh
- [x] Smooth mesh
- [x] Wireframe mesh
- [x] Apply material indexes
- [x] Show / hide data by node tree request
- [x] Topology changes test option
___
- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.


Durman:
Final UI:

vicdoval:
A small suggestion:
Leaving the  label as "Mesh Viewer" and only changing it when the node is hidden so it is clearer in every screenshot what node is being used

Durman:
2.5 times speed up if input topology is changed

```Text
[DEBUG] sverchok.core.update_system: Processed  Viewer BMesh in: 0.3033
```

```Text
[DEBUG] sverchok.core.update_system: Processed  Mesh viewer in: 0.1215
```

5.5 times speed up if topology is unchanged:

```Text
[DEBUG] sverchok.core.update_system: Processed  Viewer BMesh in: 0.1707
```

```Text
[DEBUG] sverchok.core.update_system: Processed  Mesh viewer in: 0.0315
```

**21 times speed up if input vertices are float 32 array:**

```Text
[DEBUG] sverchok.core.update_system: Processed  Mesh viewer in: 0.0080
```

speed test 2

speed test 1

Durman:
Show in viewport with tree property changes.

show viewport

Durman:
Unlock origin -
unlock origin

vicdoval:
@Durman  A little detail :)
The matrix socket: the quik_link button does not disappear and it does not provide list length information. (makes it look broken) 
image

Durman:
I will have a look.

vicdoval:
@Durman when you did this refactor did you test this?
```
me.edges.add(count=len(verts))
me.edges.foreach_set("vertices", edges.ravel())
```

Taken from here https://blender.stackexchange.com/questions/190547/mesh-edges-foreach-set-doesnt-update

Durman:
Look at the `generate_mesh` function in this file - https://github.com/nortikin/sverchok/blob/mesh_data_structure/nodes/mesh/viewer_mesh.py

I found out that using bmesh approach is faster and much more robust. Adding mesh data right inside mesh data block is fragile solution which can very eaily cause a Blender crash.

I was posting some tests here - https://github.com/nortikin/sverchok/issues/2767#issuecomment-586945916

vicdoval:
I see.. 
The only thing is that bmesh does not like numpy arrays. Does it?
And if you have numpy edges it is slower to turn them to python and setting  them than to set them directly in numpy
```
import bpy
from timeit import timeit
import numpy as np

NUM = 1000000

def get_mesh():
    me = bpy.context.active_object.data
    me.clear_geometry()
    me.vertices.add(NUM)
    me.edges.add(NUM - 1)
    return me

edges = [[i, i + 1] for i in range(NUM - 1)]
np_edges = np.array(edges, 'i')

def test1(me, edges):
    for me_e, e in zip(me.edges, edges):
        me_e.vertices = e
        
def test2(me, edges):
    me.edges.foreach_set('vertices', [i for e in edges for i in e])
    
def test3(me, np_edges):
    me.edges.foreach_set('vertices', np_edges.flatten())
    
def test4(me, np_edges):
    me.edges.foreach_set('vertices', [i for e in np_edges for i in e])

def test5(me, np_edges):
    me.edges.foreach_set('vertices', [i for e in np_edges.tolist() for i in e])

print("Test 3", timeit("test3(me, np_edges)", "from __main__ import test3, np_edges, get_mesh; me = get_mesh()", number=1))
print("Test 4", timeit("test4(me, np_edges)", "from __main__ import test4, np_edges, get_mesh; me = get_mesh()", number=1))
print("Test 5", timeit("test5(me, np_edges)", "from __main__ import test5, np_edges, get_mesh; me = get_mesh()", number=1))
print("Done")


```

```
Test 3 0.1472976000000017
Test 4 0.7561349000000064
Test 5 0.2072122000000718
```

Durman:
Probably this is one of the reasons why we don't use numpy edges.))

But if the Mesh Viewer node would use the `me.edges.foreach_set` function how to make the fast test that the input mesh is correct? 
When you reconnect the node from one mesh to another you cant reconnect `vertices`, `edges` and `polygons` simultaneously. So there can be the cases when given edges point to non existent points and this causes the crash.

vicdoval:
maybe `np.amax(np_edges)>NUM_VERTS` could do a fast check

Anyway accepting numpy edges with a tolist in the mesh viewer wouldn't be that bad...

PR #3481, opened 2020-08-25
Circle SVG outputs curves

vicdoval:
Circle SVG outputs curves objects.
image

- [x] Ready for merge.

### DIFF
--- a/docs/nodes/svg/circle_svg.rst
+++ b/docs/nodes/svg/circle_svg.rst
@@ -20,6 +20,7 @@ Outputs
 -------
 
 - SVG Objects
+- Curves: Generated Shapes as Sverchok Curve objects
 
 
 Examples
--- a/nodes/svg/circle_svg.py
+++ b/nodes/svg/circle_svg.py
@@ -17,7 +17,7 @@
 # ##### END GPL LICENSE BLOCK #####
 
 from math import sin, cos, pi, degrees, radians
-
+from mathutils import Matrix
 import bpy
 from bpy.props import BoolProperty, IntProperty, FloatProperty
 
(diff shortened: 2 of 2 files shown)


Durman:
What's differences?

I does not like curve output of the node. I thought that purpose of SVG nodes are create SVG documents. But with `curve` socket they do something more. It would be much better if the node could get curve object and convert it into svg but not vice versa.

vicdoval:
The new output is  meant to easily visualise the the output in the 3d view.
Is true it could accept `Circle Curve`  and `Ellipse curve` objects, not clear how to make the user aware about which curves are accepted and which are not (maybe it could also accept a  `<<Matrix vector field>(<Circle curve>) curve>`)  I may add a Curve mode to the node that reports an error if it is not a valid curve.

Another little thing is that ellipse (curve) accepts matrix input not a vector input ( with the vectorization issues)

image

portnov:
It would be probably good to have a node which (at least tries to) convert Curve objects to SVG:
* If Curve is SvCircle, output SVG circle
* If Curve has `to_bezier_segments()` method, then call it to convert the curve to list of SvBezierCurve, and output SVG bezier curve
** :warning: at the moment there is no `to_bezer_segments()` method; I planned to implement it some time later. If you want, I can implement it sooner, just ask.
* otherwise, you can either fail, or output SVG path by rendering the curve point-by-point. Probably you will want to create a toggle for this.
* obviously, in all cases you will have to discard one of 3 coordinates, or make a projection to some plane.

I'm not sure if this will be actually useful, though :) I myself do not have applications for svg output at the moment.

vicdoval:
@portnov I was also thinking about that. 
It would be awesome to have `to_bezier_segments()` method and a soon as you implement it I will include it in the near future `Curve SVG` node 😃 

Another idea that you may like is to have a curve output in the `Path SVG` that has a user mode to create complex paths mixing curve types with letter commands...

portnov:
@vicdoval I've added `to_bezier_segments()` method where it is possible. It will return list of curves of the same degree as source curve degree was; so you have to do something like

```python
if hasattr(curve, 'to_bezier_segments'):
    for segment in curve.to_bezier_segments():
        if segment.get_degree() not in {2, 3}: print("SVG supports only curves of 2nd and 3rd degree")
        bezier_points = segment.get_control_points()
````

PR #2024, opened 2018-01-08
Implicit data conversion solution.

portnov:
Was discussed in #2020. Has relation to #1951 (though most probably I will make a separate PR for use of this one).

This introduces a notion of "implicit data conversion policy": a class that is responsible for conversion of data from one socket type to another. There are the following implicit conversion policy classes defined:
* `NoImplicitConversionPolicy`: base one, prohibits any implicit conversion.
* `DefaultImplicitConversionPolicy`: implements the same logic we had before: do some implicit conversions. For `StringsSocket` socket class, just passes data as they were (this is actually the same logic we had before; this is done because we have some number of nodes that consume any type of data into StringsSocket inputs - all the list processing nodes, they do not need any data conversion).
* `LenientImplicitConversionPolicy`: do not check anything, pass any data as they are.

To all `sv_get()` methods, added a new `implicit_conversions` argument (defaulting to `DefaultImplicitConversionPolicy`).
If specific node wants to describe specific data conversion rules for specific socket, it should define a policy class and pass it to `sv_get()`.

As a side effect, this will have better error reporting, for example:

```
2018-01-08 21:30:30,289 [ERROR] sverchok.core.update_system: Node Viewer Draw.001 had exception: Implicit conversion from socket type MatrixSocket to socket type StringsSocket is not supported for socket edg_pol of node Viewer Draw.001. Please use explicit conversion nodes.
```

Please review and test.

PR #3483, opened 2020-08-25
Node tree module refactoring

Durman:

I'm thinking about next steps of cod improving:

- [x] Remove unused cod
- [x] Write documentation to methods (if possible)
- [x] Move out from module what is not relevant
- [x] Combine methods and properties in `mixin` classes responsible for one feature or another

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.


node_tree.py
```diff
@@ -558,17 +505,6 @@ def error(self, msg, *args, **kwargs):
...
-        """
-        Set implicit conversion policy to be used by default for specified input socket.
-        This policy will be used by default by subsequent .sv_get() calls to this socket.
-        Policy can be passed as direct reference to the class, or as a class name.
-        """
-        if isinstance(policy, str):
-            policy = getattr(sverchok.core.socket_conversions, policy)
-        #self.debug("Set default conversion policy for socket %s to %s", input_socket_name, policy)
-        self._implicit_conversion_policy[input_socket_name] = policy
-
```
portnov, line 571: "implicit convertions" are actually used; maybe "set_implicit_conversions" is not used at the moment, but since the whole thing is used, it can appear it will be used later; I'd suggest to leave the method for now.

node_tree.py
```diff
@@ -523,15 +479,6 @@ def sv_setattr_with_throttle(self, prop_name, prop_data):
...
 
-    def mark_error(self, err):
-        """
-        marks the with system error color
-        will automaticly be cleared and restored to
-        the old color
-        """
-        ng = self.id_data
-        update_error_nodes(ng, self.name, err)
-
```
portnov, line 534: I'm not sure that this is not used. Did you check?

node_tree.py
```diff
@@ -403,13 +368,6 @@ def process(self):
 
         self.has_changed = False
 
-    def get_nodes_supporting_draft_mode(self):
-        draft_nodes = []
-        for node in self.nodes:
-            if hasattr(node, 'does_support_draft_mode') and node.does_support_draft_mode():
-                draft_nodes.append(node)
-        return draft_nodes
-
```
portnov, line 412: I'm not sure that this is not used. Did you check?

node_tree.py
Durman: Everything what I was removing I surely checked.

node_tree.py
Durman: Everything what I was removing I surely checked.

core/monad.py
```diff
@@ -400,7 +400,9 @@ def poll(cls, context):
     @property
     def instances(self):
         res = []
-        for ng in self.sv_trees:
+        sv_trees = [ng for ng in bpy.data.node_groups
+                    if ng.bl_idname in {'SverchCustomTreeType', 'SverchGroupTreeType'}]
+        for ng in sv_trees:
```
zeffii, line 405: why are you making this more complicated?

node_tree.py
```diff
@@ -457,24 +342,6 @@ def absolute_location(self):
         """ does not return a vactor, it returns a:  tuple(x, y) """
         return recursive_framed_location_finder(self, self.location[:])
 
-
-    def ensure_enums_have_no_space(self, enums=None):
```
zeffii, line 461: this exists to support older existiing blend files made before enums where forced to have no spaces. I'm not sure you want to remove this ..

node_tree.py
```diff
@@ -705,7 +563,7 @@ def migrate_from(self, old_node):
         pass
 
     def sv_init(self, context):
-        self.create_sockets()
+        pass
```
zeffii, line 566: 👍

node_tree.py
```diff
@@ -334,21 +241,6 @@ def on_draft_mode_changed(self, context):
                 default = False,
                 update=on_draft_mode_changed)
 
-    tree_link_count: IntProperty(name='keep track of current link count', default=0)
```
zeffii, line 337: cool. these can definitely go now. 👍

node_tree.py
```diff
@@ -160,27 +103,6 @@ def tree_id(self):
     def build_update_list(self):
         build_update_list(self)
 
-    def adjust_reroutes(self):
```
zeffii, line 163: interesting : ) last time this was touched was 6 years ago.   good to go : )

core/monad.py
Durman: This strings was migrated from `sv_trees` property of `SvNodeTreeCommon` class.



This is only place where this property was used. I don't think it worth to create separate method for this especially in base class. 

Even more I don't think that there should be such method in any NodeTree classes. More properly would be to write something like this:

```Python
bpy.data.node_groups.sv_trees = property([ng for ng in bpy.data.node_groups if ng.bl_idname in {'SverchCustomTreeType', 'SverchGroupTreeType'}])
```

Or atleast to create separate function in `utils` or where ever.

zeffii: or 
```python
# some node tree util
def get_nodetrees_in_blend(kind=["SverchCustomTreeType", "SverchGroupTreeType"]):
    return [ng for ng in bpy.data.node_groups if ng.bl_idname in set(kind)]


for ng in get_nodetrees_in_blend():
    ...
```
i'd prefer to have the logic of the nodetree kind filtering away from that area.

Durman: The defaults values does not fit with function name. With such name `kinds` argument should not have any default values.

```Python
for ng in get_nodetrees_in_blend():
```
This code does not say about what is actually happening.

zeffii: you're right,  but seen as it would be part of a sverchok utility, it is fine to provide a default which covers all current sverchok groups by default.

Durman:
Funny that it looks like even this method is not called from any parts of Sverchok. At least I have commented this method and Sverchok is working.

https://github.com/nortikin/sverchok/blob/59354d366a3a51f6d39a25d7b5135924598a95f2/node_tree.py#L385-L404

nodes/number/scalar_mk4.py
```diff
@@ -270,6 +272,23 @@ def process(self):
 
             self.outputs[0].sv_set(result)
 
+    def ensure_enums_have_no_space(self, enums=None):
```
zeffii, line 275: cool. this is better.

Durman:
@vicdoval Can you give description of below method? As far as I understood it returns some special groups but it definitely is doing something more under hood.

https://github.com/nortikin/sverchok/blob/f87864506eb79d59ea128ec555c8b42f3d281a42/node_tree.py#L221-L230

vicdoval:
- Get monads of node tree, 
- Update them  (the sv_update method will check if anything changed inside the monad and will change the monad outputs in that case) 
- Return the monads that have changed (to inform the caller function that the nodes downstream have to be updated with the new data)

node_tree.py
Durman: It have gone two years since this PR: #2024 and only policy is used now is DefaultImplicitConversionPolicy. Probably we could simplify below method in this way:
```diff
class SvSocketCommon:
    def convert_data(self, source_data, implicit_conversions=None):
        if not self.needs_data_conversion():
            return source_data
        else:
-            policy = self.node.get_implicit_conversions(self.name, implicit_conversions)
-            self.node.debug(f"Trying to convert data for input socket {self.name} by {policy}")
-            return policy.convert(self, source_data)
+            return DefaultImplicitConversionPolicy.convert(self, cource_data)
```

And simplify node class as well:

```diff
class SverchCustomTreeNode:
-    _implicit_conversion_policy = dict()
-    def set_implicit_conversions(self, input_socket_name, policy): ...
-    def get_implicit_conversions(self, input_socket_name, override=None): ...
```

It won't impact on anything but will make code easier. 

And I hardly can imagine how different nodes can use different conversion policy. It will be difficult to predict what output to expect from different nodes.

portnov: At least Lenient policy is used somewhere, as far as I remember... In SvStringsSocket?...

Durman: Nowhere

But I'm not going to delete this class at least not in this PR.

nodes/number/scalar_mk4.py
```diff
@@ -273,6 +272,23 @@ def process(self):
...
+        https://github.com/nortikin/sverchok/pull/3483#discussion_r478389849
+        due to changes in EnumProperty definition "laws" individual enum identifiers must not
+        contain spaces. This function takes a list of enums that the node currently holds, and
+        makes sure the stored enum has no spaces.
+        """
+        for enum_property in enums:
+            current_value = getattr(self, enum_property)
+            if " " in current_value:
+                with self.sv_throttle_tree_update():
+                    setattr(self, enum_property, no_space(current_value))
```
zeffii, line 290: this should have used `self.sv_setattr_with_throttle` too.
```diff
-                with self.sv_throttle_tree_update():
-                    setattr(self, enum_property, no_space(current_value))
+                self.sv_setattr_with_throttle(enum_property, no_space(current_value))
```

Durman: Don't understand what do you mean. The method actually is using throttling method. And I didn't add any changes to this method just moved it from base class.

Issue #3293, opened 2020-05-30
Documentation for Grasshopper which is applicable to Sverchok


portnov:
It is already possible to reproduce many tutorials for GH with Sverchok nodes. Some of documentation is applicable too. I don't know what our approach will be, let's start with just collecting links, and maybe someone will be willing to adapt these tutorials/documentation.

For example, this explains many fundamental things, which are applicable in Sverchok as well (parts about NURBS — with Sverchok-Extra only): https://developer.rhino3d.com/guides/general/essential-mathematics/parametric-curves-surfaces/

zeffii:
_that's_ some beautiful documentation.

Issue #2814, opened 2020-01-11
Render Gallery 2020 #1


portnov:
Previous gallery ticket - #2666.

![adaptive_polys_12](https://user-images.githubusercontent.com/284644/72210404-a42bba00-34dc-11ea-8b24-863e7645789a.png)

portnov:
https://www.reddit.com/r/blender/comments/enb9ai/everyones_favourite_doughnut_but_more_sverchok/

BennyGovaerts:
![Basic_noise_options](https://user-images.githubusercontent.com/49908008/72402689-32f43d00-3750-11ea-90c6-636c79586410.jpg)

BennyGovaerts:
![Basic_noise_options_phi](https://user-images.githubusercontent.com/49908008/72438920-2e5a7380-37a6-11ea-801c-9e218942aa6c.jpg)

portnov:
This I decided to put into sverchok-extra repo, because that's not easy to do with plain Sverchok:
https://github.com/portnov/sverchok-extra/issues/1

AG89z:
![render cycles](https://user-images.githubusercontent.com/26380268/72716771-15b9d700-3b73-11ea-80f2-155ca58c1344.png)
http://www.codeplastic.com/2020/01/20/parametric-wall-with-blender-and-sverchok/

enzyme69:
<img width="765" alt="Screen Shot 2020-02-17 at 11 06 29 pm" src="https://user-images.githubusercontent.com/3952687/74652474-34df6f00-51da-11ea-91de-0fd40a199f56.png">

Shoebox / Containers. WIP.

enzyme69:
Some kind of creature.

<img width="411" alt="Screen Shot 2020-02-22 at 11 43 13 pm" src="https://user-images.githubusercontent.com/3952687/75092599-7e123300-55cd-11ea-8e50-fe407d3f0df7.png">

portnov:
![bend_spiral_1](https://user-images.githubusercontent.com/284644/79682717-e122ed80-823d-11ea-8e27-3ebc350fdccf.png)

portnov:
![arc_surface_2](https://user-images.githubusercontent.com/284644/80255178-ddf08d00-8695-11ea-8cfc-b0c3763fcd31.png)
https://gist.github.com/e9f9d2a2aa11c5993fc2b6ecb43ec19e

portnov:
[...]

PR #3489, opened 2020-08-27
Add a general description of NURBS support

portnov:
Add a page into documentation, which describes support of NURBS curves and surfaces in Sverchok.
I'm adding it via PR, so that probably more people will read it :)

### DIFF
--- a/docs/main.rst
+++ b/docs/main.rst
@@ -17,6 +17,7 @@ Contents:
    geometry
    curves
    surfaces
+   nurbs
    fields
    solids
    induction
--- a/docs/nurbs.rst
+++ b/docs/nurbs.rst
@@ -1,0 +1,150 @@
+
+NURBS: Curves and Surfaces
+--------------------------
+
+Among others, Sverchok supports a special kind of curves and surfaces, called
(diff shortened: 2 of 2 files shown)


enzyme69:
@portnov where to find this documentation after I update to latest sverchok? :)

I found NURBS nodes quite intimidating there are many curves and surfaces and fields, fields is especially unusual, like how to use Voronoi? etc 

If possible, if you have examples node tree for NURBS basics, please let me know! Thanks~

portnov:
@enzyme69 
Unfortunately we still do not have an automated process of updating documentation somewhere on official site, someone has to publish it manually. But, you can always can read the most actual documentation using github, in the `docs` folder; in this case, it is here: https://github.com/nortikin/sverchok/blob/master/docs/nurbs.rst

Considering curves and surfaces, and especially nurbs, you may find it useful to watch / read tutorials or documentation for Grasshopper: concepts are taken from there, some nodes are even have the same names. See also #3293 .

portnov:
Also see: https://github.com/nortikin/sverchok/issues/2814#issuecomment-682645469
and https://www.youtube.com/watch?v=J-meWs5inu8 :)

PR #3490, opened 2020-08-28
scriptnode lite crash investigation

zeffii:
this starts a snlite tidy up

### DIFF
--- a/nodes/generator/script1_lite.py
+++ b/nodes/generator/script1_lite.py
@@ -24,12 +24,13 @@ import traceback
 import numpy as np
 
 import bpy
-from bpy.props import StringProperty, IntVectorProperty, FloatVectorProperty, BoolProperty
+from bpy.props import (
+    StringProperty, IntVectorProperty, FloatVectorProperty, BoolProperty, PointerProperty
+)
 
 from sverchok.utils.sv_update_utils import sv_get_local_path
 from sverchok.utils.snlite_importhelper import (
-    UNPARSABLE, set_autocolor, parse_sockets, are_matched,
-    get_rgb_curve, set_rgb_curve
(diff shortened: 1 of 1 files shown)


nodes/generator/script1_lite.py
```diff
@@ -324,6 +320,18 @@ def process(self):
...
+        self.n_id = node_id(self)
+        return self.node_dict.get(self.n_id)
+
+    def reset_node_dict(self):
+        self.node_dict[self.n_id] = {}
+
+    def set_node_dict_keys(self, **kwargs):
+        ND = self.node_dict.get(self.n_id)
+        for k, v in kwargs.items():
+            ND[k] = v
```
Durman, line 333: What about using of standard Python dictionary API?
```suggestion
    @property
    def node_dict(self):
        self.n_id = node_id(self)
        return self._node_dict.get(self.n_id)
```

```Python
self.node_dict.clear()
self.node_dict['key'] = my_value
```

Durman: Node_dict can use `defaultdict` if needed.

```Python
class Node:
    _node_dict = defaultdict(dict)
```

https://docs.python.org/3/library/collections.html#defaultdict-examples

zeffii: i like to make small changes first.

zeffii:
you can ignore this PR @Durman . it's going to move slowly.

nodes/generator/script1_lite.py
zeffii: what you propose there is in retrospect maybe preferable,  and i will address this later. it looks better, but for now i'm looking for a bug in the prop_search

zeffii:
using the Text pointerProperty does appear to solve this crashing problem, it is needed anyway. I do not have time at the moment to fix all the places that would be effected by switching from StringProperty (script_name) to PointerProperty (script_pointer).

PR #3496, opened 2020-08-30
add string to vector conversion

Durman:
What about adding "string to vector" conversion ? I think it would let to rid out of extra `vector in` nodes in many layouts.

string to vector

### DIFF
--- a/core/socket_conversions.py
+++ b/core/socket_conversions.py
@@ -279,7 +279,8 @@ class DefaultImplicitConversionPolicy(NoImplicitConversionPolicy):
 
     @classmethod
     def string_to_vector(cls, socket, source_data):
-        return [[(v, v, v) for v in obj] for obj in source_data]
+        # it can be so that socket is string but data their are already vectors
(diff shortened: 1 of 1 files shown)


portnov:
It's not obvious how should it work :/
Should it map X -> (X, X, X), or X -> (X, 0, 0), or X -> (0, 0, X), or whatever? So It seems to me the layout with explicit conversion node will be easier to understand.

Durman:
The idea has come from animation node. They have special node which convert value to vector. The mapping is always (x, x, x).

image
https://animation-nodes-manual.readthedocs.io/en/latest/user_guide/nodes/vector/vector_from_value.html

Since I was aware of this little feature I started to notice more and more cases where it is really convenient to use.

If we will add this convention there is even no need in adding the node I guess.

portnov:
With #3312 it will be possible to add a drop-down menu to inputs to select between different conversion logics...

nortikin:
all auto conversion are great

Issue #3437, opened 2020-08-04
Automatization of node creation


Durman:
I have continued experiments with optimization API of node creation and reach much further last attempt. 
Next code is working:

```Python
from itertools import cycle

import bpy
from sverchok.data_structure import updateNode

from sverchok.node_tree import SverchCustomTreeNode
from sverchok.utils.handling_nodes import NodeProperties, SocketProperties, SockTypes, WrapNode, initialize_node

node = WrapNode()

node.props.x = NodeProperties(bpy.props.FloatProperty(update=updateNode))
node.props.y = NodeProperties(bpy.props.FloatProperty(update=updateNode))

node.inputs.verts = SocketProperties(name='Vertices', socket_type=SockTypes.VERTICES, mandatory=True)
node.inputs.faces = SocketProperties(name='Faces', socket_type=SockTypes.STRINGS)
node.inputs.x_axis = SocketProperties(name='X axis', socket_type=SockTypes.STRINGS, prop=node.props.x)
node.inputs.y_axis = SocketProperties(name='Y axis', socket_type=SockTypes.STRINGS, prop=node.props.y)

node.outputs.verts = SocketProperties(name='Vertices', socket_type=SockTypes.VERTICES)
node.outputs.faces = SocketProperties(name='Faces', socket_type=SockTypes.STRINGS)


@initialize_node(node)
class SvTestNode(bpy.types.Node, SverchCustomTreeNode):
    bl_idname = 'SvTestNode'
    bl_label = 'Test node'

    def process(self):
        node.outputs.faces = node.inputs.faces
        node.outputs.verts = [(v[0] + x, v[1] + y, v[2]) for v, x, y in zip(node.inputs.verts, cycle(node.inputs.x_axis), cycle(node.inputs.y_axis))]
```

![test](https://user-images.githubusercontent.com/28003269/89287855-bbdbb300-d665-11ea-890a-a0e09c6dd158.gif)

It should looks magically.

Most advantage of current approach is it impacts on nothing. It does not change/add any node base classes. But new nodes can gain from it.

Also I have tried to use autocomplete as much as possible. At present most names will be hinted by IDE. Access to sockets can be get via dot notation like that `node.inputs.verts`

It is possible to set which sockets should be vectorized and all other works will be done automatically. The `process` function should be coded as if Sverchok does not has vectorization.

In the end I expect to get less bugs and more new nodes coded, in the same time interval.

zeffii:
[...]

PR #3507, opened 2020-09-02
Sockets refactoring

Durman:

#2772 #3483

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done.
- [x] Ready for merge.


core/sockets.py
```diff
@@ -65,16 +50,15 @@ def process_from_socket(self, context):
...
+
     use_prop: BoolProperty(default=False)
 
-    use_expander: BoolProperty(default=True)
-    use_quicklink: BoolProperty(default=True)
     expanded: BoolProperty(default=False)
     custom_draw: StringProperty(description="For name of method which will draw socket UI (optionally)")
     prop_name: StringProperty(default='', description="For displaying node property in socket UI")
 
-    quicklink_func_name: StringProperty(default="", name="quicklink_func_name")
```
zeffii, line 76: ok.. but this was added because it was useful to allow this operator to spawn not just one Node + attach, but spawn several as a product of some function defined in the node itself. It was intentionally broad.

Durman: We can bring it back any moment but in current branch no one node use this method.

It is possible to write next node now.

```Python
def sv_init(self):
    sock = self.inputs.new('SvVerticesSocket', 'Verts')
    sock.quick_link_to_node = 'bl_idname_of_node_which_should_be_linked'
```

And the socket of particular node instance will have custom quick link.

Durman:
I'm eager to delete this code because sockets should not know about node types. 

https://github.com/nortikin/sverchok/blob/c69efa6624433332d8255d47036cc6e9c0516587/core/sockets.py#L227-L231

Sockets should grant API in general and any (node) which would like to use sockets should use them via API and don't inject their parts of code into sockets classes. 

It looks like I have a solution which can be useful not only for SNL nodes but most likely it will require MK2 version. I'll have a look what can be done.

portnov:
I'd say the next candidate for refactoring should be deprecation of `fullList` method (and replacing it with `repeat_last_for_length`). The problem with `fullList` is that it modifies the list in-place, and as far as I understand wide usage of this method has very big share of responsibility of why do we have to do deepcopy each time when transferring data in sockets. If we get rid of `fullList`, it seems to me (but still is to be confirmed) there will remain only a few places where we modify input data in-place; when we fix those, we will be able to disable deepcopy at last....
well, dreams... :)

Durman:
I'm usually using cycle from itertools to full fill data.
Also if this approach  #3437 will be wider used it won't be the problem.

portnov:
For new nodes yes, but fullList is widely used in existing nodes, especially in list-manipulation nodes...

Durman:
Using node properties in socket and its own looks redundant. I think it would be possible to use only socket properties and custom draw feature sometimes. Most likely why `ptop_name` was created is to ability to add custom update function to property. At the present we could force an update function of socket property to  call node update function if needed with next code:

https://github.com/nortikin/sverchok/blob/979172d84807af9cb37a48621589a8f90c2ac2a4/data_structure.py#L920-L939

This code will let to put extra arguments to update function for example a name of updated property.

```Python
class Socket

    def prop_update(self, context, prop_name=None):
         if hasattr(self.node, update_socket_prop):
             self.node.update_socket_prop(self, context, prop_name):
        else:
            updateNode()

    value: bpy.props.FloatProperty(update=update_with_kwargs(prop_update, 'value'))


class Node:

    def update_socket_prop(self, socket, context, prop_name):
        if prop_name == 'Values':
            ...
        elif prop_name == 'Another prop':
           ...

```

Durman:
API of new socket creating will be next after the PR:

```Python

class Socket:
    color = (1,1,1,1)  # color of new socket (only mandatory attribute)
    quick_link_to_node = 'GenVectorsNode'  # id name of a node which will be connected by quick link 
    default_property: FloatProperty(update=update_from_socket)  # if use_prop is True and socket is input and socket is unconnected this property will be shown
    default_conversion_name = 'FieldImplicitConversionPolicy'  # if sockets need another default conversion

    def draw_proerty(): ...  # if socket property has advance UI like Vector or Color sockets have now
```
In most cases it will be enough.
All properties can be Python type or Blender property. Blender property will support individual values per instance. Python values are like class variables.

It is available now to use default property for Strings socket. It supports float and int types but other types can easily be added.

```Python
class Node:
    def sv_init():
        sock = self.inputs.new('SvStringsSocket', 'Numbers')
        sock.use_prop = True
        sock.default_property_type = 'int'
        scok.default_int_value = 10
```

Draw performance is better now it will be noticeable after 100 - 200 nodes in a tree.

portnov:
@Durman Please check, didn't this break existing blend files? I have a blend file where values of SvVerticesSocket's "built-in" properties (those with `p.use_prop=True`) are gone, but i'm not sure, maybe it is not related...

Durman:
Yes in did, now I realize it. I think I know how this can be fixed. I will open new PR.

PR #3511, opened 2020-09-03
Move out SNL node code from sockets module

Durman:

https://github.com/nortikin/sverchok/pull/3507#issuecomment-686616474
Looks like it won't be trivial task - separate PR. The main idea of solving problem is add socket default properties (float/int) which can be used and by SNL node.

## Preflight checklist

- [x] Ready for merge.


zeffii:
i'm glad you're doing this. it is necessary!

it would be cool to display dynamically created bpy...Properties (maybe in Scene, or somewhere more logical..? ) on the socket layout, so we can fully define the socket ui (min max, soft minmax, step, precision etc..) directly from the scriptnode script. Otherwise we are always limited to properties and their arguments defined in the ScriptNode baseclass,  or using get/set functions which are not exactly user friendly.

Durman:
I think for such properties scene would more appropriate place indeed. The efforts will worth doing if to create abstract manager of such properties. In this case any node could order properties with certain parameters from this manager without knowing anything about how this is implemented. The manager should be responsible for creating, deleting properties, tracing names collisions, serialization its properties. Syntaxis could look like this:

```Python
class DynamicProperties(PropertyGroup):

    def add_property(self, *params):
       setattr(type(self), prams.name, bpy_props[params.type](*params))

bpy.types.Scene.properties_generator = PointerProperty(type=DynamicProperties)


class Socket:
    def draw(layout, *_):
        if self.dynamic_prop:  # or better name
            layout(bpy.context.scene.properties_generator, self.dynamic_prop)

    @dynamic_prop.setter
    def dynamic_prop(self, *props):
        bpy.context.scene.properties_generator.add_property(props)
        self._dynamic_prop = props.name


class Node:
    def sv_init(self, context):
        socket = self.inputs.new('SvStringsSocket', 'My value')
        socket.dynamic_prop = DynamicProperty(name='My value', type=float, default=1, min=0, max=10)
```

Durman:
Looks like it is simpler than I thought. No need in SNL MK2 version. 

Improved an export of sockets a bit. Now it exports all their non default properties.

zeffii:
https://github.com/nortikin/sverchok/pull/3511#issuecomment-687116855  
that works?  wow.  i look forward to adding this stuff to the `snlite directive` ... and pointerproprty for the Text object.

Durman:
In my mind it works. In this modules there are classes which have got more a less similar conceptions and they work.

https://github.com/nortikin/sverchok/blob/08aea87fb0069a07aa5bd533b73bc51644d2dc3f/utils/nodes_mixins/generating_objects.py#L249-L266

PR #3517, opened 2020-09-07
Sverchok N panel refactoring

Durman:

I'm trying make them look in style like standard Blender panels.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.


portnov:
As for profiling, it seems to me, it's better to leave enable/disable toggle in settings: most people do not use it, and so for them this section in N panel is just waste of space and a source of "what is that" questions.

Durman:
If someone toggles to show profiling features in addon properties panel how can he know in which panel to search them next? 

I think it is okay if profiling panel will be collapsed by default. Users at least will be aware of such feature. All Blender properties are shown excluding `experimental` features and it makes it easier to investigate them.

Durman:
Like this I guess. Do not think many user use it but it's in Blender UI.

Durman:
3d panel

nortikin:
love popups

zeffii:
image

no thanks.

Durman:
While refactoring I was trying to stick next pattern which I think is close to Blender style:

- If some features should be hideable use sub panel. 
- Don't use buttons just for showing hints. Hints should be shown on UI element related with it. 
- One element per row. 
- Bool properties should with checkboxes otherwise it will be difficult to discern them from operators.

zeffii:
i'm ok with the tickbox, but i don't get why the text must be centered that way, cutting off the words. it's not necessary.

Durman:
Agree. In this way it should looks a little bit better:



 I think that position of checkbox is something like UI convention in Blender:

Durman:
Final view:

zeffii:
i would prefer you keep the term `Archive .blend (zip/gz)` i want that to be a one button event, i don't want to have to look around in the File handler menu for a "zip" or "gz" option.

Durman:
> i don't want

Good argumentation for your own addon not for Sverchok.

True argumentation I think should answer the question: why this should be useful for other users?

zeffii:
i made it the way i wanted it, because it did not exist before i made it. and now you tell me there is some mythical styleguide in Blender that we should follow? (even if there was.. blindly following it in all cases is "goblins of small minds / kafka" )

it's fine that you are doing a refactoring, but maybe your choices are also just choices.

Durman:
I believe that what I'm doing in this PR will improve work in Sverchok of ~~all~~ most users. To make convenient UI only for one person is not my target.

zeffii:
whatever.  this was fine.
image

nortikin:
> I believe.

nortikin:
personally i don't agree that formula 1 was excluded. I returned it in old nodes and call it from console. And i want it's resurrection as formula_oldfag node.
Benefits are - it interacts between floats/verts verts/verts. I like that behaviour

Durman:
run tests
Hope at least @portnov won't complain that he is used runing all tests by pressing only one button.

portnov:
Huh :) No, I usually run tests from command line. But if there will be selection of tests to run in GUI, I will probably do it from gui from time to time :)

utils/modules_inspection.py
```diff
@@ -32,17 +33,19 @@ def iter_classes_from_module(module, base_types):
...
+    :param depth: 0 means it returns all submodules, if 1 it returns only modules from given directory, if 2 ... so on
     :param root: for internal work
+    :param current_depth: for internal work
     :return: module name - str
     """
     for _, module_name, is_package in pkgutil.iter_modules([str(path)]):
-        if is_package:
-            sub_path = path / module_name
+        if is_package and (depth == 0 or current_depth < depth):
+            sub_path = f'{path} // {module_name}'
```
zeffii, line 47: is path variable there a string or a Path object?

zeffii: why was it ever `path / module_name` ? : )  wow.

Durman: It got path object before but I decided to change it to string.

zeffii: and `//`  will work on all operating systems? or should it be `os.path.join(path, module_name)` to be sure ?

Durman: Did not know. I will think.

Durman:
new 3d panel

Durman:
@portnov Do you know what can this mean?
```
'str' object has no attribute 'wasSuccessful'
The command "bash ./run_tests.sh" exited with 1.
```

zeffii:
nice work @Durman 😎

Durman:
Thanks =D

nortikin:
```bash
ERROR:sverchok.utils.sv_IO_panel_tools:TypeError('Object of type IDPropertyGroup is not JSON serializable')
  File "/home/ololo/.config/blender/2.90/scripts/addons/sverchok/ui/sv_IO_panel.py", line 145, in execute
    write_json(layout_dict, destination_path)
  File "/home/ololo/.config/blender/2.90/scripts/addons/sverchok/utils/sv_IO_panel_tools.py", line 124, in write_json
    node_tree.writelines(m)
UnboundLocalError: local variable 'm' referenced before assignment
```

Durman:
This error does not related with this PR. It does not impacting on exporter or tree or node or socket structures.

PR #3524, opened 2020-09-10
return back 'prop' attribute of vector socket

Durman:

https://github.com/nortikin/sverchok/pull/3507#issuecomment-689675989

### DIFF
--- a/core/sockets.py
+++ b/core/sockets.py
@@ -331,11 +331,19 @@ class SvVerticesSocket(NodeSocket, SvSocketCommon):
 
     default_property: FloatVectorProperty(default=(0, 0, 0), size=3, update=process_from_socket)
 
+    # this property is needed for back capability, after renaming prop to default_property
+    # should be removed after https://github.com/nortikin/sverchok/issues/3514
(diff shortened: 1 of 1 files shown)


Durman:
@portnov Can you check the branch on your old layouts?

portnov:
I tried with the file I attached earlier. In GUI, the node looks like it remembers the correct value, but the node does not see it, it receives (0,0,0) from sv_get().

portnov:
@Durman as far as I understand, the problem is that old implementation stores the property value in the *node* object, while your new implementation works with properties of *sockets*...

Durman:
No. The problem is that I renamed this property. I forgot about sv_get, will fix.

Durman:
@portnov Now should work

portnov:
Yes, works for me.

portnov:
@Durman it seems setting the default value of such property in `sv_init` still does not work :/ Nodes are created with (0,0,0) as default value.

Durman:
I hope that now it will work as expected.

PR #3525, opened 2020-09-10
Pulga physics modular (WIP)

vicdoval:
Annotation 2020-09-06 101658
image

Annotation 2020-09-06 112750
untitled

New Features:
- Force Chain Control
- Use Vector Fields as a Vector Force
- Timed Force (restrict force based on iteration number)
- Align Force (flocking behaviour)
- Pins can be fixed only in some axis
- Angle Force: Resistance to change angles at edges or faces
- Boundaries Force: Now simulation can happen inside a mesh, over a mesh surface, iside a solid, over a solid surface, over a plane....
- Scipy + cython superfast collision system
- Easier upgrade system...


Durman:
@portnov What type of property the socket should contain?

@vicdoval  Can be useful: https://github.com/nortikin/sverchok/pull/3507#issuecomment-687786293

Durman:
The socket should look like this now:

```Python
class SvForceSocket(NodeSocket, SvSocketCommon):
    color = (int, int, int, int)
    default_property: FloatVectorProperty(default=(0, 0, 0), size=3, subtype='DISTANCE', update=process_from_socket)

    use_prop: BoolProperty(default=True)  # just in case if it should be True by default

    def draw_property(self, layout, prop_origin=None, prop_name='default_property'): ...
        # here you can draw what ever you like in socket layout
        # it is not mondatory
```

```Python
    def sv_init(self):
        sock = self.inputs.new('SvForceSocket', 'Name')
        sock.use_prop = True  # if it is not by default true
        scok.defaul_property = (1, 1, 1)  # in case if node should have another default value
```

vicdoval:
I will look into the new refactored sockets classes :) Thanks for the reminder.

About performance... I think is pretty fast for a numpy engine (but i could not find any other) for hundreds of vertices is usually fine, self aware forces (self collision, self attraction and fit force) are the slowest, and if you use 10000 verts things start getting pretty slow afer some iterations...

In this scenario 729 verts and 2765 edges tool 0.55 seconds to run 400 iterations (0.001375 secs per iteration)
image
But if you add a self collision force  jumps to 11 seconds... 0.0275 secs per iteration. I tried scipy kdtree but it was slower..

I'm sure it could be improved but is pretty usable in many cases

vicdoval:
@Durman is it possible to define the description attribute of the default property?

Durman:
Yes, but this description will be shared among all users of the socket.

vicdoval:
So is not possible to do it per socket (as with the regular props)? (i was trying with a `SvStringsSocket`)

Durman:
Property is like class variable only actual values can be different among class instances. 

I had a look at animation nodes and it looks like sockets properties do not have any tooltips. Probably the properties are initialized in sockets.

vicdoval:
Good news for performance: Cython is pip_instalable and cKDtree of scipy is super fast (up to 40x)

portnov:
@vicdoval Did you manage to overcome the necessity of messing with python's `*.h` files to install Cython?

vicdoval:
not any file messing, only clicking on the pip install button and start to use cKDTree (that is 100 times faster than the regular KDTree in the same scenario with 26000 vertices)

nortikin:
> not any file messing, only clicking on the pip install button and start to use cKDTree (that is 100 times faster than the regular KDTree in the same scenario with 26000 vertices)

than adding to extra nodes deps?

vicdoval:
Reminder: Don't trust numpy vectorize (creating my own for loop i got a 2x boost)

portnov:
Yeah, `numpy.vectorize` seems to be just a wrapper that creates a "vectorized" function by making a simple for-loop, it does not actually vectorize any computations. If you need performance, you have to actually write your function so that it processes many numbers at once.

vicdoval:
Also if you have to create a array for every output number (for the vectorize method) is slower that creating the array first and fill it with the values

vicdoval:
- Angles force : preserve angles at edges
- Boundaries: Mesh surface, mesh volume, Solid surface, solid volume and some other strategies to contain the simulation 
- Scipy + Cython collisions and attractions
image

vicdoval:
Has anybody experimented with the `multiprocessing ` module in Sverchok? sounds pretty cool

portnov:
I did not, but I do not have too many hopes for it: Blender clearly expects all python scripts to be "sequential", so I expect a lot of concurrency problems / locks / wtfs. But worth a try.

Durman:
It looks like that I tried. I remember each process was launching new Blender.

onerawartist:
How can i try the pulga nodes? I can't find them in the master branch 
''ps: i know nothing about coding ^^, im just an artist who came here to see and try amazing nodes''
Thank you all for the amazing work 🙏

vicdoval:
@onerawartist this is a work in progress so it is still not merged with the master. If you want to try it in its actual state (some things may change and there is not full documentation) you can get it from this branch https://github.com/vicdoval/sverchok/tree/pulga_physics_2

PR #3532, opened 2020-09-11
fix displaying default property of Vector type

Durman:
https://github.com/nortikin/sverchok/pull/3524#issuecomment-691238262


portnov:
:/
Try creating "Extrude Curve along Vector" node. It is supposed to have "Vector" parameter of (0, 0, 1) by default. Now it's (0,0,0).

Maybe it's simpler to add getter and setter to `prop` so that they refer to `default_property`?

PR #3533, opened 2020-09-12
fix displaying default property of Vector type

Durman:

https://github.com/nortikin/sverchok/pull/3532#issuecomment-691304721


Durman:
@portnov have a look at this, it looks working.

portnov:
Works for me.

PR #3535, opened 2020-09-13
"Quad Grid Sort" node

portnov:

Consider you want to use a quad grid, i.e. a mesh made of quads only, arranged in topology equivalent to subdivided Plane object. If you make such mesh in other way than by use of Sverchok's "Plane" node, for example, by using Blender's standard "Subdivide" or "Extrude" operations, then most probably the order of vertices in such mesh will be near to random. On the other hand, sometimes you need these vertices to be ordered according to grid topology, row by row, column by column. For example, it can be useful for creating Nurbs surfaces with "Build NURBS Surface" node.

## Solution description

Make a new node, "Sort Quad Grid". It takes a mesh (vertices, edges and faces) and outputs sorted vertices. It can output a separate list of vertices for each row, or join all rows into one list.

Implemented with `bmesh`'s `BMLoop` API.

Screenshot_20200913_121853

Screenshot_20200913_124937

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/vector/quad_grid.rst
+++ b/docs/nodes/vector/quad_grid.rst
@@ -1,0 +1,73 @@
+Sort Quad Grid
+==============
+
+Functionality
+-------------
+
+Consider you want to use a quad grid, i.e. a mesh made of quads only, arranged
+in topology equivalent to subdivided Plane object. If you make such mesh in
+other way than by use of Sverchok's "Plane" node, for example, by using
+Blender's standard "Subdivide" or "Extrude" operations, then most probably the
+order of vertices in such mesh will be near to random. On the other hand,
+sometimes you need these vertices to be ordered according to grid topology, row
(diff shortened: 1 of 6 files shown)


Durman:
How the node chouses first polygon and direction for starting sorting?

portnov:
A bit randomly:
* off course, it considers only 4 corners as candidates
* of those, for each corner vertex, we have one "BMLoop" object, which are actually edges connected to this vertex; BMesh API gives us only one of two edges connected to the corner vertex as "BMLoop", because only one of them is considered as "starting" at this vertex, another one is "ending" there.
* from those 4 edges, it tries to select one which is directed along X axis (more precisely, `other_vert.x > corner_vert.x`).
* it can appear that there are no such; then select one which is directed along Y (`other_vert.y > corner_vert.y`).

PR #3544, opened 2020-09-15
Multi cache node

vicdoval:
This node  allows to store numerical lists in different "memory buckets" then it can output one or many buckets at the same time.
The lists will be stored in a string property so they will e preserved when closing the file.
It can be used to bake particles or to store different states and compare them easily without recomputing de data 

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


enzyme69:
Screen Shot 2020-09-29 at 11 42 24 am

A bit weird in my testing I need to unwrap the particle cache data twice and use List Split before UV Connection

enzyme69:
[Uploading svcache_002.blend.zip…]()

vicdoval:
@enzyme69 the start value of the `number range` is set to 0 and there is nothing in the bucket 0 (the frames start at 1) if you set it to 1 everything should work as in the example

PR #3550, opened 2020-09-16
Do not update whole tree upon `show` attribute switching of a node tree

Durman:

fix show tree

- [x] Ready for merge.

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -180,7 +180,6 @@ class SverchCustomTree(NodeTree, SvNodeTreeCommon):
                 node.show_viewport(self.sv_show)
             except AttributeError:
                 pass
-        process_tree(self)
 
     def on_draft_mode_changed(self, context):
         """
--- a/nodes/viz/vd_draw_experimental.py
+++ b/nodes/viz/vd_draw_experimental.py
@@ -637,6 +637,17 @@ class SvVDExperimental(bpy.types.Node, SverchCustomTreeNode):
     def sv_free(self):
         callback_disable(node_id(self))
 
+    def show_viewport(self, is_show: bool):
+        """It should be called by node tree to show/hide objects"""
+        if not self.activate:
+            # just ignore request
+            pass
+        else:
+            if is_show:
+                self.process()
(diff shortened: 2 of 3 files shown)


portnov:
Btw, there is Screen.is_scrubbing property (https://docs.blender.org/api/current/bpy.types.Screen.html). As far as I understand, it was introduced in 2.90 (https://wiki.blender.org/wiki/Reference/Release_Notes/2.90/Python_API). It would be good to not recalculate the tree during such "scrubbing".

Durman:
I did not catch about which scrubbings? We does not use sliders for known reasons.

portnov:
I'm talking about when you're take some parameter in the node and drag it with the mouse to change the value. Currently we update the tree on each small change of the parameter during such drag; if the tree is slow, this looks terrible. It would be ideal to do not try to recalculate on each small change during drag, only not more often than 0.5-1second (should relate to tree recalculation time).

Durman:
I'm not sure that this will hit desirable effect. I guess it will look like it is always lagging.

portnov:
Well, in systems that support asynchronous work, this works as follows:
* on any small change, UI is checking if previous recalculation is currently in progress
* if it is, then interrupt it and post a new recalculation job, and accept next actions from the user (so that UI does not freeze)

As far as I understand, we have no such luxury as possibility to implement asynchronous processing (but maybe we have? may be worth investigation). But maybe we can imitate similar behaviour with some kind of timeouts tricks.

Durman:
Have a look at this, probably you can learn something. I did not experiment with that module.
https://docs.blender.org/api/current/bpy.app.timers.html#use-a-timer-to-react-to-events-in-another-thread

rendetto:
May I suggest if we do:

> ...imitate similar behavior with some kind of timeouts tricks

to have it as an option of switch it on or off, or have a controllable parameter which sets the amount of those timeouts. 
I heavily rely on the "almost realtime" calculated node trees for some of my setups and any forced timeouts will ruin the process.

Thanks

Issue #401, opened 2014-09-10
Viewer Draw MK2


zeffii:
this should not be in my todo thread :)

this node is not intended to replace ViewerDraw, but instead offer a complementary set of useful features. The following table should show the current state

| feature | done? | description |
| --- | --- | --- |
| vertex color | :black_small_square: | allows vertex colors to be configured, not individually |
| vertex size | :black_small_square: | allows range 1 - 10 px |
| edge color | :black_small_square: | allows edge colors to be configured, not individually |
| edge width | :black_small_square: | allows range 0 - 10 px |
| polygon light direction | :black_small_square: | faux light source uses rate of incident to normal to brighten the polygons |
| show edges | :black_small_square: | allows to hide edges, not individually |
| show vertices | :black_small_square: | same deal |
| show polygons | :black_small_square: | same deal |
| bake | :black_small_square: | shall allow vertex only baking (minor modifications to svBakeObj) |
| preprocess |  | test if `matrix*vector` ahead of time is faster.. |

I have time today to experiment with the existing draw code, i want less if statements and function calls in tight loops.

You are welcome to fork and experiment independently, for a while it might be messy :)

PR #3552, opened 2020-09-16
prestub of blf viewer

zeffii:
This addresses a long standing feature vacuum with respect to allowing users to display arbitrary size text, orientation, and alignment onto their 3dview drawing. Font selection is not a priority.

- size (per word), and a global scaler
- alignment (this feature is not built into blf, it must be calculated using blf's "optical length or string" feature
     - each strings width and height are known so usual text anchor points can be implemented
     - push away offset (takes into account anchor type..)
     - general x,y offset (offset ontop of pushaway)
- blf has a screenspace rotation feature, which may be useful for axis/graphing labels
- color per word
- background per word (and maybe automatic feature to generate a companion colour to ensure maximal contrast? )

A lot of this node will look similar to idxviewer, but subtle difference etc will be more convenient to place in their own node. There will be some acceptable duplication.


Durman:
If you wish this node react on changes of show attribute of a tree owner you should implement extra method `show_viewport` in the node.
Example here 3550

nodes/viz/viewer_blf_text.py
```diff
@@ -0,0 +1,375 @@
...
+    bl_icon = 'INFO'
+    sv_icon = 'SV_INDEX_VIEWER'
+
+    def make_color_prop(name, col):
+        return FloatVectorProperty(
+            name=name, description='', size=4, min=0.0, max=1.0,
+            default=col, subtype='COLOR', update=updateNode)
+
+    n_id: StringProperty(default='', options={'SKIP_SAVE'})
+
```
Durman, line 190: redundant

nodes/viz/viewer_blf_text.py
```diff
@@ -0,0 +1,375 @@
...
+            'args': (geom, config)} 
+
+        callback_enable(n_id, draw_data, overlay='POST_PIXEL')
+
+    def sv_free(self):
+        callback_disable(node_id(self))
+
+    def sv_copy(self, node):
+        ''' reset n_id on copy '''
+        self.n_id = ''
```
Durman, line 357: redundant

nodes/viz/viewer_blf_text.py
```diff
@@ -0,0 +1,370 @@
...
+        it will crash blender on undo
+        """
+
+        return {
+            'background_color': self.background_color[:],
+            'font_text_color': self.font_text_color[:],
+            'draw_background': self.draw_background,
+            'anchor': self.anchor_direction,
+            'scale': self.text_node_scale
+        }.copy()
```
Durman, line 250: Did not catch sense of `copy` here

zeffii: we established early on that passing references from a node into a callback made things unstable, sporadically. i'm not prepared to drop the copy. worry about bigger things.

Durman: Sound like magic. Magic is not good in coding.

zeffii: if it isn't needed it can of course go. the docstring explains why it's there (or why it was initially added to such callbacks). it may not be needed anymore, but you might want to dig through the issue tracker to see when it was first added and why and by who.

not the focus of this PR.

Durman: You probably know that copy only create copy of a dictionary and not its contents. Your dictionary is initialized in the method. So every time the method will be called the new dictionary will be returned. 

I know if a programmer treats code as magic it leads to unpredictable results.
https://docs.python.org/3/library/stdtypes.html#dict.copy

zeffii: thank you for your insight.

zeffii: i'll remove `copy`, and also other similar instances in other viewers. This hasn't been discussed since 2014

https://github.com/nortikin/sverchok/issues/401#issuecomment-55401403

Issue #3553, opened 2020-09-16
export to json doesn't work


nortikin:
After some new commits there is error wile jsoning:
```bash
ERROR:sverchok.ui.sv_IO_panel:Object of type IDPropertyGroup is not JSON serializable
Traceback (most recent call last):
  File "/home/ololo/.config/blender/2.90/scripts/addons/sverchok/ui/sv_IO_panel.py", line 285, in execute
    gist_body = json.dumps(layout_dict, sort_keys=True, indent=2)
  File "/home/ololo/Downloads/INSTALL/blender-2.90.0-linux64/2.90/python/lib/python3.7/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/home/ololo/Downloads/INSTALL/blender-2.90.0-linux64/2.90/python/lib/python3.7/json/encoder.py", line 201, in encode
    chunks = list(chunks)
  File "/home/ololo/Downloads/INSTALL/blender-2.90.0-linux64/2.90/python/lib/python3.7/json/encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/home/ololo/Downloads/INSTALL/blender-2.90.0-linux64/2.90/python/lib/python3.7/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/home/ololo/Downloads/INSTALL/blender-2.90.0-linux64/2.90/python/lib/python3.7/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/home/ololo/Downloads/INSTALL/blender-2.90.0-linux64/2.90/python/lib/python3.7/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  [Previous line repeated 1 more time]
  File "/home/ololo/Downloads/INSTALL/blender-2.90.0-linux64/2.90/python/lib/python3.7/json/encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "/home/ololo/Downloads/INSTALL/blender-2.90.0-linux64/2.90/python/lib/python3.7/json/encoder.py", line 438, in _iterencode
    o = _default(o)
  File "/home/ololo/Downloads/INSTALL/blender-2.90.0-linux64/2.90/python/lib/python3.7/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type IDPropertyGroup is not JSON serializable
[...]

PR #3557, opened 2020-09-17
add fixes for iojson

zeffii:
adds storage functions and skiplist

### DIFF
--- a/nodes/network/file_path.py
+++ b/nodes/network/file_path.py
@@ -71,6 +71,8 @@ class SvFilePathNode(bpy.types.Node, SverchCustomTreeNode):
         subtype='DIR_PATH',
         update=updateNode)
 
+    properties_to_skip_iojson = ['files'] 
+
     def sv_init(self, context):
 
         self.outputs.new('SvFilePathSocket', "File Path")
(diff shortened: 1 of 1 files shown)


zeffii:
fixes #3553

nodes/network/file_path.py
```diff
@@ -111,6 +116,27 @@ def process(self):
...
+        self.id_data.unfreeze(hard=True)
+
+    def storage_get_data(self, node_dict):
+        '''function to set data for exporting json''' 
+
+        local_storage = {
+            'filenames': [{"name": file_elem.name} for file_elem in self.files], 
+            'directory': self.directory
+        }
+        node_dict['string_storage'] = json.dumps(local_storage)
```
Durman, line 139: You love doing things over complicated, why don't just do this:
```Python
node_dict['string_storage'] = local_storage
```

zeffii: > You love doing things over complicated

that's "projection" @Durman .   But i did ` = json.dumps(local_storage)`  and not ` = local_storage` , because i referenced older _working_ implementations. At some point in the past i must have had a reason to use that.

PR #3560, opened 2020-09-18
Import export json refactoring

### LINKED ISSUE #3542: JSON serialization (creating new or refactoring existing implementation)
Will see if it is possible to make it in more OOP way.

```Python
class A:
    def __init__(self, v):
        self.v = v
    def to_json(self):
        return {type(self).__name__: {'v': self.v, }}

def json_encoder(o):
    try:
        return o.to_json()
    except AttributeError:
        raise TypeError
```

```Python
json.dumps([1,2,3,A(1)], default=json_encoder)
'[1, 2, 3, {"A": {"v": 1}}]'
```

Durman:

Closes #3542

- [x] Import/export refactoring
- [x] add import/exprt report
- [x] unit-tests refactoring
- [x] new unit-test: export - import each availabel node ?
_
- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.


Durman:
I think the branch is ready for manual tests if someone interested. It has completely new code so I expect that some import export cases was not covered. Probably importing existing presets and json files should be tested at first place. In debug mode it is possible to get more information about importing.

Durman:
After the PR merging be aware of that: 

- Standard json export saves all properties including collection and nesting collection and pointers. 
- For now only data block name of pointer property will be saved. 
- During import pointers will be searched in current scene, if there is no data blocks with current name nothing will be assigned to the pointer. 
- It is strongly not recommended to save pointers for viewer nodes. For skipping property to save use: `BoolProperty(options={'SKIP_SAVE'}` it will impact only on unsaving property into json file. 
- It is possible to add `save_to_json(node_data: dict)` and `load_from_json(node_data: dict, import_version: float)` method to a node for add extra logic into import export, but I would not recommended add them to new nodes.
- Improvements for next PRs is to support saving pointer data, after the step `save_to_json` and `load_from_json` will be considered like depricated.

Issue #3555, opened 2020-09-17
Proposal to standardise the nesting level of strings


Moult:
I see that most nodes return this pattern `[[n]]`, where `n` is an integer, a list of three coordinates for a vertex, or a list of two indices for an edge. I also see a lot of this pattern, which makes sense due to the nested list:

```
x_out = []
for a, b, c in zip_long_repeat(...):
    ...
    x_out.append()
```

However, then I see some examples like the `File Path` node which gives `['foo bar']`. Shouldn't this be `[['foo bar']]`?

Another example is the `Note` node, which gives `[[['foo'], ['bar']]]`. Shouldn't this also be `[['foo bar']]`, and then another node should split by whitespace, which can then convert it into `[['foo', 'bar']]`?

Durman:
#2761

Moult:
@Durman thanks - but #2761 talks about values, edges, vertices, and faces ... but it does not talk about strings. My question is about the inconsistency in strings. I believe it should be the same as values.

zeffii:
note node and file path node are exceptions , and we have currently no standardized protocol for passing strings around.

zeffii:
a case could be make for a variety of approaches.

sentences
```python
[
   [
      "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore.",
      "et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip"
   ]
]
```
and words..
```python
[
   ["Lorem", "ipsum", "dolor", "sit", "amet,", "consectetur", "adipiscing", "elit,", "sed", "do", "eiusmod", "tempor", "incididunt", "ut", "labore."],
   ["et", "dolore", "magna", "aliqua.", "Ut", "enim", "ad", "minim", "veniam,", "quis", "nostrud", "exercitation", "ullamco", "laboris", "nisi", "ut", "aliquip"]
]
```

Moult:
@zeffii I am working on generating Sverchok nodes for IFC data and Ladybug tools (Ladybug - note the Ladybug node currently in Sverchok is not the full Ladybug, which I am working on, plus Honeybee and Dragonfly). To do this, standardising the protocol for passing strings would be extremely useful. There will be hundreds of nodes generated at the end of this.

[...]

PR #3565, opened 2020-09-20
Standardise text nesting

Moult:
Referring to #3555 - after playing around with ported Grasshopper nodes, I think that strings should use the same convention as numbers. E.g., they should nest like this: `[['foo bar baz']]`. I have tested this convention with the ladybug nodes I am working on, and it works extremely well :)

Currently, the note node splits text like this: `[[['foo', 'bar', 'baz']]]` and the filepath node doesn't nest, like this: `['foo bar baz']`. This PR standardises both nodes to output `[['foo bar baz']]`.

If this PR is approved, I can build a new node that does text `split()`, in case people prefer the splitted text.

### DIFF
--- a/nodes/network/file_path.py
+++ b/nodes/network/file_path.py
@@ -114,7 +114,7 @@ class SvFilePathNode(bpy.types.Node, SverchCustomTreeNode):
         for file_elem in self.files:
             filepath = os.path.join(directory, file_elem.name)
             files.append(filepath)
-        self.outputs['File Path'].sv_set(files)
+        self.outputs['File Path'].sv_set([files])
 
     # iojson stuff
 
--- a/nodes/solid/export_solid.py
+++ b/nodes/solid/export_solid.py
@@ -23,7 +23,7 @@ else:
 
             if not all(s.is_linked for s in node.inputs):
                 return {'FINISHED'}
-            folder_path = node.inputs[0].sv_get()[0]
+            folder_path = node.inputs[0].sv_get()[0][0]
             solids = node.inputs[1].sv_get()
             base_name = node.base_name
             if not base_name:
--- a/nodes/solid/import_solid.py
+++ b/nodes/solid/import_solid.py
@@ -27,7 +27,7 @@ else:
             if not any(socket.is_linked for socket in self.outputs):
                 return
 
-            files = self.inputs[0].sv_get()
+            files = self.inputs[0].sv_get()[0]
 
             solids = []
             for f in files:
--- a/nodes/svg/svg_document.py
+++ b/nodes/svg/svg_document.py
@@ -69,7 +69,7 @@ class SvSvgServer(bpy.types.Operator):
 
             return {'FINISHED'}
 
-        save_path = node.inputs[0].sv_get()[0]
+        save_path = node.inputs[0].sv_get()[0][0]
         file_name = node.file_name
         bpy.ops.node.svg_write(idtree=self.idtree, idname=self.idname)
         spawn_server(save_path, file_name)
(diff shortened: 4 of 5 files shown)


vicdoval:
Do you really need the File Path node to output `[[file_path]]` ?
Although it is a string it is very rare a case where you need a list of lists of file paths like
 `[[file_path, file_path,..], [file_path,..]]`

When I coded the File Path node I thought about how usually the data is paired and file_path per list of numbers seems more common that  file_path per number…

Also outputting [[file_path]] will need at least `SVG Document` Node to be modified and also it may break other existing layouts so it would need or a Wrap option or a mk2 version

zeffii:
these are breaking changes, that can be easily resolve by either 
1. making new versions of the Node
2. adding a toggle on each node, to output current mode vs new mode.

i can imagine some people not being too happy about this as it is currently.

Durman:
> adding a toggle on each node, to output current mode vs new mode.

I'm against it. It will make it unnecessary complex.

It is possible make `file_path` produce such output: [['file1'], ['file2'], ['file3'], ...]
And `SVG Document` should be able to handle such shape without problems.

Moult:
Looking at the bigger picture, I see benefits of standardisation of nesting for strings, similar to what you expect for other data types.

I am OK with @vicdoval and @Durman 's suggestion of `[['file1'],['file2'],...]`. Happy to modify the PR for this.

I think adding new nodes for string split and join can provide a more flexible output for those who want other things, whilst keeping the defaults more consistent.

vicdoval:
What is the advantage of [[file1], [file2]] vs [file1, file2] that justifies the extra [] and the nodes changes?

Moult:
@vicdoval the only benefit is for standardisation, as other strings are more useful in the form `[['foo']]`. Nodes which expect numeric input have consistency, and it would be good for strings to have that too. My personal preference is for  `[['file1','file2']]`, as I do see use for looping through sets of files. I think it is less rare once there are nodes for Ladybug and IFC - where it is very valid to have lists of file paths (IFC model document associations, material external definition associations...)

vicdoval:
Ok, then the `Import Solid `node, `Export Solid ` node and `SVG Document` node need to get adapted to this change. 
Also and in case some is using their own Scripted Node Lite Scripts with a file_path socket a new version of the node would be needed.
Creating a new version means:
- moving old file to old_nodes folder
- adding `mk2 `to the class name, the bl_idname, the .py file and  the .rst file of the new version and update `network_index.rst` (in case of the SvFilePathNode)
- adding to the old node `replacement_nodes = [('SvFilePathNodeMk2', None, None)]` to activate the replacement operator

Durman:
@vicdoval Why new version of the nodes is needed?

vicdoval:
@Durman if you have a nodetree using a Scripted Node with a file_path input socket (which is possible) it will break with the nesting change, not a very big deal for me but maybe for others...

Durman:
I quite sure that nobody from users did not used such socket in script node yet. I can't even imagine usecase of it.

vicdoval:
Well I guess if someone has used the scripted node that way they will be able to change it to the new behaviour.

Then it would be enough with changing the default nodes mentioned before

Moult:
So, just to confirm, the commit I've done already is OK, but in addition, I need to modify Import Solid node, Export Solid node and SVG Document node? Should I create a new version of those three nodes, or just modify them directly?

Durman:
You can modify them without creating new version. They should be able to handle new shape of data without errors I guess.

Moult:
@Durman cheers - updated the nodes. Let me know if you think it's ready to merge :)

Durman:
`File path` and `SVG document` are woking correctly. I have not dependencies to check `import and export soldis` nodes. Probably someone else interested can check them.

vicdoval:
`File Path` is working nicely with `Import solid` and `Export solid` so I guess if this PR is ready to be merged (unless anyone has anything against it)

nortikin:
there is also index viewer, using [['a','b','c','d','e']] for index subscribtion

PR #3572, opened 2020-09-22
Add a button to open browser with PyDoc

portnov:
* visible under "Common Utils" in the N panel, only if "Developer mode" is enabled in addon preferences
* Runs pydoc's server at localhost:8080 and opens web browser at this url; if pydoc is already running, just opens the browser.

### DIFF
--- a/ui/development.py
+++ b/ui/development.py
@@ -20,6 +20,7 @@ import os
 from os.path import exists, isfile
 import subprocess
 import webbrowser
+import socket
 
 import bpy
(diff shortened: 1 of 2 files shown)


nortikin:
than we should make some text and import bpy?

portnov:
Currently pydoc doesn't show help for `bpy` and other blender's internal modules, idk why. I did this mostly for documentation on FreeCAD API, but documentation for other modules you install as dependencies (geomdl, scipy, etc) is available too.

nortikin:
thanks, it helps.

Issue #3514, opened 2020-09-06
Problem of creating new node version if socket name was changed


Durman:
The problem is that some minor changes in node code force us to create its new version. It's totally related with sockets. It is possible to fix node property without problems meanwhile one can't say this about sockets. There is hint of a solution in animation nodes. 

This articles describes the problem: https://wiki.blender.org/wiki/Source/Nodes/NodeInterfaceFramework#Existing_Approaches

One of possible approaches is re initialize node on each tree update event as article describes. And we already have one node which share similar conception. It's SNL node.

https://github.com/nortikin/sverchok/blob/979172d84807af9cb37a48621589a8f90c2ac2a4/nodes/generator/script1_lite.py#L328-L334

This method is calling each time when `process` method of the node is called and it updates sockets state. It is why my last changes in initialization sockets code does not required creating new version of the node. 

We could think whether it is possible to bring this logic to the reast of the nodes.

zeffii:
> This method is calling each time when process method of the node is called and it updates sockets state

just to be a bit more exact in language:  the `make_new_locals` function only makes changes to the node if it must, but it wil always inject locals to be available for the `execution script` (the thing that snlite loads from TextEditor)

Durman:
As far as I understand it calls `update sockets` each time when new file is opened. It's enough for renaming sockets for example without necessity of creating new version of a node.

zeffii:
yes, but the point is that new sockets are not created every time process is called.

zeffii:
there are a limited set of scenarios where snilte updates sockets 
- opening the blend, 
- python reload f8, or 
- node script update from TextEditor
- (maybe) copying node

all other times process is called, without invoking "update_sockets".      i think :)

Durman:
Probably it's enough for renaming socket without breaking capability with old files.

Durman:
The main problem I see is renaming sockets. Since a socket is renamed it is loosing its relationships with its old versions.

First what comes to mind is to create attribute like `sv_identifier` but in this case `sv_initi` methodes of all nodes should be rewrite and in old files sockets won't be able to have this attribute. 

So another solution can be create special attribute for names and use socket names as identifiers instead. The attribute can have name like `draw_name`. In this case sockets of old files can be renamed just assigning new value to `draw_name` attributes.

Durman:
[...]

PR #3578, opened 2020-09-23
add socket labels

### LINKED ISSUE #3577: number range - wrong label after connecting a socket
this happens in several modes, f.ex  Float / Count mode, has a Stop slider, but it's displayed as "Step" once it's connected.

count_range_stop_step

Durman:

Close #3577

#3514

socket labels

### DIFF
--- a/core/sockets.py
+++ b/core/sockets.py
@@ -188,7 +188,7 @@ class SvSocketCommon:
                 getattr(node, self.custom_draw)(self, context, layout)
 
         elif self.is_linked:  # linked INPUT or OUTPUT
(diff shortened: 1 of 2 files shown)


Durman:
After this PR it will be possible to use `label` socket field. 

It will be drawn instead of real socket name if given. It can used for renaming sockets of existing nodes without creating new versions of them in new layouts. If label is set in `sv_init` method sockets will be displayed with socket name in old layouts because label will be empty.

Another usage creating dynamic socket which can have role of several parameters like the node in this PR. It can switch properties and its name but not type.

PR #3580, opened 2020-09-23
Solid Boolean node: upgrade

portnov:
* Support input nesting level 1 or 2.
* add ability to calculate which edge and face of the resulting Solid came from which source object(s). This is done via `generalFuse` method.

Screenshot_20200923_210004

Screenshot_20200923_220931

The following illustrates how **EdgeSources** output is calculated:

Screenshot_20200923_213528

Here we have two cubes, 0 (plugged into Solid A input), and 1 (plugged into
Solid B input). Purple edges came from cube 0, for them EdgeSources output
contains ``[0]``. Orange edges came from cube 1, for them EdgeSources output
contains ``[1]``. Edges marked with cyan came from both cubes, for them
EdgeSources output contains ``[0, 1]``.

FaceSources output is calculated similarly, but for faces instead of edges.

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

@vicdoval you might want to make a review.

### DIFF
--- a/docs/nodes/solid/solid_boolean.rst
+++ b/docs/nodes/solid/solid_boolean.rst
@@ -1,29 +1,120 @@
 Solid Boolean
 =============
 
+Dependencies
+------------
+
+This node requires FreeCAD_ library to work.
+
+.. _FreeCAD: ../../solids.rst
+
 Functionality
 -------------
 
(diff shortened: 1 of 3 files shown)


vicdoval:
Looks awesome

PR #3582, opened 2020-09-24
Sverchok-Extra nodes in search menu

vicdoval:
Improved #3359
Add sverchok-extra and other sv-extra addons to the search menu
- [x] Ready for merge.


vicdoval:
@Moult Now you nodes should appear in the search menu, if they don't please report.

PR #3585, opened 2020-09-24
fix id prop name, drop punctuation

### LINKED ISSUE #3583: Sverchok Not working anymore with Blender daily build 2.91.0
Sverchok Not working anymore with Blender daily build 2.91.0

image

zeffii:
IDProps can now no longer include punctuation.

fixes #3583

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -105,13 +105,13 @@ class SvNodeTreeCommon(object):
     def freeze(self, hard=False):
         """Temporary prevent tree from updating nodes"""
         if hard:
-            self["don't update"] = 1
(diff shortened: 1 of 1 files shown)


portnov:
I've run this on 2.90, Line node works as expected :) What else to check? Export/import?

portnov:
Export/import seems to be working too..

Durman:
Import/export does not deal with IDProps

zeffii:
only nodes that call Freeze/ unfreeze i think..

Durman:
I'm having fun with import/export about a week and didn't see using ID properties yet.

PR #3589, opened 2020-09-25
Update lesson_01.rst

JitQL:

What I did was to clarify some trigonometric concepts for people that might be rusty at them.

## Solution description

A note clarifying why we are using 0.5PI for dividing the circle in 4 vertexes.
A phrase clarifying why we are rotating the square.


JitQL:
That hurray is cool @Durman . Thanks for your help.

Durman:
@JitQL Thanks.

PR #3590, opened 2020-09-25
Update nesting.rst

JitQL:
These changes are, more or less, a conclusion of what's been learned before, that I think help understand the info on the page. I'm not absolutely sure they are right, but I do think we need them as an introduction to the comprehensive info that comes next. It helps get the overall picture without exhaustive info.

It's really fundamental knowledge and one tends to get lost on these concepts here, as well as on lesson 02.

I think the need for having 3 brackets as input for viz nodes in lesson 02 needs better explanation and that explanation can be either there on Lesson 02 or here in this page as the final example is a specific example of a single face, and I found it hard to understand at first without comparing it with a regular example side by side (I might still haven't got it right though). I will deal with that later.

Please describe what problem does this PR solve. If it was already was discussed in an issue, it is enough to give a link as #number.
If this adds a new feature (a new node, for example), please state why (for what cases) do you think it is needed.

## Solution description

Please describe how do you intend to solve the problem: change some method, or introduce new class, or whatever.

## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/docs/nesting.rst
+++ b/docs/nesting.rst
@@ -7,15 +7,17 @@ Sverchok sockets can contain the elements of any number of objects. The number o
 
 Each collection is a list of objects and these objects are themselves lists down to the most basic level which is the list of coordinates of each vertex.
 
-So a vertex is a list of coordinates (X,Y,Z): 
+A vertex is also a list of coordinates (X,Y,Z): 
 
 ``v0 = (1.0, 3.8, 2.5) v1 = (0.0, 0.0,0.0) v2 = (2.1, 3.1, 4.1)`` 
 
-A list of vertexes is made of several coordinates: 
+A list of vertexes is made of several lists of coordinates per vertex: 
 
 ``vertices = [(1.0, 3.8, 2.5), (0.0,0.0,0.0), (2.1,3.1,4.1)]``
 
-And each list element has an index starting by "0" and ending in "n", from left to right. So, in the above vertices list the indices for each vertex are:
+Each list element has an index starting by "0" and ending in "n", from left to right.
+
+In the above vertices list the indices for each vertex are:
 
 ``v0 = 0, v1 = 1, v2 = 2``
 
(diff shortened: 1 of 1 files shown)


docs/nesting.rst
```diff
@@ -3,18 +3,49 @@ Nesting
...
+
+``edges_list = [(0, 1), (1, 2), (2, 3)]``
+
+So basically, sverchok works by using nodes to creat lists of values, then using other nodes to evaluate those lists and generate new lists. Eventually the results of the final list will be used to generate the final output that will be passed to the viewport with a viewer node.
+
+This is why the nesting concept is fundamental in order to use Sverchok.
+
+Let's look at some examples, (not necessarily outputting just one thing): 
+
+NOTE to zeffii, what do you mean with this: (not necessarily outputting just one thing)?
```
zeffii, line 36: num things | descriptor
----|----
"one thing" | would be a socket that outputs a single collection, so `[[item1, item2, item3]]`
"two things" | would be a socket that outputs two collections `[[item1, item2, item3], [item1, item2, item3]]`

zeffii:
these are fine, both PR as OK.  

My only concern is the use of the words "so" (at the start of any sentence) and "basically" (anywhere), drop those and we're golden.  

https://wynnnetherland.com/linked/2014010902/words-to-avoid-in-tech-writing

JitQL:
That is a very useful comment and advice. I think it applies to much more than tech and learning. I will remove the words and try to think on that advice in other contexts too. Thanks!

I understood the explanation sbout one or two things too.

zeffii:
I like to stick information in a table when it has a chance to remove the complexity of information being covered. But hey, the more examples people read the more they will pick up on patterns, even if we fail in our explanation of individual parts/examples. When people recognize the patterns, that's when they've learned what we're trying to get across.

keep going!

docs/nesting.rst
zeffii: fair comment btw, if that line didn't make sense on the first reading. The comment in parens could be dropped, i don't think it was helpful the way i phrased it originally.

PR #3609, opened 2020-10-03
Join and apply matrix refactoring

Durman:

There are a lot of places of applying matrixes and joining meshes in code. This PR is intended to add one module responsible for all this operations. This module will be able to deal as with Py mesh and with Numy mesh. Highlight of this module is that you even should not think about with what kind of meshes you are dealing with.

```Python
import meshes as me

mesh = me.to_mesh(vertices, edges, faces)  # create mesh Numpy or Python
mesh.apply_matrix(matrix)
mesh2 = me.to_mesh(vertices2, edges2, faces2)
mesh = mesh.add_mesh(mesh2)  # join meshes
```

Type of mesh will be dependent on which type has vertices. If it is list it will convert data to `PyMesh` if array then to `NpMesh`.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.


portnov:
Some time ago I tried to write a "simple mesh class", expecting that it would be faster than bmesh because it was much simpler. But in fact it appeared much slower. So performance worth consideration too, among other things.

Durman:
Yes I understand what do you mean. But it is slower if creates separate classes for vertices, edges, faces. I'm creating one class for mesh in general. Methods of this class has standard code which we usually use for manipulating with mesh. So it is not slower than code which we already have.

The idea is using factory method.
https://refactoring.guru/design-patterns/factory-method
It get data and convert it into mesh relevantly on its type. Independently which type of mesh did you get you can use certain methods which are supported by all mesh types. So there is no need even to know which type of mesh you get (numpy or python)

Durman:
I was using extend list method for merging meshes but it can change source data which has come from other socket. It looks that generating new list is not too slower and can be used instead of extend method.

```Python
timeit('l + l + l + l', 'l = list(range(100000))', number=1)
0.012739089987007901

timeit('l = l + l; l = l + l; l = l + l', 'l = list(range(100000))', number=1)
0.012937668012455106

timeit('l.extend(l); l.extend(l), l.extend(l)', 'l = list(range(100000)); ll = []', number=1)
0.00971808799658902
```

Also I'm trying implement simple way of handling mesh attributes as well. Probably next syntax can be used

```Python
mesh = PyMesh(verts, edges, polygons)
mesh.polygons.attrs['materials'] = materials

mesh2 = PyMesh(verts, edges, polygons)
mesh2.polygons.attrs['materials'] = materials2

merged_mesh = mesh.add_mesh(mesh2)
merged_materials = merged_mesh.polygons.attrs['materials']
```

Issue #3618, opened 2020-10-06
"Standardise text nesting" #3565 broke a lot of my setups


rendetto:
So lots of my files are broken now. 
Those use lots of strings with "Viewer Index +", "Note",  "Switcher" etc.
What can I do after those changes to fix them back on? 
Thanks

vicdoval:
By the moment if you use `List Del Levels` after the Note  node and delete the level 1 everyrthing should be as  before. About the switcher and the viewer index i guess something has to be changed in the code...

Moult:
I think `List Del Levels` does the opposite of what you're looking for. I think you need split text. See PR #3619.

rendetto:
Yes, adding the "Text Split" node solved the problem. Waiting for it to be merged in "master" soon.
Thanks

rendetto:
Just found this setup on https://vk.com/sverchok_b3d, which utilizes split() with "Formula+" node. It solved the problem I had, so I'm closing this issue.

![TextSplit_with_Formula](https://user-images.githubusercontent.com/66558924/103487615-63c0fc80-4e0f-11eb-9ad2-23b97bfe7c46.jpg)

Great thanks to the person who did this setup :)

PR #3619, opened 2020-10-07
Add text split node

Moult:

In #3565 it was identified that text nesting was inconsistent, and that the Note node actually implicitly split text. As part of the proposal to standardise text nesting, I also suggested that for those who preferred the old behaviour of split text, a new split node would be added. As seen in #3618 - users need this split node :) so here it is!

## Solution description

In the old Note node, if you typed `'foo bar'`, it would come out as `[[['foo', 'bar']]]`. Now it comes out as `[['foo bar']]`. If you then apply this new explicit split mode, it will split (by whitespace, but maybe we can add a configurable delimiter in the future?) into the original behaviour.

## Preflight checklist

- [x] Code changes complete.
- [x] Code documentation complete. (Not sure what to document, quite straightforward)
- [ ] Documentation for users complete (Will add if people are happy with behaviour).
- [x] Manual testing done. 
- [ ] Unit-tests implemented. (What's the process for this?)
- [ ] Ready for merge.

### DIFF
--- a/index.md
+++ b/index.md
@@ -569,6 +569,7 @@
     SvTextOutNodeMK2
     ---
     NoteNode
+    TextSplitNode
     SvGTextNode
 
 ## BPY Data
--- a/nodes/text/split.py
+++ b/nodes/text/split.py
@@ -1,0 +1,53 @@
+# ##### BEGIN GPL LICENSE BLOCK #####
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 2
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software Foundation,
+#  Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# ##### END GPL LICENSE BLOCK #####
+
+import bpy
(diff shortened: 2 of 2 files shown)


Moult:
image

Durman:
What about just adding option to the `Note` node which will switch between old and new string formants?
For me what should do `text split` node is a little bit unobvious.

Moult:
The advantage is an explicit split node can be used on any text input, not just note.

Split effectively does `.split()` in Python. When parsing external datasets this is quite useful.

nortikin:
maybe make string manipulation nodes stack? `join`/`split`/`to_string`/`evaluate_from_string` ...
+ `list pop node` needed also for lists.

vicdoval:
I would be nice if you could set the separation character, space, coma, user defined...

Moult:
@nortikin agree the full stack is needed. I will do in a separate PR, as this one has a higher priority since users are waiting for it.

@vicdoval yes, agree.

Can somebody guide me if there is anything I should be aware of on creating a unit test for this, as well as how nodes are documented? If people are happy with the implementation I'll add test / documentation (and add the delimiter feature too).

portnov:
We do not have a very good technology for testing *nodes*. What we do have is something like https://github.com/nortikin/sverchok/blob/master/tests/simple_generator_tests.py. It looks a bit heavy: the test will programmatically create the node tree and so on. I think, general idea should be that all "interesting" functionality should be enclosed in separate functions / modules, and tests should test those functions, not nodes. There will be some uncovered functionality that remained in the node class, but that should be more or less "trivial" (like data matching).

rendetto:
Umm... what is the state of this node. Will it be in master any time soon? For me it works without problems, and it is essential node in SV.

vicdoval:
@rendetto for the moment you can use this
image

rendetto:
Thank you @vicdoval !

zeffii:
i think this can be closed, the `String Tools` node encapsulates this behaviour.

image

Issue #3319, opened 2020-06-05
Node groups instead of monads


Durman:
Monads are very complex and difficult to support. They does not use dedicated for creating node groups Blender tools. Last Sorcar updates proves that this tools can be used.

![2020-06-05_08-31-13](https://user-images.githubusercontent.com/28003269/83837233-ee6d3c00-a706-11ea-931e-caa6123c96da.png)

Here is the code of `node group` [node](https://github.com/aachman98/Sorcar/blob/master/nodes/utilities/ScNodeGroup.py).
And here is the code of creating node group [operator](https://github.com/aachman98/Sorcar/blob/master/operators/ScGroupNodes.py). It is nearly all code I guess which is needed for handling node groups.
How had seen the code of monads know that this code is much simpler.

What I don't like in current implementation in Sorcar is that sockets in node group node do not have parameters. They should be plugged to other nodes for managing node group node.

@aachman98 Did not you try to use this class [NodeSocketInterface](https://docs.blender.org/api/current/bpy.types.NodeSocketInterface.html) for creating sockets of node group trees? Probably it could solve the problem.

I Had tried it but faced with Blender crash. Here is [bug report](https://developer.blender.org/T76591) but it has low priority and probably will be fixing for years.

zeffii:
i welcome a simpler implementation of node groups in Sverchok. It _must_ be possible.

aachman98:
> Here is the code of `node group` [node](https://github.com/aachman98/Sorcar/blob/master/nodes/utilities/ScNodeGroup.py).
> And here is the code of creating node group [operator](https://github.com/aachman98/Sorcar/blob/master/operators/ScGroupNodes.py). It is nearly all code I guess which is needed for handling node groups.

The current method is incomplete, and is somewhat unstable (frankly the whole of Sorcar is, at the moment xD). It relies on copying selected nodes to the new group and making the connections with the default `NodeGroupInput` & `NodeGroupOutput` classes. They handle the interface connections internally, and I can't seem to figure out how to add custom classes inherited from `NodeSocketInterface` (see more info below)


> What I don't like in current implementation in Sorcar is that sockets in node group node do not have parameters. They should be plugged to other nodes for managing node group node.

[...]

PR #3624, opened 2020-10-08
Node groups UI

Durman:

#3319
In this PR I'm going to add `Group node` and all functionality responsible for drawing its user interface. Adding group nodes is too big functionality I would be able to add it in one PR.

## Preflight checklist

- [x] Add group menu
- [x] Add separate trees for groups
- [x] Handling sockets creation
- [x] Add tree description
- [x] Add `group in` and `group out` nodes into menu
- [x] Opertaor for creating group from selected nodes
- [x] Operator for ungrouping
- [x] Add the check if selection can be converted into a group
- [x] Take into account frames during grouping / ungrouping (rerouters ?)
- [x] Add import error
  _
- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.


Durman:
description

Durman:
It turned out that coding grouping and ungrouping operators are harder than I thought. Even performance is the issue. I think main problem is array like structure of tree nodes and tree links. Also that straight references should not be keepped. I think I have to create module which will convert tree in more convenient data structure what make much easier manypulate with it.

Durman:
Funny thing is that most nodes have `Node` as first base class but in this case `poll` method does not work.
```Python
class VectorsOutNode(bpy.types.Node, SverchCustomTreeNode):
```

PR #3648, opened 2020-10-15
Check fake user

vicdoval:
Fixes a long standing issue of being able to lose your work if you forget to click in the fake user toggle.

- [x] Ready for merge.

### DIFF
--- a/core/update_system.py
+++ b/core/update_system.py
@@ -208,7 +208,7 @@ def separate_nodes(ng, links=None):
             node_set_list[-1].add(n)
 
     found_node_sets = [ns for ns in node_set_list if len(ns) > 1]
-  
+
     if hasattr(ng, "sv_subtree_evaluation_order"):
         sorting_type = ng.sv_subtree_evaluation_order
         if sorting_type in {'X', 'Y'}:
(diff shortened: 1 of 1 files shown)


core/update_system.py
```diff
@@ -476,6 +479,7 @@ def process_from_nodes(nodes):
             print("Something not very important happend in Blender memory", node, type(node))
 
     ng = nodes[0].id_data
+    check_fake_user(ng)
```
Durman, line 482:
```suggestion
    ng.use_fake_user = True
```

Durman:
And there is no much sense to have this buttons then I guess.

core/update_system.py
vicdoval: Are you sure that would be something to do on every update? thats why a i added the check, I dont know if something else happens inside blender when use_fake_user is set

Durman: I think it is static property which is taking into account when Blender makes decisions of clearing data.

Durman: Probably even ban could be done. We can override creating new tree operator.

Or we can try to subscribe to `node_tree` attribute changes of a windows manager.

Durman: Yes subscribesion works
```Python
bpy.msgbus.subscribe_rna(
        key=(bpy.types.SpaceNodeEditor, 'node_tree'),
        owner=object(),
        args=(),
        notify=lambda *args: print(f'Tree was changed, probably new one was created')
    )
```

vicdoval: @Durman Should I close this PR and let you do it in a more efficient way?

PR #3650, opened 2020-10-15
Set use_fake_user attribute of Sverchok trees

Durman:

#3648

- [x] Ready for merge.

### DIFF
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -245,6 +245,12 @@ handler_dict = {
 }
 
 
+@persistent
+def call_user_functions_on_post_load_event(scene):
+    for function in data_structure.post_load_call.registered_functions:
+        function()
+
+
 def register():
 
     app_handler_ops(append=handler_dict)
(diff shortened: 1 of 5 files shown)


Durman:
@vicdoval Would you like to test?

vicdoval:
Seems to work nicely :)

Durman:
There is only one problem. After F8 and loading new file it does not work.

vicdoval:
> There is only one problem. After F8 and loading new file it does not work.

This is only a problem developers will find and in a small set of cases so to me its not very important. If you find it important enough we can always go back to the previous solution

Durman:
I also think it can stay as known issue but definitely should be fixed someday.

Durman:
The problem somewhere there. If comment the line Sverchok can be even reloaded once and in a new file the new feature works. So in reload event the module starting to be like another instance of themself and if another module is trying to wright something in it it does not see.
https://github.com/nortikin/sverchok/blob/7a6ea8e5fe08433ffc0c27052bd87a0e3452ff3b/__init__.py#L78

Durman:
http://pyunit.sourceforge.net/notes/reloading.html
https://stackoverflow.com/questions/437589/how-do-i-unload-reload-a-python-module

Durman:
"Good" return for decorator 😑 

https://github.com/nortikin/sverchok/blob/65d2ee97653e1c49268573c630db10f822a9a902/data_structure.py#L1035

PR #3668, opened 2020-10-23
frame info bugfix

### LINKED ISSUE #3663: Frame info : Evaluate
frame info : Evaluate
it generate 0.0 only when start frame = 0
if start frame not 0 result not 0.0-1.0

thank you.

vicdoval:
Closes #3663
image

### DIFF
--- a/nodes/scene/frame_info_mk2.py
+++ b/nodes/scene/frame_info_mk2.py
@@ -79,7 +79,7 @@ class SvFrameInfoNodeMK2(bpy.types.Node, SverchCustomTreeNode, SvAnimatableNode)
         outputs['Current Frame'].sv_set([[frame_current]])
         outputs['Start Frame'].sv_set([[frame_start]])
         outputs['End Frame'].sv_set([[frame_end]])
-        outputs['Evaluate'].sv_set([[frame_current / num_frames]])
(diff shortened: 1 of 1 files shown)


zeffii:
run this @vicdoval 

```python
def remap(frame_current, frame_start, frame_end):
    return (frame_current - frame_start) / (frame_end - frame_start - 1)

def remap2(frame_current, frame_start, frame_end):
    num_frames = max(1, frame_end - frame_start)
    return (frame_current - frame_start) / num_frames

frame_start = 1
frame_end = 250
for i in range(frame_start, frame_end):
    print(remap(i, frame_start, frame_end), remap2(i, frame_start, frame_end))

```
one will output a range `0.....1`  the other will output `0 .... 0.9959` (which is just as wrong as `0.004...1.0`

vicdoval:
Yes, but in blender the last frame exists
so it would need to be 

```
frame_start = 1
frame_end = 250
for i in range(frame_start, frame_end+1):
    print(remap(i, frame_start, frame_end), remap2(i, frame_start, frame_end))
```
So it would be `0-1.0040322580645162` and `0-1.0`

zeffii:
i see :) that's probably why i was confused!

PR #3673, opened 2020-10-25
Vector In and Vector Out to Multilevel recursive

vicdoval:
Vector In and Vector Out to multilevel recursion fashion.
Vector In with list matching options.
Both nodes with "Simplify output" option to keep lists in line.

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/vector/vector_in.rst
+++ b/docs/nodes/vector/vector_in.rst
@@ -25,6 +25,12 @@ In the N-Panel (and on the right-click menu) you can find:
 
 **Implementation**: Python or NumPy. Python is the default and is usually faster if you input regular lists and want to get regular list. The NumPy implementation will be faster if you are using/getting lists of NumPy arrays.
 
+**Match List**: Define how list with different lengths should be matched ('Short', 'Repeat Last' or 'Cycle')
+
+**Simplify Output**: Method to keep output data suitable for most of the rest of the Sverchok nodes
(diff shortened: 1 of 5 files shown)


docs/nodes/vector/vector_in.rst
```diff
@@ -25,6 +25,12 @@ In the N-Panel (and on the right-click menu) you can find:
 
 **Implementation**: Python or NumPy. Python is the default and is usually faster if you input regular lists and want to get regular list. The NumPy implementation will be faster if you are using/getting lists of NumPy arrays.
 
+**Match List**: Define how list with different lengths should be matched ('Short', 'Repeat Last' or 'Cycle')
+
+**Simplify Output**: Method to keep output data suitable for most of the rest of the Sverchok nodes
+  - None: Do not perform any change on the data. Only for advanced users
```
Durman, line 31: Interesting how this is going to work with socket menus #3312. As far as I understand they are going perform the same thing.

vicdoval: i understand socket menus as a extra step that can perform split and warp. they can overlap in some cases but they can also give more possibilities... in a perfect world all the sverchok nodes would be recursive and `simplify output` would not be necessary at all... but for the moment is a _necessary evil_

PR #3684, opened 2020-10-30
Inputs auto links

portnov:

Good node tree usually has most of it's "vital" parameters defined in "input" nodes, such as "A Number", "Vector In" and so on. It's better to have separate nodes for meaningful node parameters, because you can explicitly name them ("Building Height" instead of "Z"). But linking all inputs of all nodes to separate input nodes and naming those nodes is tiresome process.

Another thing is that I usually tend to group input nodes in specific places of my node trees; those places can be far from places where parameters are used, so there are very long noodles. WiFi nodes can make node trees cleaner, but manually creating and linking them is tiresome too.

## Solution description

For input sockets, that have good "input" node counterparts ("A Number" for Strings socket, "Vector In" for Vertices socket),  add a dropdown menu, that allows:

* create new input node and link it to this socket; the node will be automatically named with the name of the socket.
* create new input node and link it to this socket, but instead of direct link, create a pair of WiFi nodes and link input node via them;
* link this input to one of existing parameter nodes of corresponding type;
* link this input to existing WiFi In node.

The menu button is shown only for sockets that have associated node property. But maybe it worth to replace all existing "quicklink" buttons with this new menu?

Also I think I'll show these new buttons only when "Show socket menus" checkbox in the N panel of the tree is checked.

Screenshot_20201030_222847

I know that @Durman has reservations about WiFi nodes being too implicit, and I partially understand his concerns. But also I think that WiFi nodes can be very useful when used wisely. So I wait for input from @vicdoval, @zeffii , @nortikin : is it good idea to simplify creation of WiFi node pairs? Will we continue to support them?

## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/core/sockets.py
+++ b/core/sockets.py
@@ -1001,6 +1001,7 @@ class SvCurveSocket(NodeSocket, SvSocketCommon):
 class SvScalarFieldSocket(NodeSocket, SvSocketCommon):
     bl_idname = "SvScalarFieldSocket"
     bl_label = "Scalar Field Socket"
+    quick_link_to_node = 'SvNumberNode'
 
     color = (0.9, 0.4, 0.1, 1.0)
     default_conversion_name = ConversionPolicies.FIELD.conversion_name
@@ -1014,6 +1015,7 @@ class SvScalarFieldSocket(NodeSocket, SvSocketCommon):
 class SvVectorFieldSocket(NodeSocket, SvSocketCommon):
     bl_idname = "SvVectorFieldSocket"
     bl_label = "Vector Field Socket"
+    quick_link_to_node = 'GenVectorsNode'
 
     color = (0.1, 0.1, 0.9, 1.0)
     default_conversion_name = ConversionPolicies.FIELD.conversion_name
--- a/core/sockets.py
+++ b/core/sockets.py
@@ -208,6 +208,7 @@ class SvSocketCommon(SvSocketProcessing):
     color = (1, 0, 0, 1)  # base color, other sockets should override the property, use FloatProperty for dynamic
     label: StringProperty()  # It will be drawn instead of name if given
     quick_link_to_node = str()  # sockets which often used with other nodes can fill its `bl_idname` here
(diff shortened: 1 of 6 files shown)


vicdoval:
I have used a lot of wifi nodes and I think I would use them more with this menus. 
I like them and for me they ease the reading of the nodetree in many cases (specially with many parameters and too many links caming out from them)
I understand that if one 'parameter node' has already a 'wifi in' node  your proposal would not create another (it would re-use the existing one)

zeffii:
>  manually creating and linking them is tiresome

agreed, "quicklink" was a first attempt at alleviating some of these repetitive processes, like what "auto complete / intellisense" is for code, having something similar for sockets will speed to tree creation 

i do see some issues with overpopulating the list of options in that dropdown, as the tree grows will that list turn into something that must be waded through?

i like it though. i like physical connections that can be seen, i'm not a massive fan of wifi-nodes. I only wish node-noodles could be styled.

Durman:
Just an idea of another possible solution. We could have properties bound to a each to tree on N panel like properties on 3D panel but not the same. 
There should be UI for creation any type of properties right on N panel of a tree. When a property is created any socket in this tree should have opportunity to subscribe to its changes. So if tree property will be changed it will trigger changes of al its subscribed sockets. I think this solution could be more logical then manually creating set of nodes at the beginner of a tree but it looks like that it will share the problems of wifi nodes unless only to implement somehow search of all subscribed properties for each tree property.

portnov:
> i do see some issues with overpopulating the list of options in that dropdown, as the tree grows will that list turn into something that must be waded through?

That's why that menu has a searchbox :)

portnov:
> Also I think I'll show these new buttons only when "Show socket menus" checkbox in the N panel of the tree is checked.

I ended up with adding a new option to Sverchok Preferences for that:
* do not show input menus / buttons at all
* show only old Quicklink buttons (default behaviour, so people will not be surprised with suddenly new options)
* show new menus always.

PR #3709, opened 2020-11-10
Solid Bounding box node

portnov:
## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/solid/bound_box.rst
+++ b/docs/nodes/solid/bound_box.rst
@@ -1,0 +1,44 @@
+Solid Bounding Box
+==================
+
+Dependencies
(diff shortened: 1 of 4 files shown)


vicdoval:
I was thinking it would be cool to have nodes that are geometry "agnostic" that could accept meshes, solids, curves and surfaces...
Like a bounding box, move, rotate, matrix apply...

Just one idea....

portnov:
It would be.
But even Grasshopper has the same zoo as we do :)

Issue #3660, opened 2020-10-20
Node groups fail on duplicate


vicdoval:
When a copied group is added (via duplicate or via adding new node and assigning exiting group):

```
 line 84, in socket_id
    return str(hash(self.node.node_id + self.identifier))
AttributeError: 'SvGroupTreeNode' object has no attribute 'node_id'
```

Durman:
Thanks. It is known. I was going to deal with it on second stage.

PR #3713, opened 2020-11-13
Group nodes - Alpha

Durman:

This PR should bring life into group nodes.
It's not a final stage and future updates can break work of saved trees with group nodes.
For now every node which connected to group output node or to other nodes which connected to group output node will be evaluated. Also Debug print and Stethoscope nodes are supported.

Editing the same node groups from different windows is unsupported.
Viewer nodes inside ndoe groups are not supported.

fixing #3660
#3319

## Preflight checklist

- [x] Ready for merge.


Durman:
Playing around canceling group tree update. It's not efficient to color each node which is currently updating in case if updation is going quite rapidly. But if to skip coloring some nodes when really hard node will be get it's quite likely that it won't be colored as updating node. So probably it would be better to color already updated nodes instead. Or not coloring nodes at all but potentially this functionality can be very user friendly and explorative about what is actually going on.
abort updating

PR #3715, opened 2020-11-14
"Unwrap" option for socket menus

portnov:
This option removes one pair of square brackets from a nested list, if it is
possible. If output data is empty, or output data is a list with more than one
item, this will raise an exception, and processing will stop. For example, if
the socket contains ``[[1, 2, 3]]``, then with this option enabled you will
have ``[1, 2, 3]``. But, if the socket contains ``[[1, 2], [3, 4]]``, then you
will have an exception.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


Durman:
Strange option. What is it useful for?

portnov:
Some nodes are designed to work with nested input, so they produce nested output in general case. But then it can appear that you do not want so high nesting level, if you pass only one item to the input.

Durman:
Which node expect such input `[1, 2, 3]`?

portnov:
most expect ``[[1,2,3]]`, for example. But another nodes output ``[[[1,2,3]]]`` instead.

Durman:
> most expect `[[1,2,3]]`, for example.

But you decided to convert it into `[1, 2, 3]` even if no one node expects such format and even raise an error in some cases. Looks like this option is breaking our vectorization conception.

portnov:
This option removes *one*, outermost, pair of square brackets. User enables it when he knows that the output list contains exactly one object (one nested list) at the outer level.

Issue #3721, opened 2020-11-18
Curve Mapper loses its curve when performing specific actions


mnn:
## Problem statement

When duplicating a curve mapper or creating a monad with a curve mapper, only a blank curve mapper node is created, the curve is lost.

![Peek 2020-11-18 18-20_sverchok_vanishing_curve_in_curve_mapper](https://user-images.githubusercontent.com/1975567/99564530-dfcfc400-29ca-11eb-96d8-f83f547c0017.gif)

## Steps to reproduce

1. Create Curve Mapper node with custom curve
2. Duplicate the node

## Expected result

The curve is duplicated along with other node data.

## Actual result

Curve Mapper without curve (actually even the default curve is not shown, you have change some other value in the node to get it to show, so you can edit it).

## Sverchok version

0.6.0.0

I believe it's the "stable Oktober 2020" v2, installed from inside Blender (`Add-ons` -> `Install...`)

zeffii:
yeah, curvemapper is missing a copy() function. (to tell the node how to respond to user using node duplication)

zeffii:
something like 
```python
    def sv_copy(self, other):
        ''' self is the new node, other is the old noe '''
        curve_dict = {}
        other.save_to_json(other, curve_dict)
        self.load_from_json(self, curve_dict)
        updateNode(self, None)

```
but this makes the new curvemapper linked to the original, which is probably not what you want.

zeffii:
if someone else doesn't fix it by tomorrow, i'll have a closer look.

zeffii:
i didn't manage to solve in my avaliable time.

zeffii:
https://github.com/nortikin/sverchok/pull/3726  maybe? anyone confirm

zeffii:
should be OK now , pushed a fix https://github.com/nortikin/sverchok/pull/3726

PR #3726, opened 2020-11-19
Curve mapper sv_copy

zeffii:
addresses #3721

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -277,7 +277,7 @@ class UpdateNodes:
         """
         pass
 
-    def sv_copy_pos(self, original):
+    def sv_copy_post(self, original):
         """
         Override this method to do anything node-specific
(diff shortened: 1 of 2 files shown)


node_tree.py
```diff
@@ -346,6 +353,7 @@ def copy(self, original):
             self.use_custom_color, self.color = settings
         self.sv_copy(original)
         self.n_id = ""
+        self.sv_copy_post(original)
```
Durman, line 356:
```suggestion
        self.n_id = ""
        self.sv_copy(original)
```

zeffii: if we never need to know the current node_id before copying, then yes, perhaps `sv_copy_post` is not needed.

zeffii: ok , lets do that.

zeffii (commented):
agreed

PR #3745, opened 2020-11-29
Loop in out nodes

vicdoval:
New nodes to create node loops. It seems very stable (much more than monads) although the inside logic is pretty similar, also pretty basic.
The loop is performed in the Loop Out node, this creates less interference with the actual update system.
Also they can do nested loops and use any viewer
image
image

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/logic/loop_in.rst
+++ b/docs/nodes/logic/loop_in.rst
@@ -56,6 +56,6 @@ For Each mode example, Skip input used to mask the results.
 
 .. image:: https://user-images.githubusercontent.com/10011941/101334215-e047d600-3877-11eb-89df-cfaaf73dd427.png
 
-You can change socket the socket labels in the N-Panel
+You can change the socket labels in the N-Panel
 
 .. image:: https://user-images.githubusercontent.com/10011941/101360702-519a7f80-389e-11eb-826d-0e1c5a7152d1.png
--- a/docs/nodes/logic/loop_in.rst
+++ b/docs/nodes/logic/loop_in.rst
@@ -48,6 +48,14 @@ Data0, Data1... output sockets will be created when the last input is linked
 Examples
 --------
 
+Range mode example, Break used to control the maximum vertices.
+
 .. image:: https://user-images.githubusercontent.com/10011941/101332093-22234d00-3875-11eb-819a-68e86ef8c2c2.png
 
+For Each mode example, Skip input used to mask the results.
+
 .. image:: https://user-images.githubusercontent.com/10011941/101334215-e047d600-3877-11eb-89df-cfaaf73dd427.png
+
+You can change socket the socket labels in the N-Panel
+
+.. image:: https://user-images.githubusercontent.com/10011941/101360702-519a7f80-389e-11eb-826d-0e1c5a7152d1.png
--- a/docs/nodes/logic/loop_in.rst
+++ b/docs/nodes/logic/loop_in.rst
@@ -1,0 +1,53 @@
+Loop In
+=======
+
+This node in conjunction with the Loop out node can create loops with nodes
+
+Offers two different modes 'Range' and 'For Each'
+
+Range
+-----
+
+In this mode the all the data inputted to the Loop In node will be processed every iteration.
+
+For Each
+--------
+
+In this mode the inputted data will be splitted before being processed and there will be one loop per every level 1 object.
+
+Operators
+---------
+
+**Create Loop Out**: creates a Loop out node and links the Loop In - Loop Out socket.
+
+
+Inputs
+------
+
+**Iterations**: Number of repetitions (only in Range mode).
+
+Data0, Data1... inputs will be created when the last one is linked
+
+Options
+-------
+
+**Max Iterations**: Maximum iterations (in N-panel and Contextual Sverchok Menu)
+**Socket Labels**: To change sockets names (in N-panel)
+
+Outputs
+-------
+
+**Loop Out**: Socket to link with the Loop Out node.
+
+**Loop Number / Item Number**: Actual Repetition / Item.
+
+**Total Loops / Total Items**: Total Repetitions / Items.
+
+Data0, Data1... output sockets will be created when the last input is linked
+
+Examples
+--------
+
+.. image:: https://user-images.githubusercontent.com/10011941/101332093-22234d00-3875-11eb-819a-68e86ef8c2c2.png
+
+.. image:: https://user-images.githubusercontent.com/10011941/101334215-e047d600-3877-11eb-89df-cfaaf73dd427.png
--- a/nodes/logic/loop_in.py
+++ b/nodes/logic/loop_in.py
@@ -35,12 +35,33 @@ class SvCreateLoopOut(bpy.types.Operator):
         tree = bpy.data.node_groups[self.idtree]
         node = bpy.data.node_groups[self.idtree].nodes[self.idname]
 
-        new_node =tree.nodes.new('SvLoopOutNode')
+        new_node = tree.nodes.new('SvLoopOutNode')
         new_node.parent = None
         new_node.location = (node.location.x + node.width + 400, node.location.y)
         tree.links.new(node.outputs[0], new_node.inputs[0])
         frame_adjust(node, new_node)
         return {'FINISHED'}
+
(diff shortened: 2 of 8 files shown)


portnov:
Will there be "for each" mode? Like, for each item in the input list, run the following nodes...

vicdoval:
@Durman with the event system is possible to get from one node which nodes have been updated in the last update wave?

vicdoval:
"For each" mode added :) 
image

Durman:
Events do nothing. They are only means for communication with update system. I had added them for future. At this moment they are only used in `group` update system.

Durman:
Debugging loops is the issue.

Durman:
thing generator

vicdoval:
Better Exceptions for debugging
image

nodes/logic/loop_out.py
```diff
@@ -24,6 +24,12 @@
 from sverchok.core.update_system import make_tree_from_nodes, do_update
 from sverchok.data_structure import list_match_func
 
+def process_looped_nodes(node_list, tree_nodes):
+    for node_name in node_list:
+        try:
+            tree_nodes[node_name].process()
+        except Exception as e:
+            raise type(e)(str(e) + ' @ %s node' % node_name)
```
Durman, line 32: I think this syntax could be used.
https://docs.python.org/3/tutorial/errors.html#exception-chaining

vicdoval: That seems to change totally the exception and what we need is to add information to the existing exception https://stackoverflow.com/questions/6062576/adding-information-to-an-exception

Durman:
Another idea. To add `while` mode and boolean socket to output loop node. Output node should looping until the socket will get False value.

vicdoval:
**Break** input to break the loop
image
**Skip** input to skip that object in the output
image

mnn:
I haven't tried it, but I am wondering, is it / would it be possible to name all those "data" input/output sockets (similarly like you can give custom names to sockets in group/monad)? I am not sure if it's that important feature, but it might make orientation in a node graph a bit easier - I wouldn't need to visually track `Data 0` to a node connecting to an input of a loop, if I would have appropriately named the socket in the loop node (e.g. seeing an output socket in the loop node named as  `HatVerts` means I don't have track from where `Data 0` is coming).

PS: Sorry for my English (those damn tenses), it's not my native language, but I hope it's understandable what I meant.

vicdoval:
@mnn  that is a good point... i have just done a very simple implementation of that idea exposing `socket.label` in the properties 
of the Loop In node. The only downside is that the names dont spread (update) until the user creates a new link (or tries to)
image
@Durman you think it could be possible (and valuable) to being able to capture the changes in the socket.label from the owner node?

Durman:
Probably it is possible but I don't think that it would be simplest solution. I would add code of changes of socket labels into `loop in` node. Then this code will be called each time new link will be connected to this node. Overhead of having such code in update method would not be noticeable I think.

vicdoval:
I added a little operator to update the sockets label, looks less buggy image

docs/nodes/logic/loop_in.rst
```diff
@@ -0,0 +1,61 @@
...
+
+Range mode example, Break used to control the maximum vertices.
+
+.. image:: https://user-images.githubusercontent.com/10011941/101332093-22234d00-3875-11eb-819a-68e86ef8c2c2.png
+
+For Each mode example, Skip input used to mask the results.
+
+.. image:: https://user-images.githubusercontent.com/10011941/101334215-e047d600-3877-11eb-89df-cfaaf73dd427.png
+
+You can change socket the socket labels in the N-Panel
```
mnn, line 59: First `socket` word looks redundant.

vicdoval: thanks!

nodes/logic/loop_in.py
```diff
@@ -0,0 +1,204 @@
...
+        self.inputs.new('SvStringsSocket', 'Iterations').prop_name = "iterations"
+        self.inputs.new('SvStringsSocket', 'Data 0')
+
+        self.outputs.new('SvLoopControlSocket', 'Loop Out')
+        self.outputs["Loop Out"].link_limit = 1
+        self.outputs.new('SvStringsSocket', 'Loop Number')
+        self.outputs.new('SvStringsSocket', 'Total Loops')
+
+    def draw_buttons(self,ctx, layout):
+        if not (self.outputs[0].is_linked and self.outputs[0].links[0].to_socket.node.bl_idname=='SvLoopOutNode'):
```
Durman, line 124: To call `socket.links` in draw method can be quite expensive.

nodes/logic/loop_in.py
```diff
@@ -0,0 +1,204 @@
...
+        if node.outputs[0].is_linked:
+            if node.outputs[0].links[0].to_socket.node.bl_idname == 'SvLoopOutNode':
+                loop_out_node = node.outputs[0].links[0].to_socket.node
+                for inp, self_inp, self_outp in zip(node.inputs[1:], loop_out_node.inputs[2:], loop_out_node.outputs):
+                    self_outp.label = inp.label
+                    self_inp.label = inp.label
+        return {'FINISHED'}
+
+
+class SvLoopInNode(bpy.types.Node, SverchCustomTreeNode):
```
Durman, line 66:
```suggestion
class SvLoopInNode(SverchCustomTreeNode, bpy.types.Node):
```

nodes/logic/loop_in.py
```diff
@@ -0,0 +1,204 @@
...
+
+    bl_idname = "node.create_loop_out"
+    bl_label = "Create Loop Out"
+
+    idtree: bpy.props.StringProperty(default='')
+    idname: bpy.props.StringProperty(default='')
+
+    def execute(self, context):
+        tree = bpy.data.node_groups[self.idtree]
+        node = bpy.data.node_groups[self.idtree].nodes[self.idname]
```
Durman, line 36:
```suggestion
    def execute(self, context):
        node = context.node
```

nodes/logic/loop_in.py
```diff
@@ -0,0 +1,212 @@
...
+        self.inputs.new('SvStringsSocket', 'Data 0')
+
+        self.outputs.new('SvLoopControlSocket', 'Loop Out')
+        self.outputs["Loop Out"].link_limit = 1
+        self.outputs.new('SvStringsSocket', 'Loop Number')
+        self.outputs.new('SvStringsSocket', 'Total Loops')
+
+    def draw_buttons(self,ctx, layout):
+        if not self.linked_to_loop_out:
+            self.wrapper_tracked_ui_draw_op(layout, "node.create_loop_out", icon='CON_FOLLOWPATH', text="Create Loop Out")
```
Durman, line 129:
```suggestion
            self.operator("node.create_loop_out", icon='CON_FOLLOWPATH', text="Create Loop Out")
```

Durman:
It also would be useful to use loops for vectorization purposes. Because some times it is really hard to setup whole tree to proper vectorization.

vicdoval:
The "For Each" mode can be used for this purpose: 
image

portnov:
@vicdoval is there a way to pass different values of some parameter to internal node for each iteration? Like, extrude faces 3 times, first with height = 1.0, then 0.5 and then 0.25 ?

vicdoval:
> @vicdoval is there a way to pass different values of some parameter to internal node for each iteration? Like, extrude faces 3 times, first with height = 1.0, then 0.5 and then 0.25 ?

like this?
image

Durman:
@vicdoval 
 Ah, yes. This part is  a beat inconvenient only.

vicdoval:
Its true the loop out node could be able to create its own sockets... But should this sockets appear in the "loop in node" too as outputs and inputs? that would requiere changing the actual socket logic...

Durman:
Looks like I'm not understand `for each` mode. Should not the loop generate 5 boxes?

vicdoval:
For each (Level 1) object
image

Durman:
Ok. This looks is working. But this solution is about 6 times slower than similar solution without loops. Don't see why.

Next layout is 6 times faster.

PR #3752, opened 2020-12-04
Default presets

portnov:

We used to have "custom defaults" mechanism for nodes. Now it is removed, but even when it worked it had a problem of being too hard to use. 

## Solution description

Now that we have Presets and per-node presets, let's use this mechanism to replace old custom defaults mechanism. How it works:

* You configure the node as you wish
* You go to it's Presets menu (in the N panel, or in the node's context menu, or you can press Shift-P), and select "Save current settings as node preset"
* Enter **Default** (case sensitive) as the name of the preset.

Now, when you create node of the same type by any usual way (from Shift-A menu, from T panel, from Search menu), this Default preset will be automatically applied. This preset will not be applied when the node is created as part of another (multi-node) preset being imported.

These Default presets are available in the Presets panel under T panel; you can rename them, or remove as well as others.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/panels.rst
+++ b/docs/panels.rst
@@ -118,6 +118,11 @@ categories are distinguished with a prefix "/Node/" in their titles. For
 example, a category with name "/Node/ Box" can contain only settings of the
 "Box" node. Such categories are created automatically.
 
+Among per-node presets, you can define a preset with special name ``Default``.
+Such preset will be automatically applied to the node of this type when you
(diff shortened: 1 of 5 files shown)


onerawartist:
This would be a great feature. ,looking forward to try it

PR #3767, opened 2020-12-10
KDT Edges fast mode

vicdoval:
Fast Mode
---------
This mode requires Scipy and Cython dependencies. It can be from 3 to 10 times faster but lacks of 'maxNum' and 'Skip' properties 

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/analyzer/kd_tree_edges_mk2.rst
+++ b/docs/nodes/analyzer/kd_tree_edges_mk2.rst
@@ -33,17 +33,28 @@ Parameters
 Fast Mode
 ---------
 
-This mode requires Scipy and Cython dependencies. It can be from 3 to 10 times faster but lacks of 'maxNum' and 'Skip' properties 
+This mode requires Scipy dependency. It can be from 3 to 10 times faster but lacks of 'maxNum' and 'Skip' properties
+
+.. image:: https://user-images.githubusercontent.com/10011941/101905604-017e2e80-3bb8-11eb-894d-67a3c6c51e93.png
+
+
+Max Queries Mode
+---------------
+
+This mode requires Scipy dependency. In this mode the maxNum property is used to determine how many points will be verified so it will produce less connections that the complete mode
+
+.. image:: https://user-images.githubusercontent.com/10011941/101905663-15c22b80-3bb8-11eb-8625-4e48085a2dfe.png
+
+No Skip Mode
+------------
+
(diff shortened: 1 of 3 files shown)


portnov:
@vicdoval I created `sverchok.utils.kdtree` module, which presents common interface for different KdTree implementations. You may want to use it.

vicdoval:
@portnov  i see... there is two files utils.sv_KDT_utils and and utils.kdtree possibly would be nice to merge them... i will check. Did you see 
```
n_jobs: int, optional
Number of jobs to schedule for parallel processing. If -1 is given all processors are used. Default: 1.
```
in https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.query.html#scipy.spatial.cKDTree.query

portnov:
Yeah, good thing. You can pass it via `**kwargs` thingy.

portnov:
sv_KDT_utils.py contains "non-trivial" algorithms that work with KDT, but it works with mathutils.kdtree only. It would be good to adapt these algorithms to work with arbitrary SvKdTree, yes.

vicdoval:
The problem/advantage Im finding is that the set of features are not identical, and translating the algorithm can make it even slower for a identical output.. i added some more modes that are faster than the actual solution with different outcome:
Fast mode:
image
Max Query: the numMax is used to determine the amount of queried points, this is leads to less number of generated edges.
image
No Skip: This is similar to the existing mode but the way the maximum connections is coded produces different results sorting the filter by minimum vertex index
image

and finally added the functions to the sv_KDT_utils.py to leave clean the kdtree file

portnov:
Do you mean that some functions will work faster with one implementation, and others will work faster with another? I think it's okay. You can add a parameter to the node for the user to select implementation. Or if you know in advance that some implementation works faster for this particular node, you can do `if better_implementation is available : use it; else: use whatever is available`.

vicdoval:
In some cases the best implementation is always present (kdtree) so it would not make sense to keep the scipy implementation

portnov:
for such cases you can either just use mathutils.kdtree interface, or use `SvKdTree.new(implementation='BLENDER')` just to use the same interface (so that parts of code can be copy-pasted between different nodes, for example). Whatever you like more :)

vicdoval:
I couldn't  find a suitable way to repeat the behaviours with both implementations, so I'm leaving it like this, the Scipy users will have more options :) by the moment at `sv_KDT_utils.py` maybe it can change in the future...

Issue #3719, opened 2020-11-16
Plane : Change parameter follow selected plane?


rwny:
for Plane node. is possible to change parameter follow plane that we select?
it alway X,Y

![a451](https://user-images.githubusercontent.com/32074126/99210858-7cdb0300-27f9-11eb-932e-af44042ac12c.PNG)

thank you very much

Durman:
I think the properties could be just renamed

PR #3773, opened 2020-12-12
more properties

vicdoval:
Properties change when changing direction, I added new properties to the node as they can be renamed dynamically
Closes https://github.com/nortikin/sverchok/issues/3719
image

- [X] Ready for merge.


portnov:
Isn't it enough to just do `layout.prop('size_x', text = "Size X" or "Size Y")` ?

Durman:
The same can be applied here I guess. #3578

vicdoval:
You are right, it was to complex to be the right path :)

PR #3776, opened 2020-12-13
Matrix Math Vectorize

vicdoval:
Matrix Math Node accepts nested lists of matrices (multi-level recursion fashion)
image

- [X] Ready for merge.


vicdoval:
There is something that makes me doubt: 
The actual version of the node outputs X,Y,Z vectors in the Basis mode with a wrong level of nesting (missing [ ]) .
This PR fixes that, but that fix may break some layouts. Does it worth a new mk version?

Durman:
Probably it's not very popular output of the node. I don't remember using those outputs myself.

PR #3781, opened 2020-12-17
Unique Items node

vicdoval:
Unique items Node, mimics (and uses when possible) the np.unique function.

Deconstruction and reconstruction of a list:
image


nortikin:
extended set()?

Issue #3783, opened 2020-12-18
Loop node performance


Durman:
https://user-images.githubusercontent.com/28003269/102585514-fb410280-4121-11eb-9baf-4f65d0e4b363.mp4

<details> 
  <summary>loop node stats</summary>  

```
   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    19574    0.389    0.000    0.389    0.000 {built-in method nt.stat}
     6878    0.036    0.000    0.076    0.000 ntpath.py:450(normpath)
     6348    0.035    0.000    0.501    0.000 inspect.py:760(findsource)
    13226    0.033    0.000    0.401    0.000 inspect.py:680(getsourcefile)
     6348    0.025    0.000    0.739    0.000 inspect.py:1444(getframeinfo)
    13226    0.021    0.000    0.048    0.000 inspect.py:642(getfile)
   186611    0.019    0.000    0.019    0.000 {built-in method builtins.isinstance}
     6878    0.014    0.000    0.017    0.000 ntpath.py:122(splitdrive)
    26760    0.013    0.000    0.036    0.000 {built-in method builtins.any}
     6878    0.013    0.000    0.146    0.000 inspect.py:714(getmodule)
     6348    0.011    0.000    0.127    0.000 linecache.py:53(checkcache)
```

```
   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      3/1    0.000    0.000    0.891    0.891 update_system.py:361(do_update_general)
        1    0.000    0.000    0.865    0.865 loop_out.py:149(process)
        1    0.001    0.001    0.865    0.865 loop_out.py:168(for_each_mode)
      530    0.002    0.000    0.797    0.002 logging.py:155(wrapper)
      530    0.000    0.000    0.758    0.001 inspect.py:1511(stack)
      530    0.009    0.000    0.757    0.001 inspect.py:1483(getouterframes)
     6348    0.025    0.000    0.745    0.000 inspect.py:1444(getframeinfo)
      715    0.001    0.000    0.602    0.001 sockets.py:398(sv_set)
      715    0.004    0.000    0.600    0.001 socket_data.py:84(SvSetSocket)
     6348    0.035    0.000    0.507    0.000 inspect.py:760(findsource)
      101    0.002    0.000    0.411    0.004 box_mk2.py:369(process)
[...]

Issue #3644, opened 2020-10-13
Annoying prints


Durman:
Annoying prints in the tests and console.

```Text
set render_scale to: 1.0416666269302368
set render_location_xy_multiplier to: 1.0416666269302368
3
3
3
3
3
3
3
3
3
3
3
3
3
3
3
3
3
3
3
Unexpected error: <class 'TypeError'>
on line:  115
```
I'm suspecting this print:
https://github.com/nortikin/sverchok/blob/8afad585749f25f5149a7ac6b3c6a69fbbcb707e/data_structure.py#L1215

Durman:
https://github.com/nortikin/sverchok/blob/ae82a9e4347b6021b4f6008bfbea57b12b320f8f/nodes/vector/vector_out.py#L44

Durman:
https://github.com/nortikin/sverchok/blob/f0cc42112d6664d34bd6204a35fbeac1ae6135a3/nodes/matrix/matrix_track_to.py#L164

Durman:
https://github.com/nortikin/sverchok/blob/b6583bf41141e603f4eb64d72b275dbc9f2b485b/core/socket_data.py#L90-L91

Durman:
https://github.com/nortikin/sverchok/blob/fffcc5916fc22e90046fe1ccb06e0dc677ed3ea3/core/sockets.py#L504

Durman:
https://github.com/nortikin/sverchok/blob/4ff9c71fea9aeb0f395d72bc2763906a4ab17892/utils/sv_extended_curve_utils.py#L58

PR #3785, opened 2020-12-18
Remove slowing statement

vicdoval:
In this node tree
image
the removed debug statement was causing this nodetree to be executed in 2.54 secs and after removal 0.68 sec  (+1.86 s)  even when debugging was disabled.
Reminder: Use debug statements with caution
#3783
- [X] Ready for merge.

### DIFF
--- a/core/sockets.py
+++ b/core/sockets.py
@@ -42,7 +42,7 @@ from sverchok.utils.curve import SvCurve
 from sverchok.utils.curve.algorithms import reparametrize_curve
 from sverchok.utils.surface import SvSurface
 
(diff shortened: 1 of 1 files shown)


Durman:
Also relevant to this #3644

portnov:
Was it slowing down only when logging is set to debug, or always?
Also about logging: it's in general better to wrte `debug("Hello: %s", variable)` than `debug(f"Hello: {variable}")`, because in the former case logging library will not perform string formatting in case debug is not enabled.

vicdoval:
@portnov it was slowing down the tree always

PR #3787, opened 2020-12-20
Sv get performance

vicdoval:
Optimisation of sv_get:
From profiling this nodetree:
image
I discover some performance leaks:
socket.other (get_other_socket) was being called 2-6 times every sv_get.
also the basic concept _every dot pays a price_  becomes  obvious when the function is triggered 8000 times every update.
So I have rewritten some parts of the sv_get functionality including implicit conversions.

With this make up changes sv_get was boosted 2x (from 0.2s to 0.1 s for 9053 calls), this will be hard to appreciate on a simple tree but reduces the distance between the nodes.
#3783
- [x] Ready for merge.

### DIFF
--- a/core/sockets.py
+++ b/core/sockets.py
@@ -35,6 +35,7 @@ from sverchok.data_structure import (
 
 from sverchok.settings import get_params
 
+from sverchok.utils.socket_utils import format_bpy_property, setup_new_node_location
 from sverchok.utils.field.scalar import SvScalarField
 from sverchok.utils.field.vector import SvVectorField
 from sverchok.utils.curve import SvCurve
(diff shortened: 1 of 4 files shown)


core/sockets.py
```diff
@@ -264,6 +265,15 @@ def draw_menu_items(self, context, layout):
...
             layout.prop(self, 'use_wrap')
 
+def format_property(prop):
+    if isinstance(prop, (str, int, float)):
+        return [[prop]]
+    if hasattr(prop, '__len__'):
+        # it looks like as some BLender property array - convert to tuple
+        return [[prop[:]]]
+
+    return [prop]
```
Durman, line 275:
```Python
from sverchok.utils.handle_blender_data import BPYProperty
prop_value = PBYProperty(socket, prop_name).value
```
I don't think this is the place for the function anyway, probably move into some utils module.

vicdoval: I dont get what are you proposing

Durman: I think it's better to move the function into some another (utility) module for keeping sockets module clean. It's actually already too large now.

PR #3790, opened 2020-12-21
removing deep copies from list_struct and modifier_change

vicdoval:
added some deepcopy=False to nodes of list_struct (they didn't need the deepcopy)

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/nodes/modifier_change/extrude_separate.py
+++ b/nodes/modifier_change/extrude_separate.py
@@ -24,7 +24,7 @@ from bpy.props import IntProperty, FloatProperty, EnumProperty
 from mathutils import Matrix, Vector
 
 from sverchok.node_tree import SverchCustomTreeNode
-from sverchok.data_structure import updateNode, match_long_repeat, throttle_and_update_node, make_repeaters
+from sverchok.data_structure import updateNode, match_long_repeat, throttle_and_update_node, make_repeaters, repeat_last_for_length
 from sverchok.utils.sv_bmesh_utils import bmesh_from_pydata, pydata_from_bmesh, fill_faces_layer
 
 vsock, toposock = 'SvVerticesSocket', 'SvStringsSocket'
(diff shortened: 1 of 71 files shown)


nodes/modifier_change/extrude_separate.py
```diff
@@ -195,16 +195,13 @@ def process(self):
...
-        for vertices, edges, faces, masks, heights, scales, matrixes, face_data in zip(*meshes):
+        for vertices, edges, faces, masks_, heights_, scales_, matrixes_, face_data in zip(*meshes):
 
             new_extruded_faces = []
             new_extruded_faces_append = new_extruded_faces.append
-            fullList(heights, len(faces))
-            fullList(scales, len(faces))
-            fullList(matrixes, len(faces))
-            fullList(masks, len(faces))
+            heights, scales, matrixes, masks = make_repeaters([heights_, scales_, matrixes_, masks_])
```
portnov, line 202: Shouldn't this depend on len(faces)?

portnov:
Do I understand correctly that you've removed all usages of `fullList` from nodes?
If so, then I suggest marking the method with `@deprecate` (from sverchok.utils.decorators), so that people who use the method in scripts would see warnings. After some time this method can be removed.

nodes/modifier_change/extrude_separate.py
vicdoval: It creates iterators that later are used in zip(faces, repeater), so depends of faces but does not create the full list
```
def make_repeaters(lists):
    chain = itertools.chain
    repeat = itertools.repeat
    out =[]
    for l in lists:
        out.append(chain(l, repeat(l[-1])))

    return out
```

vicdoval:
At this point (when I'm thinking to merge this PR) there are still 30 nodes using `fullList` function so maybe we are not ready to depreciate the function

portnov:
ok

PR #3796, opened 2020-12-27
polygon to edges improved

vicdoval:
Improved NumPy implementation that handles polygons lists with variable sides number.
Removed "Regular polygons" property (handled by the node)
In this scenario:
image
Before 0.17 seconds now 0.085 seconds (2X)
In this one
image
Before 0.077 now 0.066 (1.16X)
- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


vicdoval:
I have just realised that although the outputted edges are the same the order of them is not identical, this could potentially break some setups where edges had been selected by index.... Should i  make a mk2 version? 
This boost could also be applied to all the nodes that use polygons_to_edges function but also would have the changed order of edges... Any thoughts on this?

vicdoval:
This nodes could benefit from moving from polygons_to_edges to polygons_to_edges_np but they would require new mk version image

PR #3798, opened 2020-12-27
Adaptive Polygons mk3

vicdoval:
Numpy implementation for the Adaptive polygons node.
In this scenario: 
image
Before 0.793 seconds. Now 0.058 seconds (13X)
I left the previous implementation because is faster when the donor has less than 12 vertices for quad transform and 50 for tris transform. (approx.)
The default is a Auto mode that switches implementation depending on the donor vert count.
Also added a Donor Matching property that can be Repeat Last, Cycle or By Index (controlled by input)
image

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/nodes/modifier_make/adaptive_polygons_mk3.py
+++ b/nodes/modifier_make/adaptive_polygons_mk3.py
@@ -31,7 +31,7 @@ from numpy import (array as np_array,
 from numpy.linalg import (norm as np_norm,
                           inv as np_inv)
 import bpy
-from bpy.props import FloatProperty, EnumProperty, BoolProperty, IntProperty
+from bpy.props import FloatProperty, EnumProperty, BoolProperty, IntProperty, FloatVectorProperty
 from mathutils import Vector, Matrix
 from mathutils.geometry import barycentric_transform
 
@@ -81,7 +81,6 @@ def prepare_source_data(src_verts):
 
     return inverted_matrix_s, tri3_src
 
-
 def prepare_dest_data(dst_verts):
     '''Create Transformation Matrix and 4th point of the tetrahedron'''
     matrix_trasform, tri3_dest = matrix_def(dst_verts)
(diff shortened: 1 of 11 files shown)


Durman:
I have tested it. In my layout the new node is about 2 times faster and whole layout is about 1.5 times faster.

Durman:
By the way in my case I would not mind to have opportunities to set `use normals` parameter per face.

face normals

nodes/modifier_make/adaptive_polygons_mk3.py
```diff
@@ -57,6 +57,14 @@
 sqrt_3_3 = sqrt_3/3
 sqrt_3_2 = sqrt_3/2
 
+def normalize_v3(arr):
+    ''' Normalize a numpy array of 3 component vectors shape=(n,3) '''
+    lens = np_sqrt( arr[:,0]**2 + arr[:,1]**2 + arr[:,2]**2 )
+    arr[:,0] /= lens
+    arr[:,1] /= lens
+    arr[:,2] /= lens
+    return arr
```
portnov, line 66: I would suggest
1. Change this to something like
```python
result = np.zeros(arr.shape)
norms = np.linalg.norm(arr, axis=1, keepdims=True)
nonzero = (norms > 0)[:,0]
result[nonzero] = arr[nonzero] / norms[nonzero][:,0][np.newaxis].T
```
this 1) uses vectorization more effectively, 2) takes care of division by zero.

2. Move it to `sverchok.utils.math`, because I think I saw similar functions in other modules, at least next time there will be "standard" function to use.

vicdoval: Ok.
I did some performance tests
and this was 10% faster:
```
def np_normalized_v3(vecs):
    result = np.zeros(vecs.shape)
    norms = np.linalg.norm(vecs, axis=1)
    nonzero = (norms > 0)
    result[nonzero] = vecs[nonzero] / norms[nonzero][:,np.newaxis]
    return result
```
And this like 2.5x faster. Does the normalisation in-place (good for me in this case)
```
def np_normalize_v3(vecs):
    norms = np.linalg.norm(vecs, axis=1)
    nonzero = (norms > 0)
    vecs[nonzero] = vecs[nonzero] / norms[nonzero][:,np.newaxis]
```
I added both to `sverchok.utils.math`

vicdoval:
Added:
Normal mapping mode per face

New SubQuads mode, 
Top: Frame Transform Bottom: SubQuads Transform 
image

Properties refactor with possibility to control transform mode with mask input.
Transform options available per Tris, Quads and Ngons (last ones produce interesting/strange results):
image

Mask can be used to control transformation mode:

image

vicdoval:
Could anyone give me a hint of why the CI is failing? i found an error in the bl_idname of trim_and_tesselate (@portnov) but maybe I should leave it like was to avoid layout breaks because does not seem the cause

portnov:
@vicdoval did you re-save json files, or just replaced Mk2 with Mk3?
I tried to import `Adaptive_Voronoi_panel.json` from this PR, it lacks some node links, and also the node is not configured properly. When I reconnected nodes correctly and re-exported the tree, this test stopped failing...

portnov:
The error message is still confusing...

vicdoval:
Thanks @portnov. You were right the key was in the examples, they broke with the socket renaming.
Btw the error in the bl_idname of trim_and_tesselate is still there

vicdoval:
Custom Normals
image
Edges channel
image

Durman:
Also if you are interesting in adding new stuff in the node it would be useful to make the node dealing with UW mappings.

vicdoval:
With "As is" in Coordinates it does UV mapping but for -0.5 to 0.5
image

enzyme69:
@Durman Do you have the building example Blend with adaptive polygon? I am curious about the building and the logic.

Durman:
[This file](https://github.com/nortikin/sverchok/files/5820370/Simple.building.creation.zip) I used in [this some sort of a tutorial](https://youtu.be/BYgLlVJWN9c).

enzyme69:
Screen Shot 2021-01-18 at 12 53 30 pm

Try SV preferences and click "Check for new version" let it updates and restart Blender / F8 refresh addon.

PR #3800, opened 2020-12-28
This should fix #3799.

### LINKED ISSUE #3799: Presets menu Bug
Failing the tests in #3798 made me check that it may be related to the Presets in the shift+a menu 
Every node appears there with a empty menu. In this the expected behaviour?
image

portnov:


vicdoval:
This fixes the issue, one idea would we to display in the menu the node name in stead of the bl_idname

portnov:
I wanted to do it from the beginning, but ran into some problem related to order of imports. Something like menu class being initialized before node classes are loaded.

vicdoval:
I don't know where to place it but this could do the job

```
def bl_idname_to_name(s):
    s1= ''.join([s,s[-1]])
    name =''.join([' '+c if (c.isupper() and c1.islower()) else c for c,c1 in zip(s,s1[1:])]).lstrip('Sv ')
    v = name.find('Ex ')
    if v == 0:
        name= name.lstrip('Ex ')
    v = name.find(' Mk')
    if v !=-1:
        name= name[:v]
    v = name.find('MK')
    if v !=-1:
        name= name[:v]
    v = name.find(' Node')
    if v !=-1:
        name= name.replace(' Node', '')
    return name

print(bl_idname_to_name("SvBoxNodeMk2"))
print(bl_idname_to_name("SvExSurfaceFormulaNode"))
print(bl_idname_to_name("SvExtrudeEdgesNodeMk2"))
print(bl_idname_to_name("SvOBJRayCastNodeMK2"))
```
prints:
```
Box
Surface Formula
Extrude Edges
OBJ Ray Cast
```

Durman:
Why not doing this?
```Python
import bpy
from typing import Optional

def bl_name_to_name(bl_name: str) -> Optional[str]:
    cls = bpy.types.Node.bl_rna_get_subclass_py(bl_name)
    return cls.bl_label if cls else None

print(bl_name_to_name('SvBoxNodeMk2'))
```
```
Box
```

PR #3805, opened 2021-01-02
A proposal to add examples files and relative links to local imgs and files

theoryshaw:
A proposal to add...
- examples files
- relative links to local imgs and files


vicdoval:
The problem with this idea is that it adds 1.8 MB per node (and we have 657 nodes)

nortikin:
jsons will be nice. anyway good proposal

theoryshaw:
maybe a submodule?  Defer to you.

PR #3807, opened 2021-01-04
"Relax mesh" node

portnov:
This node moves vertices of the input mesh, in order to make it more "relaxed",
i.e. for mesh elements to have more even distribution in some sense. There are
several algorithms supported, each of which has different definition of what "relaxed" is:

* Lloyd-based algorithm. This is created after well-known Lloyd algorithm. For
  each vertex, find centers of all incident faces, and then find the average of
  those points; this will be the new location of the vertex. Effectively, this
  algorithm tries to make each face as close to circle as possible. This
  algorithm shows it's best for meshes that consist of Tris.
* Edge lengths. Scale each edge up or down, trying to make all edges of the
  same length. Target edge length can be minimum, maximum or average of all
  lengths of edges of the source mesh.
* Face areas. Scale each face up or down, trying to make all faces of the same
  area. Target face area can be minimum, maximum or average of all areas of
  faces of the source mes.

These algorithms can change the overall shape of the mesh a lot. In order to
try to preserve original shape of the mesh at least partially, the following
methods are supported:

* "Linear" method is supported for Lloyd algorithm only. When the algorithm has
  found the new location of the vertex, put it at the same distance from the
  plane where centers of incident faces lie, as the original vertex was. This
  method can be slow for large meshes.
* "Tangent": move vertices along tangent planes of original vertexes only (i.e.
  perpendicular to vertex normal).
* "BVH": use BVH tree to find the nearest point to the newly calculated vertex
  on the original mesh.

Screenshot_20210104_140240

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/nodes/modifier_change/triangulate.py
+++ b/nodes/modifier_change/triangulate.py
@@ -22,6 +22,7 @@ import bmesh.ops
 
 from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode, match_long_repeat, repeat_last_for_length
+from sverchok.utils.sv_mesh_utils import get_unique_faces
 from sverchok.utils.sv_bmesh_utils import bmesh_from_pydata, pydata_from_bmesh
 
 
@@ -124,6 +125,8 @@ class SvTriangulateNode(bpy.types.Node, SverchCustomTreeNode):
                 new_vertices, new_edges, new_faces = pydata_from_bmesh(bm)
                 new_face_data = []
             bm.free()
+
+            new_faces = get_unique_faces(new_faces)
 
             result_vertices.append(new_vertices)
             result_edges.append(new_edges)
--- a/docs/nodes/modifier_change/modifier_change_index.rst
+++ b/docs/nodes/modifier_change/modifier_change_index.rst
@@ -10,6 +10,7 @@ Modifier Change
    subdivide_lite
    unsubdivide
    smooth
+   relax_mesh
    delete_loose
    edges_intersect_mk2
    poke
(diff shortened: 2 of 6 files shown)


nodes/modifier_change/triangulate.py
```diff
@@ -125,6 +126,8 @@ def process(self):
                 new_face_data = []
             bm.free()
 
+            new_faces = get_unique_faces(new_faces)
```
vicdoval, line 129: @portnov why did you add this line?, it slows a lot the node

portnov: I don't remember exactly, there was some node which created meshes with duplicate faces, which broke the node...

vicdoval: So you want me to leave it as an option or just delete it?

portnov: Probably the correct option would be to remove it from here, and to create a separate node for "fixing any mesh"...

Durman: > Probably the correct option would be to remove it from here, and to create a separate node for "fixing any mesh"...

Yes, there is even request for such node. #3989

zeffii: i think such node might want to also output the indices from the original face list, which it has discarded

Durman: I thought it should remove only repeated faces and edges. Bmesh module should be able to handle repeated vertices (with the same coordinates). Such vertices can be removed by 'Merge by distance' if needed.

PR #3809, opened 2021-01-05
Changing from fraction import gcd (obsolete) from math import gcd pyt…

victorcalixto:
…hon 3.9

`from fractions import gcd` in /nodes/number/scalar_mk4.py breaking the sverchok installation on blender 2.9.1 with python 3.9

## Solution description
Changing to  `from math import gcd` solved the problem

### DIFF
--- a/nodes/number/scalar_mk4.py
+++ b/nodes/number/scalar_mk4.py
@@ -16,8 +16,7 @@
 #
 # ##### END GPL LICENSE BLOCK #####
 
-from math import pi, e
-from fractions import gcd
+from math import pi, e, gcd
 
 import bpy
 from bpy.props import EnumProperty, FloatProperty, IntProperty, BoolProperty


nortikin:
is it official blender 2.90.1 with python 3.9?

vicdoval:
The default python is 3.7.7 but `from math import gcd` works.
The problem is, as stated [here](https://stackoverflow.com/questions/49981286/replacement-for-deprecated-fractions-gcd-function), `math gcd` does not support rational numbers and that could break some layouts.

victorcalixto:
> is it official blender 2.90.1 with python 3.9?

The default blender python in GNU/Linux repositories is the default python of the system.

vicdoval:
Maybe this is the best solution

```
try:
    from fractions import gcd
except ImportError:
    from math import gcd
```

vicdoval:
@victorcalixto thanks for the PR 😄 
Finally solved by adding the original gcd function to sverchok.utils.math at #3825

PR #3814, opened 2021-01-08
Fixing CI

vicdoval:
Looking into the failed CI


vicdoval:
The CI is passed but there are still many errors that could be checked

PR #3829, opened 2021-01-14
Edge split mk3

portnov:
This provides Mk3 of "Edge Split" node.

"Mirror" option is replaced with "Mode" parameter:

* Simple - corresponds to  old "mirror" parameter unchecked, this is the default option
* Mirror - corresponds to old "mirror" parameter checked
* Multiple - divides each edge in arbitrary number of pieces, defined by new "Cuts" input.

Screenshot_20210114_224155

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/modifier_change/edge_split.rst
+++ b/docs/nodes/modifier_change/edge_split.rst
@@ -4,12 +4,13 @@ Split Edges
 Functionality
 -------------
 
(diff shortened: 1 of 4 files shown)


Durman:
Another way

PR #3839, opened 2021-01-22
example handler augmentation

zeffii:
adds the possibility to perform a state test in script notes / exec nodes, useful when the user wants to execute bpy.ops

then you would write
```python
import sverchok
if sverchok.core.update_system.is_postload_complete():
    bpy.ops.render.render(use_viewport=True)
```

untested..

maybe it needs also some throttling..
```python
import sverchok
if sverchok.core.update_system.is_postload_complete():
    ng = self.id_data
    with ng.throttle_update():
        try:
            bpy.ops.render.render(use_viewport=True)
        except:
            pass
```

nope.. also doesn't work,  at least.. not here.

and also not..
```python
import sverchok
if sverchok.core.update_system.is_postload_complete():
    ng = self.id_data
    with ng.throttle_update():
        try:
            bpy.ops.render.render(use_viewport=True)
            print("rendering new image")

        except:
            print(" ")
```

### DIFF
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -6,7 +6,7 @@ from bpy.app.handlers import persistent
 from sverchok import old_nodes
 from sverchok import data_structure
 from sverchok.core import upgrade_nodes, undo_handler_node_count
-from sverchok.core.update_system import set_first_run, clear_system_cache
+from sverchok.core.update_system import set_first_run, clear_system_cache, set_postload_completion_state
 from sverchok.core.events import CurrentEvents, BlenderEventsTypes
 from sverchok.ui import color_def, bgl_callback_nodeview, bgl_callback_3dview
 from sverchok.utils import app_handler_ops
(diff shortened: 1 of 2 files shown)


vicdoval:
maybe 
`sverchok.handlers.pre_running `
or 
`sverchok.update_system.sv_first_run`
could do the job?

Durman:
It is also possible to use timers. The group nodes actually don't need throttling at all.
Update system of groups works similar how update system of Sverchok was working before 2.8 Blender version. In that time we didn't have need in using the throttling mode.
If I ever starts refactoring current update system I'm going to remove throttling mode.

The function below is listening changes in group trees and reevaluating them if nessecary.

https://github.com/nortikin/sverchok/blob/ddb124246cfecba7994487e490ec6642377129a9/core/group_handlers.py#L81-L93
Hope this information can be useful.

PR #3842, opened 2021-01-23
string to color and vector to color conversions

vicdoval:
String to Color and Vector to color conversion:
image
image

- [x] Ready for merge.


nortikin:
lovely

PR #3849, opened 2021-01-26
solidify mk2

vicdoval:
Faster than the previous version (1.5x)
Adds variable thickness
Uneven/Even thickness 
and Offset (standard -1 to 1 like blender modifier)

image

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


nodes/modifier_make/solidify_mk2.py
```diff
@@ -0,0 +1,169 @@
...
+def create_edges(bm, v_len):
+    edges =[]
+    for edge in bm.edges[:]:
+        edges.append([v.index for v in edge.verts[:]])
+
+    edges_new = [[e[0]+v_len, e[1]+v_len] for e in edges]
+    side_edges = [[v.index, v.index+v_len] for v in bm.verts if v.is_boundary]
+    return edges + edges_new +side_edges
+
+def solidify(vertices, faces, thickness, offset=[-1], even=True):
```
portnov, line 37: Use of mutable objects as default values of method parameters can lead to unexpected behaviour in python.

nodes/modifier_make/solidify_mk2.py
```diff
@@ -0,0 +1,169 @@
...
+
+        vertices_out = vertices +  new_verts
+    else:
+        if even:
+            new_verts=[(Vector(v.co) + v.normal * t * v.calc_shell_factor() *((-o/2)+0.5))[:] for v, t, o in zip(bm.verts, cycle(thickness), cycle(offset))]
+            old_verts=[(Vector(v.co) - v.normal * t * v.calc_shell_factor() * ((o/2)+0.5))[:] for v, t, o in zip(bm.verts, cycle(thickness), cycle(offset))]
+
+        else:
+            new_verts=[(Vector(v.co) + v.normal * t * ((-o/2) + 0.5))[:] for v, t, o in zip(bm.verts, cycle(thickness), cycle(offset))]
+            old_verts=[(Vector(v.co) - v.normal * t * ((o/2) + 0.5))[:] for v, t, o in zip(bm.verts, cycle(thickness), cycle(offset))]
```
portnov, line 64: 1. I don't think Vector(v.co) does something useful comparing to just v.co
2. Maybe use of numpy will speed up calculations here?

portnov:
I'd suggest to add an option to use old logic. Probably it behaves differently in some edge cases.

nodes/modifier_make/solidify_mk2.py
```diff
@@ -0,0 +1,169 @@
...
+
+            if not res:
+                return
+            verts_out.append(res[0])
+            edges_out.append(res[1])
+            polys_out.append(res[2])
+            new_pols.append(res[3])
+            rim_pols.append(res[4])
+            pols_groups.append(res[5])
+            new_verts_mask.append(res[6])
```
Durman, line 153: Just more short variant. I'm using it in some nodes.
```Python
        res = []

        for v, p, t, o in zip_long_repeat(verts, polys, thickness, offset):
            res.append(solidify(v, p, t, o, self.even))

        verts_out, edges_out, polys_out, new_pols, rim_pols, pols_groups, new_verts_mask = zip(*res)
```

vicdoval: Have you done any performance test? Just wondering because append is such a slow operation...

nodes/modifier_make/solidify_mk2.py
vicdoval: 1. Totally agree, i thought it was giving a tuple, not a vector
2. I was testing it but it seems that acceding many times to bm.verts to create the arrays takes out the benefit of having the arrays

nodes/modifier_make/solidify_mk2.py
Durman: I don't know. Getting maximum speed of every single piece of code is not the way of program optimization anyway. It should be done only in bottleneck places all others should be written with priority of readability.

vicdoval:
For Docs
image

PR #3879, opened 2021-02-06
Scripting nodes to Script category

vicdoval:
image
Anyone against this PR?


zeffii:
maybe `mesh eval` too?
`profile mk3`

vicdoval:
I was thinking about those nodes too...
image
Any other candidate?

portnov:
Do we really support "SN Functor B"? Maybe it should go to Alpha, or even to Old nodes?

zeffii:
snfunctor b, should indeed perhaps be demoted to Alpha. I never got around to pumping real time into that.

nortikin:
cool. it makes more sence now
move recieve from sorcar to exchange section!

PR #3886, opened 2021-02-08
socket menus to rclick menu

vicdoval:
On my experience with socket menus I find a couple inconveniences:
Generally In a nodetree I use them in one or two nodes per setup and i don't like how they fill the working space.
Also being a nodetree property makes that you have to enable them per nodetree...

My idea/proposal is to have them accesible in the right-click menu.
image

But i cant insert the menus  in the submenu...
I have gotten this far:
image
Which works but is not ideal specially for nodes with many outputs... 

What do you guys think? 

Anyone could give me hint about how to insert each socket menu in a submenu?

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/core/sockets.py
+++ b/core/sockets.py
@@ -85,8 +85,10 @@ class SV_MT_AllSocketsOptionsMenu(bpy.types.Menu):
 
         for s in node.outputs:
             if hasattr(s, 'draw_menu_items'):
-                layout.label(text=s.name)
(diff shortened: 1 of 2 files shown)


vicdoval:
Now it is working as expected:
image

PR #3888, opened 2021-02-09
Rotation Field

vicdoval:
image

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/field/field_index.rst
+++ b/docs/nodes/field/field_index.rst
@@ -7,6 +7,7 @@ Field
 
    coordinate_scalar_field
    attractor_field_mk2
+   rotation_field
    scalar_field_formula
    vector_field_formula
    noise_vfield
@@ -32,4 +33,3 @@ Field
    scalar_field_eval
    vector_field_eval
    vector_field_apply
-
--- a/docs/nodes/field/rotation_field.rst
+++ b/docs/nodes/field/rotation_field.rst
@@ -1,0 +1,64 @@
+Rotation Field
+==============
+
+Functionality
+-------------
+
+This node generates a Vector Field which is calculated as
+force rotating points around some axis. Several physics-like falloff laws are supported.
+Falloffs similar to standard proportional editing mode are supported too (they
+are marked with `(P)` in the name).
+
+Inputs
+------
+
+This node has the following inputs:
+
+* **Location**. Point on the rotating axis
+
+  It is possible to provide several location points. The default value is `(0, 0, 0)` (origin).
+
+* **Direction**. Direction of the rotation axis.  The default value is `(0, 0, 1)` (Z axis).
+
+* **Amplitude**. The attracting force amplitude. The default value is 0.5.
+* **Coefficient**. The coefficient used in the attracting force falloff
+  calculation formula. The exact meaning of this input depends on fallof type:
+
+   * If **Falloff type** is set to **Inverse exponent** or **Gauss**, then this
+     is the coefficient K in the corresponding formula: ``exp(-K*R)`` or
+     ``exp(-K*x^2/2)``.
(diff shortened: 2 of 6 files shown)


portnov:
Do I understand correctly that this rotates always by 90 degrees?

vicdoval:
The rotation force is always perpendicular to the line attraction force, there is not angular control, to rotate by a angle then a "Matrix Field" is more convenient. The amplitude and coefficient is related to the fallof...

portnov:
So, Amplitude = 1 will mean 45 degree rotation?

nortikin:
> So, Amplitude = 1 will mean 45 degree rotation?

on what distance?

summereasy:
> The rotation force is always perpendicular to the line attraction force, there is not angular control, to rotate by a angle then a "Matrix Field" is more convenient. The amplitude and coefficient is related to the fallof...

Interesting 👍 
Since it is a Field, I think it should represent some **Rotation Speed Vector** depends on the fall of the Distance? Correct me if I'm wrong.
The thing is, in physics, if you apply some Attraction Force at the Center Point, say, **Gravity** or **Electrical-Magnetic Force** (which follows an Inverse-Square Law) you should get an "angular speed" or say "Rotation Vector" proportional to r^(2/3) if I am right. So, my question is, is that r^(2/3) fall off even in the option list ... 👯👯👯

portnov:
Usual physics law is "inverse square", i.e. "1/r^2". I believe there is such an option.

summereasy:
> Usual physics law is "inverse square", i.e. "1/r^2". I believe there is such an option.

Yes, but Inverse-Square is more of a Radial Loss, like Energy/Gravity/Electrical Field Type of thing. **But it is really rare for rotational vector fall-offs.**

I just did a quick check. Due to the Inverse Square Law of Gravity :

equation-4828

Which T is the period and **ω is the angular speed**, or in our case, the actual **Rotational Vectors**.
Reference Link: [Kepler's Motion](https://en.wikipedia.org/wiki/Kepler%27s_laws_of_planetary_motion)

portnov:
Well then, @vicdoval would you add this R^(-3/2) fallof option? :)

vicdoval:
I'm glad we have some expertise feedback around :)

I always thought that Inverse Exponent was `1/(R^Exponent) ` now, checking the code I see it is `E^(-Exponent*R)` so ok I will implement R^(-3/2).

How should it be called: Kepler? Rotational Fall-off? Planetary motion?? :)

Durman:
Am I understand correctly that such node is stepping in physical simulation domain?

vicdoval:
@Durman I think all the Fields nodes step in physical simulation

summereasy:
@Durman 
> Am I understand correctly that such node is stepping in physical simulation domain?

Actually ... I just found the rotation field idea is interesting, so I just asked ... :)

The Kepler Motion is only an example of Rotation Falloff guiding by an Inverse-Square-Law Force. 
Since the math is identical, any energy-like or force-like radial falloff should present the same Rotation Field falloff.
And if you expend the idea, since it is mathematically related, Say, you got a Vector Field input for some kind of Arbitary Surface, its Divergence / Curl / Laplacian will just follow the same way as Gravity did, as long as your input is **Inverse-Square-like**. 

So I don't think it is just for physics. It is more of math.
I think there should be more examples out there, but sadly I'm not an expert and can only leave it to others.

@vicdoval 
> How should it be called: Kepler? Rotational Fall-off? Planetary motion?? :)

That is too much though 😄 , Thanks for considering it. As to the name, I think It's your choice.
If you ask me, I will just keep R^(-3/2) for its general use purpose.

nortikin:
some time ago found out corelian forces, maybe it is concern

PR #3903, opened 2021-02-14
Data block list

### LINKED ISSUE #3873: bpy.data list node
## Problem statement

I'm thinking of creating new node with list UI with elements of Blender data blocks like objects, materials, collections and cetera. There are already several nodes which use list UI like that - `objects in` and set `assign material list` nodes. It would be useful to generalize this idea in the separate node. The node won't have any inputs. The only node parameter should be type of `data blocks` collection from which to choose elements.

```text
+-----------------+
|            data o
|                 |
| type of data ^  |
| --------------- |
| | item 1    - | |
| | item 2    - | |
| |      +      | |
| --------------- |
+-----------------+
```

Right now I'm not sure which behavior will be more convenient in case of changing type of data block when list of elements is not empty. It is possible to reset whole list, to mark elements with wrong type as invalid or do nothing.

Durman:

closes #3873

data block list npde

Welcome suggestions if any

## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Ready for merge.

### DIFF
--- a/index.md
+++ b/index.md
@@ -622,6 +622,7 @@
     SvPointOnMeshNodeMK2
     SvOBJRayCastNodeMK2
     SvSCNRayCastNodeMK2
+    SvBlenderDataListNode
 
 ## Scene
     SvObjectsNodeMK3
--- a/nodes/object_nodes/blender_data_list.py
+++ b/nodes/object_nodes/blender_data_list.py
@@ -1,0 +1,132 @@
+# This file is part of project Sverchok. It's copyrighted by the contributors
+# recorded in the version control history of the file, available from
+# its original location https://github.com/nortikin/sverchok/commit/master
+#
+# SPDX-License-Identifier: GPL3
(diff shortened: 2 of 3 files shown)


Durman:
The problem of the Object socket is that it keeps straight reference to a data block which can become invalid in any moment. We have quite a few input nodes with such socket this is probably why it's not big problem now. It looks like that Blender devs did something and calling method of such objects doesn't cause a crash as it was before. Now it rises the references error. 

https://user-images.githubusercontent.com/28003269/109138279-a1893400-7773-11eb-9dee-cabdd5d7b31f.mp4

Solution could be instead of keeping straight references to keep meta information for searching a data block. Obvious disadvantage of this is that the search will take O(len(data block collection)) each time. To make it more efficient will be tricky task. Probably, a socket ,before `sv_get` method, should check whether reference to an object is valid and if not search the data block in appropriate collection.

PR #3909, opened 2021-02-17
Mesh viewer ensure list

vicdoval:
Mesh Viewer fails when inputting numpy arrays so i added a check to convert the inputs to list if needed
- [x] Ready for merge.

### DIFF
--- a/data_structure.py
+++ b/data_structure.py
@@ -235,7 +235,7 @@ def repeat_last_for_length(lst, count, deepcopy=False):
     else:
         for i in range(count - n):
             result.append(x)
-            
+
     return result
 
 def cycle_for_length(lst, count):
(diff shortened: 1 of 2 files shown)


Durman:
It's actually was supporting np.arrays of vectors which increased performance significantly. See the comment.

https://github.com/nortikin/sverchok/pull/3475#issuecomment-680010778

vicdoval:
But was failing if merge was enabled.
This solution is not ideal but more robust... feel free to remove the check if you manage to do it numpy-ready

Durman:
It would be better to create bug report instead if you are not ready to supply normal solution.

vicdoval:
> It's actually was supporting np.arrays of vectors which increased performance significantly. See the comment.
> 
> [#3475 (comment)](https://github.com/nortikin/sverchok/pull/3475#issuecomment-680010778)

I was not aware of this feature and i could only see the node was failing when numpy arrays were given so thought this was a "normal solution"

Issue #3922, opened 2021-02-22
Blender 2.93 Alpha af940c68cbed (Linux)  cannot activate Sverchok Addon


meikasahara:
The following error appears upon attempting to activate the latest Sverchok (from sverchok-master.zip from github,) on the latest Blender build:

`/home/vdb/.config/blender/2.93/scripts/addons/sverchok-master/settings.py:16: RuntimeWarning: Use 'sys.executable' instead of 'binary_path_python'!
  PYPATH = bpy.app.binary_path_python
sv: import all modules
/home/vdb/.config/blender/2.93/scripts/addons/sverchok-master/settings.py:16: RuntimeWarning: Use 'sys.executable' instead of 'binary_path_python'!
  PYPATH = bpy.app.binary_path_python
Traceback (most recent call last):
  File "/home/vdb/locald/blender-2.93.0-af940c68cbed-linux64/2.93/scripts/modules/addon_utils.py", line 351, in enable
    mod = __import__(module_name)
  File "/home/vdb/.config/blender/2.93/scripts/addons/sverchok-master/__init__.py", line 75, in <module>
    imported_modules = init_architecture(__name__, utils_modules, ui_modules)
  File "/home/vdb/.config/blender/2.93/scripts/addons/sverchok-master/core/__init__.py", line 115, in init_architecture
    import_all_modules(imported_modules, mods_bases)
  File "/home/vdb/.config/blender/2.93/scripts/addons/sverchok-master/core/__init__.py", line 100, in import_all_modules
    import_modules(mods, base, imported_modules)
  File "/home/vdb/.config/blender/2.93/scripts/addons/sverchok-master/core/__init__.py", line 70, in import_modules
    im = importlib.import_module('.{}'.format(m), base)
  File "/home/vdb/locald/blender-2.93.0-af940c68cbed-linux64/2.93/python/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/home/vdb/.config/blender/2.93/scripts/addons/sverchok-master/core/sockets.py", line 1463, in <module>
    register, unregister = bpy.utils.register_classes_factory(classes + list(socket_interface_classes()))
  File "/home/vdb/.config/blender/2.93/scripts/addons/sverchok-master/core/sockets.py", line 1447, in socket_interface_classes
    prop_func, prop_args = socket_cls.__annotations__['default_property']
[...]

PR #3925, opened 2021-02-23
Prop and args change

zeffii:

blender 2.93 (feb 21) changed the behaviour of an undocumented portion of the bpy API. this patch aims to let sverchok work in Blender 2.93 before _and_ after that change, by providing a function that detects if the new API is present or not. The function uses the old undocumented "unpacking" method if it doesn't find the new api, and it uses direct property lookup `.function` and `.keywords` if it can.


zeffii:
this resolves https://github.com/nortikin/sverchok/issues/3922

zeffii:
just this part needs to be dynamic too..

`handling_nodes.py`

```python
blender_properties = {
    # properties are functions which return tuples with themselves as first argument
    # it should help to rebuild properties with new arguments
    bpy.props.BoolProperty().function: bpy.props.BoolProperty,
    bpy.props.BoolVectorProperty().function: bpy.props.BoolVectorProperty,
    bpy.props.CollectionProperty().function: bpy.props.CollectionProperty,
    bpy.props.EnumProperty().function: bpy.props.EnumProperty,
    bpy.props.FloatProperty().function: bpy.props.FloatProperty,
    bpy.props.FloatVectorProperty().function: bpy.props.FloatVectorProperty,
    bpy.props.IntProperty().function: bpy.props.IntProperty,
    bpy.props.IntVectorProperty().function: bpy.props.IntVectorProperty,
    bpy.props.PointerProperty().function: bpy.props.PointerProperty,
    bpy.props.StringProperty().function: bpy.props.StringProperty
}
```

zeffii:
i've tested this now on 2.92 and the latest 2.93 seems to work OK.

PR #3929, opened 2021-02-25
attempt regex style parsing of snlite directive

zeffii:
makes it possible to use single or double qoute for the directive, at least .. that's the plan.


zeffii:
ahha!  it's not finding a directive in some of the tests!

zeffii:
looks like users also want to include import statements before the directive.  This is not a big problem, but i'll need to make the regex aware. .. f.ex
```python
from sverchok.utils.geom import LinearSpline, CubicSpline
from sverchok.utils.math import supported_metrics
from sverchok.utils.curve import SvSplineCurve

"""
in numbr s d=[[]] n=0
out splin C
out verts v
"""

# '''use note node to define toe
# l - left, r - right, t - top,
# bake, convert to curve and
# extrude 0.3
# Start always from l finish with t
# and finish inside out
...

```

portnov:
It would be nice to write a parser based on `sverchok.utils.parsec`. It would be more flexible (easily allow any number of spaces, for example), and more readable. And probably error handling would be better (not that it would be very good, but better than none).

zeffii:
@portnov  OK, but not yet :(  on top of all the crazy stuff at the moment, my mother fell and broke her shoulder, this means even less time to concentrate on this.  

I did however seem to find a simple `directive` regex. The content of the directive would be indeed nice to formalize instead of a if else.. but if else also has its benefits for simple people like me :)

zeffii:
```regex
\+in\s+(\w+)\s+(\w+)\s+d=(.+)\s+n=(\d)(\s+#\sname=.+)?
```
captures: 
```console
+in named_socket v d=[[0, 0, 0]] n=2 # name="Some Name"
+in named_socket s d="some content" n=2
+in named_socket v d=[[0, 0, 0]] n=2
```

using the `+` qualifier can use an alternative socket info parser,, which will allow for more lenient defaults,,  but it will force the convention of using `d=<  >`, `n=number` and optional `(# name="display name")`

PR #3950, opened 2021-03-04
Update raycaster_lite.py

vktcl:

Added ray detection maximum distance limit parameter. Used to control the length of each light

### DIFF
--- a/nodes/analyzer/raycaster_lite.py
+++ b/nodes/analyzer/raycaster_lite.py
@@ -31,7 +31,7 @@ class SvRaycasterLiteNode(bpy.types.Node, SverchCustomTreeNode):
     bl_icon = 'OUTLINER_OB_EMPTY'
     sv_icon = 'SV_RAYCASTER'
     
-    MaxDistance: bpy.props.FloatProperty(name='MaxDistance', description='MaxDistance', default=1.0, update=updateNode)
+    max_distance: bpy.props.FloatProperty(name='MaxDistance', description='MaxDistance', default=1.0, update=updateNode)
     start: bpy.props.FloatVectorProperty(default=(0,0,0), size=3, update=updateNode)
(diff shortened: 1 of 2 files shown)


nodes/analyzer/raycaster_lite.py
```diff
@@ -30,7 +30,8 @@ class SvRaycasterLiteNode(bpy.types.Node, SverchCustomTreeNode):
     bl_label = 'Raycaster'
     bl_icon = 'OUTLINER_OB_EMPTY'
     sv_icon = 'SV_RAYCASTER'
-
+    
+    MaxDistance: bpy.props.FloatProperty(name='MaxDistance', description='MaxDistance', default=1.0, update=updateNode)
```
Durman, line 34:
```suggestion
    max_distance: bpy.props.FloatProperty(name='MaxDistance', description='MaxDistance', default=1.0, update=updateNode)
```

vktcl: max_distance Variable already exists

Durman: `max_distance` or `MaxDistance` is not a variable it is an attribute of the node class. Also this class has `start`, `direction` and no more other attributes. According Python convention names of a class should be in snake case. Sverchok code is not totally written according to this convention but it's better always to do for new code.

nodes/analyzer/raycaster_lite.py
```diff
@@ -40,6 +41,7 @@ def sv_init(self, context):
 
         si('SvVerticesSocket', 'Verts')
         si('SvStringsSocket', 'Faces')
+        si('SvStringsSocket', 'MaxDistance').prop_name = 'MaxDistance' #添加输入最大距离接口
```
Durman, line 44: This name also should be fixed

vktcl: There are some errors and I need to close the merge again

Issue #3960, opened 2021-03-07
WISHLIST: SV Custom SET and GET Attribute for Geometry Nodes instancing


enzyme69:
Snapshot:
<img width="853" alt="Screen Shot 2021-03-08 at 12 40 30 am" src="https://user-images.githubusercontent.com/3952687/110241805-f2383280-7fa6-11eb-8382-7c6e62435a2c.png">

As shown by Bradley Animation with Animation Nodes here:
https://www.youtube.com/watch?v=EmMyxSRGhRc

I am wondering if SV can support the creation of Custom Attribute (Points) and fill it values so that Geometry Nodes can understand the value and using it for instancing.

vicdoval:
If it can be done in python then it can be done in sverchok

enzyme69:
Yes, I am trying to dig more into it. Bradley Animation is using AN version with some extra nodes from Kuldeep Singh to set custom attribute.

enzyme69:
In the past, we have already Attribute "Col". But in recent Blender there is also "Geometry" etc which I hardly used. But probably related to Geometry Nodes instancing.

<img width="935" alt="Screen Shot 2021-03-08 at 1 08 48 am" src="https://user-images.githubusercontent.com/3952687/110242579-d898ea00-7faa-11eb-82d2-be484027ffb3.png">


https://docs.blender.org/manual/en/latest/render/shader_nodes/input/attribute.html

enzyme69:
Currently reading this:
https://docs.blender.org/manual/en/dev/modeling/geometry_nodes/attributes_reference.html

Durman:
This is looks like [original code](https://github.com/3DSinghVFX/animation_nodes/blob/extranodes/animation_nodes/nodes/object/set_custom_attribute.py) of the node.

<details> 
  <summary>code</summary>  

```py

        if object is None: return object
        if object.type != "MESH": self.raiseErrorMessage("Object should be Mesh type.")
        if attName == "": self.raiseErrorMessage("Attribute name can't be empty.")

        attribute = object.data.attributes.get(attName)
        if attribute is None:
            attribute = object.data.attributes.new(attName, self.dataType, self.domain)
        elif attribute.data_type != self.dataType or attribute.domain != self.domain:
            object.data.attributes.remove(attribute)
            attribute = object.data.attributes.new(attName, self.dataType, self.domain)

        if self.domain == "POINT":
            amount = len(object.data.vertices)
        elif self.domain == "EDGE":
[...]

PR #3969, opened 2021-03-13
Set attribute

Durman:

#3960

test

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done.
- [x] Ready for merge.


Durman:
There are some problems at the current stage. 

- If rename attribute a mesh will get two attributes with old and new names. 
- If change input object old object has remained with node attribute. 
- If node overrides attribute which was set by another `Set attribute` node it remains overridden until that node reevaluation.

Not sure yet whether it worth efforts to fix this. It won't be easy. The only solution I can think of for now is to keep pointers to input objects in each node to compare them with new input objects and fixing there attributes if necessary.

PR #3991, opened 2021-03-27
matrix_out: nesting  fix + recursive node

vicdoval:
fixes matrix_out_mk2 nesting and adds recursiveNode mixin to accept lists of list of matrixes

- [x] Ready for merge.

### DIFF
--- a/nodes/matrix/matrix_out_mk2.py
+++ b/nodes/matrix/matrix_out_mk2.py
@@ -21,6 +21,7 @@ from bpy.props import EnumProperty
 from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode
 from sverchok.utils.sv_transform_helper import AngleUnits, SvAngleHelper
+from sverchok.utils.nodes_mixins.recursive_nodes import SvRecursiveNode
 from mathutils import Matrix
 
 
@@ -37,7 +38,7 @@ output_sockets = {
 }
 
 
-class SvMatrixOutNodeMK2(bpy.types.Node, SverchCustomTreeNode, SvAngleHelper):
+class SvMatrixOutNodeMK2(bpy.types.Node, SverchCustomTreeNode, SvAngleHelper, SvRecursiveNode):
     """
     Triggers: Matrix, Out
     Tooltip: Convert a matrix into its location, scale & rotation components
@@ -79,7 +80,7 @@ class SvMatrixOutNodeMK2(bpy.types.Node, SverchCustomTreeNode, SvAngleHelper):
         items=mode_items, default="AXISANGLE", update=update_mode)
 
     def sv_init(self, context):
-        self.inputs.new('SvMatrixSocket', "Matrix")
+        self.inputs.new('SvMatrixSocket', "Matrix").is_mandatory =True
         # translation and scale outputs
         self.outputs.new('SvVerticesSocket', "Location")
         self.outputs.new('SvVerticesSocket', "Scale")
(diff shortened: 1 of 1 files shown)


summereasy:
I got this error message just after the update.
temp

vicdoval:
I can't reproduce it, i think there was a period in which the matrix out (mk2) node was producing a wrong output and if your node is from that  time  it may complain now , just replace it with a new matrix out and all should be fine
image

Durman:
Does not vertices have too much nested levels now?

vicdoval:
The Matrix node of the example had the Flat Output un-checked

Durman:
I would not expect to get more nested levels any way. I think list of matrixes should produce plain list of vertices.

vicdoval:
If [m, m] produces [[v], [v]] (most usual case)
Then [[m,m]] should produce [[[v], [v]]]. Any other output can be generated with socket post-processing

vicdoval:
@Durman I see your point... i will change it then

summereasy:
> I can't reproduce it, i think there was a period in which the matrix out (mk2) node was producing a wrong output (like a couple of weeks) and if your node is from that time it may complain now , just replace it with a new matrix out and all should be fine

Thanks for your quick response!
I'm pretty sure I have the latest version (just had re-download the zip again)
I am not a good programmer though, but as far as I can see, the problem I encountered is from here:

**1. The newer version 3 days ago has changed a bit for the input of matrix_out Node:**
```python
[removed]  input_M = self.inputs['Matrix'].sv_get()
[added]    input_M = params[0]
```
**2. I printed out 2 different input_M as a test:** 
- I tried several modes like "Flat/Unwrapped/Wrap" etc
- The original `input_M = params[0]` can give me correct nest-level of input of Matrix
But the newer version cannot - it stays at 2-level brackets:
```python
# FLAT MODE
#[NEW VERSION] input_M = params[0] output:
input_M = [[Matrix(((1.0, 0.0, 0.0, 0.0),
        (0.0, 1.0, 0.0, 0.0),
        (0.0, 0.0, 1.0, 0.0),
        (0.0, 0.0, 0.0, 1.0)))]]

#[OLD VERSION] input_M2 = self.inputs['Matrix'].sv_get()
input_M2 = [Matrix(((1.0, 0.0, 0.0, 0.0),
        (0.0, 1.0, 0.0, 0.0),
        (0.0, 0.0, 1.0, 0.0),
        (0.0, 0.0, 0.0, 1.0)))]
```
- so as the data pass into the for loop I got this wired error message
```
2021-03-31 08:58:39,157 [ERROR] sverchok.core.update_system: Node Matrix Out had exception: 'list' object has no attribute 'decompose'
  File "/Users/zhaowei/Library/Application Support/Blender/2.92/scripts/addons/sverchok-master/core/update_system.py", line 384, in do_update_general
    node.process()
  File "/Users/zhaowei/Library/Application Support/Blender/2.92/scripts/addons/sverchok-master/utils/nodes_mixins/recursive_nodes.py", line 125, in process
    result = process_matched(params, self.process_data, self.list_match, input_nesting, len(self.outputs))
  File "/Users/zhaowei/Library/Application Support/Blender/2.92/scripts/addons/sverchok-master/utils/sv_itertools.py", line 173, in process_matched
    result = main_func(matching_f(params))
  File "/Users/zhaowei/Library/Application Support/Blender/2.92/scripts/addons/sverchok-master/nodes/matrix/matrix_out_mk2.py", line 126, in process_data
    T, R, S = m.decompose()
AttributeError: 'list' object has no attribute 'decompose'
```
**Since I don't have such experiences in programming, I'm guessing that ...
May this error come from a python version problem?
Or does it come from a system OS>?
My Environment: 
OSX - 10.15.7
blender - 2.92 (think python version is 3.7.7)**

Maybe the question is stupid but ... Thanks for your help again

PR #3994, opened 2021-03-28
Control Points Output to Curve Mapper

vicdoval:
Control Points Output for Curve mapper node.
It can work as a XY slider :)
image

- [x] Ready for merge.


iarga:
Thank you very much Victor.

PR #4003, opened 2021-03-31
matrix out format fix

vicdoval:
Format fix for matrix_out mk2.
Flat Quaternions Output option to keep them in the original format that can be disabled or wont be used when inputing list of list of matrices

image

also in this PR:

a method added to the nodes to assing properties to the input sockets via dictionary, (suggestion from @zeffii) 
now you can use as example:
 `self.sv_new_input('SvVerticesSocket', 'Vectors', dict(prop_name='Vector', is_madatory=True))`

- [x] Ready for merge.

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -322,7 +322,7 @@ class UpdateNodes:
                 sys.stderr.write('ERROR: %s\n' % str(err))
             self.set_color()
 
-    def sv_new_input(self, socket_type, name, attrib_dict):
+    def sv_new_input(self, socket_type, name, **attrib_dict):
         socket = self.inputs.new(socket_type, name)
         for att in attrib_dict:
             setattr(socket, att, attrib_dict[att])
--- a/node_tree.py
+++ b/node_tree.py
@@ -184,7 +184,7 @@ class SverchCustomTree(NodeTree, SvNodeTreeCommon):
     sv_show_error_details : BoolProperty(
             name = "Show error details",
             description = "Display exception stack in the node view as well",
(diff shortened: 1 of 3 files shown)


node_tree.py
```diff
@@ -322,6 +322,11 @@ def init(self, context):
                 sys.stderr.write('ERROR: %s\n' % str(err))
             self.set_color()
 
+    def sv_new_input(self, socket_type, name, attrib_dict):
```
Durman, line 325:
```suggestion
    def sv_new_input(self, socket_type, name, **attrib_dict):
```
As to me it would be a little bit more flexible. 
It will be still possible to use dictionary `self.sv_new_input(**{'attribute': value})`
and in the same time you have option to pass attributes as normal parameters `self.sv_new_input(attribute=value)`

nodes/matrix/matrix_out_mk2.py
```diff
@@ -80,7 +85,7 @@ def update_mode(self, context):
         items=mode_items, default="AXISANGLE", update=update_mode)
 
     def sv_init(self, context):
-        self.inputs.new('SvMatrixSocket', "Matrix").is_mandatory =True
+        self.sv_new_input('SvMatrixSocket', "Matrix", dict(is_mandatory=True, nesting_level=2))
```
Durman, line 88:
```suggestion
        self.sv_new_input('SvMatrixSocket', "Matrix", is_mandatory=True, nesting_level=2)
```
I see this as more convenient API

PR #4025, opened 2021-04-07
update bpy for 2.93+

### LINKED ISSUE #4024: Monad is probably not working as expected right now
Screen Shot 2021-04-07 at 2 52 45 pm

In the past when using Monad, when I connect in attribute, it will create extra additional connect input node, but it does not do that anymore, so no input can be created. 

Is there already alternative to Monad? Like Group is actually working or not yet?

zeffii:
addresses failed property lookup in 

resolves #4024


zeffii:
i'm not certain that the `("", {})` fallback will ever be set in that try/except, but let's see people break it first.

PR #4028, opened 2021-04-08
duplicate image found

stansu:
as title

### DIFF
--- a/docs/nodes/list_struct/levels.rst
+++ b/docs/nodes/list_struct/levels.rst
@@ -96,6 +96,7 @@ By enabling "Flatten" at the deepest available level, we concatenate vertices da
 
 By flattening at the outermost level, we concatenate lists of vertices into a single list of vertices:
 
+-- this image is duplicate with above image, should be changed.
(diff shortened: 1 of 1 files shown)


Durman:
Do you think you can generate right picture?

stansu:
If the author did not have original file, maybe I can.

Durman:
The layout seems pretty simple to recreate it from scratch.))

stansu:
> 
> 
> The layout seems pretty simple to recreate it from scratch.))

how can i upload image?

Durman:
You can attach an image to a message in this thread, this will generate a link which can be used to put into the documentation.

stansu:
圖片
圖片
圖片
圖片
圖片
圖片
圖片
圖片

stansu:
@Durman is this ok to you?
https://github.com/stansu/sverchok/blob/patch-1/docs/nodes/list_struct/levels.rst

Durman:
Looks like this is repeated twice now.

stansu:
圖片

stansu:
> 
> 
> Looks like this is repeated twice now.
> 

my bad. now should be ok.

Durman:
Thanks. Now changes are in docs - http://nortikin.github.io/sverchok/docs/nodes/list_struct/levels.html

Issue #4045, opened 2021-04-12
Add the "To Control Panel" button to "Bezier in" and "NURBS in"


maxim1967:
Add the "To Control Panel" button to "Bezier in" and "NURBS in"

zeffii:
which parts of these nodes? 

![image](https://user-images.githubusercontent.com/619340/114775488-a466d500-9d71-11eb-8aab-f3312dabac79.png)

maxim1967:
Add this for these nodes
![Снимок](https://user-images.githubusercontent.com/62519357/114779828-3de8b380-9d7f-11eb-8d8b-37f1b94dc96c.JPG)

zeffii:
yes, but which buttons from the node UI ?

zeffii:
![image](https://user-images.githubusercontent.com/619340/114836016-7cf42480-9dd2-11eb-8edb-395b074f1d07.png)

zeffii:
i don't know if it makes sense to add all those to the 3d panel  

opinions @portnov , @nortikin @vicdoval @Durman  ?

zeffii:
how about just  the "Get selection" button, and maybe a small "focus nodetree on this node" button, so user can tweak the rest on the node? or maybe all buttons on one row ` GET | SORT | APPLY MATRICES`

maxim1967:
Only "Get selection" , and for NURBS still the choice of the method "Sverchok" or " Geomdl"

Durman:
`Get selection` could be definitely useful, not sure about the rest. 

>  "focus nodetree on this node"

Oh, this can be tricky to code. @DolphinDream was trying to do something like this.

maxim1967:
Perhaps yes, it is better to "Get selection", and for NURBS still the choice of the method "Cricket" or " Geomdl"

zeffii:
there is "view2d.zoom_border" operator, which we should be able to context-fake..
```
        # this can maybe be cached...
        op2 = row.operator("view2d.zoom_border", text="", icon="TRACKER_DATA")
        node_x, node_y = node.absolute_location
        border_width = node.width * 4
        border_height = node.height * 3
        op2.xmin = node_x - border_width / 2
        op2.xmax = node_x + border_width / 2
        op2.ymin = node_y - border_height / 2
        op2.ymax = node_y + border_height / 2
        op2.wait_for_input = False
        # xmin=0, xmax=0, ymin=0, ymax=0, wait_for_input=True, zoom_out=False)
```

[...]

PR #4048, opened 2021-04-15
Bezier nurbs in to3dview

zeffii:
addresses #4045


nodes/exchange/bezier_in.py
```diff
@@ -14,31 +14,32 @@
...
 
 class SvBezierInCallbackOp(bpy.types.Operator):
 
     bl_idname = "node.sv_bezier_in_callback"
     bl_label = "Bezier In Callback"
     bl_options = {'INTERNAL'}
 
-    node_name: StringProperty(default='')
-    tree_name: StringProperty(default='')
+    idname: StringProperty(default='')
```
zeffii, line 25: @portnov i complete want to convert all `idname` and `idtree` to `node_name` and `tree_name`.  after this :)

portnov: Then why are you renaming tree_name to idtree here?

zeffii: because i want to focus on this stupid operator first, there are many references to idname and idtree.

PR #4060, opened 2021-04-21
Inset Special Mk2

vicdoval:
Inset Special Mk2:
Speed test done with a 5000 faces voronoi sphere
Upgrades:
- Numpy implementation (between 3,5 times faster than old implementation)
- Multi-level recursive node

New modes:
- Sides : Works like the inset faces node but around 1,5x faster for constant inset and 16x for variable inset  (*only partial support for concave polygons)
- Matrix: Works like Extrude Faces  and Extrude Faces Lite in Matrix Mode but 8 times faster

New Input:
- **Custom Normals** - custom normals for inset displacement

New outputs: 
- **Original Vert Idx** - the index of the original vertex. Can be used to pass  Vertex Data to new vertices. In case of Zero inset in Fan mode the index of the  new vertex will be the first index of the polygon
- **Original Face Idx** - the index of the original face. Can be used to pass Face Data to new faces 
- **Pols Group** - Outputs a list to mask polygons from the modified mesh,
  0 = Original Polygon
  1 = Side polygon
  2 = Inset Polygon.
- **New Verts Mask** - Mask of the new vertices

Custom Normal Example
image

Original Face idx and Original Vert Idx Example
image

New verts mask used to bevel only inset vertices
image

Pols group output to filter the outputted polygons
image

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


zeffii:
very nice indeed. super fast :)

Durman:
Is it faster know than `inset faces`?

vicdoval:
>Sides : Works like the inset faces node but around 1,5x faster for constant inset and 16x for variable inset (*only partial support for concave polygons)
Yes is faster but not so robust for concave polygons. Even with "Concave support" enable fails for polygons with weird shapes 

>

zeffii:
> Even with "Concave support" enable fails for polygons with weird shapes

 fails?  as in   flipping normals?

vicdoval:
The normals are ok in all cases. This is what i call 'fail' 
image

zeffii:
ah, it doesn't know what's inside or outside, in the case of the concave. and therefor it doesn't know which direction to move the new vertex towards.

vicdoval:
Yes, i was also thinking about some method @Durman suggested, something like: 

pick the most left vertex of the polygon, 
if the generated vertex is even more in the left then it means is outside.
An use that point to check the parallelism of the rest edges..

PR #4068, opened 2021-04-30
Vectorization decorator

### LINKED ISSUE #3988: recurse_f_level_control function
## Problem statement

@vicdoval 

Could you explain what the function is doing? Some example probably would help.

https://github.com/nortikin/sverchok/blob/4c9c1259476e8565ed20744684db8b308bc940c1/utils/sv_itertools.py#L121-L141

Durman:

This adds experimental decorator which solves vectorization problem.

This is how it works.

```py
from sverchok.utils import vectorize

def main_node_logic(*, prop_a: List[float], prop_b: Matrix, mode_a: str) -> Tuple[list, list]:
    ...
    return data1, data2

class MyNode:
    ...
    def process(self):
        input_a = self.inputs[0].sv_get(default=None)
        input_b = self.inputs[1].sv_get(default=None)

        main_node_logic = vectorize(main_node_logic, match_mode=self.match_mode)
        out1, out2 = main_node_logic(input_a, input_b, mode_a = self.mode_a)

        self.outputs[0].sv_set(out1)
        self.outputs[1].sv_set(out2)
```

In this example `input_a` can be one float value, list of floats or list of list of floats of arbitrary shape (`[1, [[2, 3], 4], 5, 6]`).
This is true and for all other parameters of the function.

If the shape of an input data is different from each other it will be matched according `match_mode` parameter which can repeat last, cycle etc.

Closes #3988

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/nodes/matrix/apply_and_join.py
+++ b/nodes/matrix/apply_and_join.py
@@ -59,6 +59,29 @@ def apply_matrices(
     return out_vertices, out_edges, out_polygons
 
 
+def apply_matrix(
+        *,
+        vertices: SvVerts,
+        edges: SvEdges,
+        polygons: SvPolys,
+        matrix: Matrix,
+        implementation: str = 'Python') -> Tuple[SvVerts, SvEdges, SvPolys]:
+    """several matrices can be applied to a mesh
+    in this case each matrix will populate geometry inside object"""
+
+    if not matrix or (vertices is None or not len(vertices)):
+        return vertices, edges, polygons
+
+    if implementation == 'NumPy':
+        vertices = np.asarray(vertices, dtype=np.float32)
+
+    _apply_matrices = matrix_apply_np if isinstance(vertices, np.ndarray) else apply_matrix_to_vertices_py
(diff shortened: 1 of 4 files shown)


nortikin:
self.mpde_a?

Durman:
What is it?

nortikin:
> What is it?

result = main_node_logic(input_a, input_b, mode_a = self.mode_a)
self.mode_a

Durman:
I more or less satisfied with the performance. It definitely slower than just iterate over a list of values but if a node is doing significant job overhead of vectorization is hardly noticable. But I think in future it should be optimized better because some data is copied during matching process what would be better to avoid. In the speed test example it is seen that even `sv_set` takes more time than vectorization.

**Copy Suzane with 1000 matrixes**
output

One more way for optimization would be to keep meta information with Sverchok data about its type.

PR #4075, opened 2021-05-04
add usability feature to dupli_mk4

zeffii:
dupli instancer has a few issues which can be addressed here

i'm in favour of moving a few controls directly onto the node UI, 
- show (parent) in viewport
- show (parent) in render
- show child in viewport

there is no corresponding need to toggle "show child in render" because blender doesn't render the child by default.

### DIFF
--- a/nodes/scene/dupli_instances_mk4.py
+++ b/nodes/scene/dupli_instances_mk4.py
@@ -92,6 +92,13 @@ class SvDupliInstancesMK4(bpy.types.Node, SverchCustomTreeNode):
                 row4.enabled = ob.use_instance_faces_scale
                 row4.prop(ob, "instance_faces_scale", text="Factor")  #float
 
+                """
+                to hide the child object from the viewport only, use
+                >>> bpy.data.objects['AFUWCR'].hide_set(True)  # toggle bool
+                
+                the child is hidden by default during render
+                """
+
         finally:
             pass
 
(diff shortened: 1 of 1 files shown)


vicdoval:
i was working on it too 
image
I could not achieve 'hide base child in viewport' because setting off the visibility from python seems to disable the instances too. i'll make tomorrow the PR

zeffii:
see the comment i added in the code

zeffii:
use a new boolproperty, and do the toggle inside its update function.

vicdoval:
but to do the toggle inside the update function i would need to call sv_get inside the update function and it would not be consistent if the user changes the input child...

vicdoval:
And doing it in the process func wold have the downside that if you try with different children they will stay hidden after releasing them

zeffii:
ensure the state(s) from process, but trigger process from the boolproperty update function so it looks responsive.

> children they will stay hidden after releasing them

yep, but for most of the usecases i don't see this as a huge problem.   in the order of "curvemapper won't update automatically, but not having curvemapper is worse"

zeffii:
also time for some `numpy`

vicdoval:
Also it is a little absurd to convert the vectors to matrices and then back to vectors (in verts mode)

zeffii:
maybe 'verts' mode should have a vertex socket instead

zeffii:
but that might be a `mk5` change..

vicdoval:
> also time for some `numpy`

I have been trying to improve performance via numpy but there is a bottleneck when converting the matrices to arrays or when converting the transformed verts to arrays...

It makes me wonder how much we could win if we slightly moved tje whole Sverchok from mathutils Matrix to Numpy Matrices...

Durman:
It's hardly possible to rewrite all nodes, that uses bmesh, to use numpy. So switching to numpy matrices can effect performance of those nodes. Probably some speed tests can say more on the matter. I imagine that most intensively matrices are used for move, rotate and scale operations so perhaps this operations should be optimized in the first place.

PR #4080, opened 2021-05-07
Load handler extra push

zeffii:
test 2 -  augment post_load handler with explicit refesh for animatable nodes.

### DIFF
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -207,7 +207,7 @@ def sv_post_load(scene):
             except:
                 traceback.print_exc()
 
-            for node in nd.nodes:
+            for node in ng.nodes:
                 if hasattr(node, "is_animatable"):
                     node.refresh = True  # this reverts back to False afterwards.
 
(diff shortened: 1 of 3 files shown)


core/handlers.py
```diff
@@ -217,6 +216,11 @@ def sv_post_load(scene):
 
     for ng in sv_trees:
         if ng.bl_idname == 'SverchCustomTreeType' and ng.nodes:
+
+            for node in ng.nodes:
+                if hasattr(node, "is_animatable"):
+                    node.process()
```
Durman, line 222: From my point of veiw only the update system can call the process method because the method of nodes should be called in proper order to get input data, otherwise the method can raise NoData error easily.

zeffii: i know.

zeffii: this PR is not intended to be pushed. at all. ever.

Durman: I wonder what is purpose of such PRs?

PR #4088, opened 2021-05-12
blender mesh read upgrade

vicdoval:
Objects In and Object ID Out now are faster thanks to `foreach_get` and can output NumPy arrays (boost of 1.5x for python lists to 4x with numpy arrays) 

Image for docs:
image

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/object_nodes/blenddata_to_svdata2.rst
+++ b/docs/nodes/object_nodes/blenddata_to_svdata2.rst
@@ -1,0 +1,49 @@
+Object In
+=========
+
+Functionality
+-------------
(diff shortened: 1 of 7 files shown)


vicdoval:
This brings me back the idea that this to nodes do almost the same but the name seem opposite...

also they should be together but that would make more obvious the contradiction in the names...

Any thoughts on this?

zeffii:
A "Get Object Data" node to combine the two nodes?

yes, it makes little sense to have this kind of duplicity in two nodes.

vicdoval:
Good idea @zeffii

PR #4091, opened 2021-05-14
New Get objects Data Node

vicdoval:
Joins Objects In and Object Id Out in a single node with all the functionalities and with a performance boost of 1.5x to 5x
Also adds the possibility to output Numpy arrays making the preformance even better (up to 10x)
image
image

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


zeffii:
`git --add <your node>.py`  : )

vicdoval:
There is a slight ui trouble, the object socket does not trigger the process of the node when you hit the little X in the name field. This makes the get selection button not to appear until you click in the update button of the node.
I dont know if that is the desired behaviour of the object socket and I should add a toggle to the node "by socket/by selection" or try to fix the obiect socket

Durman:
Does buttons in the property panel switch between numpy and python data type per socket? Probably it can be converted into single option for simplicity?

Durman:
By the way for me the object socket works as expected. Probably the problem somewhere else?
obj socket

nodes/scene/get_objects_data.py
```diff
@@ -0,0 +1,394 @@
...
+            else:
+                outputs['Object'].sv_set([data_objects.get(o.name) for o in self.object_names])
+
+
+    def save_to_json(self, node_data: dict):
+        node_data['object_names'] = [o.name for o in self.object_names]
+
+    def load_from_json(self, node_data: dict, import_version: float):
+        for named_object in node_data.get('object_names', []):
+            self.object_names.add().name = named_object
```
Durman, line 390: I belife that current JSON module can handle it by itself

Issue #4092, opened 2021-05-15
Switcher problem


piotrkopiczynski:
Hi,
when i change the source data, Switcher do not refresh. It only remembers old data - out of bound. Reconect do not help. 
![01](https://user-images.githubusercontent.com/25440884/118354829-47775e00-b56d-11eb-9435-afea3b4028ef.jpg)
![02](https://user-images.githubusercontent.com/25440884/118354862-64139600-b56d-11eb-911c-44bbeece8903.jpg)

zeffii:
maybe the elements that aren't visible anymore, after updating the range node, are still marked internally as selected?

zeffii:
yep
![GIF 5-15-2021 11-55-53 AM](https://user-images.githubusercontent.com/619340/118356272-a55b7400-b574-11eb-981a-516354c24aaf.gif)

zeffii:
@Durman   https://github.com/nortikin/sverchok/pull/4093  is OK?

zeffii:
i can imagine a scenario where the user does not want to so easily "lose" the toggle states of previous user list states.

zeffii:
that's updated @piotrkopiczynski

PR #4093, opened 2021-05-15
possible fix

zeffii:
addresses #4092


nodes/logic/custom_switcher.py
```diff
@@ -102,8 +109,17 @@ def process(self):
         else:
             self.string_values.clear()
 
+        self.ensure_user_list_sync()
         self.outputs['Item'].sv_set([[i for i, b in enumerate(self.user_list) if b]])
```
Durman, line 113: Looks like fixing the string is enough. And in this cases selection still will be active when number of input elements increases.
```suggestion
        self.outputs['Item'].sv_set([[i for i, b in enumerate(self.user_list[:self.string_values]) if b]])
```

zeffii:
yeah. much better

PR #4115, opened 2021-05-22
update gl drawing info from nodetree

zeffii:
this is a backend only task.

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -618,14 +618,6 @@ class SverchCustomTreeNode(UpdateNodes, NodeUtils):
             return False
         return prefs.over_sized_buttons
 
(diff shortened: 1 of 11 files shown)


node_tree.py
```diff
@@ -120,6 +120,22 @@ def throttle_update(self):
         finally:
             self.skip_tree_update = previous_state
 
+    def update_gl_scale_info(self, origin=None):
+        """
+        the nodeview scale and dpi differs between users and must be queried to get correct nodeview
+        x,y and dpi scale info.
+
+        this is instead of calling `get_dpi_factor` on every redraw.
+        """
```
Durman, line 129: Why this method is in the node tree class?

Issue #4100, opened 2021-05-18
JSON export does not preserve socket menu choice


wassimj:
## Problem statement

Exported NodeTrees to JSON do not preserve the socket menu choice (WRAP, UNWRAP, FLATTEN, etc)

## Steps to reproduce

1. Create a new NodeTree
2. Drop a Number node
3. Select Show socket menus from the Tree Properties/Sverchock tab
4. Set the output socket menu to "WRAP"
5. Export to a JSON
6. Create a new NodeTree
7. Import the JSON file that you save in step 4
8. Examine the socket menu. It has lost the WRAP choice.

## Expected result

The node output socket menu choice should be the same as before (e.g. WRAP)

## Actual result

The menu choice has disappeared.

## Sverchok version

0.6.0.0

PR #4118, opened 2021-05-22
New structure for JSON files and new approach of adding changes into the JSON format

### LINKED ISSUE #4096: Import / exprot json for new node groups
I'm ready to reimplement serialization into JSON to include data of node groups.
Also this version won't support import of monads therefore whenever exported tree includes monads old import version will be used.
Also I see sence to consider approach for exporting data blocks (text, mesh).
Previous step #3560

Here is how current structure of JSON looks like:
```py
{"export_version": "0.10",
"framed_nodes": dict(),
"groups": dict(),
"nodes": dict(),
"update_lists": []}
```

I'm thinking to change it in the new version and make it looks more like structure of Blender `bpy.data` module.

```py
{"export_version": "0.20",
"node_groups": {  # it can include more than one node tree (main node tree + node group trees)
    "My tree": {
        nodes: dict(),
        links: dict()}
"texts": {
    "My script": str()}
"meshes": {
    "Some mesh": dict()}}
```

With this format it will be easy to add new data blocks support if there will be an interest in it.

Durman:

This PR should make it easier to add new features into JSON format like saving mesh data blocks, text script files etc.
Also it adds supporting serialization of `node groups` into JSON

closes #4096 #4100

What's new:
- JSON file format
- Group tree export/import
- Node presets for group nodes
- Node presets inside group trees

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.


ui/sv_IO_panel.py
```diff
@@ -199,6 +185,7 @@ def draw(self, context):
             col = self.layout.column()  # old syntax in <= 2.83
 
         col.use_property_split = True
+        col.prop(self, 'compact')
```
zeffii, line 188: OK. i like that!

PR #4121, opened 2021-05-23
contour 2d optimize

vicdoval:
Optimisation of Contour  2D node (1,5 faster)

- [x] Ready for merge.

### DIFF
--- a/nodes/modifier_make/contour2D.py
+++ b/nodes/modifier_make/contour2D.py
@@ -42,27 +42,25 @@ intersec_mode_items = [
     ("Poligonal", "Poligonal", "Intersecction dependent from num. of vertices (Faster)", 1)]
 
 
-def check_dist_to_verts(v, or_verts, or_radius, net, poli_ang_list, modulo, mask_t):
+def check_dist_to_verts(v, or_verts, or_radius, net, poli_ang_list, poly_ang_cos_list, mask_t):
     d = 0
     min_dist = 2.0e-5
-    for j in range(modulo):
-        rad = or_radius[j]
+    for rad, vo, polig_ang, polig_ang_cos, v_net in zip(or_radius, or_verts, poli_ang_list, poly_ang_cos_list, net):
         if rad < min_dist:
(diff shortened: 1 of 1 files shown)


nodes/modifier_make/contour2D.py
```diff
@@ -122,14 +125,14 @@ def mask_by_distance(verts, parameters, modulo, edges, mask_t):
     net = parameters[3]
     mask = []
 
-    sides_space = sides_space_limits(or_verts, or_radius, net, edges, mask_t, modulo)
+    sides_space, sides_space_bbox = sides_space_limits(or_verts, or_radius, net, edges, mask_t, modulo)
     poli_ang_list = [pi/v for v in or_vert_num]
-
+    poli_ang_cos_list = [cos(ang) for ang in poli_ang_list]
```
portnov, line 130: np.cos(poli_ang_list).tolist()
?
or maybe even without tolist()
Need to test if it will be faster, off course.

nodes/modifier_make/contour2D.py
```diff
@@ -42,58 +42,59 @@
...
             continue
 
-        polig_ang = poli_ang_list[j]
-        offset = net[j][1][0]
-        if len(net[j][0]) == 3:
-            offset += net[j][0][0] - net[j][0][1]
+        offset = v_net[1][0]
+        if len(v_net[0]) == 3:
+            offset += v_net[0][0] - v_net[0][1]
         v_vo_ang = (normal_angle(atan2(vfy, vfx) - offset) % (2 * polig_ang)) - polig_ang
```
portnov, line 62: Maybe it is possible to get rid of trigonometry functions here? Like, cos(atan2(vfy,vfx)) = vfx
If not, then vectorization with numpy can probably help.

vicdoval: Thanks for the hints, I am doing a deeper rewrite to improve the performance of this node because it is still very slow and I will consider your ideas :)

PR #4123, opened 2021-05-24
timegraph tweaks

zeffii:
more tweaks to timegraph

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -37,10 +37,6 @@ from sverchok.utils.logging import debug
 from sverchok.ui import color_def
 from sverchok.ui.nodes_replacement import set_inputs_mapping, set_outputs_mapping
 from sverchok.utils.exception_drawing_with_bgl import clear_exception_drawing_with_bgl
-from sverchok.utils.nodeview_time_graph_drawing import (
-    configure_time_graph, disable_time_graph,
-    configure_node_times, disable_node_times)
-
 
 class SvNodeTreeCommon(object):
     '''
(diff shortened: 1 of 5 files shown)


zeffii:
@vicdoval  the code as it is currently expressed feels the most convenient of my iterations.

i don't know if it's useful to have the graph drawing simultaneously in multiple views. but doing so certainly requires more changes to parts of the update system which i prefer not to touch.

if at some later stage the update system is "rebooted", i would like it to be possible to use a "`graph_dict`"
```python
graph_dict[tree_name] = {sub_graph_0: [event list],  sub_graph_1: [event list], ..... }
```

rather than the current
```
graphs = [event_list, event_list, ...]
```

anyway. i need to do something else now :)

vicdoval:
Works great, thanks again :)

Maybe the start button could reset after F8 so it is clear you have to activate it again (just a little detail)
Another idea for the perfect world would be to separate it from the 'printing update timings in the console' or that they are both activated/deactivated from the same spot of the UI,  when developing nodes this prints are a little annoying. Aren't they?

zeffii:
instead of using F8, i use alt+space -> `Nuke Python++` , it resets the state boolean for that operator back to False. Ideally we would have something like an event handler that triggers after `bpy.ops.script.reload()` completes, to shake sverchok's drawing callbacks up, as interaction is returned to the user.

i agree, it's not optimal. Activating / deactivating "printing update timings in the console" i'll have a look at it, but i hadn't noticed them printing, else it may have annoyed me already : )

zeffii:
@vicdoval  so currently that debug timings are visible if you have the logging state set to "debug", but they don't display when it's just set to "information". it should be possible without too much codeshuffle to spresses prints in debug mode, if timegraph/etc is drawing.  that's what you're looking for right?

vicdoval:
Yes you are right turning the debug level to 'information' solved my problem, I thought that profiling the tree was not possible in Information mode (and I was wrong) thanks!

Btw the In the 'Tree profiling' menu the 'Start/Stop Profiling' Button resets on the usual F8, maybe the same code could be used to reset the Time graph Start/End Button

zeffii:
> Btw the In the 'Tree profiling' menu the 'Start/Stop Profiling' Button resets on the usual F8

it does? indeed! that can be harvested i think : )

zeffii:
OK. i think i'll have that working tomorrow.  it's a little late and my brain is mushed.

zeffii:
cool, that's fixed in the next push. of  https://github.com/nortikin/sverchok/pull/4132

zeffii:
@vicdoval   maybe it would be useful to add a toggle to only display the info for _selected_ nodes

vicdoval:
Imho is enough with only showing the info of the last  updated nodes (as it is now).
Otherwise you would need some way to change the selection of nodes.... 
Or of it is dynamic selection it would only show the node you are changing... that  could work for the node timings but would no make sense for the time graph...

Something that I think it could be improved is removing the global update that happens when you click on the start button. I would prefer to trigger it manually as with the 'tree profiling' so you can decide if you want a global update or a partial update

zeffii:
> Something that I think it could be improved is removing the global update that happens when you click on the start button.

i think i agree. feel free to comment out line 82,  from 

https://github.com/nortikin/sverchok/blob/873b4564800cb10961c137e425d6ca2869c9eea8/ui/nodeview_operators.py#L82

PR #4137, opened 2021-05-31
change 282 to 292

zeffii:
2.9.2 or 2.9.3..

seen as 2.9.3 is now in LTS, maybe we can use 2.9.3 as the test suite.
i'm not sure if the tests will run this yaml before it hits master? @vicdoval  @portnov  ?

right... it still uses 2.82 untill it's pushed to master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ install:
     #set -x
     #set -e
     if [ $TRAVIS_BRANCH = "master" ]
-    then BLENDER_VERSION=2.82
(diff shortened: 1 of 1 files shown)


portnov:
I'm all for updating tests to 2.9+. Although I'm not sure, maybe it's better to fix them first.

zeffii:
the tests are only failing because Sverchok is loading a bl_icon which doesn't exist in 2.82.

i'm not sure we can sanely test for that a priori, unless we grep or regex findall (bl_icon.?=.?.+) or something horrible like that

zeffii:
```python
functions = bpy.types.UILayout.bl_rna.functions
icons = functions["prop"].parameters["icon"].enum_items.keys()
print(icons)
```

PR #4141, opened 2021-06-03
Update test-sverchok.yml

zeffii:
testing yaml blindly.     eat me.

ok a few notes then.. for the `bash` leppers such as myself :) 

### basename from an archive

```bash
BLENDER_URL=https://ftp.nluug.nl/pub/graphics/blender/release/Blender2.93/blender-2.93.0-linux-x64.tar.xz
BLENDER_DIR=$(basename $BLENDER_URL .tar.xz)
echo $BLENDER_DIR
#  blender-2.93.0-linux-x64
```
this is the equivalent to python 
```python
BLENDER_URL = "https://ftp.nluug.nl/pub/graphics/blender/release/Blender2.93/blender-2.93.0-linux-x64.tar.xz"
BLENDER_DIR = os.path.basename(BLENDER_URL).replace(".tar.xz", "")
print(BLENDER_DIR)
>>>  "blender-2.93.0-linux-x64"
```

### DIFF
--- a/.github/workflows/test-sverchok.yml
+++ b/.github/workflows/test-sverchok.yml
@@ -38,11 +38,11 @@ jobs:
 
         echo "blender basename from blender url: ${BLENDER_DIR}"
 
-        echo "ls installation/blender/*..."
-        ls installation/blender/*
-        echo "...end ls"
+        # echo "ls installation/blender/*..."
+        # ls installation/blender/*
+        # echo "...end ls"
 
-        if [ ! -f installation/blender/$RELEASE ]
+        if [ ! -f installation/blender/blender ]  # installation/blender/$RELEASE .. if first time.
         then
             mkdir -p installation
             cd installation
--- a/.github/workflows/test-sverchok.yml
+++ b/.github/workflows/test-sverchok.yml
@@ -51,8 +51,8 @@ jobs:
             tar xf $BLENDER_TAR -C $RELEASE --strip-components 1
 
             rm $BLENDER_TAR
-            echo "removing content of blender dir"
-            rm -r blender/*
+            # echo "removing content of blender dir"
+            # rm -r blender/*
             mv $RELEASE blender
 
             pushd blender/$RELEASE/
--- a/.github/workflows/test-sverchok.yml
+++ b/.github/workflows/test-sverchok.yml
@@ -34,25 +34,25 @@ jobs:
         SVERCHOK_DIR=scripts/addons/sverchok
         BLENDER_TAR=$(basename $BLENDER_URL)
         BLENDER_DIR=$(basename $BLENDER_URL .tar.xz)
-        STABLE_RELEASE=blender-2.93.0-stable
+        RELEASE=blender-2.93.0-stable+blender-v293-release.84da05a8b806-linux.x86_64-release
 
         echo "blender basename from blender url: ${BLENDER_DIR}"
 
-        if [ ! -f installation/blender/$STABLE_RELEASE ]
+        if [ ! -f installation/blender/$RELEASE ]
         then
             mkdir -p installation
             cd installation
             echo "Current directory before downloading and installing blender: $(pwd)"
 
             wget -nv $BLENDER_URL  #not verbose, but not quiet.
-            tar xf $BLENDER_TAR $STABLE_RELEASE
-            
+            tar xf $BLENDER_TAR
+
             rm $BLENDER_TAR
             echo "removing content of blender dir"
             rm -r blender/*
-            mv $STABLE_RELEASE blender
+            mv $RELEASE blender
 
-            pushd blender/$STABLE_RELEASE/
+            pushd blender/$RELEASE/
             PYTHON=${BLENDER_VERSION}/python/bin/python3.9
             $PYTHON -m ensurepip
             $PYTHON -m pip install --upgrade pip setuptools wheel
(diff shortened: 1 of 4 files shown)


zeffii:
@portnov i have movement.

image

we probably don't want to hardcore `blender-2.93.0-linux-x64` anywhere, so how does one rename 
```installation/blender/blender-2.93.0-linux-x64 to installation/blender/blender```

something like
```
mv "installation/blender/" + ${BLENDER_DIR} installation/blender/blender
```
i don't know how to mix literals with strings.

**Comment by portnov** [comment]
@zeffii `$VARIABLE` (or `${VARIABLE}`, which is the same) is just replaced with value of VARIABLE. You can think about it in terms of python's f-strings. 
Python:
```
print(f"Hello {variable}")
```
Shell:
```
echo Hello $variable
```
The difference is, in shell you do not have any special "context markers" for f-strings (like `f""`), everything in shell is one big f-string :)

**Comment by zeffii** [comment]
@portnov   OK. it's testing, but stail failing the actual tests : )

**Comment by zeffii** [comment]
OK. this is working now.  but @portnov  forgive me if i've completely misunderstood this file..

it seems that this produces an image somehwere in the cloud of a filesytem?  and at the moment it is always download blender and installing python dependancies because this line's `then` clause is executed

```
        if [ ! -f installation/blender/$RELEASE ]
        then
            mkdir -p installation
```

**Comment by zeffii** [comment]
anyway, there's other bugs to fix now  : ) 
```
File "/home/runner/work/sverchok/sverchok/installation/blender/blender-2.93.0-stable+blender-v293-release.84da05a8b806-linux.x86_64-release/2.93/scripts/addons/sverchok/old_nodes/vd_attr_node.py", line 80, in SvVDMK3Properties
    prop_func, kw_args = SvVDExperimental.__annotations__[key]
```
 : ) 
these i can handle.

**Comment by zeffii** [comment]
now this.. 
```python
======================================================================
ERROR: test_intersect_edges (intersect_edges_tests.IntersectEdgesTest2)
----------------------------------------------------------------------
  File "/home/runner/work/sverchok/sverchok/installation/blender/blender-2.93.0-stable+blender-v293-release.84da05a8b806-linux.x86_64-release/2.93/scripts/addons/sverchok/tests/intersect_edges_tests.py", line 14, in test_intersect_edges
    result_verts = get_output_socket_data(node, "Verts_out")
  File "/home/runner/work/sverchok/sverchok/installation/blender/blender-2.93.0-stable+blender-v293-release.84da05a8b806-linux.x86_64-release/2.93/scripts/addons/sverchok/core/socket_data.py", line 165, in get_output_socket_data
    raise SvNoDataError(socket)
sverchok.core.socket_data.SvNoDataError: No data passed into socket 'Verts_out'

----------------------------------------------------------------------
```

zeffii:
i've got a bit of a headache after this,  but I think my next step is to look at the original `test-sverchok.yaml` and see if there's a more minimal change possible.. like `tar` _to a named directory_ instead of the directory given in the `tar` .

zeffii:
can someone figure out the caching?  
https://github.com/actions/cache

because, in the `run` section, if I do `ls installation/*` , it shows always that there is an 2.82,  so it is should be possible to get 2.93 to stay in that image too

zeffii:
ok don't know if related..

zeffii:
@portnov @vicdoval   why is the newly downloaded blender/python dependancies gone when this job is triggered again?

that caching is a little beyond my ability to figure out,..   unless ofcourse i start throwing all kinds of mad stuff at the actions until something clicks.  but maybe someone knows.

zeffii:
we make some kind of a key file using `md5sum blender`

portnov:
Maybe you should change cache key? For example, "empty" -> "b29"

Minoru:
Hi! Portnov shared a link to this PR, so I took a peek.

The main problem here is that the cache key doesn't depend on anything, it's a fixed value. So you're using the cache that you created some time ago, and GitHub doesn't update it because the key at the end of the run is the same as at the start of the run.

You can fix this temporarily by changing the key as @portnov suggested, but you'll have to do it every time you change anything about Blender installation (e.g. Blender version or release).

A better approach would be to include Blender version in the key, something like this:

```yaml
jobs:
  build:
    env:
      blender_version: "2.93"
    steps:
      - uses: actions/cache@v2
        with:
          path: ./installation
          key: blender-${{ env.blender_version }}
```

Now every time you change `blender_version`, the key will change, and the job will start without a cache. At the end of the job, GitHub will pack up `./installation` and use it in all future jobs until you change `blender_version` again.

The cache can even depend on the workflow file itself: `key: blender-${{ env.blender_version }}-${{ hashFiles('.github/workflows/test-sverchok.yml') }}`. That way, when you change your Blender installation script, the cache key changes too, and GitHub is forced to create a new cache file.

Hope this helps.

Minoru:
Ah, and you don't need to manually add a "post cache" step, GitHub updates the cache automatically at the end of each run (if the key changed).

zeffii:
@Minoru that is very helpful!  the syntax is a new to me, but i'll have time tomorrow to complete this.  if someone else doesn't do it already.

Minoru:
You're welcome! ae3a434 and cae2c43 don't work because the `blender_version` is supposed to be under `env` — it's an environment variable ;)

zeffii:
i still have no idea what i'm doing.  but this is progress.

zeffii:
thank you for your time @Minoru ! works as hoped !

zeffii:
i'll keep poking around in this, and document it.

PR #4149, opened 2021-06-07
general backend cleanup

zeffii:
This PR will include miscellaneous modifications to increase test readability. Have patience.

- filename from search_enum produces awkward data.

- at the end of a sverchok session, there's a massive printout of the exception.
```python
def sv_unregister_modules(modules):
    clear_system_cache()
    for m in reversed(modules):
        if hasattr(m, "unregister"):
            # print("Unregistering module: {}".format(m.__name__))
            if m.__name__ == "sverchok.old_nodes":
                import bpy
                old_node_names = sverchok.old_nodes.old_bl_idnames
                loaded = any([hasattr(bpy.types, node_class_name) for node_class_name in old_node_names.keys()])
                if not loaded:
                    print("skipping unregister of old nodes..they are not registered")
                    continue
            try:
                m.unregister()
            except RuntimeError as e:
                print("Error unregistering module: {}".format(m.__name__))
                print(str(e))
```


Durman:
What is it for?

zeffii:
i was going through the output of the `run_tests.sh` and i think a few nodes that store `self.filename` have an issue recalling the file. It may be down to the json .  I modified `run_tests_win.bat` so now i can test these things offline, and that's why i closed it.

node_tree.py
```diff
@@ -508,6 +508,15 @@ def get_bpy_data_from_name(self, identifier, bpy_data_kind):  # todo, method whi
...
         return None
 
+    def safe_socket_remove(self, kind, key, failure_message=None):
+        with self.sv_throttle_tree_update():
+            sockets = getattr(self, kind)
+            if key in sockets:
+                sockets.remove(sockets[key])
+            else:
+                canned_msg = f"{self.name}.{kind} has no socket named {key} - did not remove"
+                print(failure_message or canned_msg)
```
Durman, line 518: Definitely not print here. And I would appreciate a few string of documentation to such basic method.

core/__init__.py
```diff
@@ -29,6 +29,13 @@ def sv_unregister_modules(modules):
     for m in reversed(modules):
         if hasattr(m, "unregister"):
             # print("Unregistering module: {}".format(m.__name__))
+            if m.__name__ == "sverchok.old_nodes":
```
Durman, line 32: Are you sure that old modules visit this method? I guess that there register unregister methods are only called in `load_post` handler.

zeffii: i am , because otherwise `m.__name__ == "sverchok.old_nodes"`  would not be visisted, but it is.

but i think the solution to fix the `unregister` function inside old_nodes, so it checks if a node.is_registered_node_type or is `hasattr(bpy.types, node.bl_idname)`

it should be able to call unregister, without raising an exception -- even if all nodes are unregistered. -- maybe it can print a simple message, that only few of us would ever see, but we can search the code base if we see it.

node_tree.py
zeffii: thanks for your concern, again durman, but i will do this as i like. I will change to self.debug or something if I decide to push. relax man.

zeffii:
i think we need to find a way to clear `bpy.data.texts` between tests.

zeffii:
either that. or forbid test modules/json/blend to use "Text.{0...n}" as their text datablocks. ...and that's just not practical.

zeffii:
Tie not storing shape data in Objects In lite.

zeffii:
@Durman 
```
Error: Python: Traceback (most recent call last):
  File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\ui\sv_IO_panel.py", line 299, in execute
    is_tree_exportable, msg = self.can_be_exported(ng)
  File "C:\Users\zeffi\Downloads\blender29\2.93\scripts\modules\bpy_types.py", line 720, in __getattribute__
    return super().__getattribute__(attr)
AttributeError: 'SvNodeTreeExportToGist' object has no attribute 'can_be_exported'

```
where is `can_be_exported` defined.

node_tree.py
Durman: The thing is that leaving prints everywhere in code is a common mistake. And I don't see why it can't be coded correctly from the beginning.

core/__init__.py
Durman: Can you say how the problem can be reproduced? When I use reload (F8) I can't see any prints about errors that a module can't be unregistered.

zeffii: look at the output of 'test it` https://github.com/nortikin/sverchok/runs/2764532114   at the very end

PR #4153, opened 2021-06-08
add draw_label to SvSetDataObjectNodeMK2

zeffii:
adds self.formula to draw_label when `is.hide`


zeffii:
it;s a pitty minimized nodes don't show their icon.

PR #4162, opened 2021-06-11
remove curvemapper from snlite

zeffii:
some snlite pruning.
- removing old directive (curvemapper is now a proper node)
   - also removing all associated functions from helper files, they are now part of the curvemapper files.
- adding alternative parsing for required sockets 
    ```python
    """
    >in socketname type   # for required linked sockets  (no default, no nested specification)
    ...etc
    """
    ```
    this `>` is also a hint to the reader that these sockets are necessary for the code to function. Not all script node scripts will need this, but for the scripts you include in json/examples you might want to add this, it also keeps the testing suite happy.


zeffii:
there are a few ways to do this via the `directive`, i could allow the user to write

i like this, it's really short.
```python
>in socketname type   # for required linked sockets
```
this is more descriptive, so nice for newcomers, but verbose after the first encounter with the concept.
```python
in socketname type (required)
in socketname type (r)   # shorthand
```
in these options, you don't need to keep track of renaming sockets.
another option would be
```python
in socketname type
in socketname type
in socketname type
out socketname type
...etc
required=(socketname, socketname.... )
```
and i sure as hell am not going to be typing that.

PR #4164, opened 2021-06-12
New update system

### LINKED ISSUE #3058: Update system requirements

### LINKED ISSUE #3795: tree.sv_links.get_nodes(tree)
```
(Pdb) ll
 60     def start_exception_drawing_with_bgl(ng, node_name, error_text, err):
 61         """ start drawing the exception data beside the node """
 62         try:
 63             node = ng.nodes[node_name]
 64         except:
 65             breakpoint()
```

```
(Pdb) w
  sverchok\node_tree.py(231)update()
-> self.sv_update()
  sverchok\node_tree.py(104)sv_update()
-> process_from_nodes(self.sv_links.get_nodes(self))
  sverchok\core\update_system.py(484)process_from_nodes()
-> do_update(update_list, ng.nodes)
  sverchok\core\update_system.py(424)do_update()
-> do_update_general(node_list, nodes)
  sverchok\utils\profile.py(104)wrapper()
-> return func(*args, **kwargs)
  sverchok\core\update_system.py(409)do_update_general()
-> start_exception_drawing_with_bgl(ng, node_name, error_text, err)
> sverchok\utils\exception_drawing_with_bgl.py(66)start_exception_drawing_with_bgl()
```

```
(Pdb) ll
 95         def sv_update(self):
 96             """
 97             the method checks if anything changed inside the normal tree or monad
 98             and update them if necessary
 99             """
100             self.sv_links.create_new_links(self)
101             if self.sv_links.links_have_changed(self):
102                 self.has_changed = True
103                 build_update_list(self)
104  ->             process_from_nodes(self.sv_links.get_nodes(self))
105                 self.sv_links.store_links_cache(self)
106             else:
107                 process_from_nodes(self.get_groups())
(Pdb) self.sv_links.get_nodes(self)
[bpy.data.node_groups['NodeTree'].nodes["("]]
```
It means (`[bpy.data.node_groups['NodeTree'].nodes["("]]`) that the `sv_links.get_nodes` method sometimes returns objects with invalid link in memory.
The question is how to reproduce this.

Durman:

This is third attempt - #3179 #2393
I fill this will be last one.))

Closes #3058
Closes #3795

### New features

- [x] It will be possible to cancel node tree updating
But not always. It can cancel only between nodes but not when a node is working. It is more flexible with node groups. It will possible to evaluate them one by one internal node so in this way work of a group node can be canceled. Evaluation can be canceled by pressing escape but also it is possible by changing a node value before a tree is fully evaluated. In this case the change of the property will cancel current evaluation and start new one. Also canceling is possible during animation (the same way as with property changed) but probably it is not always expected and there should be an option.
cansel updating by draging a property

- [x] If a node tree has more then one disconnected node graph they will be evaluated independently.
It means that an error in one subgraph does not prevent updating of another. Also error messages will be more persistent and wont disappear until the cause to be resolved. But there is also another side effect. Nodes will be never reevaluated if there is a node with an error before. In current update system in some cases it's possible and I don't know whether I should change the logic here somehow.
new error handler

- [x] Supporting muted links
muting links

- [x] show progress in the header of node tree editor

- [x] support tree annotations inside node groups

- [x] cumulative update time

cum time

- [x] Changes in panels

### Supporting old features

- [x] wifi nodes

- [x] loop nodes. Actually I would prefer not to touch them but I know that they reuse some code from the update system module and it may appear that they expect the module to be in some state to work correctly.
I've checked loop nodes seems there is no need in any changes.

- [x] print timings in a node editor. Just show time of a last update. Can be hidden in the panel of tree properties.

- [x] add profiling

### Unsupported old features

- Evaluation order according nodes position. The approach looks doubtful to me. If an order has metter it should be set explicitly with links I think. And would prefer not to overcomplicate the update system. But if some one interested it could be implemented ofcause. In this case new sorting method should be added to the utils/tree_walk.py module. And the event should get extra option I guess.

- Show error stack in a node tree editor. It looks like useless for users and is more  needed for devs. I expect last ones use system console for that since there is other useful information there. 

- Showing time graph. I would not mind to connect it to new update system but it's not connectable. It uses the update system module directly. Proper place to connect it would be in `tree.update_ui` method. This is new method called by new update system when the job is done. Updating time can be get from `tree.handler.get_update_time()`. The only problem will be with evaluation order because it is not kept in statistic results currently but could be added.

- Heat map. It is tightly bound to current update system and can't be reused. Showing update time in a tree editor partly covers this functionality. In the future I have plans of improving debugging tools as well.

### Plans for next PRs

After this PR the `throttle` decorator and context manager will be deprecated. Evaluation of a tree will be in another Python session and it wont be effected by extra updates of a tree. It will be possible to manipulate with sockets more easily for the same reason. When a tree is evaluated the sockets will already be with the correct `is_linked` state.

Add more functionality for debugging. I'm thinking of adding list widget to the tool panel. It can show list of all nodes in a tree and their updating time. List widget has sorting and filtering UI which will be quite useful to search, for example, slow nodes. A list item is selectable so it will be possible to give extra information about selected node. Probably it could have an operator - "go to selected node".

Add vectorization loop nodes which would include last ideas about how the vectorization should work. Also it should be possible to cancel their work. It should be possible to debug each iteration. Implementing them could be simpler then past the vectorization into all existing nodes.

Think how the work of the `socket_data` module could be delegated to update system.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.


Durman:
### Additional features

- [ ] support tree annotations inside node groups

Durman:
The code is more or less ready. I don't have plans of adding anything but bug fixes now. If someone want to be sure not to have problems with new update system it's time to test it.

vicdoval:
@portnov sverchok-extra fails with this PR due: 
`from sverchok.core.update_system import process_from_node` in triangular_mesh.py  `process_from_node` can't be imported

Another little issue: I have a node-tree in my default file, this node-tree is not evaluated until I interact with it (click on toggle, make a link, even if it is just start  the link and cancel it). This does not happen if I open blender by clicking in a file, only with the default file. With the actual update system does not happen.

portnov:
@Durman is there a method in new update system to be used instead of `process_from_node`?
The problem is: some node(s) is too slow to process data "online", when you change anything in the tree. So it has explicit button "process!" to trigger processing.

Durman:
@portnov `node.process_node` is it not what you want?
@vicdoval loading startup files fixed

portnov:
will try `process_node`.

vicdoval:
@portnov you could also check how I solved that problem in the "Evolver" node, in which the `process` function is only in charge of passing the data from a dictionary to the next node

vicdoval:
@Durman great, now loads perfectly.
Some ideas:
Should node timings be activate by default? maybe is to technical and at the beginning everything is just 0ms

Does it really improve performance that the timings position only changes on nodetree update? The actual solution (moving when you move nodes) looks much more responsive.

Durman:
Yes, probably the timings can be disabled by default but we can leave it as it is for a while so more people could be aware of this feature.


Problem with tracking position of a node is that we can't put its reference into a draw function. So the function each redraw should search the node. According that all nodes have a massage to draw it it takes n2 time. It's not such a problem if only one node has massage or a node tree has a few nodes but with big trees (don't know probably handred nodes) the workflow becomes realy slow.
https://discord.com/channels/745273148018262158/806643543182802954/854253625806815242

Another solution could be to draw messages right in nodes. This is what is being done in GN.

zeffii:
> Problem with tracking position of a node is that we can't put its reference into a draw function.

as soon as we click into nodeview or drag a node, each "_draw_text_handler"  will be triggered, there's a lot of stuff happening anyway. but what is more overhead a single callback draw function for 80 nodes or 80 separate callbacks.

> So the function each redraw should search the node. According that all nodes have a massage to draw it it takes n2 time

we are talking not much time here in relation to the execution of a good demo sverchok layout. really. it seems to me that when we squeeze as many inefficiencies as possible in the rest of the code it allows us to make conscious choices to allow moderate inefficient features to be more pleasant to use. Most users may never even need to know execution times.

isn't there some kind of a `node_id -> node` hashmap ?

zeffii:
can  `dynamic_location` be the default for the timings, even if it is inefficient. if a nodetree is so heavy that a user might have some benefit from static timer text location, then maybe leave that up to the user.

Durman:
It will be easy to add extra property to the Node timings panel which make timing prints stick with nodes.

PR #4177, opened 2021-06-20
Nurbs birail update 2

portnov:
Add a parameter which controls rotation of profile curves.

   * **Path Normal Average**. The node will try to place profile curves so that
     they will lie in normal planes of both path curves. Since normal planes of
     two path curves can differ, the node will calculate average normal plane.
   * **Path 1 Normal**. The node will place profile curves so that they lie in
     normal plane of the first path curve.
   * **Path 2 Normal**. The node will place profile curves so that they lie in
     normal plane of the second path curve.
   * **By profile**. The node will try to place profile curves so that they be
     parallel to initial location of the path curve. This is not always
     possible, but the node will try to keep it as parallel as possible.

   The default option is **Path Normal Average**.

Screenshot_20210620_112703
Screenshot_20210620_112632
Screenshot_20210620_112548
Screenshot_20210620_112521

## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/docs/nodes/surface/nurbs_birail.rst
+++ b/docs/nodes/surface/nurbs_birail.rst
@@ -28,7 +28,9 @@ curve.
 By default it is supposed that initially the provided profile curve(s) lie in
 XOY plane. However, there is an option to instruct the node to try to figure
 out correct rotation of profile curve(s). Note that this option may result in
-precision loss in some cases.
+precision loss in some cases. Or, in other cases, it can guess desired profile
+rotation incorrectly. In such cases, you will have to place your profile curves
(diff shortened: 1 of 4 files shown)


rendetto:
Thank you for this cool update! I'll start exploring the options. 
What settings will recreate the behavior of the node before the update?

portnov:
The default ones. We always try to preserve backward compatibility.

rendetto:
Perfect! Thank you!

Issue #3603, opened 2020-10-01
Gordon NURBS Surface node


portnov:
Also known as "surface from net of curves". We have a number of curves along U direction, another number of curves along V, they are intersecting at some number of points. Discussed in #3571.

Algorithm is looking very simple:
1. Find intersection points of each of M U-Curves with each of N V-Curves, to have MxN intersection points. I expect this to be the trickiest part. In many cases the curves will just pass near one another at very small distance instead of intersecting. First idea how to do this is to use solvers from scipy, similar to how it is implemented in existing "intersect *" nodes.
2. Build Nurbs surface L1 = loft between all U-curves. Use v-degree = number of V-curves - 1.
3. Build Nurbs surface L2 = loft between all V-curves. Use u-degree = number of U-curves - 1.
4. Build Nurbs surface I = interpolation between all intersection points. It will have degrees (M-1) x (N-1), the same as L1 and L2 by construction.
5. Unify u- and v- knot vectors of L1, L2, I surfaces by adding new knots. Probably, have an option to do not add knots if existing knots have very similar values - use average value of knot in such cases.
6. Calculate resulting surface control points = L1 + L2 - I.

and then fight bugs for some time :)

rendetto:
A small suggestion... It could be convenient if the user have the option to manually input the net intersection points, in case the numerical methods lack of precision on a particular setup. @portnov what's your opinion?

![net_from_curves](https://user-images.githubusercontent.com/66558924/94869983-21260780-044f-11eb-872b-fcba66701876.jpg)

Here is the file shown above in case you find it useful for testing:
[net_from_curves.zip](https://github.com/nortikin/sverchok/files/5315132/net_from_curves.zip)

Thanks

portnov:
Agreed.

onerawartist:
Agreed 100%

rendetto:
I just finished some testing how the Gordon implementation in FreeCAD/CurvesWB  will deal with the test file above.
The net is a bit too wavy, but shows an extreme case where the curves have arbitrary intersection/connection points.

![FreeCAD_Gordon](https://user-images.githubusercontent.com/66558924/100765259-f7ae3b80-33ff-11eb-8c12-5a2a9e8a3753.jpg)

It seems that this type of surface did a good job of interpolating all the curves from the net. So it's definitely worth having.

I'm sharing the file from the above screenshot in case you need a closer look: 
[...]

PR #4183, opened 2021-06-24
Surface from Net of Curves Node (Gordon Surface)

portnov:
refs #3603

Given a net of intersecting curves, this node generates a surface which passes
through all given curves.

A net of curves is two sets of curves, one set along one direction and one set
along another. One set of curves is called "U-curves" (curves along U
direction) and another set of curves is called "V-curves" (curves along V
direction).

Apart of curves grid, this node requires intersection points of curves provided
explictly. Intersection points can be calculated by use of "Intersect NURBS
curves" node. Note: that node uses numeric algorithms to calculate
intersections, so it can fail to find intersections, or give imprecise results.
So, if you have intersection points in advance, it's better to use them.

This node generates a NURBS surface from NURBS curves according to algorithm
known as "Gordon Surface".

Screenshot_20210626_000816

Screenshot_20210704_115911

Screenshot_20210627_213402

Screenshot_20210627_192737

Screenshot_20210627_193025

Gordon surface algorithm has a number of restrictions:
* All input curves must be NURBS or NURBS-like. More specifically, only non-rational curves are supported - i.e., without weights (or all weights equal).
* All U-curves must have "the same" direction; all V-curves must also have "the same" direction.
* There must be N curves along one direction, and M curves along another direction, which must *exactly intersect* at `N x M` points.
* intersection points must be located evenly in parameter spaces of curves. For example, V-Curve `C_v` must intersect *all* U-curves `C_u` at the same value of U parameter.
* U-curves must be ordered along direction of V-curves, and vice versa. For example, if U-curve `C_u1` intersects V-curve `C_v` at `v = v1`, and another U-curve `C_u2` intersects the same V-curve `C_v` at `v = v2`, and `v1 < v2`, then in input curves list, `C_u1` must be provided before `C_u2`.

One relatively easy way to build a correct curves net is to draw a number of Bezier curves, which intersect exactly at their node points.

Planned developments:

- [x] Add "implementation" parameter to "Interpolate NURBS Surface" node. Currently this node depends on Geomdl. Now two implementation will be available: Sverchok (built-in) and Geomdl.
- [x] New "Intersect NURBS Curves" node. It will have two implementation: SciPy (algorithm in Sverchok, which uses SciPy minimization routines) and FreeCAD, and some settings about precision / tolerance.
- [x] New "NURBS Surface from Curves Net" node. It will have mandatory "Intersections" parameter, because finding curve intersections is a separate task, for which there will be specialized node with it's own parameters.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.


portnov:
@rendetto initial implementation of "NURBS Surface from Curves Net" node is more or less ready. It still lacks documentation, and there may be some tunings... but you may already try it from this branch.

portnov:
> intersection points must be located evenly in parameter spaces of curves. For example, V-Curve C_v must intersect all U-curves C_u at the same value of U parameter.

The following is the most trivial approach to removing this restriction. Given a curve and a set of T-values at which it has intersections with other curves from the net, reparametrize the curve so that intersection points will be located evenly in curve's parameter space.

```python
"""
in curve_in C
out curve_out C
"""

from sverchok.data_structure import zip_long_repeat, ensure_nesting_level
from sverchok.utils.curve import SvCurve
from sverchok.utils.curve.nurbs import SvNurbsCurve

def process(curve):
    t_values = [0.2, 0.3, 0.7, 0.8]
    t_min, t_max = curve.get_u_bounds()
    t_values = [t_min] + t_values + [t_max]
    
    segments = []
    for t1, t2 in zip(t_values, t_values[1:]):
        segment = curve.cut_segment(t1, t2, rescale=True)
        segments.append(segment)
    
    result = segments[0]
    for segment in segments[1:]:
        result = result.concatenate(segment)
    
    return result

curve_out = []
for curve in curve_in:
    c = process(curve)
    curve_out.append(c)
```

rendetto:
Great work! I did a simple test - 4x4 straight Bezier curves - forming a simple 4x4 grid. Found a funny thing  - when I use SciPy for calculating intersections the resulting Gordon have 9x9 control vertices and the U and V knot vectors are (0, 0, 0, 0, 0.5, 0.5, 0.5, 1, 1, 1, 1).
When using FreeCAD for intersections or manual ones the Gordon predictably have 4x4 control vertices and (0, 0, 0, 0, 1, 1, 1, 1) knot vectors in U and V.
Here is the file I'm testing with:
[SVGordon_Test001.zip](https://github.com/nortikin/sverchok/files/6722121/SVGordon_Test001.zip)

rendetto:
Did a second test - this time using 2 rails and 5 sections. All generation curves are BSplines of degree 3 with 5 control vertices each. 
SVGordon_Test002

Several things appeared:
1) The intersection node did not solved completely this setup - some points were missing resulting an empty places in the output list.
SVGordon_Test002_intersect_fail
2) After I gave manual intersection points I got the surface. But I've noticed that the surface did not interpolate good enough though the given curves. The "paths" were interpolated well and also the start and end "profiles" but the inner curves were not.
SVGordon_Test002_InterpolationIssue

3) Using Points metric I got 5x5 surface, 
SVGordon_Test002_PointsMetric

but using the other metrics I got 5x7 control vertices
SVGordon_Test002_CentripetalMetric

I guess we could get more control if we have similar generation parameters like for the Mono- and the Birail like "Out Number of U/V Sections" and "Out U/V Degree"...if it's applicable here.

portnov:
@rendetto 
> 1. The intersection node did not solved completely this setup - some points were missing resulting an empty places in the output list.

Well, as here you're using FreeCad implementation, there is nothing I can do. FreeCad's `intersectCC` method says there is no intersection, and I do not see any means in it's documentation how it would be possible to provide some amount of tolerance.
You can try "SciPy" implementation, it has "Precision" parameter. In my examples, I often have to set "precision" ridiculously high, to `0.02`, which technically means that the node should say that curves are intersecting when they are in fact passing at 0.02 distance of one another. I think all of this is just an unfortunate consequence of floating-point numbers usage.

2.
For original Gordon algorithm, it is required, that intersections are located "evenly" in curve's parameter spaces. I.e. one U-curve intersects all V-curves at the same T parameters of those V-curves. If this restriction is not met exactly, then the generated surface will be only "approximately" Gordon surface.

I've just pushed updates to "Intersect curves" and "gordon surface" nodes: Intersection node outputs T values of intersection points, and Gordon node can consume them. If you provide these T values of intersections, Gordon node will automatically reparametrize curves so that in new parametric space intersection would occur "evenly". This reparametrization algorithm is somewhat rude, so it can produce unwanted additional control points, or create "bad" parametrization of resulting surface. But it works for my example :)

3. I'm not sure if I should leave "metric" option. Theoretically, this metric should calculate T values of intersection points, based on location of these points, and do that in such a way that resulting T values would be "even" in a sense I mentioned above. But in my examples, I do not see cases when any other option apart of "Points" gives better results.

portnov:
Screenshot_20210628_203111

rendetto:
> that intersections are located "evenly" in curve's parameter spaces

Does "evenly" means like placing 5 sections exactly at (0.0, 0.25, 0.5 ,0.75 1.0)?

portnov:
It is the simplest way to meet the restriction, but the restriction is not that strict.
It only means that each U-curve must intersect V-curves at the same value of parameter of V-curves, and each V-curve must intersect all U-curves at the same value of parameter of U-curves.

Maybe it will be simpler to illustrate.
Good parametrization: (first number is parameter of U-curve at the intersection point, the second number is parameter of V-curve at the intersection point):
Screenshot_20210628_204916

"Bad" parametrization:
Screenshot_20210628_205316

rendetto:
OK got it. So my setup is fine - I placed the sections at the same parameters on both paths (0.0 ,0.3, 0.52, 0.76, 1.0) and still the interpolation is off.
I tried feeding the exact T values but I got GEOMDL error: 
SVGordon_Test002_ExactT

portnov:
Woh, interesting. Please post console output here.
Meanwhile, you can play with other implementations of curves.

portnov:
Can you share blend file?

rendetto:
Sure, I'm preparing it now.

rendetto:
As it appears FreeCAD also gives an error, and Sverchok implementation gives strange result.
[SVGordon_Test002c.zip](https://github.com/nortikin/sverchok/files/6727862/SVGordon_Test002c.zip)

rendetto:
> I'm not sure if I should leave "metric" option

I guess you can leave it as it is. It could be possible that in some cases those other metrics could give more suitable (not exactly better) results. From my experience so far - having more options create more opportunities, sometimes not predicted by design :)

portnov:
Well, I understood "Geomdl error". Currently the node is not "lenient" to shape of input data for T1,T2 inputs. It expects the data in exactly the same shape as "Intersect Curves" node outputs it. In your case it is possible to achieve with "List Repeater" node.

Screenshot_20210628_215443

However, I still do not understand the reason why the surface goes not precisely by the curves in your case...

portnov:
My example:
[gordon9.zip](https://github.com/nortikin/sverchok/files/6728050/gordon9.zip)

rendetto:
Thanks, noted :) Also thanks for the example - it does give me an error though:

gordon9_error

portnov:
That's odd. Does it give you the same T-values?
Screenshot_20210628_222335

rendetto:
Umm.. not exactly:
gordon9_Tvalues

rendetto:
In my case this datas hape works:
SVGordon_Test002c_datashape
In your case the data shape is 
gordon9_datashape1
but even if I simplify it to be as in my case it does throw an error:
gordon9_datashape2

portnov:
What is going on.
Try this one, it's simpler.
[gordon5.zip](https://github.com/nortikin/sverchok/files/6728349/gordon5.zip)

rendetto:
gordon5 does work, but it does not use the Exact T values:
gordon5

when I plug the T values the same kind of error appears:
gordon5_usingT

rendetto:
Nope, it does not want to cooperate. I rebuilt all the nodes from scratch with the same settings you used, just in case if some node had some bad old data - but got this error again.

portnov:
I'm using blender 2.93.1 from tar.xz on blender.org, on debian unstable.
What is your Blender / OS version?

rendetto:
I'm using one of the last py37 blender builds: blender-2.92.0-a9092768c003-windows64

rendetto:
Yes, just tested blender 2.93.1 and the error is gone.

portnov:
Interesting. I reproduced your error ("cannot broadcast...") on Blender 2.93.0.
I wonder why such small change of blender version triggers algorithmic error... most probably floating-point precision issue again...

rendetto:
Also I tried to open one of my last files made on 2.92 and blender2.93.1 crashed.

portnov:
I added some rounding. This should remove "cannot broadcast" issue for both blender versions.

rendetto:
Does this rounding affects only knot vector accuracy for Gordon specifically, or it is applied globally?

portnov:
For Gordon specifically.
It rounds knotvector values of intermediate loft and interpolated surfaces to N decimal digits. So, for example, if one surface has knot at u=0.100001, and another has knot at v=0.100002, we can say that it is the same knot u=0.1, instead of producing a surface with two knots very close to one another.

rendetto:
Great, got it! 
Meanwhile I'm downloading the 2.93.2 candidate and will test it also. 

A small off topic - seems on 2.93.1 + the latest FreeCAD0.20 the Mefisto method started producing edges and polys again...but exporting curves still throws the same error reported. I'll update the issues accordingly after some more testing because I want to check 2.93.2 also.

rendetto:
Just tested Blender2.93.2(win64) with gordon9.zip and there are no errors.

rendetto:
I tested the Explicit T values on my test file. Now the surface goes through the sections really well, but there is brake in the internal continuity. I used the "zebra" matcap to illustrate the effect.

SVGordon_Test002_ExplicitT

Maybe there is no way to interpolate the sections so perfect and in the same time to maintain the initial degree and control points structure of the source curves. Nevertheless I like the initial result being really tight on the contour curves and in the same time identical to the source curves structure and degree. In many cases that's more important than interpolating exactly through the sections.

I wanted to compare this particular setup as being identical to the Birail case. Mainly because I have it set with the Birail node where I get somewhat better (but again not exact) interpolation of the sections. The curious thing with the Birail node is that the best result is achieved using exactly 5 VSections(an equal number to the source curves in V direction) as control parameter and everything above created somewhat bigger deviation (but better parametrization regarding the initial placement of the sections).

portnov:
> Maybe there is no way to interpolate the sections so perfect 

AFAIK there is a way, but it's too complicated for me to implement in nearest future :)

portnov:
As far as I understand, it is possible to do reparametrization as follows:
* write down pairs of T-values: `(existing T, wanted T)`
* compose a smooth function which interpolates these pairs. Most probably with Legendre polynoms.
* Apply methods described in p.6.4 "Reparameterization of NURBS Curves and Surfaces" of "The NURBS Book" to reparametrize the curve with that interpolating function.

I did not dig into TiGL / FreeCAD implementation, but I suspect that they use the same general approach. It looks simple when described in general words, but it involves quite significant amounts of mathematics :)

The good thing about this approach is that it uses smooth interpolation. My current approach, in turn, is much simpler, but it uses picewise-linear interpolation:
* Cut the curve into segments at original T values
* Use (trivial) linear reparametrization for each segment
* Concatenate resulting segments into new curve.

rendetto:
Thanks a lot. I'll do a search on those topics.

As for TiGL, this is the paper I know of https://arxiv.org/pdf/1810.10795.pdf . Hope you can find it useful in some way.

portnov:
Interesting. It uses the following approach:
* To make a smooth interpolation function (`sigma`), a B-spline is used.
* Instead of exact reparametrization "a la NURBS book", it samples composed function `f(sigma(t))` at number of points and approximates resulting points with a new nurbs curve.
* for this approximation, "a hybrid approximation / interpolation technique" is used.

rendetto:
The control it gives is by ramping up some precision tolerance value and a target maximum of control points (which could be not the best approach since you can never get exactly what you want). The "precise enough" result can work in certain cases, but when the goal is high precision/quality surface modelling it hits its limits because of the oscillation and the unpredictable control vertices structure/degree/knot vector.

FCGordon

Also it could be a real disaster, but that I guess is a matter of source curves distribution.

FCGordon_waves

portnov:
> I tested the Explicit T values on my test file. Now the surface goes through the sections really well, but there is brake in the internal continuity. I used the "zebra" matcap to illustrate the effect.
> 
> SVGordon_Test002_ExplicitT
> 

By the way, I'm not sure at this scale, but it looks like you have curve intersection point very near of "even" T parameter value. Due to cut-and-join approach I described above, this leads to two rows of control points very near one another.

rendetto:
It's not just near - the sections touch the paths precisely at the provided T values. I was assuming that this is the way to construct the curve net.

portnov:
I investigated this issue for a while. It seems, with current implementation of reparametrization, it is not possible to automatically have smooth continuity every time.

In some cases it is possible to achieve good results by "manually" removing excessive knots. In current state of this branch, you can pass result of Gordon node through Formula node with the following formula: `surface.remove_knot('V', 0.25, 2)`. It means remove knot V=0.25 two times. In your example setup, this gives good result for me. But in general, there is no guarantee, so I don't think it's good idea to do so automatically.

rendetto:
Thanks a lot! It looks a lot better now. I was planning to make a feature request of removing knots, but saw you already started implementing it :) 
SVGordon_Test002_ExplicitT_RemovedKnots

I agree that the manual intervention is the appropriate approach. An automatic one could solve a number of cases but never all the cases, so tweaking the things manually is a very good option.

PR #4190, opened 2021-06-29
Fix list levels node

### LINKED ISSUE #4184: List levels
## Problem statement

With the same input the output of the node is not persistent. When the shape of input data is changed the node can forgot about it's settings and when the shape returns to initial state the output is different.

List levels bug

Durman:

Fixes #4184

## Preflight checklist

- [x] Code changes complete.


Durman:
Fixed node can work with old Blender files, with old JSON files but if to save fixed node and open it in Sverchok before the fix the sttings will be lost. The question is should I create next MK version then?

Durman:
Also UI refactoring

image

PR #4196, opened 2021-07-01
introduce node bound caching (for SNLite at least...)

zeffii:
- [x] static_caching  (basic, one per node)
- [ ] responsive_caching
     - some kind of combination tuple as hashmap key:

         ```python
            component_id = self.node_id
            component_function_text = hash(inspect.getsource(function_to_use))
            component_variables_hash = hash(str(variables))

            cache_key = (component_id, component_function_text, component_variables_hash)
         ```

     - so if the cache is invalidated if 
         - user has code changes to the input function
         - user changes input variables
     
     - considerations:
         - do i want automatic cache culling?
             - include function name as part of the hash key so as to make it more convenient to delete cache's where keys contains the first two components
- [x] make it a mixin
     - altho, because not all scripts will use this maybe having it as a default mixin is not a good idea...

and see how that goes

### DIFF
--- a/nodes/script/script1_lite.py
+++ b/nodes/script/script1_lite.py
@@ -21,6 +21,7 @@ import sys
 import ast
 import json
 import inspect
+import textwrap
 import traceback
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


zeffii:
which let's us write cache a potentially heavy processing step,
```python
"""
in frame_nr s d=1 n=2
out points v
out pressure s
"""

def make_dict():
    true = True
    false = False
    return eval(bpy.data.texts.get("gp_boy.json").as_string())
    
if bpy.data.texts.get("gp_boy.json"):
    
    cache = self.get_static_cache(make_dict, ())
    
    # add your own frame range testing here, exercise to the user.
    frame_offset = -1
    this_dict = cache[frame_nr + frame_offset]

    np_points = np.array(this_dict['vertices']).reshape((-1, 3)).tolist()
    points.append(np_points)
    pressure.append(  np.array(  this_dict['stroke']['force']))

```

PR #4205, opened 2021-07-03
Fix source typos

luzpaz:
Follow-up to 165a9bf4c however may include breaking API changes. All similar typos have been put in to separate commits so that they may be cherrypicked if necessary.

### DIFF
--- a/data_structure.py
+++ b/data_structure.py
@@ -423,7 +423,7 @@ iter_list_match_func = {
 # NOTE, these function cannot possibly work in all scenarios, use with care
 
 def dataCorrect(data, nominal_dept=2):
-    """data from nasting to standart: TO container( objects( lists( floats, ), ), )
+    """data from nesting to standard: TO container( objects( lists( floats, ), ), )
     """
     dept = levelsOflist(data)
     output = []
@@ -432,11 +432,11 @@ def dataCorrect(data, nominal_dept=2):
     if dept < 2:
         return data #[dept, data]
     else:
-        output = dataStandart(data, dept, nominal_dept)
+        output = dataStandard(data, dept, nominal_dept)
         return output
 
 def dataCorrect_np(data, nominal_dept=2):
-    """data from nasting to standart: TO container( objects( lists( floats, ), ), )
+    """data from nesting to standard: TO container( objects( lists( floats, ), ), )
     """
     dept = levels_of_list_or_np(data)
     output = []
@@ -445,11 +445,11 @@ def dataCorrect_np(data, nominal_dept=2):
     if dept < 2:
         return data #[dept, data]
     else:
-        output = dataStandart(data, dept, nominal_dept)
+        output = dataStandard(data, dept, nominal_dept)
         return output
 
 def dataSpoil(data, dept):
-    """from standart data to initial levels: to nested lists
+    """from standard data to initial levels: to nested lists
      container( objects( lists( nested_lists( floats, ), ), ), ) это невозможно!
     """
     __doc__ = 'preparing and making spoil'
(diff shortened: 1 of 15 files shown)


nodes/object_nodes/closest_point_on_mesh2.py
```diff
@@ -43,7 +43,7 @@ def sv_init(self, context):
         si('SvVerticesSocket', "point").use_prop = True
         si('SvStringsSocket', "max_dist").prop_name = "Mdist"
 
-        so('SvStringsSocket', "succes")
+        so('SvStringsSocket', "success")
```
Durman, line 46: Names of sockets should definitely not be changed without creating MK next node version.

zeffii:
all proposed changes are valid, but as you state there are some breaking changes, but particularly this is limited to code inside `sv_init` where we initialize sockets. ( some interpretation of `self.inputs.new(socket_type, name)` ) .  Socket names tend to be bound to the socket object instance when the nodetree is stored inside a `.blend` , and changing socket names will invalidate those .blend files that store such nodes.   There will be some resistance to those typo corrections, even if they are correct.

zeffii:
in these cases maybe it's worth using `socket.identifier` again,  to fix the typo in the UI (user facing being the most important).

luzpaz:
I defer to you on how you choose to proceed

zeffii:
also words like `dept` and `nominal_dept`

luzpaz:
how should I proceed?

Durman:
You can just remove fixes in `sv_init` methods.

nodes/object_nodes/closest_point_on_mesh2.py
luzpaz: Reverted in 27d9b24

data_structure.py
```diff
@@ -471,13 +471,13 @@ def Spoil(dat, dep):
     return out
 
 
-def dataStandart(data, dept, nominal_dept):
-    """data from nasting to standart: TO container( objects( lists( floats, ), ), )"""
+def dataStandard(data, dept, nominal_dept):
```
Durman, line 474:
```suggestion
def data_standard(data, dept, nominal_dept):
```

utils/dummy_nodes.py
```diff
@@ -76,7 +76,7 @@ def create_dummy_class(bl_id):
         {
             'bl_idname': bl_id,
             'bl_label': label,
-            "missing_dependecy": dependency_name,
+            "missing_dependency": dependency_name,
```
Durman, line 79: @vicdoval What do you think, can this break backward compatibility?

vicdoval: At a first glance I think it wont break anything. Maybe if you save a nodetree with a dummy node and you try to open it again it may complain, but it will be a ui complain

data_structure.py
```diff
@@ -432,11 +432,11 @@ def dataCorrect(data, nominal_dept=2):
     if dept < 2:
         return data #[dept, data]
     else:
-        output = dataStandart(data, dept, nominal_dept)
+        output = dataStandard(data, dept, nominal_dept)
```
Durman, line 435:
```suggestion
        output = data_standard(data, dept, nominal_dept)
```

data_structure.py
```diff
@@ -445,11 +445,11 @@ def dataCorrect_np(data, nominal_dept=2):
...
```
Durman, line 448:
```suggestion
        output = data_standard(data, dept, nominal_dept)
```

data_structure.py
```diff
@@ -471,13 +471,13 @@ def Spoil(dat, dep):
...
-def dataStandart(data, dept, nominal_dept):
-    """data from nasting to standart: TO container( objects( lists( floats, ), ), )"""
+def data_standard(data, dept, nominal_dept):
+    """data from nesting to standard: TO container( objects( lists( floats, ), ), )"""
     deptl = dept - 1
     output = []
     for object in data:
         if deptl >= nominal_dept:
-            output.extend(dataStandart(object, deptl, nominal_dept))
+            output.extend(dataStandard(object, deptl, nominal_dept))
```
Durman, line 480:
```suggestion
            output.extend(data_standard(object, deptl, nominal_dept))
```

Durman (approved):

Issue #511, opened 2015-01-03
Poison Disc distribution for rectangle.


zeffii:
As per: 
http://www.jasondavies.com/poisson-disc/

![image](https://cloud.githubusercontent.com/assets/619340/5602344/23f0786c-9347-11e4-9e49-23972aa7b159.png)

would be neat to have first as a script node, then as a node with multimode, to take any input geometry and place vertices + edges at even distributions over their surfaces

and best candidate:
http://bl.ocks.org/mbostock/1893974

enzyme69:
I always like this kind of algorithm that space objects evenly. Also nice
when work with delaunay triangulation.

In fact if we have point scatter node, that works together with this.

nortikin:
tryed to make so, but forgot to continue.
It is scripten node in nikitron folder, https://github.com/nortikin/sverchok/blob/master/node_scripts/templates/nikitron/circles_on_points.py
It does not work

zeffii:
I have a few other things to do first, but will complete an implementation soon

zeffii:
further inspiration is "Spectral sampling of manifolds"
https://graphics.ethz.ch/publications/papers/paperOzt10.php

zeffii:
https://gist.github.com/anonymous/754b8effd46791304d49
translation from js to py isn't easy here.. will try again later

zeffii:
found a nice python implementation: http://nbviewer.ipython.org/github/HyperionAnalytics/PyDataNYC2014/blob/master/poisson_disc_sampling.ipynb

zeffii:
```python
# BSD License
# http://nbviewer.ipython.org/github/HyperionAnalytics/PyDataNYC2014/blob/master/poisson_disc_sampling.ipynb

import numpy as np
import math
import random
from random import random

def poisson_disc_select(img_width, img_hight, r, n_try):
    """
    Select points from Poisson disc
    Input:
    img_width - integer, 1 to n
    img_hight - integer, 1 to n
    r - minimum didtance between two points, float
    n_try - number of randomly sampled points per try, integer, 1 - n
    Output:
    sample_pts_array - floats array, shape[img_width*img_hight, 2]
    """
    r_square = r**2
    A = 3*r_square
    cell_size = r/math.sqrt(2)
    grid_width = int(math.ceil(img_width/cell_size))
    grid_hight = int(math.ceil(img_hight/cell_size))
    grid = [None]*grid_width*grid_hight
    queue = list()
    queue_size = 0
[...]

PR #4210, opened 2021-07-05
Snlite corrections 8

zeffii:
cherrypicking from older commits.

- Functions that can be compiled using `numba` (using `njit`), must be cached for snlite specific reasons. 
- The name of the `function` is the `cache key` _under the hood_

```python
sv_njit_clear(make_hadley3)  # <-- if the cached function needs to be cleaned, comment this out when happy

# here pz are the results of the calculations.
pz = sv_njit(make_hadley3, parameters)
```
allowing this to be written

```python
"""
in  alpha   s d=0.2 n=2
in  beta    s d=4.0 n=2
in  delta   s d=8.0 n=2
in  gamma   s d=1.0 n=2
in  factor  s d=7.0 n=2
in  dif     s d=0.005 n=2
in  t       s d=20000 n=2
in  step    s d=5 n=2
in  mod     s d=4 n=2
out verts   v
""" 

import sverchok
import numba
from numba import njit

@njit
def make_hadley3(alpha=0.2, beta=4.0, delta=8.0, gamma=1.0, factor=7.0, dif=0.005, t=20000, step=5, mod=4):
    Points = []
    p = Points.extend

    x = 0.1
    y = 0.0
    z = 0.0
    AxD = (alpha * delta)
    for i in range(1, t):
        for j in range(step):
            x += dif * (-(y**2) - (z**2) - (alpha*x) + AxD)
            y += dif * ((x*y) - (beta*x*z) - y + gamma)
            z += dif * ((beta*x*y) + (x*z) - z)
        if i % mod == 0:
            p([x*factor, y*factor, z*factor])

    return Points

# sv_njit_clear(make_hadley3)
parameters = (alpha, beta, delta, gamma, factor, dif, t, step, mod)
pz = sv_njit(make_hadley3, parameters)

verts.append(np.array(pz).reshape((-1, 3)).tolist())
```

### DIFF
--- a/nodes/script/script1_lite.py
+++ b/nodes/script/script1_lite.py
@@ -107,7 +107,15 @@ class SvScriptNodeLiteTextImport(bpy.types.Operator):
 
 
 class SvScriptNodeLite(bpy.types.Node, SverchCustomTreeNode, SvAnimatableNode):
-    ''' snl SN Lite /// a lite version of SN '''
+
+    """
+    Triggers: snl
+    Tooltip: Script Node Lite
+    
+    This code represents a conscious weighing of conveniences to the user, vs somewhat harder to understand
+    code under the hood. This code evolved as design specs changed, while providing continued support for
+    previous implementation details.
+    """
 
     bl_idname = 'SvScriptNodeLite'
     bl_label = 'Scripted Node Lite'
--- a/nodes/script/script1_lite.py
+++ b/nodes/script/script1_lite.py
@@ -121,23 +121,20 @@ class SvScriptNodeLite(bpy.types.Node, SverchCustomTreeNode, SvAnimatableNode):
     bl_label = 'Scripted Node Lite'
     bl_icon = 'SCRIPTPLUGINS'
 
-    def custom_enum_func(self, context):
+    def return_enumeration(self, enum_name=""):
         ND = self.node_dict.get(hash(self))
         if ND:
-            enum_list = ND['sockets']['custom_enum']
+            enum_list = ND['sockets'][enum_name]
             if enum_list:
                 return [(ce, ce, '', idx) for idx, ce in enumerate(enum_list)]
 
         return [("A", "A", '', 0), ("B", "B", '', 1)]
+
+    def custom_enum_func(self, context):
+        return self.return_enumeration(enum_name='custom_enum')
 
     def custom_enum_func_2(self, context):
-        ND = self.node_dict.get(hash(self))
-        if ND:
-            enum_list = ND['sockets']['custom_enum_2']
-            if enum_list:
-                return [(ce, ce, '', idx) for idx, ce in enumerate(enum_list)]
-
-        return [("A", "A", '', 0), ("B", "B", '', 1)]
+        return self.return_enumeration(enum_name='custom_enum_2')
 
 
     def custom_callback(self, context, operator):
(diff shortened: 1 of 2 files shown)


zeffii:
@gerroon didn't see the message in the other thread:  https://github.com/nortikin/sverchok/issues/511

digging up some old scripts `poisson disc`  , modified minimally to use `numba` , but I don't notice the different so much.  Will only work if you have the `numba` dependency installed (see preferences, if it's not there you have an older sverchok probably)

```python
"""
in img_width s d=3.0 n=2
in img_height s d=4.0 n=2
in radius s d=0.23 n=2
in retries s d=30 n=2
out verts_out v
"""

from numba import njit
import numpy as np
import math
import random
from random import random

@njit
def poisson_disc_select(img_width, img_height, r, n_try):
    """
    Select points from Poisson disc
    Input:
    img_width - integer, 1 to n
    img_height - integer, 1 to n
    r - minimum didtance between two points, float
    n_try - number of randomly sampled points per try, integer, 1 - n
    Output:
    sample_pts_array - floats array, shape[img_width*img_hight, 2]
    """
    r_square = r**2
    A = 3*r_square
    cell_size = r/math.sqrt(2)
    grid_width = int(math.ceil(img_width/cell_size))
    grid_height = int(math.ceil(img_height/cell_size))
    grid = [None] * grid_width * grid_height
    queue = list()
    queue_size = 0
    sample_size = 0

    def distance(x, y):
        x_idx = int(x/cell_size)
        y_idx = int(y/cell_size)
        x0 = max(x_idx-2, 0)
        y0 = max(y_idx-2, 0)
        x1 = min(x_idx+3, grid_width)
        y1 = min(y_idx+3, grid_height)

        for w in range(y0, y1):
            p = w*grid_width
            for h in range(x0, x1):
                if grid[p+h]:
                    s = grid[p+h]
                    dx = s[0]-x
                    dy = s[1]-y
                    if dx**2 + dy**2 < r_square:
                        return False
        return True

    def set_point(x, y):
        nonlocal queue, grid, queue_size, sample_size
        s = [x, y]
        queue.append(s)
        grid[grid_width*int(y/cell_size) + int(x/cell_size)] = s
        queue_size += 1
        sample_size += 1
        return s

    # Set first data point
    if sample_size == 0:
        x = random()*img_width
        y = random()*img_height
        set_point(x, y)

    while queue_size:
        x_idx = int(random()*queue_size)
        s = queue[x_idx]

        # Generate random point in annulus [r, 2r]
        for y_idx in range(0, n_try):
            a = 2*math.pi*random()
            b = math.sqrt(A*random() + r_square)

            x = s[0] + b*math.cos(a)
            y = s[1] + b*math.sin(a)

            # Set point if farther than r from any other point
            if 0 <= x and x < img_width and 0 <= y and y < img_height and distance(x, y):
                set_point(x, y)

        del queue[x_idx]
        queue_size -= 1

    sample_pts = list(filter(None, grid))

    sample_pts_array = np.asfarray(sample_pts)
    N = sample_pts_array.shape[0]
    padded_pts = np.c_[ sample_pts_array, np.zeros(N) ] 

    return padded_pts.tolist()

params = (img_height, img_width, radius, retries)
pts_poisson = sv_njit(poisson_disc_select, params)
# pts_poisson = poisson_disc_select(*params)
verts_out.append(pts_poisson)


```

PR #4212, opened 2021-07-08
Internal snlite njit processor

zeffii:
lets snlite drop the decorator
lets snlite reload the cache after blend reload


zeffii:
this no longer requires the import of `numba` or `numba.njit` , there's just two features now
```
sv_njit_clear(function_to_njit)
sv_njit(function_to_njit, parameters)
```

zeffii:
this `sv_njit` approach is not useful for the scenario where there are f.ex 6 short functions that can be njitted, it will work but it will get in the way of clarity.

PR #4241, opened 2021-07-23
Debug print socket bool

zeffii:
resolves longstanding desire to be able to toggle printing per socket from the node ui,  without `draw_buttons_ext`

image

### DIFF
--- a/nodes/text/debug_print.py
+++ b/nodes/text/debug_print.py
@@ -35,9 +35,6 @@ class SvDebugPrintNode(bpy.types.Node, SverchCustomTreeNode):
     base_name = 'Data '
     multi_socket_type = 'SvStringsSocket'
 
-    # I wanted to show the bool so you could turn off and on individual sockets
-    # but needs changes in node_s, want to think a bit more before adding an index option to
-    # stringsockets, for now draw_button_ext
     print_socket: BoolVectorProperty(
         name='Print', default=defaults, size=32, update=updateNode)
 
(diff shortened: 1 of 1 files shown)


nodes/text/debug_print.py
```diff
@@ -55,9 +53,19 @@ def draw_buttons_ext(self, context, layout):
...
+        layout.label(text=text)
+        icon = ("HIDE_ON", "HIDE_OFF", )[self.print_socket[socket.index]]
+        layout.prop(self, "print_socket", icon=icon, index=socket.index, text="")
+
     def sv_update(self):
         multi_socket(self, min=1)
 
+        for socket in self.inputs:
+            if not socket.custom_draw:
+                socket.custom_draw = "draw_socket_boolean"
```
Durman, line 67: This looks a bit suspicions to apply every update,

zeffii: because `sv_init` is only called once, and sockets are added dynamically by `multi_socket`. and they won't get the custom_draw string.  

this does nothing if the socket has a custom draw.

zeffii: putting that code inside `process()` is possible, but UI does not update immediately.. the user must interact with the blender UI in a way to triggers the process function. 

maybe i'm not understanding what's happening here, but it does seem that the node's process function is called if I hover over the label area of a socket (no clicking or anything..)

socket_showing_activity_in_process

PR #4252, opened 2021-07-25
Fix: Enums in Surface from Curves node

### LINKED ISSUE #4237: Enums in Surface from Curves node
## Problem statement

https://github.com/nortikin/sverchok/blob/a611d3ed8459737d71ae4614d90a6549f7c951fe/nodes/surface/interpolating_surface.py#L29-L37

https://github.com/nortikin/sverchok/blob/a611d3ed8459737d71ae4614d90a6549f7c951fe/nodes/surface/interpolating_surface.py#L45-L54

https://docs.blender.org/api/current/bpy.props.html#bpy.props.EnumProperty
> There is a known bug with using a callback, Python must keep a reference to the strings returned by the callback or Blender will misbehave or even crash.

`D.node_groups['NodeTree'].nodes['Surface from Curves'].bl_rna.properties['nurbs_implementation'].enum_items`
`<bpy_collection[0], EnumProperty.enum_items>` but should be `<bpy_collection[1]`

Solution is to move items outside of the functions.

Patil2099:
Fixes  #4237

### DIFF
--- a/nodes/surface/interpolating_surface.py
+++ b/nodes/surface/interpolating_surface.py
@@ -25,6 +25,7 @@ class SvInterpolatingSurfaceNode(bpy.types.Node, SverchCustomTreeNode):
     bl_label = 'Surface from Curves'
     bl_icon = 'OUTLINER_OB_EMPTY'
     sv_icon = 'SV_SURFACE_FROM_CURVES'
+    items = []
 
     def get_interp_modes(self, context):
         modes = [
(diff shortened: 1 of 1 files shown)


Durman:
Probably this can be done in this way but in this case in the beginning of the function the list should be cleared otherwise the list will grow each call. Also I would prefer the list to have more accurate name outside of the function. Probably it can take some part from node attribute where the enum is used. And this `get_interp_modes` method is waiting for care too.

PR #4255, opened 2021-07-29
Improve Live update mode

### LINKED ISSUE #4254: "CancelError()" is blocking the tree live processing
This error occurred when I tried to Live Update a tree.
Live_Update_Issue
The same tree is processed without problems when animated.
The tree was running with no errors with the previous update system.

The other weird thing is that this error appears very erratic and on different nodes. Is that normal? For example the next time I executed Live Update the left node didn't throw an error or red color, but the tree sill doesn't update.
Live_Update_Issue_no_red
Here on another run of the Live Update I got this(tree still doesn't update):
Live_Update_Issue_both_green

I'm also a bit lost with those flashing colors of the nodes. What means red flashing and what green flashing? Why are those errors and how I have to read this info so I can debug the problems?
Thanks

Durman:

Instead of cancelling tree evaluation the mode should skipp events if evaluation of the tree is still in progress.
It should make the work in this mode to be more pleasant.

As a side effect there is slight unsynchronize between user action and a tree response.

Live update

Closes #4254

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.

### DIFF
--- a/core/events.py
+++ b/core/events.py
@@ -34,10 +34,11 @@ class TreeEvent:
     FORCE_UPDATE = 'force_update'  # rebuild tree and reevaluate every node
     FRAME_CHANGE = 'frame_change'  # unlike other updates this one should be un-cancellable
 
-    def __init__(self, event_type: str, tree: SverchCustomTree, updated_nodes: Iterable[SvNode] = None):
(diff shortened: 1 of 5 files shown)


rendetto:
@Durman Thanks a lot! I just tested this PR on the tree in question and it started running interactively. I don't mind the lag in the viewport.

PR #4268, opened 2021-08-07
"Bevel a Curve (Surface)" node et al

portnov:
* "Taper & Sweep" node gets a "NURBS" mode. In this mode, it will process only NURBS curves, and output NURBS surface.
* Old "Bevel a Curve" node is renamed to "Bevel a Curve (Mesh)"
* New node is introduced: "Bevel a Curve (Surface)". In general, it is similar to "Bevel a Curve (Mesh)", but instead of working with mesh objects, it takes Curve objects and outputs a Surface object. Depending on settings, it can either work with arbitrary Curve objects and generate a generic Surface object, or work with NURBS curves and output a NURBS surface.

Screenshot_20210807_123146
Screenshot_20210805_214517
Screenshot_20210807_221243
Screenshot_20210807_223851
Screenshot_20210807_223550
Screenshot_20210807_225058

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


nodes/surface/bevel_curve.py
```diff
@@ -0,0 +1,268 @@
...
+        taper = SvLine.from_two_points(p1, p2).to_nurbs()
+        taper = taper.elevate_degree(target = max(profile_degree, 2))
+        return taper
+
+    def _get_orient_axis_idx(self):
+        return 'XYZ'.index(self.orient_axis)
+
+    def process(self):
+        if not any(socket.is_linked for socket in self.outputs):
+            return
```
Durman, line 196: I don't think we need this check now. It makes update system more complex. It can be even less efficient because some nodes update even when there are no output links but update system have to update them anyway. Update system does not know whether the node has this check or no.
https://github.com/nortikin/sverchok/blob/3d76f45c9c2c6e423335c42f6ebc33392e9aad2b/core/main_tree_handler.py#L391-L396

portnov: I'm afraid I don't understand.
If the node does not have connected outputs, why spend time doing complex calculations?

Durman: Some nodes does not have such optimization. How update system should understand whether a node has this check or not to update the node properly?

portnov: As far as I understand, the only case when the node has to be updated without having connected outputs - is the case when it has "side effects", such as drawing in the viewport or generating objects in scene. We have only few of such node classes.
So, such nodes could have special class property (`has_side_effects = True`), so that update system could recognize them and trigger their update even when they do not have connected outputs.
For other nodes, update should not be triggered when they do not have connected outputs.
If we will have such logic in update system, we can remove the check from the nodes.

portnov: But for now, do I correctly understand that the update system triggers `process()` of the node even if does not have connected outputs?
In this case nodes which do complex calculations have to do the check themselves.

Durman: Yes the `process` is called even there are no output connection of a node. I guess there are case when such update are expected for example SNL node could do something without passing any result to next nodes.
The thing is that such checks do not let make optimization in update system and adding this checks in new nodes make future improvements in this direction harder.

portnov: But without this check node trees which include complex nodes do excessive computations. And at least for some nodes it means a lot of time.

PR #4276, opened 2021-08-10
add wrapper to sv_extra

zeffii:
a short wrapper to automatically call self.process_node(context) after the code of the `func` completes


zeffii:
this can be part of sv-extra instead

PR #4299, opened 2021-08-20
Scene handler

### LINKED ISSUE #4283: Upgrade Live update operator
## Problem statement

https://github.com/JacquesLucke/animation_nodes/pull/1769#issuecomment-890803459
>  I used the scene update handler, but due to unknown reasons this handler was always executed. We only got a "real" scene changed handler reeelatively recently, using it sounds like a good idea.

Probably instead of timer the operator could use the scene changed handler. It would help to avoid useless updates (when nothing was changed). At least it's worth to experiment.

Durman:

Closes #4283. It should update nodes reading data from Blender upon changes in the scene.
I feel this make work with Sverchok more comfortable in case when workflow include reading data from Blender.
Update live operator can be removed.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.


Durman:
I removed `AnimatableNode` class. Now it's possible make a node to be animatble in this way.

```py
class Node:
    is_animation_dependent = True
```

Also, instead of `draw_buttons`, `sv_draw_buttons` should be used now. It will let to draw extra buttons in the node UI.

Also it's possible to add `is_scene_dependent = True` property. In this case the node will be update on user changes in the scene.

rendetto:
I just did a quick test on one of my heaviest scenes and it works perfect.
Great solution that makes a ton of difference for my workflow.

Thank you @Durman!

Durman:
@rendetto I'm glad that it's useful =)

PR #4301, opened 2021-08-21
Console node tokenizer api change

zeffii:
- fixes PEG lexer changes in `tokenize` module for 3.9
- adds background color change (may be better to mix but this is all i found time for)

### DIFF
--- a/utils/sv_extra_search.py
+++ b/utils/sv_extra_search.py
@@ -59,9 +59,14 @@ def ensure_short_description(description):
 
 def ensure_valid_show_string(nodetype):
 
-    loop_reverse[nodetype.bl_label] = nodetype.bl_idname
-    description = nodetype.bl_rna.docstring.get_shorthand()
-    return nodetype.bl_label + ensure_short_description(description)
+    try:
+        loop_reverse[nodetype.bl_label] = nodetype.bl_idname
+        description = nodetype.bl_rna.docstring.get_shorthand()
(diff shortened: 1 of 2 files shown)


zeffii:
altho it wasn't my intention to modify
```python
def ensure_valid_show_string(nodetype):

    loop_reverse[nodetype.bl_label] = nodetype.bl_idname
    description = nodetype.bl_rna.docstring.get_shorthand()
    return nodetype.bl_label + ensure_short_description(description)
```
with this PR.   the change should not interfere with the search feature, and will result in a more interesting `exception`

utils/sv_extra_search.py
```diff
@@ -59,9 +59,14 @@ def ensure_short_description(description):
...
-    loop_reverse[nodetype.bl_label] = nodetype.bl_idname
-    description = nodetype.bl_rna.docstring.get_shorthand()
-    return nodetype.bl_label + ensure_short_description(description)
+    try:
+        loop_reverse[nodetype.bl_label] = nodetype.bl_idname
+        description = nodetype.bl_rna.docstring.get_shorthand()
+        return nodetype.bl_label + ensure_short_description(description)
+    except Exception as err:
+        print(f"{nodetype}")
+        print(err)
```
Durman, line 68:
```py
except Exception as err:
    error(f'Nodetype "{node_type}" is not valid - {err}')
```

zeffii: yes ok.

Issue #468, opened 2014-11-11
Speed up the execution


JacquesLucke:
Hi,
I just browsed a bit through your code and found something that really slows down the node tree execution.
While I speeded up the Animation nodes I noticed that everytime I call _socket.links_, internally it iterates through every link in the node tree. (I actually found the code where this happens but forgot where it is). 
[EDIT] found the line: https://developer.blender.org/diffusion/B/browse/master/release/scripts/modules/bpy_types.py;e43c5fa005c941aa773f35fb6b8398afe37b25e9$816 [/EDIT]

This seams like a small thing but the effect of changing these lines is huge. By removing this in several areas in my kernel the script generation progress became around 10 times faster! And here you seam to call this in every single node.
this is one example commit of 5 or so:
https://github.com/JacquesLucke/animation-nodes/commit/399dae85aaae38c395879361d0df0437bb5b15d1 

The trick I used is that everytime I need to know where the data comes from I iterate through all the links myself AND write the 'origin sockets' in a dictionary.

https://github.com/JacquesLucke/animation-nodes/blob/master/mn_network_code_generator.py#L554-L593

You just have to be sure that you clear this dictionary after one update process. Otherwise Blender can easily crash. This method also help to deal with Reroute nodes, because a function can define where the data comes from instead of the actual link. I do this here:
https://github.com/JacquesLucke/animation-nodes/blob/master/mn_utils.py#L115-L126

I hope that helps to improve the performance. 
btw: the good thing is that this is only a back-end change (an additional abstraction layer), so it is 100% compatible with existing nodes. Another good thing is that you can do this change step by step. No need to do a whole rewrite until it works again.

PR #4323, opened 2021-09-02
Sv get bottleneck

### LINKED ISSUE #4227: Sv_get bottleneck
## Problem statement

Before starting process a node is reading data from previous node and other node, thanks to Blender data structure, can be get only via search. So in big trees and in trees with loop nodes this search takes significant time. I have two benchmark examples. First example has loops and the search takes 50% of all execution time. The second example does not have loop nodes but it's big. The search takes 10% of time there.

<details> 
  <summary>with loop nodes</summary>  

output

</details>

<details> 
  <summary>just big tree (400 nodes)</summary>  

output (1)

</details>

The solution is to make them not to search the previous nodes.

- First what come to mind is to use udpate system for the search. Update system has its own more efficient representation of a tree. But the thing is that it is not always has updated state. It's reliable only during update event. But nodes also can use search before update event (for example to find what is socket type of previous node)
- Another solution is to make update system to copy link to output of nodes to input of next nodes. The storage of data of input sockets can have similar implementation as it's now for output sockets.

Also there is another problem which probably should be fixed together. Output sockets inside node groups can have multiple outputs and current storage system of sockets data does not support multiple values. To solve the problem I made a trick in node groups. When a node group is evaluated all node IDs are changed and it helps to storage system to put values in appropriate place for current group. But it's really nasty thing to do. The node ids should be always constant.

Durman:

This PR changes data management of tree layout. Input sockets will have their data too to which a node will have access. Update system will pass data from output sockets to input. It's improve animation performance, performance of large trees and loop nodes (if they will be fixed) significantly.

Closes #4227

It lett remove performance penalty of Blender tree data structure entirely what was known quite long ago. #468

Modules refactoring:
execution process

## Todo

- [x] Fix groop nodes

Unfortunately it brokes work of some nodes which executes group of nodes in a tree. Probably author of those nodes would like to join to fix them?

- [x] Loop node
- [x] Evolver

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.


Durman:
Actually the PR became more than just solving the Sv Get problem. The update system was rewritten (again). But I suppose it's now more easy to understand because I simplified handling node groups. Now node groups does not have separate socket data catch per group node instance. It lets to apply the same approach to them as to main trees. 

Nodes development is remains unchanged except nodes which execute a tree by their own like Loop and Evolver nodes. Now input and output sockets storage data. Update system paths data from output to input sockets so when a node grubs data from input socket it does not have to perform any searches. Statistics now is kept on node objects directly and can be read via get item interface. It will be simply to have several update systems (if there is need of that) because all events path through single function which can deliver them father anywhere.

The tests show next performance improvements on my test files:
- Animation - about **x2.85**
- Loop nodes - **x1.35-1.6**
- Big trees (400 nodes) - **x1.25**
- Big trees but with nodes distributed in many node groups - no difference.

nodes/viz/viewer_draw_mk4.py
```diff
@@ -757,7 +754,7 @@ def migrate_from(self, old_node):
     def draw_property_socket(self, socket, context, layout):
         drawing_verts = socket.name == "Vertices"
         prop_to_show = "point_size" if drawing_verts else "line_width"
-        text = f"{socket.name}. {SvGetSocketInfo(socket)}"
+        text = f"{socket.name}. {str(socket.objects_number)}"
         layout.label(text=text)
```
zeffii, line 758: maybe this can be simplified?
```python
        text = f"{socket.name}. {str(socket.objects_number)}"
        layout.label(text=text)
```
to
```python

    <socketclass>
    def label_and_count(self, name=None):
        return f"{name or socket.name}. {self.objects_number}"        


    <some layout>
        layout.label(text=socket.label_and_count())
```

Durman: Yes. But I'd rather add such changes in separate PR because this one does not make any conceptual changes nor suggest any improvements in this area.

Probably nicer way would be to add number of objects outside of the custom draw function if the socket is connected. It seems is always expected behavior.

PR #4334, opened 2021-09-09
Revert: SvRecursiveNode mixin (#3954)

Durman:
If data with vertices was connected to mask socket it changed shape fo all output data.

## Preflight checklist

- [x] Code changes complete.

### DIFF
--- a/nodes/modifier_change/merge_by_distance.py
+++ b/nodes/modifier_change/merge_by_distance.py
@@ -24,9 +24,10 @@ from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode, Vector_generate, repeat_last, zip_long_repeat
 from sverchok.utils.sv_bmesh_utils import bmesh_from_pydata, face_data_from_bmesh_faces, vert_data_from_bmesh_verts
 from sverchok.utils.logging import info, debug
-from sverchok.utils.nodes_mixins.recursive_nodes import SvRecursiveNode
+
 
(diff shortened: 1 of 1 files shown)


vicdoval:
Why would you connect vertices data to the mask input? Is makes sense if you input a mask with a level 3 (as the vertices lists) the matching of data changes the output compared to if you input the usual level 2 mask

Durman:
I used mask node for filtering vertices data. The node does not provide another way.

I'm not against new functionality but it should extend existing functionality or create next version of a node if it changes existing behaviour.

vicdoval:
> I used mask node for filtering vertices data. The node does not provide another way.
> 
> I'm not against new functionality but it should extend existing functionality or create next version of a node if it changes existing behaviour.

This behaviour look to me more like a bug than a feature (what vectors are a True mask (1,1,1), (0,2,1)...)

Durman:
Hm.. it doesn't look like a bug to me. Any way it's useful functionality and there is no any alternative to it. I used it in many layouts and I guess that other users may do this too. My opinion is that this changes should cause another node version to be created.

vicdoval:
But in which cases the vector in interpreted as false value? Only when is (0,0,0)?

Durman:
I guess never.

vicdoval:
This pr may break layouts made after the SvRecursiveNode mixin to avoid breaking layouts (with a weird feature use) that existed previously. I think it should be reversed.

PR #4337, opened 2021-09-09
example upgrade

zeffii:
delete loose hot upgrade

### DIFF
--- a/nodes/modifier_change/delete_loose.py
+++ b/nodes/modifier_change/delete_loose.py
@@ -40,6 +40,11 @@ class SvDeleteLooseNode(bpy.types.Node, SverchCustomTreeNode):
         self.outputs.new('SvStringsSocket', 'VertsMask')
 
     def process(self):
+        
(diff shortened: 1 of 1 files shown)


nodes/modifier_change/delete_loose.py
```diff
@@ -64,8 +64,15 @@ def process(self):
...
-            self.outputs['VertsMask'].sv_set(verts_mask_out)
+        
+        # potentially older layouts have nodes where sv_init has not added this socket yet.
+        vm_socket = self.outputs.get('VertsMask')
+        if vm_socket and vm_socket.is_linked:
+            vm_socket.sv_set(verts_mask_out)
+        
+        # this autoupgrades older nodes.
+        if not vm_socket:
+            self.outputs.new('SvStringsSocket', 'VertsMask')
```
Durman, line 75: Probably the order should be opposite ))

zeffii: yep

zeffii:
it seems we don't have test coverage of `delete_loose.py`

PR #4347, opened 2021-09-11
Macro objectout to get objects data

zeffii:
should maybe be a rightclick option for nodes that have an Objects output. Using macro for now.


Durman:
It's not clear what is it and how to use.

PR #4385, opened 2021-12-06
Fix typo tanget->tangent

luzpaz:


nodes/vector/interpolation_mk3.py
```diff
@@ -104,8 +104,8 @@ def sv_init(self, context):
         self.inputs.new('SvVerticesSocket', 'Vertices')
         self.inputs.new('SvStringsSocket', 'Interval').prop_name = 't_in'
         self.outputs.new('SvVerticesSocket', 'Vertices')
-        self.outputs.new('SvVerticesSocket', 'Tanget')
-        self.outputs.new('SvVerticesSocket', 'Unit Tanget')
+        self.outputs.new('SvVerticesSocket', 'Tangent')
```
zeffii, line 107: `socket.name` renames should be avoided, they will break existing layouts in a very unpleasant way.

luzpaz: Is there a way to softly deprecate them?

zeffii:
i think in the case of output sockets, it is possible to compromise and do the user a favour by getting a reference to the socket and setting the `socket.label = "new name"`,  this changes the text displayed on the socket ui without changing the `socket.name` in a way that breaks layouts/blends.

i'm not in favour of deprecating the node just to fix the socket name, but this is only because deprecating nodes is still a bit of a convoluted process and has side effects.

luzpaz:
Fair enough. Maybe this approach can be considered for a major version release. Closing this PR.

zeffii:
I do want to thank you for taking the time to track these typos (and all the previous PR's). MK4 of that node should use corrected socket names and variable names. It may seem weird to know that much effort goes into the backend of these nodes and yet the part that the user sees may display a typo. Shows you how accurately people read, i think the brain stops at the first 5 letters usually (entered into the fovea) and the rest of the word is judged by an assumption of lower resolution shapes.

the socket's `name` property and node classname/bl_idname are some of the limited areas in Sverchok that are bound by Blender API. At least the socket has a `label` property,

PR #4386, opened 2021-12-12
Fix source typos

luzpaz:


nodes/modifier_change/mesh_separate.py
```diff
@@ -33,16 +33,16 @@ class SvSeparateMeshNode(bpy.types.Node, SverchCustomTreeNode):
 
     def sv_init(self, context):
         self.inputs.new('SvVerticesSocket', 'Vertices')
-        self.inputs.new('SvStringsSocket', 'Poly Egde')
+        self.inputs.new('SvStringsSocket', 'Poly Edge')
```
zeffii, line 36: this must be reversed.    socket names , both inputs and outputs  should not be renamed like this.

Issue #4432, opened 2022-04-23
Write FCStd node should create a FreeCAD file.


macdroid53:
Expected:
When Write FCStd node is executed creates file with specified name.

Actual:
When node is executed no file is created. If an empty FCStd file by the specified name exists, the data is written.

zeffii:
https://wiki.freecadweb.org/FreeCAD_API
suggests a function call like ` F.newDocument(doc_name) ` would be capable of starting a new empty file.  
it likes like F.open is really only for existing Freecad docs?

zeffii:
I see the `FCStd_write` node doesn't offer a way to start a new freecad file from scratch, only a way to pick up an existing file.

zeffii:
i propose we add a button to the FCStd_write node, which lets you 
   - pick a file location and a new file name, 
   - and then the node automatically adds a FilePath Node attached to the FCStd_write node 
       - The newly created FilePath will point at the File
       - The newly created File will be initialized as a new empty freecad file. 

yay/nay ?

macdroid53:
Hmm...I thought what happened when I tried the node was I could indeed select a path and filename (note: it by default offered .blend as the extension **_not_** .FCStd).
But, if the file did not exist and the node executed, it did nothing (which was good at that point because it would have overwritten my blend file of the same name?)
IMO, it should act as would be expected of any file output. If the specified name exists, ask if it should be overwritten, otherwise create the file and output the solid.

zeffii:
yeah OK. that makes a lot more sense. FilePath node allows you to type an arbitrary filename into the FileName text box, and it will pass the `current path + filename` (or actually `os.path.join(current_path, filename)`) into the FilePath output socket (even if the file does not exist)

zeffii:
perhaps something like this
```python
    try:
        from os.path import exists
        
        Fname = bpy.path.display_name_from_filepath(fc_file)
        
        if not exists(fc_file):
            F.newDocument(fc_file)
        else:
            F.open(fc_file)

    except Exception as err:
        info(f'FCStd open error, {err}')
        return

    F.setActiveDocument(Fname)
    fc_root = F.getDocument(Fname)

```

macdroid53:
This seems good.

I don't know if there is a way to do it. When I click on the socket and it creates a File path node, it would be nice if it defaulted to .FCStd rather than .blend. But, the code you show would at least help to prevent and accidental overwrite.

zeffii:
[...]

PR #4437, opened 2022-04-25
test genererating new FCStd file on the fly

zeffii:
tests

### DIFF
--- a/nodes/exchange/FCStd_write.py
+++ b/nodes/exchange/FCStd_write.py
@@ -147,9 +147,8 @@ def fc_write_parts(fc_file, verts, faces, part_name, solid, mod):
 
         if not exists(fc_file):
             doc = F.newDocument(Fname)
-            doc.FileName = fc_file
             doc.recompute()
-            doc.save()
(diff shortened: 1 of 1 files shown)


luzpaz:
If merged should close #4432

luzpaz:
Aside, is there a detailed tutorial of how to:
* build blender and sverchok
* build freecad 
* so one can test these developments locally?

zeffii:
@luzpaz 
- building Blender is way beyond the scope of the sverchok issue tracker or wiki
- Sverchok is merely an add-on which should run on any recent Blender, on any platform, b2.9.3 and upwards
- certain (optional) Sverchok nodes have dependencies and they might expect a particular version of Python. 
    - an example of such dependency is FreeCAD which may not be available with a python interface supported by Blender 2.9.x
    - in essence you would need to match the Blender version that is built with the same version as the python FreeCAD library is buit with.

The docs do offer some dependency hints: https://github.com/nortikin/sverchok/wiki/FreeCAD

zeffii:
not sure if this is standard for FreeCAD api, but this seems to create a `FCStd` and `FCStd1`. I will assume for now that this is desired behaviour, Blender does this too.

PR #4450, opened 2022-05-02
Svg doc autowrite with framenum

### LINKED ISSUE #4448: SVG Document node Parameter control
I want to create a 2D animation in Blender without passing by the internal render of Blender. So I am creating SVG files for each frame but sadly there is no parametric control for The 'name' and 'write' parameter. Is there any way to automatically save a SVG for each frame? Thank you

zeffii:
adds a switch that can be enabled if the user wants the current frame to be included in the filename

image

resolves #4448

### DIFF
--- a/nodes/svg/svg_document.py
+++ b/nodes/svg/svg_document.py
@@ -211,7 +211,7 @@ class SvSvgDocumentNode(bpy.types.Node, SverchCustomTreeNode):
         self.inputs.new('SvFilePathSocket', 'Folder Path')
         self.inputs.new('SvFilePathSocket', 'Template Path')
         self.inputs.new('SvSvgSocket', 'SVG Objects')
-        self.inputs.new('SvStringsSocket', "File Name").prop_name = "file_name"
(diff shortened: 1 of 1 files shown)


Durman:
It seems that converting the Name property into a socket would be more general solution of the problem.

zeffii:
i don't think so

zeffii:
@vicdoval  opinions?

zeffii:
> It seems that converting the Name property into a socket would be more general solution of the problem.

there's a very good case to be made for an input socket for filename, i agree - but in this instance my focus is the automated suffixing of frame number, here i like the simplest intervention (a boolean/switch)

vicdoval:
I think that the numbered suffix property is a good idea, independently of making a socket for the name

zeffii:
hmmm or `SvTextSocket` ?

zeffii:
also if going with the socket, it seems like a more appropriate place to include the `frame number suffix` switch, and not beside the Write button.

zeffii:
i'm not finished. but this is all i can do for now

PR #4456, opened 2022-05-05
Fp config from socket

zeffii:
This makes it possible to configure a FilePath Node as a function of the kind of FilePath socket with which it is linked.
- this introduces a new property (StringProperty) of the `SvFilePathSocket` called `filepath_node_mode`, upstream the FilePath Node will look at the `filepath_node_mode` and use the mode to configure the `SvFilePathFinder` Operator.

`SvFilePathFinder` Operator can behave according to the kind of files it's used for.

i'm not sure this is the best way to do it, because this while this makes some functionality easier, it may interfere with flexibility in its current state. Another path is to add features to `draw_buttons` function of the SfFilePathFinder Operator, as a fail safe incase the automatic mode resulted in something undesirable.

### DIFF
--- a/core/sockets.py
+++ b/core/sockets.py
@@ -1060,6 +1060,9 @@ class SvFilePathSocket(NodeSocket, SvSocketCommon):
 
     color = (0.9, 0.9, 0.3, 1.0)
     quick_link_to_node = 'SvFilePathNode'
+    filepath_node_mode: StringProperty(
+        name="filepath node mode",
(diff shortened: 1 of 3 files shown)


zeffii:
not all nodes that use FilePath socket will make use of this feature, as most are currently self appending an extension, and only using the FilePath node to provide a location and `file_name` (without .ext). I'm reluctant to make changes to these nodes, even to the point where I might only commit the changes to the single FCStd write node.  Or just close the PR.

PR #4468, opened 2022-05-12
Listjoin fix1

### LINKED ISSUE #4467: List Join node
## Problem statement

The node call `links` attribute of sockets what can be quite expensive.
In the example it takes 85% percent of the time.
image

zeffii:
resolves #4467 

@Durman 
does the current update system still require stuff like 

```python
        if not self.outputs['data'].is_linked:
            return

```
?

### DIFF
--- a/nodes/list_main/join.py
+++ b/nodes/list_main/join.py
@@ -185,8 +185,9 @@ class ListJoinNode(bpy.types.Node, SverchCustomTreeNode):
         for socket in self.inputs:
             if socket.is_linked:
                 slots.append(socket.sv_get(deepcopy=False))
-        if len(slots) == 0:
+        if not slots:
             return
+
         if self.match_and_join:
             match_func = list_match_func[self.list_match]
             if self.mix_check:


zeffii:
i vaguely remember we used `.links` in early sverchok development when `is_linked` would return True while it should not have. With the update system rewritten and now relying on (from what i can tell) a layer of abstraction, `.links` can be dropped.

it passes tests and i can't seem to make it fail, so @Durman i'll commit unless you have objections?

Durman:
Yes there is known issue with `is_linked` attribute - https://developer.blender.org/T82318
But it seems to me that in this node the attribute should work correctly. It can be only broken when connected socket is replacing itself with another type.

Durman:
By the way there should not be any problem with using `is_linked` attribute at all. The tree data is modified in one Python session and it is executed in another one (via timer). Between the two sessions Blender takes place and fixes all tree attributes including `is_linked` attributes of all sockets I guess. Except animation event which is calculated in the same session but during animation we don't expect any changes in a tree anyway.

PR #4471, opened 2022-05-13
is it overhead?

zeffii:
minor syntax modifications to the CubicSpline class lets us numba.njit decorate the expensive parts of this algorithm, the speedup is about 10-15 fold with high input volumes resulting in significant relative performance improvements.

### DIFF
--- a/utils/__init__.py
+++ b/utils/__init__.py
@@ -128,7 +128,7 @@ utils_modules = [
     "snlite_utils", "snlite_importhelper", "context_managers", "sv_node_utils", "sv_noise_utils",
     "profile", "logging", "testing", "sv_requests", "sv_shader_sources", "tree_structure",
     "avl_tree", "sv_nodeview_draw_helper", "sv_font_xml_parser", "modules.edge_utils", "modules.polygon_utils",
-    "wfc_algorithm", "handle_blender_data", "nodes_mixins.generating_objects", "decorators_warpspeed",
+    "wfc_algorithm", "handle_blender_data", "nodes_mixins.generating_objects", "decorators_compilation",
     "nodes_mixins.show_3d_properties", "modules_inspection", "sv_json_export", "sv_json_import",
     "meshes", "tree_walk", "mesh_functions", 'mesh.inset_faces', 'mesh.extrude_edges', "sv_json_struct",
     "nodeview_time_graph_drawing", "modules.shader_utils", "dummy_nodes",
--- a/utils/__init__.py
+++ b/utils/__init__.py
@@ -128,7 +128,7 @@ utils_modules = [
     "snlite_utils", "snlite_importhelper", "context_managers", "sv_node_utils", "sv_noise_utils",
     "profile", "logging", "testing", "sv_requests", "sv_shader_sources", "tree_structure",
     "avl_tree", "sv_nodeview_draw_helper", "sv_font_xml_parser", "modules.edge_utils", "modules.polygon_utils",
-    "wfc_algorithm", "handle_blender_data", "nodes_mixins.generating_objects",
+    "wfc_algorithm", "handle_blender_data", "nodes_mixins.generating_objects", "decorators_warpspeed",
     "nodes_mixins.show_3d_properties", "modules_inspection", "sv_json_export", "sv_json_import",
     "meshes", "tree_walk", "mesh_functions", 'mesh.inset_faces', 'mesh.extrude_edges', "sv_json_struct",
     "nodeview_time_graph_drawing", "modules.shader_utils", "dummy_nodes",
--- a/utils/decorators_compilation.py
+++ b/utils/decorators_compilation.py
@@ -11,7 +11,7 @@ from sverchok.dependencies import numba
 
 local_numba_storage = {}
 
-def gofaster(function_to_compile):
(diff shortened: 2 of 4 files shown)


Durman:
Probably input lists should be converted into NumPy arrays (np.as_array) before calling the function. When Numba executes a function it looks into types of input parameters and compiles a function which parameters much them.

zeffii:
it seems happy to accept numpy arrays as arguments (can be flattened easily, if its trivial for me its trivial for the compiler).

it looks more like `np.newaxis` is not supported. https://github.com/numba/numba/issues/2761

zeffii:
passing a non 1D array may very well also be a problem, but i wanted to avoid changing the signature of that function so soon.

zeffii:
oh wait.. the incoming knots and locs aren't `np.array`? . yikes. ok.

zeffii:
now it doesn't like this line
```PYTHON
            for i in range(1, n - 1):
                l[i] = 2 * (tknots[i + 1] - tknots[i - 1]) - h[i - 1] * u[i - 1]
                l[i, l[i] == 0] = 1e-8   # <----------
                u[i] = h[i] / l[i]
                z[i] = (q[i] - h[i - 1] * z[i - 1]) / l[i]
```
throwing
```
[ERROR] numba.core.dispatcher:409 : Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<built-in function setitem>) found for signature:

 >>> setitem(array(float64, 2d, C), Tuple(int64, array(bool, 1d, C)), float64)

There are 16 candidate implementations:
  - Of which 14 did not match due to:
  Overload of function 'setitem': File: <numerous>: Line N/A.
    With argument(s): '(array(float64, 2d, C), Tuple(int64, array(bool, 1d, C)), float64)':
   No match.
  - Of which 2 did not match due to:
  Overload in function 'SetItemBuffer.generic': File: numba\core\typing\arraydecl.py: Line 176.
    With argument(s): '(array(float64, 2d, C), Tuple(int64, array(bool, 1d, C)), float64)':
   Rejected as the implementation raised a specific error:
     NumbaNotImplementedError: only one advanced index supported
  raised from D:\b30\3.1\python\lib\site-packages\numba\core\typing\arraydecl.py:69

During: typing of setitem at C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\utils\geom.py (259)

File "C:\Users\zeffi\Desktop\scripts\addons_contrib\sverchok\utils\geom.py", line 259:
        def perform_stage(tknots, n, locs):
            <source elided>
                l[i] = 2 * (tknots[i + 1] - tknots[i - 1]) - h[i - 1] * u[i - 1]
                l[i, l[i] == 0] = 1e-8
                ^
```

zeffii:
one more part left and i think it will oompile fully as `njit`

zeffii:
@Durman  can you try that? it's about 10 times faster here.

Durman:
It's better now but does not work in cycles mode:
`sverchok.utils.geom:214 : operands could not be broadcast together with shapes (4,3) (10000,3)`

Also I thought if we are going to use more fast implementations as alternative to pure Python we can mark it in the node label.
For example with Python implementation the node can have name `[Py] Vector Interpolation` and using Numba implementation `[C] Vector Interpolation`. Inthis particular case `CubicSpline` class can has `implementation` class attribute which returns `Py` or `C`.

The `Py` flag in a node label can suggest to user that there is faster alternative of the node.

zeffii:
@Durman yes, i think there will be a situation where the node will try to acquire the most optimal implementation, but failing that always fallback/degrade to original slower implementation.

I did not try `cyclic`, should be possible to make that work too.

zeffii:
this 
```python
import numpy as np

vertices = np.array([
    [0.0, 1.0, 2.0],
    [1.0, 2.0, 3.0],
    [2.0, 3.0, 4.0],
    [3.0, 4.0, 5.0],
    [4.0, 5.0, 6.0],
    [5.0, 6.0, 7.0],
    [6.0, 7.0, 8.0],
    ])

locs = np.array(vertices[-4:] + vertices + vertices[:4])

print(locs)
```
also throws that error, 

> ValueError: operands could not be broadcast together with shapes (4,3) (7,3) 

this is a good error , resolved by:
```python
locs = np.concatenate((vertices[-4:], vertices, vertices[:4]), axis=0)
```

zeffii:
there's a way to do Ahead-of-time compilation, i think in this scenario this will mean that the user only compiles the function once (might take a noticable amount of time), and it becomes available precompiled in the next .blend session . Looking forward to getting that to work.

zeffii:
https://numba.pydata.org/numba-doc/latest/user/pycc.html  cool i think this will be possible

zeffii:
mostly a note to self:

Ahead of time compilation requires us to pass a function signature, preferably in the form  `return_type(input_type, ...)`, but when return type is not provided it is inferred. [#numba-types](https://numba.pydata.org/numba-doc/latest/reference/types.html#numba-types)

When compiled the result is written to disk (location will possible to configure, like `sverchok.utils.modules.compiled_functions`, as a `pyd, so`

not exactly sure when to call compile, and when/how sverchok will seek compiled versions for startup.

PR #4488, opened 2022-05-22
Cubic numba proper compilation

zeffii:
solves the signature issue, does not solve auto compilation for later reuse.
- rewrites minor parts of the cubic function to match the subset of numpy that numba can reliably read
- adds decorator and faux decorator
- ~adds automatic signature generation (compilation step for diskstorage does not work yet!)~

turns out that you can generate a signature automatically ( why the docs aren't explicit about this is baffling)  by doing
```python
import numba

def make_signature_tuple(*params):
    return tuple([numba.typeof(p) for p in params])

param_1 = np.array([[0.0, 0.1, 2.0], [1.0, 0.3, 2.0], [2.0, 0.1, 2.3]])   
param_2 = 3
param_n = np.array([0, 2, 3, 4, 12, 14])
params = param_1, param_2, param_n

print(make_signature_tuple(*params)
>>> (array(float64, 2d, C), int64, array(int32, 1d, C))
```


zeffii:
with the caching argument set to `True` the function is definitely stored to disk (pickled..)

image

https://numba.readthedocs.io/en/stable/reference/envvars.html#numba-envvars-caching

zeffii:
@portnov   what do you think of this PR. any improvement ideas?

zeffii:
if we can get some consensus on this, i'm interested in applying numba to other areas too.

zeffii:
I think this is the least invasive approach. njit is attempted if numba is found, else the function just passes through the wrapper unchanged.

PR #4496, opened 2022-05-27
add ccw_angle function to geom_utils.py

zeffii:
add small util function ccw_angle

### DIFF
--- a/utils/modules/geom_utils.py
+++ b/utils/modules/geom_utils.py
@@ -8,6 +8,7 @@
 
 import math
 import random
+import numpy as np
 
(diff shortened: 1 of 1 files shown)


Durman:
As alternative =)
https://github.com/nortikin/sverchok/blob/81062ac28da6c4418994a20972afbc2edddfc411/utils/geom_2d/lin_alg.py#L89-L97

zeffii:
i was thinking about maybe `ccw_angle` function should take 3 points and a normal, taking care of the translation for the user..

PR #4503, opened 2022-06-01
add bpy console printing from snlite using cprint

zeffii:
adds `cprint` to convenience functions of snlite. like `print` 
- it will print whatever you pass into it, but rather than printing to stdout it will print directly to the first open blender-bpy-console that it finds. 
- to avoid reprinting identical content the function checks first if the last printed message is the same as the proposed new message, if they are identical the operation is skipped.

image


zeffii:
i'll add documentation as soon as i'm certain that this is the extent of this feature.

PR #4511, opened 2022-06-04
Stethoscope console

zeffii:
- [x] will add nicer looking mode  (for topology, arbitrary lists, even mixed list lengths)
     - [ ] possibly lame-lexer for arbitrary string content (something which doesn't expect syntax highlighting)
     The solution here might be to try/except a syntax-highlighter, and in the exception clause attempt a lame-lexer ( that handles explicit `[]/\[0-9]()|:-+=*&@#~;<>?`, rather than lexing something that must be compilable - just focus on certain characters. 
- [x] will increase speed of default text mode
- [x] hijack REPL to allow printing arbitrary data to, instead of stdout. with "console_print(.....)`
    ```python
    from sverchok.utils.console_print import console_print
    console_print(V1, allow_repeats=True, pretty=True, rounding=4, compact=True)
    ```


zeffii:
https://numpy.org/doc/stable/reference/generated/numpy.array2string.html

zeffii:
image

at threshold 50 items, the array2string can show an abbreviation. This is a pretty slow implementation at the moment.

zeffii:
image

zeffii:
image

zeffii:
with mixed polygon lengths, the array2string function does this..
image

zeffii:
this is now showing favourable speed increases vs the original blf / socket parsing.

zeffii:
seems to generalize reasonably well so far.. 
image

image

the overhead of 12ms can be reduced a lot a think, with a bit of a rewrite and numba.njit  , njit now does allow basic dicts (from what i can tell the character to uv lookup table would be considered a basic dict type)

zeffii:
argh!  numpy.meshgrid not supported by numba.

zeffii:
numba has no np.tile support.  ok :)

zeffii:
the typed Dict. https://numba.pydata.org/numba-doc/dev/reference/pysupported.html#id9

zeffii:
image

zeffii:
i see python `tokenizer` has changed again in recent pythons.. but that's a problem to fix in  `console_node`

zeffii:
if anyone wants to improve the chop_data function for more broad input, that would be great. but for now i think this will suffice for the most common input.

The docs will be updated shortly.

PR #4514, opened 2022-06-06
Fix looping of scene handler

### LINKED ISSUE #4510: flashes in the ui
Blender 3.3 alpha, Sverchok 1.1.0 nightly (Github)

The UI starts flashing when I add the random num gen to the Adaptive Polygons. Please see the video.

https://user-images.githubusercontent.com/8519469/171984147-c489974e-1d61-44d5-95f5-7f3025296a82.mp4

Not sure if this is a random gen problem, something about the scene updates causing this it seems.

https://user-images.githubusercontent.com/8519469/171985098-3a950d13-eaa3-40a8-810c-3c2c5612899f.mp4

Thanks

Durman:

Fixes #4510

## Preflight checklist

- [x] Code changes complete.

### DIFF
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -126,6 +126,15 @@ def sv_main_handler(scene):
 
     pre_running = False
 
+    # When the Play Animation is on this trigger is executed once. Such event
+    # should be suppressed because it repeats animation trigger. When Play
+    # animation is on and user changes something in the scene this trigger is
+    # only called if frame rate is equal to maximum.
+    if bpy.context.screen.is_animation_playing:
+        return
+    for ng in BlTrees().sv_main_trees:
+        ng.scene_update()
+
 
 @persistent
 def sv_clean(scene):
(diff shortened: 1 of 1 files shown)


core/handlers.py
```diff
@@ -126,6 +126,15 @@ def sv_main_handler(scene):
 
     pre_running = False
 
+    # When the Play Animation is on this trigger is executed once. Such event
+    # should be suppressed because it repeats animation trigger. When Play
+    # animation is on and user changes something in the scene this trigger is
+    # only called if frame rate is equal to maximum.
+    if bpy.context.screen.is_animation_playing:
```
zeffii, line 133: seems to work well

rendetto:
I was hoping that merging #4514 will fix my problem as well, but my tree does not stop updating no matter what until I turn off either the "Scene update" button or the "Live update" button.

I also noticed that another tree does not trigger an update loop, so I guess there are some specific nodes that triggers it.

The tree in question is fairly old, but I tested it (and many more) about two or three weeks ago and there were no problems. So the problem should be related to some of the last commits.

The question is if there is a specific node that triggers this which will be the next one that I didn't tested yet.

Also what if the behavior is related to usage of old nodes? For example an older version of "Get Objects Data" node.

What will be the right approach to isolate the problem on my side?

Thanks

Durman:
I don't think it has any connections with nodes oldness. It seems that one problem was fixed but others remain. I expressed my thoughts in the issue.

PR #4517, opened 2022-06-09
Add support of muting nodes functionality

Durman:

muting_support

Now it's possible to implement `sv_internal_links` property for nodes which returns iterable pairs of input and output sockets of the node. The pairs will be used by update system to define how to path data through the node. The implementation is needed only if standard Blender mapping does not fit to the node logic and it looks that in many nodes it does not.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.


zeffii:
i'll also add this muting to `snlite`, allowing user to configure what happens during mute. Looking forward to this PR.

Issue #4521, opened 2022-06-10
File name with non alphanumeric characters not handled correctly


macdroid53:
## Problem statement

When using File path and Read FCStd node, if the file name includes non alphanumeric characters (-, '(',')' ) file is not recognized.

If the file name includes a '-' or ')', the error (shown above the Read FCStd node):

```
2022-06-10 08:31:41,837 [INFO] sverchok.nodes.exchange.FCStd_read 334: FCStd read error
2022-06-10 08:31:41,837 [ERROR] sverchok.nodes.exchange.FCStd_read:337 : <class 'NameError'>
```

If the file name includes a '_', the error (shown in the terminal window): 

```  
File "/home/mac/.config/blender/2.93/scripts/addons/sverchok-master/nodes/exchange/FCStd_read.py", line 337, in LoadSolid
    F.closeDocument(Fname)
NameError: Unknown document 'RaspPi4random_4'

location: <unknown location>:-1
```

## Steps to reproduce

1. Create a FreeCAD file with the described characters
2. Attempt to specify the file in the File path node connected to a Read FCStd node.

## Sverchok version

v1.1.0-alpha.2

zeffii:
okay, let's have a look

zeffii:
that's weird, because the only file i've ever opened is called `Simple_assy_001.FCStd`,  for testing

just to be clear here, are you saying you can use these nodes without issues read/write.etc all day long, just as long as the filenames do not contain `underscores, hyphens and parenthesis` characters?  

is this something which has never been possible for you or _don't know never tried_

zeffii:
kind of peculiar that the line numbers in the error message (334, 337) are beyond the length of the current version of that node. 243 lines) . Hinting that the error might be in the FreeCad module.

macdroid53:
I'm on Debian and Manjaro.

If I remove the offending character it works fine.

Not sure it matters, but I'm am using this version of FCStd_read.py:


[FCStd_read.zip](https://github.com/nortikin/sverchok/files/8879280/FCStd_read.zip)

zeffii:
this line worries me
```
                            labels.append( (obj.Label, obj.Label, obj.Label) )
```
this appends entries to a bpy enum i think, and they must be not include any characters other than `[a-zA-Z_0-9]`

zeffii:
i;m going  to guess object labels are permitted to have spaces in FreeCAD.
```python
from sverchok.data_structure import no_spaces

...
[...]

PR #4524, opened 2022-06-11
Fcstd read debug

zeffii:
adresses discrepency between `display name` and `doc.name`, they are not interchangable.


zeffii:
fixes https://github.com/nortikin/sverchok/issues/4521

PR #4529, opened 2022-06-12
a few modifications and rst

zeffii:
adds a modification to read fcstd node, which includes experimental filtering features, and simplified object acquisition.


zeffii:
specially for @macdroid53

untested here.

PR #4531, opened 2022-06-13
Snlite doubleupdate fix

zeffii:
removes redundant update


zeffii:
hmm i think this had more to do with Sverchok being in debug mode. while i nudged the frame number using left/right.
a non-issue.

PR #4534, opened 2022-06-15
Fcstd fcad dicts

zeffii:
6+ speed increase on `raspbi initial .FCStd` , using a dict to store `(v.x, v.y, v.z): idx` pairs.  I suspect a dedicated custom dict with rounding (6 digits) before insertion/recall may be even faster.


utils/solid.py
```diff
@@ -548,3 +548,79 @@ def find_vertex(v):
...
+
+    # write FreeCAD faces as polygons when possible
+    for face in shape.Faces:
+        
+        if (len(face.Wires) > 1) or (not isinstance(face.Surface, Part.Plane)) or hascurves(face) or tessellate:
+            # face has holes or is curved, so we need to triangulate it
+            rawdata = face.tessellate(quality)
+            
+            for v in rawdata[0]:
+                if not (v1 := (v.x, v.y, v.z)) in vdict:
```
Durman, line 593:
```py
if (v1 := (v.x, v.y, v.z)) not in vdict:
```
More readable

zeffii: yeah, agreed. bad habit i picked up..

PR #4540, opened 2022-06-20
Snlite search script

zeffii:
i find it tedious to go the the N-panel and navigate through the files, a search is much more convenient. Might also add a templates dir.


zeffii:
i'm thinking of adding a template dir and "rescan nodescripts" directory feature.

Issue #4607, opened 2022-08-02
Add different sized lists matching to List Mask Join (In)


ArpegorPSGH:
Like it is done with many nodes, an option to match lists of different sizes (Shortest, Cycle, Repeat Last) should be added to the List Mask Join (In) node.

PR #4616, opened 2022-08-20
Adding inputs lengths matching and solving a few minor issues/limitations of Mask Join node

ArpegorPSGH:

This solves issue #4607. I also solved what looked like an issue and slightly expanded the node functionality.

## Solution description

I added list length matching at each level in the inputs, while not changing the node treatment on the lists.
In the get_level function, it seemed to me that the case >2 should instead be >= 2, otherwise the treatment is not consistent when you increase the number of levels.
Finally, I enabled the use of the choice option when there is 0 nesting in the data input lists.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


nodes/list_masks/mask_join.py
```diff
@@ -87,28 +103,29 @@ def apply_mask(self, mask, data_t, data_f):
         return out
 
     def get_level(self, mask, data_t, data_f, level):
+        if self.choice:
+            apply_mask = self.apply_choice_mask
+        else:
+            apply_mask = self.apply_mask
```
Durman, line 109: It's not very important but I don't agree that this declaration should be moved to the start of the function. I like the idea that variables should declared as close as possible to place of their usage.
https://softwareengineering.stackexchange.com/questions/113262/why-declare-variables-close-to-where-they-are-used

nodes/list_masks/mask_join.py
```diff
@@ -87,28 +103,29 @@ def apply_mask(self, mask, data_t, data_f):
...
+            
         if level == 1:
             out = []
-            param = (mask, data_t, data_f)
+            param = list_match_func[self.list_match_global]([mask, data_t, data_f])
             if not all((isinstance(p, (list, tuple)) for p in param)):
                 print("Fail")
```
Durman, line 115: Feel free to remove all prints inside process method of a node.
http://nortikin.github.io/sverchok/docs/contributing/node_api.html#printing-logging

nodes/list_masks/mask_join.py
```diff
@@ -87,28 +103,29 @@ def apply_mask(self, mask, data_t, data_f):
...
-            else:
-                apply_mask = self.apply_mask
 
-            for i in range(max_index):
-                out.append(apply_mask(mask[i], data_t[i], data_f[i]))
+            for m, t, f in zip(*param):
+                out.append(apply_mask(m, t, f))
             return out
-        elif level > 2:
+        elif level >= 2:
```
Durman, line 121: It seems also requires next version of the node. I doubt that else statement did the same for level 2.

nodes/list_masks/mask_join.py
```diff
@@ -87,28 +103,29 @@ def apply_mask(self, mask, data_t, data_f):
...
             out = []
-            for t, f in zip(data_t, data_f):
+            param = list_match_func[self.list_match_global]([data_t, data_f])
+            for t, f in zip(*param):
                 out.append(self.get_level(mask, t, f, level - 1))
             return out
         else:
-            return self.apply_mask(mask[0], data_t, data_f)
+            return apply_mask(mask[0], data_t, data_f)
```
Durman, line 128: Now it's not clear which particular cases this statement should handle.

nodes/list_masks/mask_join.py
```diff
@@ -63,7 +78,8 @@ def process(self):
 
     def apply_choice_mask(self, mask, data_t, data_f):
         out = []
-        for m, t, f in zip(cycle(mask), data_t, data_f):
+        param = list_match_func[self.list_match_global]([mask, data_t, data_f])
```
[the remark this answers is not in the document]
ArpegorPSGH: If we put cycle by default in match mode, could it work without creating a new node?

nodes/list_masks/mask_join.py
ArpegorPSGH: The `apply_mask` is used in both `if level == 1` and `else`, so, to declare it only once, it cannot be declared elsewhere.

nodes/list_masks/mask_join.py
ArpegorPSGH: It makes sure that true and false data have the exact same structure at each level, otherwise it is unclear how to switch between the two.

nodes/list_masks/mask_join.py
ArpegorPSGH: Well, yeah, although it is broken, if people created node tree based on broken node, it won't behave the same, they would need to change the level to apply mask. If I create a new node, the old one will be moved to deprecated, is that it?

nodes/list_masks/mask_join.py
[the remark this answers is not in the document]
Durman: Probably, with a comment that the default value should not be changed because it might break old layouts.

nodes/list_masks/mask_join.py
Durman: A node becomes and old one when its code gets into the old nodes folder.
http://nortikin.github.io/sverchok/docs/contributing/node_api.html#create-new-node-version
No need to create new node, just change version of this one and copy old code in the old nodes folder.

nodes/list_masks/mask_join.py
Durman: I mean

```
level = 1, 2, 3, ..., n

if level == 1:
elif level >= 2:
else:
```
What value should level have to reach the else statement?

ArpegorPSGH: 0

nodes/list_masks/mask_join.py
```diff
@@ -18,14 +18,14 @@
...
 from sverchok.node_tree import SverchCustomTreeNode
-from sverchok.data_structure import changable_sockets, updateNode
+from sverchok.data_structure import changable_sockets, updateNode, list_match_modes, list_match_func
 
 
-class SvMaskJoinNode(bpy.types.Node, SverchCustomTreeNode):
+class SvMaskJoinNodeMK2(bpy.types.Node, SverchCustomTreeNode):
     '''Mask Join'''
-    bl_idname = 'SvMaskJoinNode'
+    bl_idname = 'SvMaskJoinNodeMK2'
```
Durman, line 28: The new name also should be fixed in the index.md file

nodes/list_masks/mask_join.py
Durman:
```py
level: IntProperty(name="Level", default=1, min=1, update=updateNode)
```

Currently level can't be less than one according to the property declaration and I don't think that 0 level has any sense. Maybe the function should raise an error about unexpected value of given level?

ArpegorPSGH: No, no, in the `process `function, when calling `get_levels`, it is not `level `property which is passed, but `level - 1`, so everything's fine.

nodes/list_masks/mask_join.py
ArpegorPSGH: That was not said in the link you gave me.

Durman: Should be added then.

nodes/list_masks/mask_join.py
```diff
@@ -87,28 +103,28 @@ def apply_mask(self, mask, data_t, data_f):
...
+            param = list_match_func[self.list_match_global]([data_t, data_f])
+            for t, f in zip(*param):
                 out.append(self.get_level(mask, t, f, level - 1))
             return out
         else:
-            return self.apply_mask(mask[0], data_t, data_f)
+            return apply_mask(mask[0], data_t, data_f)
 
 
 def register():
```
Durman, line 130: The name of the class should be updated here too.

Durman:
I tested the node and it works much better then old version. For example old version did not take into account choice option in level 1 mode and it seems did not work properly with levels more than 2.

You can fix the class name in registration function, resolve the conflict with master and merge the PR.

ArpegorPSGH:
Yes, I did say in PR description that I enabled the use of choice when there is 0 nesting (level 1).
Same thing with the problem with levels above 2 ; before, it just progressively unwraped the list until reaching level 2, then just went to else case and used only default mask on the first element in list. However, I believe it was a mistake made by the author, as there was absolutely no logic in `>2`, but `>=2` was perfectly logical.

ArpegorPSGH:
It pretty much says (I guess?) the bl_idname of the master branch List Mask Join (In) node and my branch's version of the node do not match, but that's just normal, so how do I resolve the conflict?

Durman:
The conflict is because of this [commit](https://github.com/nortikin/sverchok/commit/1804124cdd7be8f65f81f3bd49e8b8502ae174d5#diff-3207d3976343bcc0eced7160e1b3e2fb315a957b2cf3cad0c3c0960c896b7eab)

image

So you can take all from your current node module, fix label of the node according to the commit above, and commit. This should be done via resolving conflicts interface. I prefer to use PyCharm interface for that but it can be done even in web browser on Github.

ArpegorPSGH:
I do not see a commit button when I access to conflict via 'Resolve conflicts' button. Is it okay if I just change my code and commit it like I did to solve the previosu problems?

Durman:
If you want to resolve conflict locally you have to merge master branch from Github into your local branch for the node.
In Github interface the button is here:
image

ArpegorPSGH:
Okay, done.
I did not understand that I had to remove completely the version I did not want along with all the indications added by GitHub, so that is why I couldn't mark it as resolved.

Durman:
You forgot to add `py` extension to `mask_join` file.
Also we keep the examples without old nodes. After fixing the extension the test will show which examples should be updated.

ArpegorPSGH:
What do you mean by examples?

Durman:
Didn't you know we have examples of different layouts?)
image

You can find the files in the `json_examples` folder. They just should be replaced with new ones.

Durman:
I think the best way is to import a layout, replace old nodes there, export the layout with the same name and replace the the file in the json_examples folder with new one.

ArpegorPSGH:
So, I should remove the current Sverchok version I have, and install the version of this branch, then do as you say, is that it? By the way, how do I know which examples I need to update (which use this node)?

ArpegorPSGH:
I am quite new to Github, so I am wondering : as commits are added continuously to the master branch, isn't there a risk that an incompatibility arises between the time I submitted the pull request and the moment it is merged? As the code I tested won't be exactly the same upon merging. Shouldn't committing be frozen the time to update the pull request and check everything still works, and only after merge it?

Durman:
> So, I should remove the current Sverchok version I have, and install the version of this branch, then do as you say, is that it?

Yes, you can do that in this way. Though more convenient way is to put your development folder somewhere on your disk and set its path in the Blender settings. In this case Blender will be able to use it and you does not have to reinstall the add-on every time.

> By the way, how do I know which examples I need to update (which use this node)?

You can go into the test logs and read error tests. Or you can have a look at my screenshot several messages before.

>  isn't there a risk that an incompatibility arises between the time I submitted the pull request and the moment it is merged?

You actually encountered with such case already. This is called a conflict when the same file is edited in several branches. If you are trying to submit changes in a file which was already changed you have to merge both changes by special resolving conflict commit. The final view of the file is completely up to a committer. Technically nothing prevents from giving the file absolutely new content irrelevant to changes added in both branches.

ArpegorPSGH:
> Though more convenient way is to put your development folder somewhere on your disk and set its path in the Blender settings.

I'm not sure how to set the path to a development directory on Blender. Only way I see it, Blender will only look to that folder for addons, so I won't be able to use my other addons. Would it work as well if I deleted the content of Sverchok directory, and then put the branch I want to work on into it?

> You actually encountered with such case already. This is called a conflict when the same file is edited in several branches. If you are trying to submit changes in a file which was already changed you have to merge both changes by special resolving conflict commit. The final view of the file is completely up to a committer. Technically nothing prevents from giving the file absolutely new content irrelevant to changes added in both branches.

I understand that in case of conflict, you can choose which version to keep or even create a third one to make both work. I believe you may even need to modify more than just the lines creating the conflict, so you'll have to modify the branch, check it still works, and commit again. But all that is fine, as long as the conflict is detected. However, if there is no conflict, but the code of another file your modifications depend on was modified, how to detect such a thing? I think the only way may be to test the branch right after updating it and before merging it, making sure no new commit is added in-between (blocking committing, basically).

Durman:
> I'm not sure how to set the path to a development directory on Blender.

image

The add-ons from the path and from the default folder will be loaded together. Also you can install multiple Sverchok add-ons of different versions. Just make sure before enabling another version of the add-on that previous one is disabled and Blender is reloaded because only one version of the add-on can be enabled.

> However, if there is no conflict, but the code of another file your modifications depend on was modified, how to detect such a thing?

This can be detected either through automated or manual tests. If something was added into master, what influences onto your branch, the master branch should be merged into current to resolve this. And after that extra fixes can be made.

ArpegorPSGH:
Yes, I understand, this amount to updating the pull request branch with master, then fetching it from my local repo and testing it manually. What I am concerned about, though, is that between my update and test, and the merge, more changes are added to the master, which would force me to update and test again, and so on, so that may be going on for a long time.

Durman:
Of cause such things are possible but I can assure you that in Sverchok it's not a problem. Code of nodes is quite independent from each other. Changes in one node usually does not effect others. Even if something is changed in the core it does not effect nodes too much because we have to much nodes and Sverchok extensions to fix them all.

ArpegorPSGH:
Okay, so if I undestand well, changes on the core are transparent to the nodes, and there is only a need to test again if a change is applied to a utils file used by the node (so only check utils commits). Of course there is also the case when multiple people work on the same node, but we just need to coordinate via a common issue thread.

Durman:
By the way when you are ready I would suggest to squash commits and merge into master to make the git history cleaner.

ArpegorPSGH:
I can squash, but I cannot merge as I do not have the rights. Also, I cannot resolve the new conflict, why?

ArpegorPSGH:
In fact, I cannot squash, I have the error message that a merge commit cannot exist, but there are a whole lot in my branch's history. So, how could I do what you ask of me?

ArpegorPSGH:
By the way, I see that the JSON files I exported are only one line long, which is awful for reading, and also maintaining. Isn't there a problem, as the other files are well vented on multiple lines? I am planning to develop a project using node trees stored as JSON, and maitainability will be an issue for me as well.

Durman:
The view of json files is determined by the import setting. I don't thing it is necessary to be able to read json example files. If you need to change something in an example you simply create new json file that's all.

I prefer to save example files in zip format. In this case they take less memory and this is the reason of the conflict I guess. You can't resolve the conflict via Github interface because it's sort of complex one. In your branch CNC_PANNO file has json extension, in the master it has zip one. To resolve this you have to merge `mask_join_add_global_match` branch first I guess because if you resolved the conflict via Github interface your local branch is outdated and should be synchronized with origin. After that you have to merge master to you branch also locally and push the changes as usually to Github.

I thought you already received the permission to merge but I see that you are not.

ArpegorPSGH:
> The view of json files is determined by the import setting. I don't thing it is necessary to be able to read json example files. If you need to change something in an example you simply create new json file that's all.

Err, well, being able to tell what changes were made to a node tree can be essential when they constitute a central part of your project. Moreover, the other files are well structured, so why only the new ones are one line? Is that because of an import setting I didn't set right? Where are these, by the way?
Saving it as zip takes less memory, but they aren't maintainable, right? Is it really worth the few KB saved?

> To resolve this you have to merge `mask_join_add_global_match` branch first I guess because if you resolved the conflict via Github interface your local branch is outdated and should be synchronized with origin.

Merge my branch with what? Don't you just mean to update it by fetching new commits from origin? If the problem is with zip file, I can just export node tree as zip, although it seems neater to me to replace zip by json.

> After that you have to merge master to you branch also locally and push the changes as usually to Github.

I'm a bit lost there, merge master with my branch, and then merge the result with master? Why not just merge my branch into master?

Also, you did not answer me about how to squash my commits ; how do I proceed about that?

Durman:
> Err, well, being able to tell what changes were made to a node tree can be essential when they constitute a central part of your project. Moreover, the other files are well structured, so why only the new ones are one line? Is that because of an import setting I didn't set right? Where are these, by the way?
> Saving it as zip takes less memory, but they aren't maintainable, right? Is it really worth the few KB saved?

Yes this is because of the settings. Import settings are in the same place as import settings of other tools of Blender (in file browser). It can save a few MB I guess with our amount of examples but probably it's not very important indeed. Probably the main reason is that JSONs get into Git as editable files when they have nothing with development.

> Merge my branch with what? Don't you just mean to update it by fetching new commits from origin? If the problem is with zip file, I can just export node tree as zip, although it seems neater to me to replace zip by json.

Yes, fetching should work because there should be no conflicts between local and origin of current branches, but master branch should be merged anywhere to resolve the conflict. Another way can be to remove all changes in the json file (they are not important by the way) and in this case Git wont find any conflicts. Probably copy the json file from master and commit it also can resolve the conflict but I'm not sure.

You can forget about squashing the commits (until you will get the permission to merge). I will squash and merge the branch myself when the conflict will be resolved.

ArpegorPSGH:
> Yes this is because of the settings. Import settings are in the same place as import settings of other tools of Blender (in file browser). 

Maybe this setting (what is its name?) should be set to multiline layout by default, as that is the format generally wanted for the JSON files.

> It can save a few MB I guess with our amount of examples but probably it's not very important indeed. Probably the main reason is that JSONs get into Git as editable files when they have nothing with development.

Well, they have to do with Sverchok, even if they are not meant to be edited, but being able to visualize the changes and review/undo modifications locally may be useful in some cases, that is why I think being able to read them is better.

> Yes, fetching should work because there should be no conflicts between local and origin of current branches, but master branch should be merged anywhere to resolve the conflict. Another way can be to remove all changes in the json file (they are not important by the way) and in this case Git wont find any conflicts. Probably copy the json file from master and commit it also can resolve the conflict but I'm not sure.

Didn't you say the problem was coming from my example file being a JSON while the master one is a zip? In that case none of these options will solve it ; the only way I see is for me to create a zip file instead of JSON, or for you to resolve the conflict by authorizing my JSON file to replace the master zip file.

Anyway, I will be updating all my JSON files so that they are more readable.

ArpegorPSGH:
It does not seem it is possible to import zipped JSON, so I left my examples as normal JSONs

ArpegorPSGH:
Okay, there is still a conflict, I do not know what to do anymore. Please tell me what to do next.

Durman:
Probably documentation can help - https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging#_basic_merge_conflicts

ArpegorPSGH:
I didn't see anything that could help; when I fetch the master branch locally, I have no conflicts, so I cannot do anything that is said there. Please, try to see on your end what has to be done, and tell me.

Durman:
Fetching is not enough, you have to merge master into your branch to resolve the conflict. If you wish I can try to merge this PR into temporary branch, resolve the conflict myself and merge into master.

ArpegorPSGH:
Thanks, but I think I just found how to do it. I think the whole conflic problem comes from the fact that the master branch file is a zip, and thus cannot be intepreted as code file. So, as soon as it gets updated, there is an unsolvable conflict since the only way is to replace the file. That's a pain. So, I will be solving the conflict by merging master into my branch locally, and then pushin to the PR, but putting a json file instead of the zip.

PR #4619, opened 2022-08-23
"NURBS Curve Nodes" node.

portnov:
This node calculates so-called *node* points (also known as Greville points or
average knot points) of the NURBS curve.
NURBS curve *nodes* (or Greville abscissaes) are defined as:
``t[i] = sum(u[i+j] for j from 1 to p) / p``, where ``u`` is curve's knotvector, and
``p`` is curve's degree.
NURBS curve *node points* (or Greville points) are points at the curve at
parameter values equal to node values.
The number of curve's nodes is equal to the number of curve's control points.
Node points of the curve are positioned on the curve near corresponding control
points. So in many NURBS algorithms, node points are used to define the shape
of NURBS curve, instead of control points.

Screenshot_20220823_221621

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/data_structure/nurbs.rst
+++ b/docs/data_structure/nurbs.rst
@@ -96,6 +96,16 @@ for reference.
   in which distance between neighbour knots is different in different places.
   For example, ``0 0 0.5 0.6 1.5 1.7 1.9 2 2 2`` is a non-uniform (and clamped)
   knot vector.
+* **Node values**, **Node points**. Also known as Greville abscissaes and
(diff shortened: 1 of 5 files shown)


rendetto:
That's awesome! Thank you @portnov! 
This is a simple test showing how we can recreate exactly the source curve interpolating the node points using the "Points" metric. It seems it is valid only for Bezier curves though. 
greville_to_bezier

PR #4622, opened 2022-08-25
Fix typos carthesian->cartesian 

luzpaz:
Fixes both source code and source comment


zeffii:
can we merge this already? seems good.

luzpaz:
Question here is about backward compatibility though, right?

zeffii:
so far i don't see anything in the PR that would break existing .blends / .jsons , in the enum lists ( in math module, the modes lists) only the first element is stored in the .blend

luzpaz:
alrighty then :)

portnov (approved):
LGTM

PR #4652, opened 2022-09-11
Curve Viewer Draw & Surface Viewer Draw nodes.

portnov:
Shorthand nodes for viewing curves and surfaces in 3D View.

In my experience, I tend to use 4 nodes to inspect the curve I generated: Evaluate Curve + Viewer Draw, and Deconstruct Curve + Viewer Draw. Similarly with surfaces.

These nodes allow to replace four nodes with one.

For NURBS, the nodes can show control polygon / control net; "Viewer Draw Curve" node can also show "node points" (Greville points) of the curve.
For non-NURBS, the nodes just do not try to show these NURBS-specific things.

The nodes also have "bake" capability.

Screenshot_20220911_211123
Screenshot_20220911_210840

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/docs/nodes/viz/viewer_draw_curve.rst
+++ b/docs/nodes/viz/viewer_draw_curve.rst
@@ -53,6 +53,14 @@ The parameters of the node are (in this order):
   display of curve's node points, for NURBS and NURBS-like curves. Nodes are
   not shown by default.
 
+Operators
+---------
+
+This node has the following buttons:
+
+* **BAKE**. Create mesh objects in Blender scene according to curves being displayed.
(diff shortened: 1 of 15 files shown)


portnov:
@zeffii @Durman Will it be okay if I replace existing logic of creating "Evaluate Curve / Surface" + "Viewer Draw" from context menu, with the new nodes?

Durman:
It would be logical. Also it seems that Ctrl + LMB could connect the new viewer to curve and surface sockets.

PR #4654, opened 2022-09-12
Add matching to List Mask Join (In)

### LINKED ISSUE #4607: Add different sized lists matching to List Mask Join (In)

Durman:

Continue of #4616
Closes #4607

## Preflight checklist

- [x] Code changes complete.


Durman:
@ArpegorPSGH My congratulations ))

ArpegorPSGH:
Thanks, only my other pull request on cut object by surface, now.

PR #4686, opened 2022-09-27
Fix: Correct node class mixin order to resolve poll function

HansGoudey:
Python mixin order goes from left to right here, so when looking for the node's poll function, it first finds a `poll` function from `bpy.types.Node`, then Sverchok's base node class poll, and always returns false for most nodes.

In https://developer.blender.org/T101259, this incorrect poll caused Sverchok nodes to be visible in the Geometry Nodes editor. While that would be fun, it doesn't work unfortunately. It is possible to fix that too, but the polls should be fixed anyway.

The fix is switching the order of mixins, so that the Sverchok class with the proper poll function always comes first.

An alternative solution would be to somehow avoid defining the poll returning "True" in `bpy.types.Node`, but I'm not quite sure how that would work, and it might have more unexpected consequences.


Durman:
Thanks and congratulations on first commit to Sverchok (probably not last). =)

HansGoudey:
Thanks!

Issue #4575, opened 2022-07-17
Add Catmull-Clark subdivision to Sverchok


GeneralPancakeMSTR:
## Problem statement

[Catmull-Clark Subdivision](https://en.wikipedia.org/wiki/Catmull%E2%80%93Clark_subdivision_surface)  (usually as [OpenSubdiv](https://github.com/PixarAnimationStudios/OpenSubdiv)) is an important tool standard to most (if not all) poly-based 3D modeling programs, and a core component of many users' modeling pipeline (including mine). 

However, in Sverchok, there is no way to apply Catmull-Clark subdivision directly to mesh data, which is a significant drawback of an otherwise very powerful and complete tool. 

## Solution 
A potential solution to this problem would be to somehow wrap the OpenSubdiv subdivision algorithm in Python, to make it compatible with Sverchok. 

I am pleased to say that I have achieved this, specifically using [ctypes](https://docs.python.org/3/library/ctypes.html) for wrapping the C++ code from OpenSubdiv. I have also [forked Sverchok](https://github.com/GeneralPancakeMSTR/sverchok/tree/sverchok_OpenSubdiv), written a node that implements my subdivision wrapper, and tested its functionality directly in Blender: 

![vectorized_opensubdiv_final](https://user-images.githubusercontent.com/79929629/179411519-e146d968-13cf-4962-a1b5-0cace348debb.gif)

## Implementation 
- The code I wrote for wrapping OpenSubdiv's subdivision algorithm is available at my [pyOpenSubdivision](https://github.com/GeneralPancakeMSTR/pyOpenSubdivision) repository, which includes instructions on how to build and test it. 

- My fork of Sverchok that includes the `OpenSubdiv` node is available [here](https://github.com/GeneralPancakeMSTR/sverchok/tree/sverchok_OpenSubdiv). The [modifications](https://github.com/GeneralPancakeMSTR/sverchok/commit/622fe83cee8fad3d301c119d2a9738d48152cafb) are as follows: 
    - Added `nodes/modifier_change/opensubdivide.py`: The `OpenSubdiv` node. 
    - Added `utils/modules/ctypes_pyOpenSubdiv.py`: Python code which communicates with the C++ code (using ctypes) that implements the OpenSubdiv Subdivision function and exposes it in a "friendly" way (i.e. as an easily useable function) to the `OpenSubdiv` node (`opensubdivide.py`).
[...]

PR #4691, opened 2022-10-03
Sverchok open subdiv 2

GeneralPancakeMSTR:

#4575 [Add Catmull-Clark subdivision to Sverchok](https://github.com/nortikin/sverchok/issues/4575)

- Renames "OpenSubdiv" node to "Catmull-Clark Subdivision"
- Improved stability, particularly level 0 subdivision on topology without quads is handled a lot better, along with subdivision <= 0 in general.  

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.


Durman:
Now you can't change `bl_idname` attribute it will break previous user layouts (present of undefined nodes) . I think fixing label is enough.

GeneralPancakeMSTR:
> Now you can't change `bl_idname` attribute it will break previous user layouts (present of undefined nodes) . I think fixing label is enough.

Good point, I will fix that.

GeneralPancakeMSTR:
I'm not sure the best way to do the node mute implementation. Currently I'm just mapping verts->verts and faces->faces, but this sort of breaks the Edge output if the user has that plugged in and mutes the node. I'm not sure what the best way to deal with that is, so if anyone has ideas, please let me know.

index.md
```diff
@@ -9,7 +9,7 @@
 > Failing to follow these points will break the node category parser.
 
 ## Generator
-    SvLineNodeMK4
+    SvLineNodeMK4SvOpenSubdivisionNode
```
Durman, line 12: Not fixed yet

Durman:
Probably there is no need to generate edges in your case.

index.md
GeneralPancakeMSTR: I did not even notice that. Thank you.

PR #4697, opened 2022-10-06
Remove alpha and beta categories. Mark nodes in development by frames.

Durman:

It's implementation of this proposal #4685
Now nodes in development can be marked by frames. To do this alpha or beta icon should be added to a node class.

```py
class Node:
    sv_icon = 'SV_ALPHA`  # or 'SV_BETA`
```

image

I moved some (seems to me as redundant) nodes from removed categories to old nodes. Some nodes marked as Alpha/Beta nodes. I does not have strong option about whether I've made changes correct so if someone has any suggestions they are welcome.

## Preflight checklist

- [x] Code changes complete.
- [x] Ready for merge.


portnov:
Maybe instead of changing icon it's better to have special class property for this? like
```
class MyNode:
    sv_alpha = True
```

Durman:
Using icons has extra advantages. It's clear from menu that a node is in Alpha/Beta stage.

image

PR #4699, opened 2022-10-06
Simplify adding new node docs

Durman:

It seems redundant to keep list of node documentation file names - https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-toctree

Now it's enough to add `node_name.rst` file to appropriate folder in the docs without adding the file name into a `index.rst` file.

## Preflight checklist

- [x] Code changes complete.
- [x] Ready for merge.


satabol:
Hello, @Durman 
IMHO the problem list of menu items is not equal to list in help:

image

Is it possible to align them automatically or programmatically?

PR #4703, opened 2022-10-11
Add node menu refactoring and some new features

### LINKED ISSUE #3410: Presenting nodes with dependencies in menu
As far as I understand the nodes which are not available according not installed Python modules are not presented in the `add node` menu. 

I think it makes more difficult to learn those nodes. For example if one install one of the dependencies you have to investigate all `add nodes` menu to find new nodes and most likely you will not spot them from first time among huge number of available nodes. 

It would be much easier if you first find a node read documentation and required dependencies and decide would it be useful install it.

### LINKED ISSUE #4700: Tooltips of Add node operator in Blender 3.3
## Problem statement

Blender 3.3 silently added `description` class method for `AddNodeOperator` - https://developer.blender.org/D14963 https://developer.blender.org/T101657

Now it overrides our approach of adding tooltips. It's a coincidence that we did not loos tooltips completely because their approach is similar, it also uses node doc strings.

Our current approach is a class factory of `AddNodeOperator` subclasses for each node item in the menu.
https://github.com/nortikin/sverchok/blob/b5a8ebcf59d7fb667754523d1e28c9671884f56a/menu.py#L256-L264

With the new feature we have opportunity to simplify our solution and replace the factory wit single Add operator. The question is about backward compatibility. I see 3 ways to go:

- Add `description` method to our current approach. It's simplest solution and backward compatible but it want make our approach simpler.
- Rewrite our current approach. It's more complicated, backward incompatible but it will simplify the menu code. And probably it's not a big deal to drop the feature in 3.2 3.1. 3.0 and 2.93 Blender versions.
- Add separate solution for Blender 3.3 and higer. It's most complicated but backward compatible.

Durman:

New features:

- [x] New yaml format for main add node menu config file
- [x] Generating categories from the config files
- [x] Showing dependent nodes in menus
  image

Fixes #4700
Closes #3410

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Ready for merge.


portnov:
So python-yaml is now a required dependency for sverchok?

Durman:
No need for dependencies. I made primitive parser for the file.

utils/yaml_parser.py
```diff
@@ -0,0 +1,113 @@
...
+    def is_list_value(self):
+        return self._line.strip().startswith('-')
+
+    @property
+    def is_dict_value(self):
+        line = self._line.expandtabs(1).strip()
+        return ': ' in line or ':' == line[-1]
+
+    @property
+    def indention_level(self):
```
portnov, line 75: indentation_level or just indent_level.

Durman: Probably the name is incorrect. The function predicts number of nesting to each other lists and dictionaries. For 0 indentation it's1, for 1 it's 3, for 2 it's 5 etc. This is not the property of yaml formats but of our one. This works correct because the format is list of dictionaries and dictionaries can have either one value or a list only.

portnov: I was just referring to the typo in the name :)

Durman:
@vicdoval @moult @mostaphaRoudsari @wassimj

After the PR the API of adding new categories to Sverchok's Add Node menu will be changed. Also it can lead to some problems in imports. It will be possible to add new categories in this way:

```py
import sverchok.ui.nodeview_space_menu as sm
sm.add_node_menu.append_from_config(config)

def register():
    sm.add_node_menu.register()
```

The format of config structure -

    [option, option, ..., menu item, menu item, ...]

Menu item can be one of next elements:

- `'---'` - String of a dotted line to define separator.
- `'SomeNode'` - String of node `bl_idname` to define Add Node operator.
- `{'Sub category name': [option, menu_item, ...]}` - Dictionary of
  a subcategory.
- `{'Operator name': [option, ...]}` - Dictionary of a custom opeartor
  to call. Options for operator call are not supported currently.
- `{'Menu name': [option, ...]}` - Custom menu to show.

Options have such format - `{'option_name': value}`. They can be added
to sub categories, operators and custom menus.

Categories can have next options:

- `{'icon_name': 'SOME_ICON'}` - Value can be a string of standard Blender
  icons or Sverchok icon.
- `{'extra_menu': 'menu_name'}`  - Values is a name of one of extra menus.
  Possible values:

  - "BasicDataPartialMenu"
  - "MeshPartialMenu"
  - "AdvancedObjectsPartialMenu"
  - "ConnectionPartialMenu"
  - "UiToolsPartialMenu"

Operators options:

 - `{"icon_name": "SOME_ICON"}` - Icon to show in the menu.
- `{"operator": "operator id name"}` - `bl_idname` of operator to call.

Custom menus options:

- `{"icon_name": "SOME_CION"}` - Icon to show in menu.
- `{"custom_menu": "menu id name"}` - `bl_idname` of menu to show.

The example of format can be found in the `sverchok/index.yaml` file (of current branch).

portnov:
@Durman IMHO these parameters
Screenshot_20221024_115750
should be in settings. In toolbar they just consume screen space.

wassimj:
Two questions:
1. Will the previous release (1.0) still be available for a while? I need to find time to understand and implement the needed changes to the Toplogic menu. Anyone willing to volunteer to do a PR on TopologicSverchok is much appreciated :)  http://github.com/wassimj/TopologicSverchok
2. Does this release solve the issue of re-running a parent node if a new connection is made to a new child node? This is slowing execution big time as everytime I need the output for a second or third node, it re-runs potentially a very slow node rather than re-sue the output already generated.

portnov:
Also, can we have spacers in toolbar in the same places where we have them in menu? They are used to split long menu in smaller logical groups, it is useful in toolbar too.

portnov:
3. Toolbar tooltips look like "Add SV node: description". IMHO 1) "Add SV node" phrase is not needed; 2) first line of tooltip should be the title of the node, because it is not visible anywhere in "icons only" mode.

Durman:
> should be in settings. In toolbar they just consume screen space.

I guess add-on settings should obtain properties which should be set once after add-on installation. It seemed to me that at least `showing only icons` property is useful to switch on/off depending on the filter output.

I will have a look at spaces and content of the tooltips.

@wassimj 
It's not a release. This PR has description in the first post. Last release (1.1) will work fine with all Sverchok's extensions.

Durman:
@portnov a74063a

wassimj:
I kindly need help making topologic compatible with sverchok 1.2. The current init file is at (https://github.com/wassimj/TopologicSverchok/blob/main/__init__.py). Would someone from the sverchok team be willing to suggest how to edit it with examples so it does not break the way it imports 3rd party libraries? Thank you.

PR #4727, opened 2022-10-28
More stable way to generate node documentation links

Durman:

Node modules categories and node documentation categories are not 100% consistent. So there was magical conditions to translate the modules categories to docs ones. I reconsidered the approach and now categories does not take part in search of node documentations. It's completely based on the module names and it means that all node modules should have uniq names (even if they are in differen folders).

Also, now all nodes implement new method:

```py
def get_doc_link(self, link_type='ONLINE') -> Optional[str]:
```
This method should be overridden by Sverchok's extensions if they want to use links to node documentations.

New way reviled two nodes which does not have documentation and were not marked as known problems:
- flip_surface.py
- ruled_surface.py

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.


Durman:
Since the documentation link is generated inside the node method it's possible to override it for some nodes. For example it can be useful for SNL node to implement a way to show documentation of a loaded script.

PR #4749, opened 2022-11-08
Fillet polyline - maximum radius factor  

vkter:
## maximum radius factor  
Added a maximum radius factor input.
Fixed error when radius is 0.
Fixed error when fillet Tangent


Durman:
I'm not sure I understand the purpose of that extra parameter. It seems redundant to me. The factor can be controlled by radius.

image

vkter:
It will assign a reasonable maximum radius range to each chamfer
image

Durman:
I see, probably it has sense to have the feature in this way. Did you test the clipping in cyclic mode?

vkter:
This is in cyclic mode
image

Durman:
It seems clipping does not work properly with this setup.

[untitled1.zip](https://github.com/nortikin/sverchok/files/10000237/untitled1.zip)

image

vkter:
The maximum radius limit depends on the length of the shortest adjacent edge. This simple method seems unable to cope with complex situations，It only works well on non concave polygons. Maybe I should find a better way to determine this value

vkter:
I changed the code last night, it's working properly now
image

vkter:
在连接模式和相邻圆角刚好接触时会引发错误，这似乎是曲线连接模块的问题，它无法连接这样的圆角，这并非我添加了bug
image

portnov:
Doesn't this require an Mk2 of the node?

Durman:
> Doesn't this require an Mk2 of the node?

I guess yes.

@vkter http://nortikin.github.io/sverchok/docs/contributing/node_api.html#upgrade-node

Alternative way is to add an option to make new (clamping) behavior switched off by default.

vkter:
It seems better to add a clamping option that is turned off by default. I will continue to modify the code

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "commentTranslate.source": "Bing"
```
Durman, line 2: I'm not sure we need this in Git. I guess `.vscode` should be added to the git ignore file. Everything else seems okay.

vkter: Yes, just ignore it，

Durman (approved):

Durman:
Congratulations on the first contribution =)

PR #4750, opened 2022-11-08
Update string_tools.py

KoreTeknology:
orthographic correction "brakes" > "breaks" in string_tools.py

### DIFF
--- a/nodes/text/string_tools.py
+++ b/nodes/text/string_tools.py
@@ -193,7 +193,7 @@ class SvStringsToolsNode(SverchCustomTreeNode, bpy.types.Node):
         name='Characters',
         description='Text to modify',
         default='', update=updateNode)
-    keep_brakes:BoolProperty(
+    keep_breaks:BoolProperty(
         name='Keep breaks',
         description='Text to modify',
         default=False, update=updateNode)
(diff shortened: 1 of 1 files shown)


portnov:
As far as I understand, this will break existing layouts :/

KoreTeknology:
> As far as I understand, this will break existing layouts :/

? but keep_breaks  is a unique identifier, right?
how this can breaks the layout ? i am curious here, is there any other use of this identifier in the addon ?

portnov:
You can conduct an experiment:
* Roll back to sverchok master
* create some layout using "string tools" node, and, for example, check that "keep_brakes" flag. Save the blend file.
* Switch back to your version of code
* Load saved blend file again. You will see that "keep_breaks" flag is not checked. That is because in blend file the flag was saved under it's old name, and Blender does not have idea that newly created flag should inherit the value of old property. Accordingly, the node works differently. Consider the level of confusions of users, whose layouts suddenly start to work differently :)

To avoid such sudden changes, we have mechanism of "old nodes" and new versions of nodes, see http://nortikin.github.io/sverchok/docs/contributing/node_api.html#upgrade-node
I'm just not sure if it worth to create a new version of node for such a small change.

KoreTeknology:
i see... understood now... yes, may it not worth the change right now, until a major update of the addon, thnks for your explaination... i am working on extending the text nodes category. 
by the way, is ther any dev plateform other than on github where you are discussing changes and updates ? i would like to be in touch with the core team. thx

portnov:
Apart of github issues and github discussions, we use discord https://discord.gg/pjHHhjJz8Z .

Durman:
I will close it since it seems there is no easy way to fix this.

PR #4765, opened 2022-11-17
Update nurbs_birail.rst

rendetto:
The current master version of "NURBS Sweep" node accepts only ascending order of numbers fed into V1 and V2 (otherwise it throws an error "...Cannot solve..."). The old version was accepting non-consecutive numbers though. Lost some time figuring out what was causing the error, so I propose adding a short description in the docs.


portnov:
Thanks!
And congratulations on first contribution! :)

rendetto:
Thanks! I had to overcome the initial panic, not to mess something up :)

PR #4770, opened 2022-11-22
Fix various typos (trivial, documentation, and source)

luzpaz:


nodes/modifier_make/contour2D.py
```diff
@@ -453,8 +453,8 @@ class SvContourNode(ModifierLiteNode, SverchCustomTreeNode, bpy.types.Node):
         min=-1.0, default=1.0e-5,
         step=0.02, update=updateNode)
 
-    intersecction_handle: EnumProperty(
```
Durman, line 456: This name can be changed

PR #4775, opened 2022-11-28
Embed mesh B on mesh A into mesh A, new node

vkter:
Embed mesh B on mesh A into mesh A, new node


vkter:
https://user-images.githubusercontent.com/74725748/204120874-005aa169-166a-466d-bd07-aa5e8a35ed77.png
https://user-images.githubusercontent.com/74725748/204191959-202a40ea-e7ab-4994-875f-c7160a4fdf6a.png

portnov:
Huh :)
I thought about something similar a while ago. But I did not have good applications for this idea at time.
See https://github.com/portnov/sverchok-extra/blob/master/nodes/spatial/delaunay_mesh.py (node in Sverchok-Extra) and https://github.com/nortikin/sverchok/blob/master/utils/mesh_spatial.py (implementation in Sverchok API). Probably you will find that API useful for your purpose.

vkter:
image
After about a week for me, this feature has been implemented, and indeed in the process, there have been many problems, and it will try to embed mesh B (including edges and faces) into A

vkter:
image

docs/nodes/CAD/embed mesh.rst
```diff
@@ -0,0 +1,49 @@
+Delaunay 2D Cdt
+===============
```
Durman, line 2: This name should be fixed I guess ))

vkter: Sorry, I forgot to change

vkter: now,is ok

Durman:
I've tested the node and it seems there is a case when it does not work properly. If the approach has limitations it's better to describe them in the documentation.

[untitled1.zip](https://github.com/nortikin/sverchok/files/10109773/untitled1.zip)


image

vkter:
> I've tested the node and it seems there is a case when it does not work properly. If the approach has limitations it's better to describe them in the documentation.
> 
> [untitled1.zip](https://github.com/nortikin/sverchok/files/10109773/untitled1.zip)
> 
> image

... ,, you do not specify index, and the node needs to input to determine which face each vertex is on

vkter:
, you do not specify an index, and the node needs to input to determine which face each vertex is on
image

vkter:
If the test is okay, I hope to merge it as soon as possible

Durman:
I've got a stable crash with the file. Probably it's worth rising explicit error when indexes are not supplied or indexes are out of bounds?

https://user-images.githubusercontent.com/28003269/204711866-ed6a3ee1-1334-47a8-8595-5e0c676d429e.mp4

portnov:
@vkter 
> , you do not specify an index, and the node needs to input to determine which face each vertex is on

If one of inputs is mandatory, the error which the node raises should state clear enough that it is necessary to connect it. Not a mysterious "BMVert used multiple times" :)

Durman:
I have another example which seems falls into infinite loop.

[untitled1.zip](https://github.com/nortikin/sverchok/files/10122196/untitled1.zip)

https://user-images.githubusercontent.com/28003269/204773697-dc87d305-f4a2-4f49-b2cb-c81343df676f.mp4

vkter:
has been fixed

vkter:
image
has been fixed

Durman:
It seems I'm unable to crash with this node anymore =)
I will mark the node as Alpha so you can fix it without keeping backward compatibility. Before the next release we have to remove the alpha state.

vkter:
Ok, if there are still bugs, I'll keep fixing it

docs/nodes/CAD/embed mesh.rst
```diff
@@ -0,0 +1,49 @@
...
+Inputs
+------
+
+- **VertsA** - vertices of mesh A
+- **EdgesA** - edges of mesh A
+- **FacesA** - faces of mesh A
+- **VertsB** - vertices of mesh B
+- **EdgesB** - edges of mesh B
+- **FacesB** - faces of mesh B
+- **Index**  - Specifies on which face of mesh A each point of mesh B is on
```
nortikin, line 20: Please, try to use trivial **Donor** and **Recipient** naming as for similar nodes presents

nortikin (commented):
And some behaviour with masking. maybe to output faces masked? Is it real?

docs/nodes/CAD/embed mesh.rst
vkter: Which name do you think is better? This is the node of interest
image

vkter:
> And some behaviour with masking. maybe to output faces masked? Is it real?

I thought about the possibility of implementation, which is essentially implementing circular region selection, and I intend to implement this functionality in future bmesh nodes, so I will skip it for now

docs/nodes/CAD/embed mesh.rst
nortikin: not sure

Issue #4799, opened 2022-12-07
Some problems  "cross section" at the zero point.


Gaalets:
Some problems at the zero point.  The "cross section" does not cut in the XY plane at Z=0 if it hits the edges. In any other Z - normal slice. 1) Icosphere - a slice at zero - torn, full section not work. 2) Torus - there is a ring inside, there is no external cut.  Any other slices are normal.
![image](https://user-images.githubusercontent.com/120024667/206217829-003d6bcc-1c2e-467d-b156-0557700cde0e.png)
![image](https://user-images.githubusercontent.com/120024667/206218456-d6fad077-ae13-4a26-a6cd-f4a9be126f5c.png)
![image](https://user-images.githubusercontent.com/120024667/206218641-fe88121d-94c6-4e99-9163-322298cbb5f3.png)

Durman:
Use bisect node instead. It's more efficient and it seems does not have the issue.

![image](https://user-images.githubusercontent.com/28003269/206360734-2ad17f2f-ae1d-443a-b5cf-967725378d4a.png)

PR #4803, opened 2022-12-08
CAD->bisect node. Fix fill holes on complex objects (ex. torus).

satabol:

ex. #4799 if replay with Bisect then result is unexpected too:

image

But Blender do bisect well:

image

What the problem? The part of problem is in operator **bmesh.ops.bisect_plane**. The Bisect Operator does not fill holes! (Unexpected!).
See:

https://github.com/blender/blender/blob/3780a402651aeac0caa86767f823a8517a6f8bae/source/blender/editors/mesh/editmesh_bisect.c#L322-L333

image

But how Blender do fill? It do fill later with **triangle_fill** operator (https://docs.blender.org/api/current/bmesh.ops.html#bmesh.ops.triangle_fill)! 

https://github.com/blender/blender/blob/3780a402651aeac0caa86767f823a8517a6f8bae/source/blender/editors/mesh/editmesh_bisect.c#L346-L355

image

But how Sverchok do fill on bisect [see line 49]?

https://github.com/nortikin/sverchok/blob/9ff8f36eab43ccb926840123dff58469c50bff6b/nodes/modifier_make/bisect.py#L44-L49

I think this is a problem. So I try replace **bmesh.ops.edgeloop_fill** to **bmesh.ops.triangle_fill**. And I got next result that look good:

image

Bisect 001

With joined mesh:

Bisect 002

Suzanne tomography:

Bisect 003

Suzanna has non-manifold geometry so glitches are expected.

for test this pull request: https://gist.github.com/78671a6264b9255acbb3883779fefc73

### DIFF
--- a/nodes/modifier_make/bisect.py
+++ b/nodes/modifier_make/bisect.py
@@ -46,7 +46,7 @@ def bisect_bmesh(bm, pp, pno, outer, inner, fill):
         fres = bmesh.ops.edgenet_prepare(
             bm, edges=[e for e in res['geom_cut'] if isinstance(e, bmesh.types.BMEdge)]
         )
-        bmesh.ops.edgeloop_fill(bm, edges=fres['edges'])
+        bmesh.ops.triangle_fill(bm, edges=fres['edges'], use_beauty=True, use_dissolve=True, normal=pno)
 
     edges = []
     faces = []


satabol:
It is very old function. 

image

May I ask somebody for approve or test? (I did several test and did not see unexpected results but I believe it is very important)

satabol:
Can't stop play with that:

https://gist.github.com/f397c61f7275217008fb28b29bdfb260

Bisect 010 edited

image

image

portnov:
Huh. If you will be able to make the Waffel node do it's work for this thing, you'll probably be able to CNC it :)

rendetto:
I've made a test with one of my interactive scenes and it does not affect the performance.

Durman (approved):

Issue #4722, opened 2022-10-26
Bug at the upper domain limit for unclamped splines (geomdl and Native implementations) 


rendetto:
I'm reporting the strange behavior of the unclamped splines mentioned in https://github.com/nortikin/sverchok/issues/4469 
It is still present for Native and geomdl implementations with the most recent Sverchok master.

![LatestSV_PeriodicSplineBug](https://user-images.githubusercontent.com/66558924/198097371-b65dfb59-d187-462f-99ca-1a0185ea9380.JPG)

The setup is the same as used in https://github.com/nortikin/sverchok/issues/4469 (a constructed closed unclamped /periodic/ spline using wrapped control points - a convenient method for constructing arbitrary smoothly closed curves)
The problem is present when taking a segment from a spline and the end (TMax) of this segment is exactly at the upper domain limit (which is 1.0 in this particular example)

I found that reverting to an earlier version of Sverchok (32f8f36787fcf89809d6cff6165caa662bb1ea26) this problem is not present for both Native and geomdl. I run this early version of Sverchok under blender-2.90.0-windows64 and also aligned it to the corresponding Sverchok-extra for the time being.
I also used the latest geomdl (v5.3.1) due to #3521 and https://github.com/orbingol/NURBS-Python/issues/115

![oldSV_PeriodicSpline](https://user-images.githubusercontent.com/66558924/198098581-134c124f-9e03-4198-8903-35a3020bd0c3.JPG)

As with other similar issues I got, there is a workaround that could "solve" the issue by using a value of 0.99999 instead of 1.0

portnov:
Well, what I understood for now.
1. Splitting of a Nurbs curve at parmeter `t` is done by inserting a knot in that point `p` times (`p` being the degree of the curve). After such insertion, there will be a control point coinciding with `Curve(t)` point. After that, the process of splitting is obvious: first segment's control points are all old control points at left of `t` plus new control point, second segment's control points are new control point plus all old control points at right of `t`.
2. Knot insertion always cuts a corner, or several corners, from curve's control polygon. On illustration from The Nurbs Book, P are old control points, Q are new control points:

![Screenshot_20221104_132455](https://user-images.githubusercontent.com/284644/199926751-b71d71f4-9ddc-43ab-817a-dbe5a1739f50.png)

So to calculate how exactly corners must be cut, one has to have several control points at left of `t` and several control points at right of `t`, otherwise there will be no corners to be cut.
[...]

Issue #4772, opened 2022-11-25
The performance seem dropped twice since last year


rendetto:
I'm doing a pass on all setups I worked on the past few years, searching for bugs and broken things in the current Sverchok master. 
Unfortunately it seems that for the past year the performance of Sverchok dropped twice concerning this particular setup:

![Sweep_Node_PerformanceDrop](https://user-images.githubusercontent.com/66558924/203990228-7281fd52-aac2-43b2-950a-93e3ce59ccaf.jpg)

The problem doesn't seem to be in the current code of NURBS Sweep, because I deliberately reverted it to match the one from Apr 2021 and it doesn't matter what calculations it's doing - it's just twice as slow in master. So it should be somewhere else. Maybe in the update system? @Durman I've seen you making detailed studies on the node performances, can you please take a look at this?

I'm uploading the file for examination:
[Sweep_Precision_Symmetry_TestSpeed.zip](https://github.com/nortikin/sverchok/files/10093036/Sweep_Precision_Symmetry_TestSpeed.zip)

Thanks

portnov:
Can you point a commit at which performance was "good"?

rendetto:
This is the one I used for the testing 8bef7918a6bcbb05772c040bbbb189b4b27f5eb8
I'm not sure that I can find the breaking point commit, but I can try.

portnov:
Theoretically it should be easy enough, using git bisect.
The only problem can be is that your file probably will not load on old blender :/

rendetto:
I'm using several Blender versions for those purposes. I'm on it. It just will take some more time because I do those things manually x_x (need to learn about git bisect).

portnov:
Hm. I just loaded your file on blender 2.90 with https://github.com/nortikin/sverchok/commit/8bef7918a6bcbb05772c040bbbb189b4b27f5eb8, and on 3.3.1 with master. I see no difference: for both, tree update takes about 66 ms.
What are your results?
Can you provide profiling data for "good" and "bad" cases?

rendetto:
Ummmm... thats weird. Same scene, same revision:

Blender 2.90.0 - 25fps
DEBUG:sverchok.core.update_system:Node set updated in: 0.0332 seconds

Blender2.93.9 - 11fps but again:
2022-11-25 18:18:49,567 [DEBUG] sverchok.core.update_system: Node set updated in: 0.0307 seconds

I don't know what to think :/

portnov:
@rendetto 
[...]

PR #4805, opened 2022-12-08
Do not use generic (rational) algorithm if the curve is non-rational.

portnov:
refs #4722.

## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


rendetto:
@portnov  I tested the speed of this PR with the setup from https://github.com/nortikin/sverchok/issues/4772
 and now the framerate seems the same as the one before those performance drops.

Thanks a lot!

PR #4806, opened 2022-12-09
Multiple data offset

vkter:
Offset Multiple Data
image

### DIFF
--- a/nodes/list_struct/shift_mk2.py
+++ b/nodes/list_struct/shift_mk2.py
@@ -66,8 +66,16 @@ class ShiftNodeMK2(SverchCustomTreeNode, bpy.types.Node):
             return
 
         data = self.inputs['data'].sv_get(deepcopy=False)
-        number = self.inputs["shift"].sv_get(deepcopy=False)[0][0]
+        number = self.inputs["shift"].sv_get(deepcopy=False)
+        output = []
+        for d,n in zip(data,number):
+            out = self.new_method(d, n[0])
+            output.append(out)
 
+
+        self.outputs['data'].sv_set(output)
+
+    def new_method(self, data, number):
         if self.selected_mode == 'np':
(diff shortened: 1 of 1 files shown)


Durman:
image

nodes/list_struct/shift_mk2.py
```diff
@@ -102,6 +110,8 @@ def process(self):
                         d.rotate(number)
                         sub_output.append(list(d))
                     output.append(sub_output)
+        return output
+
 
 
         self.outputs['data'].sv_set(output)
```
Durman, line 117: I guess last line can be removed.

nodes/list_struct/shift_mk2.py
```diff
@@ -66,8 +66,16 @@ def process(self):
...
+        number = self.inputs["shift"].sv_get(deepcopy=False)
+        output = []
+        for d,n in zip(data,number):
+            out = self.new_method(d, n[0])
+            output.append(out)
 
+
+        self.outputs['data'].sv_set(output)
+
+    def new_method(self, data, number):
```
Durman, line 78: The name of the new method should be something more meaningful. Something like roll list.

vkter:
Whether the applications can be combined？

Durman:
Could you add any description?

vkter:
Offset Multiple Data
image

Durman:
If you add vectorization to one input I guess it's better to add it to another input socket as well. In the example below I would expect two lists to be created by the node.

image

Also it's not clear how to make it work consistently with work of other similar nodes. For example List Slice node has on object in each socket but generate multiple ones.

image

vkter:
The internal code uses zip () to match the input data. In short, it matches the data according to the shortest, which seems to be no problem

Durman:
Nevermind, it can be fixed later.

PR #4810, opened 2022-12-10
Viewer Draw Color: indicate curvature by colors.

portnov:
## Preflight checklist

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


rendetto:
Just a suggestion. It would be more descriptive if we can map a gradient instead of a single color.

portnov:
Yes, but then we have to configure this gradient somehow. It would complicate node's UI.

portnov:
If you really want a gradient, it's already possible with a couple of additional nodes:
Screenshot_20221210_225146

rendetto:
> If you really want a gradient, it's already possible with a couple of additional nodes: Screenshot_20221210_225146

That's cool! Thanks a lot! :)

PR #4817, opened 2022-12-13
append threshold parameter to N-panel.

satabol:

Some glitches happens. Example with no fill:

image

They appears in situations when vertices are very close. But minimal distance is hardcoded in function:

https://github.com/nortikin/sverchok/blob/5773c45f964cb9604f505f029cf6a41926d2b7c8/nodes/modifier_make/bisect.py#L39-L42
image

So I append threshold parameter to the N-panel:
image

if set threshold to zero then glitches disappear and bisect fill holes good:
image

### DIFF
--- a/nodes/modifier_make/bisect.py
+++ b/nodes/modifier_make/bisect.py
@@ -32,12 +32,12 @@ from sverchok.utils.nodes_mixins.sockets_config import ModifierNode
 # but using python bmesh code for driving
 # by Linus Yng / edits+upgrades Dealga McArdle and Victor Doval
 
-def bisect_bmesh(bm, pp, pno, outer, inner, fill):
+def bisect_bmesh(bm, pp, pno, outer, inner, fill, threshold):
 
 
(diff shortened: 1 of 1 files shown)


nodes/modifier_make/bisect.py
```diff
@@ -32,12 +32,12 @@
 # but using python bmesh code for driving
 # by Linus Yng / edits+upgrades Dealga McArdle and Victor Doval
 
-def bisect_bmesh(bm, pp, pno, outer, inner, fill):
+def bisect_bmesh(bm, pp, pno, outer, inner, fill, threshold):
```
portnov, line 35: I would suggest to add a default value for the new parameter here: `threshold = 0.00001` (the same value which was earlier hardcoded). Just in case someone is doing `from sverchok.nodes.modifier_name.bisect import bisect_bmesh` somewhere in scripted node. We do not have any obligations to do not break scripted nodes, but still, if this costs nothing, it's better to preserve compatibility.

satabol: Append

PR #4819, opened 2022-12-13
CAD->bisect node. Some glitches appear on skewed quads. Fixed.

satabol:

Glitches on skewed quads:
image

Bisect_without_normal_update v002

Expected:

Bisect_with_normal_update v001 no_errors edited

and now it works. https://docs.blender.org/api/current/bmesh.types.html#bmesh.types.BMesh.normal_update

I think this patch has to be tested in other functions where **bmesh.ops.bisect_plane** used.

Test:

[Bisect_2022_12_13_23_11.zip](https://github.com/nortikin/sverchok/files/10222818/Bisect_2022_12_13_23_11.zip)

https://user-images.githubusercontent.com/14288520/207464158-75ce0e12-f4a9-4489-9e15-bc0a35d17dfd.mp4

### DIFF
--- a/nodes/modifier_make/bisect.py
+++ b/nodes/modifier_make/bisect.py
@@ -34,7 +34,7 @@ from sverchok.utils.nodes_mixins.sockets_config import ModifierNode
 
 def bisect_bmesh(bm, pp, pno, outer, inner, fill):
 
-
+    bm.normal_update()
     geom_in = bm.verts[:] + bm.edges[:] + bm.faces[:]
     res = bmesh.ops.bisect_plane(
         bm, geom=geom_in, dist=0.00001,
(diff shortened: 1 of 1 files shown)


satabol:
In the library .update_normals is used:

https://github.com/nortikin/sverchok/blob/b7c5278a52781875996737bda5fcbec261a54c96/utils/sv_bmesh_utils.py#L855-L861

bit fill used an old method of filling:

https://github.com/nortikin/sverchok/blob/b7c5278a52781875996737bda5fcbec261a54c96/utils/sv_bmesh_utils.py#L862-L866

mifth:
It's hard to get what's going on. I guess the problem is in NGons? Maybe if you triangulate NGons the issue will be solved?

satabol:
@mifth Good remark. Triangulation can be used as additional method to fix some errors that bisect can not calculate (self intersections, single points and other). It is good to denote it in docs . I try to maximize usefulness for this node. In some tests manual bisect and bisect_plane in Sverchok has different results on same geometry. I could not understand why? And now I think this is it! (with #4817 )

satabol:
More info:

For tests with current Sverchok version: https://gist.github.com/b9d4a75befd5b82f3466317bbbb8c4c1

I spend not much time to find problem with current bisect node:
image

And try to repeat it with Blender bisect:

image

No problemo. )

If set **.normal_update**:

image

then Sverchok result and blender result are equal!

PR #4824, opened 2022-12-17
Bmesh process—bmesh module

vkter:


vkter:
image
holes_fill()

vkter:
image
intersect_face_point()

vkter:
image
connect_verts()

Durman:
I see 2 architectural faults:

- You does not make copies of bmesh objects and I assume it make it impossible to use the nodes in several parallel node branches.

image

- Empty bmesh objects are very heavy. Some number of thousands empty bmesh objects can exhaust memory depending on its capacity.

image

image

If to put all 15k points in one bmesh object the memory usage will be hardly noticeable.

vkter:
The particularity of BMESH processing:
1. The uniqueness of the BMESH object
BMESH is a data that cached in memory. You always modify the same BMESH object, so when you need multiple parallel branches, you need to copy the new BMESH object
image
In addition, the BMESH object cannot be returned to the previous state after being modified. When you modify the BMESH again,  it based on the current state. There are two methods to solve this problem: you can use the global update to update the entire node, or You can use BMESH in the node group. Any changes in the node group will drive the entire node group to update
image

vkter:
> I see 2 architectural faults:
> 
> * You does not make copies of bmesh objects and I assume it make it impossible to use the nodes in several parallel node branches.
> 
> image
> 
> * Empty bmesh objects are very heavy. Some number of thousands empty bmesh objects can exhaust memory depending on its capacity.
> 
> image
> 
> image
> 
> If to put all 15k points in one bmesh object the memory usage will be hardly noticeable.

vkter:
The empty BMESH is too much occupied by memory, which is indeed the case, so I added a release button to the output, and any input BMESH will be released from the memory. This is introduced in the official document: https://docs.blender.org/api/current/bmesh.html
image
After my test, as long as a large number of empty objects are created at one time, they will crash. Perhaps a maximum number limit should be increased? Does anyone have a better way?

Durman:
> 1. In addition, the BMESH object cannot be returned to the previous state after being modified.

I guess this is a property of all objects. This is why we have deepcopy parameter in socket.sv_get, but it just works only with Python lists. I don't see the point not to copy bemesh in each node before it can modify it. This is basic principle of all Sverchok nodes. Ideally it should be implemented in sv_get method so update system could add some optimizations, for example, as you mentioned, in group nodes.

> After my test, as long as a large number of empty objects are created at one time, they will crash. Perhaps a maximum number limit should be increased? Does anyone have a better way?

If each node will create it's own copy we already have them equal to number of bmesh nodes in all Sverchok trees. What in some cases can be measured in thousands. So the only solution which I can see is not to support vectorization for input bmesh nodes (though other nodes seems can still be vectorized).

vkter:
1. The deep copy function is global and very concise. It should not be modified, so I have opened a copy button
2. I added a maximum quantity limit to the input. It is not good to remove the vectorization. The user can decide the maximum quantity by himself
3. When the user uses bmesh processing outside the node group, turn on the copy button, but when the user is in the node group, cancel the copy to save performance. I recommend that the user use bmesh processing nodes in the node group
4. The development logic of the bmesh node is completely based on the bmesh module. When you are familiar with the node, you should process the bmesh object in the form of script, which will be more flexible and save performance.
5. The ultimate goal of the bmesh node is to train users to handle geometry in scripts
image

Durman:
All nodes should copy bmesh, I assume, not only Bmesh Ops.

I believe that with these two nodes I flipped normals of a bmesh. How to visualize the bmesh? I think if a node perform some operation over data it should return its result. Node is like a function you call it with some data and it returns modified data then you can you use it in next node to modify it again and so on.
image

vkter:
I consider your thoughts, which requires a specific package based on specific functions or functions. Such a workload is huge. And SVERCHOK has packaged a large number of BMESH module functions based on the purpose of fast use, we need to make a choice between convenience and flexibility
In addition, it is not feasible for all nodes to copy the bmesh, which will break the bmesh processing process. For example, the face you use the type node to obtain will not be the face you input the bmesh. Their memory addresses are different

For basic users, the 79 functions in ops are already very large. If you want to go deep into bmesh processing, you need to use script nodes, and bmesh processing will be some threshold content

BMesh really lacks a direct preview node, which will be put on the agenda in the future

The return is not set by me. It depends on specific functions. Some functions do not return data

When you are familiar with the processing flow of the bmesh module, this will be very simple
image

Durman:
I'm trying to recreate Extrude Faces node. Is it possible with bmesh nodes? How to use return value of Bmesh Ops node? Why vec socket of scale operation can't work with normal Sverchok list of vertices? I don't think any script node should be used here, otherwise don't see reasons messing with nodes.

image

vkter:
With this example, I found bmesh replication to be pointless, I'll remove the copy button later
image
Regarding your thoughts, I gave examples:
image
The return value depends on the function, it is the basic type of python, dictionaries, tuples and the like, use the script to read it, the function in the ops node should try to give parameters, the lack of necessary parameters The function naturally does not work
Almost every socket has an explanation that will help you use it
image

Durman:
Can we do this in the Second Bmesh Ops?
```py
verts = [bmesh_B.verts[v.index] for v in bmesh_A.verts]
```

Durman:
How it's suppose to work?

https://user-images.githubusercontent.com/28003269/208375812-75f61ef3-b41c-43ad-a6e0-64eada5a4498.mp4

vkter:
Use within a group and use global updating outside the group
https://user-images.githubusercontent.com/74725748/208384041-7c120309-ea14-4bc9-8da9-7eb2cf44b3c6.mp4

Durman:
Well, it does not seem good idea to add nodes which does not work properly in main tree. We have a Sverchok extension [Sverchok-Extra](https://github.com/portnov/sverchok-extra). It includes nodes which usefulness is not proven. I can suggest to consider to add bmesh nodes there. Or you can create your own Sverchok extension and add to Sverchok any nodes you wish. In the last scenario you don't have to prove to anybody your ideas. =)

Durman:
Feel free to ask any question if you'll have any trouble with creating extension. When it will be ready I think there will be sense to remove currently existing bmesh nodes from Sverchok.

PR #4826, opened 2022-12-18
[Curves->Marching Squares on Surface] gap fix.

satabol:

In node's example of [Marching Squares on Surface] there is a gap:
http://nortikin.github.io/sverchok/docs/nodes/curve/marching_squares_on_surface.html
image

and:
image

Recreated scheme of nodes:

https://gist.github.com/a7846c7f7a3c1de01b182ae4981d79c2

image

In this fix I remove a gap:

image

image

### DIFF
--- a/nodes/curve/marching_squares_on_surface.py
+++ b/nodes/curve/marching_squares_on_surface.py
@@ -116,11 +116,20 @@ class SvExMSquaresOnSurfaceNode(SverchCustomTreeNode, bpy.types.Node):
                 u_size = (u_max - u_min)/(samples_u-1)
                 v_size = (v_max - v_min)/(samples_v-1)
 
-                # some times [(u_max - u_min)/(samples_u-1)]*(samples_u-1) > (u_max-umin) !!! need compensation:
-                if( (u_max - u_min) < u_size*(samples_u-1)):
-                    u_size -= ((u_max - u_min)/(samples_u-1)*(samples_u-1)-(u_max-u_min))/(samples_u-1)
-                if( (v_max - v_min) < v_size*(samples_v-1)):
-                    v_size -= ((v_max - v_min)/(samples_v-1)*(samples_v-1)-(v_max-v_min))/(samples_v-1)
+                # some times u_max < u_min + u_size*(samples_u-1) !!! need compensation:
+                if( u_max < u_min + u_size*(samples_u-1)):
+                    for i in range(10):
+                        _u_size = u_size - i*((u_max - u_min)/(samples_u-1)*(samples_u-1) - u_max+u_min)/(samples_u-1)
+                        if( u_max > u_min + _u_size*(samples_u-1)):
+                            u_size = _u_size
+                            break
+
+                if( v_max < v_min + v_size*(samples_v-1)):
+                    for i in range(10):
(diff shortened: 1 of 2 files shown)


nodes/curve/marching_squares_on_surface.py
```diff
@@ -113,8 +113,8 @@ def process(self):
 
                 contours = measure.find_contours(field_values, level=value)
 
-                u_size = (u_max - u_min)/samples_u
-                v_size = (v_max - v_min)/samples_v
+                u_size = (u_max - u_min)/(samples_u-1)-0.0000001
+                v_size = (v_max - v_min)/(samples_v-1)-0.0000001
```
portnov, line 117: I would suggest to make this a node setting, instead of hardcoding. Maybe hide it in node's N panel.

satabol: @portnov This is not good idea. I found real problem and wrote new fix. The description of problem:

With some T value calculation do errors:
image

The problem is (4):

u_size = **(u_max - u_min)/(samples_u-1)**
**u_size * (samples_u-1) > (u_max - u_min)** (WTF??? - float precision)

and this go to next wrong result:
image

In this situation need decrease a bit **u_size** and I propose do this with next code:
image

This take good result:
image

If this is not enough then this can be do with (for range) **u_size/v_size** to calc a value where
**u_size * (samples_u-1) <= (u_max - u_min)**

What do you think?

satabol: MarchingSquaresOnSurface

satabol: Found some glitches. I create for-range with new formula based on:

https://github.com/nortikin/sverchok/blob/b7b06ebc9b61f0002463973f3648dd30f8a807db/utils/marching_squares.py#L31-L32

https://gist.github.com/df4d5b173e4934aff740746f28450cf1
MarchingSquaresOnSurface fix 002

satabol: There is a connection in current version between these lines of code:

https://github.com/nortikin/sverchok/blob/b7b06ebc9b61f0002463973f3648dd30f8a807db/nodes/curve/marching_squares_on_surface.py#L116-L119

https://github.com/nortikin/sverchok/blob/b7b06ebc9b61f0002463973f3648dd30f8a807db/utils/marching_squares.py#L31-L32

**x_size = u_size** and on last iteration in line 31 result of **x** has to be **u_max** but with some **T** values result exceed **u_max**. This is a main problem. If refactor of **make_contours** function and send **u_max** and **v_max** as addition parameters and assign its in last [for-range] cycle like:

x = min_x + x_size * x0
y = min_y + y_size * y0

if x0==samples_u-1: x=u_max

if y0==samples_u-1: y=v_max

then problem will solved forever )))

portnov: you can always add new parameter to make_contours if you make it optional.

satabol: @portnov Thank you! It work. Can you look at that?

satabol: https://gist.github.com/faa14b885442ea6a6388e6bf103faeb5
Curves-MarchinSquares v004 edited

portnov:
Looks good to me.

PR #4830, opened 2022-12-20
Allow extra_menu: tag for nested menus.

portnov:
This allows to process `extra_menu:` tags for nested menus in `index.yaml`.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/ui/nodeview_space_menu.py
+++ b/ui/nodeview_space_menu.py
@@ -520,6 +520,10 @@ class CallPartialMenu(SverchokContext, bpy.types.Operator):
             for cat in add_node_menu:
                 if getattr(cat, 'extra_menu', '') == self.menu_name:
                     cat.draw(_self.layout)
+                elif isinstance(cat, Category):
+                    for item in cat:
+                        if getattr(item, 'extra_menu', '') == self.menu_name:
+                            item.draw(_self.layout)
 
         context.window_manager.popup_menu(draw, title=self.menu_name, icon='BLANK1')
         return {'FINISHED'}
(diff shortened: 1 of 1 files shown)


ui/nodeview_space_menu.py
```diff
@@ -520,6 +520,10 @@ def draw(_self, context):
             for cat in add_node_menu:
                 if getattr(cat, 'extra_menu', '') == self.menu_name:
                     cat.draw(_self.layout)
+                elif isinstance(cat, Category):
+                    for item in cat:
+                        if getattr(item, 'extra_menu', '') == self.menu_name:
+                            item.draw(_self.layout)
```
Durman, line 526: I guess this method can be used instead

https://github.com/nortikin/sverchok/blob/b7b06ebc9b61f0002463973f3648dd30f8a807db/ui/nodeview_space_menu.py#L272

```suggestion
            for cat in add_node_menu.walk_categories():
                if getattr(cat, 'extra_menu', '') == self.menu_name:
                    cat.draw(_self.layout)
```

Probably I preferred first variant for performance.

portnov: ok, updated.
I had to add a parameter to walk_categories to include CustomMenu items in order to include Groups and Presets into 5 menu.

Durman: I did not realize that Custom Menus are also needed. Then using walk_categories it not best choice. The idea was that the method returns only categories. The annontation also show this. And there is recursive call which makes things a bit more complicated. I can suggest than:

```py
sub_categories = [add_node_menu]
for _ in range(1000):
    cat = sub_categories.pop()
    for cat_item in cat:
        if getattr(cat, 'extra_menu', '') == self.menu_name:
            cat_item.draw(_self.layout)
        if hasattr(cat, '__iter__'):
            sub_categories.append(cat_item)
        if not sub_categories:
            break
else:
    log('There is some problem')
```

Issue #4247, opened 2021-07-25
New update system time display location error


vkter:
![V~OF4CNJI_7KZKH %GHMB%Q](https://user-images.githubusercontent.com/74725748/126884625-be49a568-4cc1-4429-8cf8-297dacc48467.png)

Communicate with other users and understand that some people display normal and some display error

Durman:
Can you elaborate the issue?

vkter:
After the node is updated, the location where the update time is displayed is not above each node, but scattered, which is obviously difficult to judge the specific time of each node

Durman:
Aaa.. I see. The thing is that there is no way to stick the text to nodes efficiently. To update position you can just make an attempt of link creation.

![triger_redraw](https://user-images.githubusercontent.com/28003269/126942533-1403c8d5-6995-4ffc-a4c3-4984b58b1969.gif)

The alternative is to put the message inside a node.

![2021-07-26_10-22-34](https://user-images.githubusercontent.com/28003269/126942651-3ecf7453-0e9d-49c5-a109-95011364d992.png)

@vkter Which message display is more convenient do you think?

vkter:
https://user-images.githubusercontent.com/74725748/126945366-d55682c8-f5e8-4966-9214-a9b9ceb495d1.mp4

My question is, it doesn't seem correct. I've asked other users about this question, and some are the same as me

Durman:
I will have a look.

zeffii:
![moving_higher_res](https://user-images.githubusercontent.com/619340/126972388-dba658b5-36b0-4d67-91f4-239a0181303e.gif)

It depends on which screen the computer is displaying on, not just timer data is drawing in the wrong place, but also the exception.

Durman:
@vkter Let me know if the problem was not solved.

vkter:
The display is normal after updating again

PR #4836, opened 2022-12-22
stethoscope offset scale fix

### LINKED ISSUE #4709: UI scale - Blender 3.4
## Problem statement

There are some changes in UI scale of Blender 3.4

https://user-images.githubusercontent.com/28003269/196092922-4bace697-3ae2-4219-a07c-56e67ee3d636.mp4

satabol:
Windows 11, Blender 3.4.1, Sverchok-master

Proposal to fix #4709 

If Blender ui scale in preferences set to non 1.0 then offset of stethoscope show text in unexpected location:

image

**Position of text look good after fix**:

image

UI_SCALE ISSUE fix 001

### DIFF
--- a/nodes/text/stethoscope_v28.py
+++ b/nodes/text/stethoscope_v28.py
@@ -248,7 +248,7 @@ class SvStethoscopeNodeMK2(SverchCustomTreeNode, bpy.types.Node, LexMixin, SvNod
                 'content': processed_data,
                 'location': get_xy_for_bgl_drawing,
                 'color': self.text_color[:],
-                'scale' : float(scale),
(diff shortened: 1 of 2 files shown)


satabol:
I found problem like this in: https://github.com/nortikin/sverchok/issues/4247 with closed in https://github.com/nortikin/sverchok/commit/2409d84ca7b1324306f2d1935820ed862ddc7317

portnov:
@satabol doesn't this fix break behaviour on blender < 3.4?
If it does, please add a check (`bpy.app.version >= ...`) before applying the scale fix.

Durman (changes requested):
It does not work when Resolution Scale is changed before Stethoscope is added.

PR #4841, opened 2022-12-25
Possibility to select a menu preset

portnov:

For now, it is not clear (yet), what should be The Right structure of Sverchok's Add Node menu. The ongoing discussion is in #4831.

## Solution description

* `index.yaml` in the root of sverchok directory stays in place. It is still the default menu, but it is no longer the menu which is actually loaded at sverchok startup!
* At Sverchok startup we load the `index.yaml` menu under user's datafiles directory - something like `~/.config/blender/3.4/datafiles/sverchok/index.yaml`. 
* If at startup Sverchok does not see that file under datafiles, it will copy the default `index.yaml` there, before loading it.
* Non-default menu presets that we provide with Sverchok are under `menus/` directory.
* For standard menu presets, which are named `full_*.yaml`, there will be a CI test which checks that they have all the same nodes as `index.yaml` has.
* There can be menu presets, which contain only some subset of nodes. The convention is to name them `partial_*.yaml`. At the moment, there are no such presets in Sverchok distribution, but they may appear in future.
* The user can create his own menu presets, and place them under datafiles directory, `menus/` subdirectory - something like `~/.config/blender/3.4/datafiles/sverchok/menus/MY_MENU.yaml`.
* In Sverchok settings, there is now an option, allowing the user to select which preset he wishes to use. When the user clicks "Set", we copy the selected preset file to `index.yaml` under user's datafiles directory. So at next startup, we will use that preset.
* If the user wishes, he may directly edit `index.yaml` under his datafiles directory, without having to touch the one which is in sverchok directory.

Screenshot_20221225_172422

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [ ] Ready for merge.


settings.py
```diff
@@ -1,12 +1,15 @@
...
+import shutil
 
 import bpy
 from bpy.types import AddonPreferences
 from bpy.props import BoolProperty, FloatVectorProperty, EnumProperty, IntProperty, FloatProperty, StringProperty
 
 from sverchok.ui.utils import message_on_layout
+from sverchok.ui.presets import get_sverchok_directory
 
 """Don't import other Sverchok modules here"""
```
Durman, line 14: You are going to initialize half of Sverchok architecture via the settings module again.

portnov: okay...
get_sverchok_directory is quite simple function, it should be somewhere in core I think...

PR #4851, opened 2023-01-01
Disable "Scene Update" by default

rendetto:
@Durman When opening old setups containing lots heavy trees it's a real pain until disabling all those flags. It's better when all are disabled by default.

### DIFF
--- a/node_tree.py
+++ b/node_tree.py
@@ -213,7 +213,7 @@ class SverchCustomTree(NodeTree, SvNodeTreeCommon):
         name="Scene update",
         description="Update upon changes in the scene",
         options=set(),
-        default=True)
+        default=False)
     """If enabled together with `SverchCustomTree.sv_process` the tree will be
     reevaluated upon changes in the scene. It will effect only nodes with
     `UpdateNodes.is_interactive` property enabled. The scene changes can be:


Durman:
I see the problem but the solution, though it's simple, is not good from my point of view. I see the option as switch on by default because it brings more convenient experience and ease in use.

There might be another solution which can solve the problem. When a file is opened a special handler/function is called where we can execute an arbitrary stuff.

https://github.com/nortikin/sverchok/blob/7623fc1c6c7c10feb82b7ae58bd6837a3da6efbc/core/handlers.py#L145-L148

I assume that in the function we can test trees whether they are old or not and switch off the option for old ones. The problem here how to define the old trees. =)
Probably we can start to mark our trees (or nodes) with current Sverchok version in this case if a tree (a node) does not have the mark then it's an old tree and the option should be switched off.

rendetto:
> Probably we can start to mark our trees (or nodes) with current Sverchok version in this case if a tree (a node) does not have the mark then it's an old tree and the option should be switched off.

Will this check(and then disabling) be executed before loading the file or after loading it?
I'm asking because before doing this PR I tried to disable all flags automatically with a script in an already loaded file - and it preformed insanely slow, like I'm disabling those manually.

Durman:
Now I'm not quite understand the problem. Scene update does not effect processing tree when a file is opening. The only option which is responsible whether a tree will be executed on startup is Live update.

https://github.com/nortikin/sverchok/blob/92354054461d07f72ae87e16f394f27081fdc612/core/update_system.py#L71-L74

rendetto:
> Now I'm not quite understand the problem. Scene update does not effect processing tree when a file is opening. The only option which is responsible whether a tree will be executed on startup is Live update

Yes it does not affect the processing while opening but after that - pressing the Scene Update buttons or disabling those with a script on an already opened file is very unresponsive. I guess your solution will check before that so it doesn't matter. 
Should I just close those this PR? Maybe open an issue instead?
Thanks

Durman:
Yes, I guess to open new issues is good idea. And probably you can attach some example to reproduce the problem more precisely.

PR #4856, opened 2023-01-02
Menu presets: automatic application.

portnov:

This is a continuation of #4841.

When we add a node, or just rearrange menu presets for some reason, user's index.yaml in datafiles directory still has to be updated. Before current PR, user had to press "Set" in preferences manually after each sverchok update. This is not obvious at all.

## Options which were considered

1. Original solution: copy the preset from sverchok directory to user's datafiles directory only at first sverchok startup, when there is no file in datafiles. - :x: Rejected: the file will not be updated when we add a new node in Sverchok, or rearrange the default menu.
2. Add a possibility to apply the selected menu preset automatically, at each blender startup. If enabled, at startup, copy the selected preset file from sverchok directory to user's datafiles. — :x: Rejected: there is no sense in having `index.yaml` in user's datafiles, if that file is never actually used, as it is overwritten at each blender startup.
3. ...If enabled, overwrite user's `index.yaml` only when the file in sverchok directory is newer than the one in datafiles. :x: Rejected: not obvious for user, when his changes are overwritten, and there is no clear enough way to notify the user of such actions (message in the log will be most probably ignored).
  * Sub-option: preserve previous version of user's `index.yaml` as `index.yaml.bak`. :x: Rejected: there is no clear or obvious way for user to merge his version of menu with the one from sverchok preset. And again, no clear way for user to know when his changes were overwritten.
4. Options 1 - 3 were designed having in mind the problem that we can not refer to preferences at the moment when we are loading the menu. But, if we need an option to enable or disable menu file overwriting, then we will have to solve that problem somehow. But, once it is solved, we can look into preferences and use menu preset file name from there. So: If the option is enabled, then do not copy anything, just use menu preset file from sverchok directory. If it is disabled, then we copy the preset file once at first Sverchok startup, and when the user explicitly presses the button. This means that if the option is disabled, the user takes responsibility of updating his `index.yaml`. :heavy_check_mark: 

## Solution description

Add a possibility to use the menu preset file from sverchok directory directly, ignoring the file in user's datafiles.

If the feature is not enabled, it means the user takes the responsibility on updating his index.yaml manually.

Screenshot_20230104_115648

Technically, for this to work, I had to implement "lazy" initialization of "add node" menu, because in previous state, when I tried to call to preferences at the place where we located `index.yaml`, it appeared that preferences were not registered yet.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/ui/nodeview_space_menu.py
+++ b/ui/nodeview_space_menu.py
@@ -29,8 +29,11 @@ from bpy.props import StringProperty
 from sverchok.ui.sv_icons import node_icon, icon, get_icon_switch
 from sverchok.ui import presets
 from sverchok.ui.presets import apply_default_preset
+from sverchok.ui.utils import get_menu_preset_path
+from sverchok.utils.context_managers import sv_preferences
 from sverchok.utils import yaml_parser
 from sverchok.utils.modules_inspection import iter_classes_from_module
+from sverchok.utils.logging import info, debug
 
 
 """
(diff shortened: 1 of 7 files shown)


ui/nodeview_space_menu.py
```diff
@@ -466,13 +469,36 @@ def draw(self, context):
...
+        info(f"Applying menu preset {default_menu_file} at startup")
+        shutil.copy(default_menu_file, menu_file)
+    debug(f"Using menu preset file: {menu_file}")
+    add_node_menu = Category.from_config(yaml_parser.load(menu_file), 'All Categories', icon_name='RNA')
+
+def get_add_node_menu():
+    global add_node_menu
+    if add_node_menu is None:
+        setup_add_menu()
+    return add_node_menu
```
Durman, line 501: I can't agree less that the function represents functionality of the module. If it returns Category instances than it can/should be a class method of the Category. The same is about setup_add_menu.

Also I don't like a laze concept here and in general - it makes things more complicated than they could be. It probably would be valid to use if initialization of the menu was really in arbitrary place in code. I assume you don't know from where the menu is called what leads to treating code as a black box / magic. From my knowledge this is not good approach.

I had a look at the places and it seems that the menu is only used in panels and operators what means that if you need the settings to be available during initialization of the menu it's possible to initialize the menu in the register function.

```py
def register():
    add_menu = Category.generate_main_menu()
    add_menu.register()
```

portnov: I will not have objections if you rewrite it according to your suggestions :)

Durman: I'm too tired to rewrite somebody code.

PR #4877, opened 2023-01-16
Introducing "Solid Section" node

rendetto:
Solid Section
=============
SectionNodeMenuV2

The node will generate Curves and Points from intersecting different Shapes including Solids, Solid Faces, Solid Edges, NURBS Surfaces and NURBS Curves.

This node can operate on pair of objects ("Shape A" intersects "Shape B") or on pair of lists of objects.

Inputs
------

The node has the following inputs:

* **Shape A**. The first object to perform operation on. This input is
  mandatory and can accept a list of Shapes*.
* **Shape B**. The second object to perform operation on. This input is
  mandatory and can accept a list of Shapes*.
  
  Curves (NURBS Curves and Solid Edges) cannot be mixed with other Shape types.  
  For example [<NURBS_Curve>, <Solid_Edge>, <Solid>] is not a valid input list.
  
  Valid inputs are [<NURBS_Curve1>, <NURBS_Curve2>, <Solid_Edge>] or [<NURBS_Curve>, <Solid_Edge>]  
  or [<NURBS_Surface1>, <Solid>, <NURBS_Surface2>] etc.

Options
-------

This node has the following parameters:

* **NURBS Output**. This parameter is available in the N panel only. If
  checked, the node will generate curves in NURBS representation. Otherwise, it
  will generate standard FreeCAD curves (Solid Edge curves). Unchecked by default.
  
Outputs
-------

The has the following outputs:

* **Vertices**. The resulting intersection (puncture) Points or endpoints of the intersection Curves.  
* **Curves**. The resulting intersection Curves.

Product variations of the Section operation:

* Solid × Curve → Point (common case) or Curve (if the curve or part of the curve overlays with the solid)

* Solid × Surface → Curve (common case) or Point (when both are only touching)

* Curve × Surface → Point (common case) or Curve (if the curve or part of the curve overlays with the surface)

* Solid × Solid → Curve (common case) or Point (when both are only touching)

* Surface × Surface → Curve (common case) or Point (when both are only touching)

* Curve × Curve → Point (common case) or Curve (if curves overlay at a certain segment)

If no intersections are found the node will output empty lists.

Examples
--------

**Curve** × **Curve** → **Points**:
SVSectionNode_CurveXCurve

**Curves** × **Surface** → **Points and Curve**:
SVSectionNode_CurveXSurface

**Surface** × **List of Surfaces** → **Curves**:
SVSectionNode_SurfaceXListOfSurfaces

**Surface** × **Solid** → **Curves**:
SVSectionNode_SurfaceXSolid


rendetto:
Hmm.. there are errors. Will try again.

PR #4878, opened 2023-01-16
Introducing "Solid Section" node

rendetto:
Solid Section
=============
SectionNodeMenuV2

The node will generate Curves and Points from intersecting different Shapes including Solids, Solid Faces, Solid Edges, NURBS Surfaces and NURBS Curves.

This node can operate on pair of objects ("Shape A" intersects "Shape B") or on pair of lists of objects.

Inputs
------

The node has the following inputs:

* **Shape A**. The first object to perform operation on. This input is
  mandatory and can accept a list of Shapes*.
* **Shape B**. The second object to perform operation on. This input is
  mandatory and can accept a list of Shapes*.
  
  Curves (NURBS Curves and Solid Edges) cannot be mixed with other Shape types.  
  For example [<NURBS_Curve>, <Solid_Edge>, <Solid>] is not a valid input list.
  
  Valid inputs are [<NURBS_Curve1>, <NURBS_Curve2>, <Solid_Edge>] or [<NURBS_Curve>, <Solid_Edge>]  
  or [<NURBS_Surface1>, <Solid>, <NURBS_Surface2>] etc.

Options
-------

This node has the following parameters:

* **NURBS Output**. This parameter is available in the N panel only. If
  checked, the node will generate curves in NURBS representation. Otherwise, it
  will generate standard FreeCAD curves (Solid Edge curves). Unchecked by default.
  
Outputs
-------

The has the following outputs:

* **Vertices**. The resulting intersection (puncture) Points or endpoints of the intersection Curves.  
* **Curves**. The resulting intersection Curves.

Product variations of the Section operation:

* Solid × Curve → Point (common case) or Curve (if the curve or part of the curve overlays with the solid)

* Solid × Surface → Curve (common case) or Point (when both are only touching)

* Curve × Surface → Point (common case) or Curve (if the curve or part of the curve overlays with the surface)

* Solid × Solid → Curve (common case) or Point (when both are only touching)

* Surface × Surface → Curve (common case) or Point (when both are only touching)

* Curve × Curve → Point (common case) or Curve (if curves overlay at a certain segment)

If no intersections are found the node will output empty lists.

Examples
--------

**Curve** × **Curve** → **Points**:
SVSectionNode_CurveXCurve

**Curves** × **Surface** → **Points and Curve**:
SVSectionNode_CurveXSurface

**Surface** × **List of Surface** → **Curves**:
SVSectionNode_SurfaceXListOfSurfaces

**Surface** × **Solid** → **Curves**:
SVSectionNode_SurfaceXSolid


rendetto:
I'm not sure how to read this failing test:

```
FAIL: test_full_menu_presets (ui_tests.UiTests) (path='/home/runner/work/sverchok/sverchok/installation/blender/2.93/scripts/addons/sverchok/menus/full_nortikin.yaml')
----------------------------------------------------------------------
  File "/home/runner/work/sverchok/sverchok/installation/blender/2.93/scripts/addons/sverchok/tests/ui_tests.py", line 100, in test_full_menu_presets
    self.assertEqual(preset_nodes, index_nodes)
AssertionError: Items in the second set but not the first:
'SvSolidSectionNode'
```

I replicated how SvSolidBooleanNode was done and didn't expected this error. Is it a show stopper?

portnov:
At last run tests are passed.

nodes/solid/solid_section.py
```diff
@@ -0,0 +1,112 @@
...
+        layout.prop(self, 'nurbs_output', toggle=True)
+
+    def sv_init(self, context):
+        self.inputs.new('SvSolidSocket', "Shape A")
+        self.inputs.new('SvSolidSocket', "Shape B")
+        
+        self.outputs.new('SvVerticesSocket', "Vertices")
+        self.outputs.new('SvCurveSocket', "Curves")
+
+    def make_solid(self, solids):
```
portnov, line 50: do we really need this method? It seems like it's possible to combine it with `make_solid_simple` into one method.

portnov: Also I'm not sure if wrapping the result into `SvBoolResult` makes sense here. In Solid Boolean that class is used to bear information about which edge / face in the resulting object come from which original object. But here, as far as I understand, such information is not available anyway.

rendetto: Thanks! Noted and updated.

portnov (approved):

portnov:
@rendetto should we merge this, or maybe you want to add something?

rendetto:
@portnov Yes, please merge it. I don't plan adding more things at this stage. Thanks a lot for the help!

portnov:
okay, then please add a comment here when you think it will be ready for merge.

rendetto:
@portnov I just did the renaming pass. I'm feeling a bit uncomfortable as I'm replacing the term "Solid" with the term "Shape". Will you approve this terminology twist?

portnov:
ok.

portnov:
@rendetto FYI.
Screenshot_20230117_234903
It seems FreeCAD finds the intersection line just point-by-point, and gives us degree-1 curve, i.e. just a polyline with a lot of points. Well, it's better than nothing :)

portnov:
This is how one can make a more nice curve.
Screenshot_20230117_235906

rendetto:
> It seems FreeCAD finds the intersection line just point-by-point, and gives us degree-1 curve, i.e. just a polyline with a lot of points. 

Yes, I just checked this with FreeCAD and that's what Section operation does there too. BUT also noticed that there is an option in the tool to approximate the result. I did a quick search and found approximate( ) method in TopoShape API. It doesn't seem to have any parameters but it looks like a valid algorithm that we can use also.

FreeCADApproximateSection

portnov:
@rendetto there is `approximate` method in `FreeCAD.Part.BSplineCurve` class. It may be of use.
But that should, I think, be used in Sverchok as yet another implementation of "Approximate NURBS Curve" node, in addition to SciPY and Geomdl.

rendetto:
> there is `approximate` method in `FreeCAD.Part.BSplineCurve` class. It may be of use.
> But that should, I think, be used in Sverchok as yet another implementation of "Approximate NURBS Curve" node, in addition to SciPY and Geomdl.

Yes, it sounds most appropriate for it to be there. I don't feel confident enough to dive into such thing yet :/ 
It looks like an interesting alternative and yet it will do similar things so it's not crucial but nice to have.
I found that CurvesWB is using the method and at first glance it works pretty good. Unfortunately the only way to compare the ones we have with this one is to have them all in.

portnov:
You can always experiment by use of scripted node.

rendetto:
Yeah, I already started gathering the info and reading about but don't know If I will succeed. Too many new things and almost zero experience :)

PR #4886, opened 2023-02-01
"Interpolate NURBS Surface" node - FreeCAD implementation

rendetto:
This PR will implement the interpolation method from FreeCAD/OCCT.

Interpolate_NURBS_Surface_Node_FreeCAD

The resulting surface is of degree 3 and continuity C2 (except we use 2x2 or 2xN grid of Vertices) .

### DIFF
--- a/nodes/surface/interpolate_nurbs_surface.py
+++ b/nodes/surface/interpolate_nurbs_surface.py
@@ -28,7 +28,6 @@ class SvExInterpolateNurbsSurfaceNodeMK2(SverchCustomTreeNode, bpy.types.Node):
     bl_idname = 'SvExInterpolateNurbsSurfaceNodeMK2'
     bl_label = 'Interpolate NURBS Surface'
     bl_icon = 'SURFACE_NSURFACE'
-    sv_dependencies = {'geomdl', 'FreeCAD'}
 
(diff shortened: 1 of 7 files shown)


nodes/surface/interpolate_nurbs_surface.py
```diff
@@ -8,22 +8,27 @@
...
     """
     Triggers: NURBS Surface Interpolate
     Tooltip: Interpolate NURBS Surface
     """
-    bl_idname = 'SvExInterpolateNurbsSurfaceNode'
+    bl_idname = 'SvExInterpolateNurbsSurfaceNodeMK2'
     bl_label = 'Interpolate NURBS Surface'
     bl_icon = 'SURFACE_NSURFACE'
-    sv_dependencies = {'geomdl'}
+    sv_dependencies = {'geomdl', 'FreeCAD'}
```
portnov, line 31: @Durman says there is no need nowadays to include optional dependencies to `sv_dependencies`. So it's better to just remove this line.

rendetto: Done! Tested that the node works without that line.

rendetto:
@portnov Thanks!

PR #4891, opened 2023-02-28
Logging refactoring

Durman:
### Fixing problems:

- Logging call was in some another module, not where actual logging was happening. It made such record attributes as `modulename` or `lineno` useless.
- Root logger was configured what could conflict with using logging in other add-ons.

### New features

- Colored logging =)
image

### How to use logging now?

Inside nodes nothing has changed `self.debug(message)`

In other modules:

```diff
- from sverchok.utils.logging import debug
- debug(message)

# option 1
+ import logging
+ logger = logging.geLogger('sverchok')
+ logger.debug(message)

# option 2
+ from sverchok.utils.sv_logging import sv_logger
+ sv_logger.debug(message)

# option 3
+ from sverchok.utils.sv_logging import get_logger
+ logger = get_logger()
+ logger.debug(message)
```

For Sverchok extensions first option is especially useful since during reloading event sverchok can reload after an extension. It has sense to add its own sub logger `logging.getLogger('sverchok.megapolis')`

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done.


portnov:
Do we really need `sv_` prefix in module name?

Durman:
I have a conflict with build-in `logging` module when I execute files in the utils folder.

PR #4902, opened 2023-04-25
Add epsilon outside button raycaster node

ArpegorPSGH:
I just added the option to set the epsilon for raycasting, as I need it for a project. I copied/pasted what was done for the overlap polygons node.


ArpegorPSGH:
There was an issue due to some conflict with the already existing version of the node. Did you solve it?

Durman:
The test fails because documentation (which was missed) to some node previously was added and name of the node was not removed from the missing documentation list. It's irrelivant to your changes.

ArpegorPSGH:
Ok.

Issue #4915, opened 2023-05-27
Sverchok not working at all in Blender 4.0 Alpha


meikasahara:
Operating System: Xubuntu 22.04.2 LTS with 5.15.0-72-generic kernel.

In any of the latest Blender 4.0 alpha builds (later than May 22nd build),  the latest Sverchok 1.2 add-on does not work at all. The add-on appears to be enabled in the Preferences, but there is no context, and it does not appear in the editor types at all.
When loading, Blender returns the following errors in the console regarding Sverchok:

 ```
File "/home/user/.config/blender/4.0/scripts/addons/sverchok-master/__init__.py", line 113, in <module>
    imported_modules, node_modules, core = import_sverchok()
  File "/home/user/.config/blender/4.0/scripts/addons/sverchok-master/__init__.py", line 106, in import_sverchok
    node_modules_ = core.import_nodes()
  File "/home/user/.config/blender/4.0/scripts/addons/sverchok-master/core/__init__.py", line 56, in import_nodes
    import_modules(names, '{}.{}'.format(base_name, category), node_modules)
  File "/home/user/.config/blender/4.0/scripts/addons/sverchok-master/core/__init__.py", line 63, in import_modules
    im = importlib.import_module('.{}'.format(m), base)
```
When trying to disable and re-enable the addon from the Preferences, the following error appears:

`addon_utils.disable: sverchok-master not disabled
addon not loaded: 'sverchok-master'
cause: cannot import name 'core' from partially initialized module 'sverchok-master' (most likely due to a circular import) (/home/user/.config/blender/4.0/scripts/addons/sverchok-master/__init__.py)
addon not loaded: 'sverchok-master'
cause: cannot import name 'core' from partially initialized module 'sverchok-master' (most likely due to a circular import) (/home/user/.config/blender/4.0/scripts/addons/sverchok-master/__init__.py)
`

**Comment by Durman** [comment]
It seems for Blender 4.0 all strings `3D_UNIFORM_COLOR` should be replaced with `UNIFORM_COLOR` according to the example in the docs - https://docs.blender.org/api/4.0/gpu.html#d-lines-with-single-color

Probably there are others incompatible changes in 4.0

**Comment by meikasahara** [comment]
Sverchok can now be installed on Blender 4.0 but Stethoscope and Viewer Draw nodes do not work. The following errors appear in the console: 
```

Traceback (most recent call last):
[...]

Issue #4895, opened 2023-03-30
Switch from BGL to GPU module


Durman:
## Problem statement

> **BGL Deprecation**
BGL is a direct wrapper around OpenGL calls. The BGL module is deprecated for and since Blender 3.4 has been replaced with the gpu module. From this release it will not work when using the Metal backend. Add-ons and scripts that uses the bgl module are warned about the deprecation and should use the gpu module in stead.

https://wiki.blender.org/wiki/Reference/Release_Notes/3.5/Python_API

Nodes require changes:

- [ ] Viewer Draw
- [ ] Stethoscope MK2
- [ ] Matrix View
- [ ] Viewer Index
- [ ] Viewer 2D
- [ ] Viewer Draw Curve
- [ ] Viewer Draw Surface
- [ ] Texture Viewer
- [ ] Texture Viewer Lite
- [ ] SvWaveformViewer
- [ ] Console Node
- [ ] Solid Viewer (did not test)
- [ ] Easing 0..1

zeffii:
oh boy : /

zeffii:
ok. this seems promising: https://docs.blender.org/api/current/gpu.state.html

for `vdmk4`
```python
bgl.glEnable(bgl.GL_BLEND)
bgl.glPolygonMode(bgl.GL_FRONT_AND_BACK, bgl.GL_LINE)
bgl.glEnable(bgl.GL_POLYGON_OFFSET_FILL)
bgl.glPolygonOffset(1.0, 1.0)
bgl.glDisable(bgl.GL_POLYGON_OFFSET_FILL)
bgl.glPolygonMode(bgl.GL_FRONT_AND_BACK, bgl.GL_FILL)
bgl.glLineWidth(config.line_width)
bgl.glPointSize(config.point_size)
```
to

```python
....no POLYGON_OFFSET_FILL ?
gpu.state.face_culling_set(culling)   # NONE, FRONT, BACK but no FRONT_AND_BACK ?
gpu.state.blend_set(mode)
gpu.state.point_size_set(size)
gpu.state.line_width_set(width)
```

Andrej730:
[...]

PR #4936, opened 2023-07-01
remove use of bgl module in all nodes. ( B3.5+ )

zeffii:
Addresses the Blender.org deprecation of the `bgl` module. This means that the nodes listed in [this issue](https://github.com/nortikin/sverchok/issues/4895) should be revised to meet the new specifications. Should be cool, if all bgl features have a `gpu.state` analogue

- addresses https://github.com/nortikin/sverchok/issues/4915 too
- also `blf` module has gained modifications between 3.4 and 3.5+/4.0

## Preflight checklist

- [x] sverchok appears in window list. and loads
     
image
 
- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


Durman:
Are you going to drop the support of B3.4- by the changes?

If so my proposal is to make release 1.2 and commit the changes in next Sv version. There are a few issue which would be good to solve before the release - https://github.com/nortikin/sverchok/milestone/11.

zeffii:
> Are you going to drop the support of B3.4- by the changes?

i would prefer to support both, so that's a slight annoyance that needs a tidy solution. good point. 

and there's changes in `blf` too..
```python
blf.size() takes exactly 2 arguments (3 given)
```
fun :)

zeffii:
point size and line width don't seem to work at the moment,  at least --- on my machine.

zeffii:
@Durman  is it sane to try to support pre b3.5?

One way i could imagine doing that is by making a wrapper for the bgl/blf functions we do use. With little code modifications we can then port over ~much~ some of the code unchanged. Rather than having forks in the code of each node, put the forks in the wrapper.

```python
from sverchok.utils import bgl_wrapper as bgl
```
and then code like this would still be valid
```python

        bgl.glLineWidth(config.line_width)
```
but in `bpy.app.version >= (3, 5, 0):`   it would call instead
```python
        gpu.state.line_width_set(config.line_width)

```

zeffii:
additionally, drawing triangles using vdmk4 ..
```
    ibo = GPUIndexBuf(type=type, seq=indices)
TypeError: IndexBuf.__new__: invalid sequence length. expected 3, got 4
```
this appears to have changed at some point.

zeffii:
i'm still looking for a way to do the `offset polygon` in gpu.state , else.... it might have to be done in the shader.

Durman:
We can drop support of B3.4 and lower. Also we can support both the new and old API. In this case it would be nice to keep code so it would be simple to remove old API some day. I think function implementation could look like this

```py
def universal_fun():
    if bl_v >= 3.5:
        fun_a()
    else:
        fun_b()
```
In this example `fun_b` calls old API and can be removed some day. Also similar thing can be done with classes which could be called something like `DrawingEngine`. One class for calling old API and another for calling new one. But that probably will take more time to code.

zeffii:
@Durman yes a `DrawingEngine` makes sense, and I think I will do that before the issue is pushed.

Issue #4947, opened 2023-07-11
Overlapping Voronoi 3D sites if points distribute over surface


satabol:
Sverchok-master, windows 11, Blender 3.6.1.

Sometimes glitch appears:

![image](https://github.com/nortikin/sverchok/assets/14288520/75d73b93-456e-4961-a363-e17d723fc392)

![image](https://github.com/nortikin/sverchok/assets/14288520/4cb8d256-355a-4442-8809-0542b8467c70)

File to test:
[Voronoi on mesh glitces if sites on surface of mesh](https://github.com/nortikin/sverchok/files/12016660/Voronoi.on.mesh.003.blend.zip)

satabol:
Working now

PR #4948, opened 2023-07-11
#4947 fixing Overlapping Voronoi 3D sites

satabol:
01. Fix some glitches of disappearing parts if spacing is growing.
02. Voronoi algorithm speed up. Has some glitches. qhull do not see some sites near each other.

---------

01.
image

02.
image

Sites 1 and 9 overlapped. QHull do not see them as overlapped.

for tests:
[Voronoi on mesh.004.glitches.overlapped sites.blend.zip](https://github.com/nortikin/sverchok/files/12016780/Voronoi.on.mesh.004.glitches.overlapped.sites.blend.zip)

---------

See also:

- https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.Voronoi.html
- http://www.qhull.org/html/qh-quick.htm#options

### DIFF
--- a/utils/voronoi3d.py
+++ b/utils/voronoi3d.py
@@ -238,15 +238,22 @@ def voronoi_on_mesh_bmesh(verts, faces, n_orig_sites, sites, spacing=0.0, fill=T
 
     def get_ridges_per_site(voronoi):
         result = defaultdict(list)
+        sites_list = []
         for ridge_idx in range(len(voronoi.ridge_points)):
             site1_idx, site2_idx = tuple(voronoi.ridge_points[ridge_idx])
+            if (voronoi.points.shape[0]-1)<site1_idx:
+                continue
+            if (voronoi.points.shape[0]-1)<site2_idx:
+                continue
             site1 = voronoi.points[site1_idx]
             site2 = voronoi.points[site2_idx]
             middle = (site1 + site2) * 0.5
             normal = site2 - site1
             plane = PlaneEquation.from_normal_and_point(normal, middle)
+            sites_list.append( (ridge_idx, (site1_idx, site2_idx), (list(site1), list(site2) ), plane, ) )
             result[site1_idx].append(plane)
             result[site2_idx].append(plane)
+        print("sites_list:", len(sites_list))
         return result
 
     def cut_cell(verts, faces, planes, site, spacing):
@@ -296,13 +303,14 @@ def voronoi_on_mesh_bmesh(verts, faces, n_orig_sites, sites, spacing=0.0, fill=T
     edges_out = []
     faces_out = []
 
-    voronoi = Voronoi(np.array(sites))
+    voronoi = Voronoi(np.array(sites, dtype=np.float64), qhull_options='TR1 QJ1e-6 Qz')#, qhull_options='TR10 QJ1e-6 QbB') #, qhull_options='QJ1e-06') #, qhull_options='Qs Qc QJ')
     ridges_per_site = get_ridges_per_site(voronoi)
     if isinstance(spacing, list):
         spacing = repeat_last_for_length(spacing, len(sites))
     else:
         spacing = [spacing for i in range(len(sites))]
     for site_idx in range(len(sites)):
+    #for site_idx in range(n_orig_sites):
         cell = cut_cell(verts, faces, ridges_per_site[site_idx], sites[site_idx], spacing[site_idx])
         if cell is not None:
             new_verts, new_edges, new_faces = cell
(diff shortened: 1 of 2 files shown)


satabol:
image

Test:

https://github.com/nortikin/sverchok/assets/14288520/0b392b1a-538f-44d7-af96-c3f89b9a5f94

satabol:
Description of optimization:
image


**1. Optimization for except some bisects**
1. if bbox_aligned is on the positive side of site then object will have no changes after bisect and all its vertices stay unchanged. One can skip that bisects.
image

2. if bbox_aligned is on the negative side of the site then object will be erased (clear) after bisect (if any bisects was before too). So one can skip all bisects and return empty object without of any bisects.
image

**2. Optimization for bisect order**
This program doesn't know anything about topology of mesh. So this algorithm observe this mesh as cloud of points with max concentration near the center_of_mass (not geometric center. geometric center do not point to the center of mass).
image
Radius has no specific value now but this is not need for this algorithm. It is very simple for now to select what plain to bisect first plain12 or plane23 when process site1? (Do you see this is not need to know the radius value?) Now obvious that bisect plane with center on negative side will remove more points from mesh geometry than plane with center on positive side. So plane12 is better plane13. So we simplify next bisects because they will be work with less geometry.
If there is two planes both with center on negative side then need first select a plane with bigger distance from center to plane:
image
here is the plane14 is better the plane12.
Positive side of planes has the same rule about distance. First plane has bigger distance from center to plane to remove maximum geometry on bisect.

**3. test is spacing is more than all dists from planes to Site.
if all distance from planes to site is less than spacing/2 then bisect will cutoff all vertices anyway:
image
so one can to skip all bisect and return nothing for this site.

satabol:
Result
-------
image

utils/voronoi3d.py
```diff
@@ -236,81 +238,271 @@ def calc_bvh_projections(bvh, sites):
...
+        rrc += means[:, np.newaxis]
+        rrc = rrc.transpose()
+        return rrc
+
     def get_ridges_per_site(voronoi):
         result = defaultdict(list)
+        sites_list = []
         for ridge_idx in range(len(voronoi.ridge_points)):
             site1_idx, site2_idx = tuple(voronoi.ridge_points[ridge_idx])
+            if (voronoi.points.shape[0]-1)<site1_idx:
```
portnov, line 271: IMHO a comment is needed here.

utils/voronoi3d.py
```diff
@@ -236,81 +238,271 @@ def calc_bvh_projections(bvh, sites):
...
-            #print(f"Plane co {plane_co}, no {plane_no}, signs {signs}")
-            if (signs <= 0).all():# or (signs <= 0).all():
-                continue
+    def get_sites_delaunay_params(delaunay):
+        result = defaultdict(list)
+        ringes = []
+        sites_pair = dict()
+        for simplex in delaunay.simplices:
+            site1_idx, site2_idx, site3_idx, site4_idx = tuple( sorted( [i for i in simplex] ) )
+            ringes+= [tuple( [site1_idx, site2_idx] ),
```
portnov, line 292: 1) what is "ringes"? did you mean "rings"? or "ridges"?

2) ringes += itertools.combinations(site_idxs, 2)

utils/voronoi3d.py
```diff
@@ -236,81 +238,271 @@ def calc_bvh_projections(bvh, sites):
...
-            signs = PlaneEquation.from_normal_and_point(plane_no, plane_co).side_of_points(current_verts)
-            #print(f"Plane co {plane_co}, no {plane_no}, signs {signs}")
-            if (signs <= 0).all():# or (signs <= 0).all():
-                continue
+    def get_sites_delaunay_params(delaunay):
+        result = defaultdict(list)
+        ringes = []
+        sites_pair = dict()
+        for simplex in delaunay.simplices:
+            site1_idx, site2_idx, site3_idx, site4_idx = tuple( sorted( [i for i in simplex] ) )
```
portnov, line 291: Do we really need [i for i in simplex]? why not simply `simplex`?

utils/voronoi3d.py
```diff
@@ -236,81 +238,271 @@ def calc_bvh_projections(bvh, sites):
...
-                surround = [e for e in res['geom_cut'] if isinstance(e, bmesh.types.BMEdge)]
-                if surround:
-                    fres = bmesh.ops.edgenet_prepare(src_mesh, edges=surround)
-                    if fres['edges']:
-                        bmesh.ops.edgeloop_fill(src_mesh, edges=fres['edges'])
-
-        if n_cuts == 0:
+        return result
+
+    summ_delta_time1 = datetime.datetime.now()-datetime.datetime.now()
```
portnov, line 315: what is this?

utils/voronoi3d.py
satabol: This code used for exclude some additional points I added to voronoi on tests and tried to except its from source point. This function "**get_ridges_per_site(voronoi)**" is not using now so this function can be removed.

utils/voronoi3d.py
satabol: 1. Ridges of Couse! :D (fixed any places with that errata)
2. I never used it! Nice. Fixed. (Commit after all finish)

utils/voronoi3d.py
satabol: You are right. I remove inline for in **sorted(...)**

utils/voronoi3d.py
satabol: Fix datetime.timedelta()

satabol:
All vector expressions aligned to direction to site_idx from site_pair_idx

PR #4955, opened 2023-07-20
normally call unregister SV_PT_ToolsMenu on exit blender

satabol:
Sverchok 1.2.0-alpha master, windows 11, Blender any version (3+).

In dev mode I get next exception and Sverchok do not start:

image
image

This appear over 1 time after good start blender - one time Blender/sverchok start good and next time start failure. Then start is good again. So I restart Blender dev mode every second time.

After fix no error on start blender and sverchok:
image


satabol:
Can somebody review this little fix?

nortikin:
seems like it works on linux. But had no such trouble before

satabol:
Thanks for test.

PR #4958, opened 2023-07-22
fix torus function torus_verts performance.

satabol:
Out some calculations from inner for.

image

Sverchok master, Windows 11, Blender 3.6.1

(manual measurement)
image

Performance up 1.75x.


Durman (approved):

Durman:
Is 1.75 boost of the particular loop or the whole work of the node?

satabol:
1.75 is for “for”. It is hard to compute so many sin, cos, and power. I try calc numpy. It can be speed up 1.5x. But it is very hard read that code.

Durman:
Yes, numpy is a functional programming. It is something to which someone have to get used. By the way Sverchok itself also is sort of functional programming. =)

PR #4963, opened 2023-07-25
torus_mk2. performance. Convert function 'sign' in formulas as inline function

satabol:
Sverchok master, Windows 11, Blender 3.6.1

image

235ms no optimisation

1 - 1x (111ms-104ms) (not usign sign function before and after optimization)
0 - 1.78x (212ms-119ms)
any - 1.65x (246ms-149ms)

with using of viewer node:
image

So, optimization is 1.5x-2.3x if do 'sign' function as inline call.


satabol:
Append cache for edges and faces for some geometry params. It can speed up some cases.

PR #4966, opened 2023-07-28
VeiwerDraw_mk4. Optimize some list extend operation by numpy (Performance)

satabol:
Sverchok master

image

### DIFF
--- a/nodes/viz/viewer_draw_mk4.py
+++ b/nodes/viz/viewer_draw_mk4.py
@@ -7,6 +7,7 @@
 
 
 from itertools import cycle
+import numpy as np
 
 from mathutils import Vector, Matrix
 from mathutils.geometry import tessellate_polygon as tessellate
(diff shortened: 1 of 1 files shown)


nodes/viz/viewer_draw_mk4.py
```diff
@@ -362,7 +363,8 @@ def polygons_geom(config, vecs, polygons, p_vertices, p_vertex_colors, p_indices
             if not config.uniform_pols:
                 p_vertex_colors.extend([p_cols for v in v_path])
         if idx_p_offset[0]:
-            p_indices.extend([[c + idx_p_offset[0] for c in p] for p in polygon_indices])
+            #p_indices.extend([[c + idx_p_offset[0] for c in p] for p in polygon_indices])
+            p_indices.extend( (np.array( polygon_indices )+idx_p_offset[0]).tolist() )
```
Durman, line 367: Will this work when polygons have different number of verts?

satabol: I don’t think so. I will think about new trick. ))) But this place too slow.

Issue #4979, opened 2023-08-14
Add output socket 'Sites' for node "Voronoi On Mesh"


satabol:
(for Pull Request)

Not all sites are cells on mesh.

![image](https://github.com/nortikin/sverchok/assets/14288520/143fa6a7-c1b8-45e4-ae5e-5328fa5b9d3a)

satabol:
See: http://nortikin.github.io/sverchok/docs/nodes/spatial/voronoi_on_mesh_mk2.html

PR #4980, opened 2023-08-14
Calculation of used sites for node "Voronoi on Mesh". 

satabol:
#4979 

Inspired by https://www.youtube.com/watch?v=Ip6JI6Qiiwg but more precise.

Some issue is in this video-tutorial - all emitted points has to be applied to source mesh. If no then result will have glitches.
Not all source sites converts to Voronoi cells. This Pull Request take capability to get only used sites:

image

Voronoi Tower 001
[Voronoi Tower.001.blend.zip](https://github.com/nortikin/sverchok/files/12332957/Voronoi.Tower.001.blend.zip)

**Is it sense to add Mask or Indices of used sites?**


portnov:
1. I'd say it is more usual for Sverchok nodes to output masks, rather than subsets of vertices.
2. What if node already exists in the scene, but does not yet have this new output? maybe this is not a problem since the new version of the node was introduced very recently.

satabol:
1. I'll do indices
2. Code to check new output socket exists in the "process" function. But I think a new version is better. )

satabol:
@portnov this is new version of file for tests: [Voronoi Tower.003.blend.zip](https://github.com/nortikin/sverchok/files/12339718/Voronoi.Tower.003.blend.zip) (press play for animation)

image

portnov:
Sites_idx should be a SvStringSocket then?

satabol:
What do you think about property to switch type socket to a vector or an index for Site_idx? With rename socket to “Site” and show this additional property in “N-panel”.

satabol:
Fix socket type to SvStringsSocket:
image

portnov:
ok with me.

Issue #4959, opened 2023-07-24
Populate Mesh node - infinite loop


Durman:
## Problem statement

When size of the box is 600 and more the Populate Mesh node falls into infinite loop.

![image](https://github.com/nortikin/sverchok/assets/28003269/e4fe479e-dea9-4cea-932f-35e6a2d59990)

The problems seems is somewhere here:

https://github.com/nortikin/sverchok/blob/44e611da468d60c3134fa8d31634d8b1c1907b00/utils/sv_mesh_utils.py#L376-L389

PR #4981, opened 2023-08-14
fix 4959. Dot product can be used to realize side of mesh.

satabol:
Sverchok 1.3-alpha, Blender 3.6.1,  windows 11.

#4959 

Description of problem:
https://github.com/nortikin/sverchok/blob/44e611da468d60c3134fa8d31634d8b1c1907b00/utils/sv_mesh_utils.py#L376-L389

Core of problem is line 386:
https://github.com/nortikin/sverchok/blob/44e611da468d60c3134fa8d31634d8b1c1907b00/utils/sv_mesh_utils.py#L386-L386

In some cases "point == location + axis * 0.00001"

Solution for box with size=600:

image

### DIFF
--- a/docs/nodes/spatial/random_points_on_mesh.rst
+++ b/docs/nodes/spatial/random_points_on_mesh.rst
@@ -39,7 +39,8 @@ This node has the following parameters:
 
   * **Surface**. Generate points on the surface of the mesh.
   * **Volume**. Generate points inside the volume of the mesh. The mesh is
-    expected to represent a closed volume in this case.
+    expected to represent a closed volume in this case. Recomend use 
+    [Modifiers->Modifier Change-> :doc:`Recalculate Normals </nodes/modifier_change/recalc_normals>`] before node.
   * **Edges**.  Generate points on the edges of the mesh.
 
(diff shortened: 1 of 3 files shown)


satabol:
if old code do not check mesh for closed and self intersections (and other validity) so drill faces is a not precise method. May be test normal is enough precise and safe?

satabol:
if old code do not check mesh for closed and self intersections (and other validity) so drill faces is a not precise method. May be test normal is enough precise and safe?

@portnov What do you think about this method?

portnov:
I'm not sure. Probably better to test both approaches on different meshes and compare.

satabol:
What to do if comparison are different?

portnov:
select the method which gives better results, i think.

satabol:
Mmm. How to test what results are better?

portnov:
create several meshes of different forms (for example concave ones). Supposedly, on some forms one algorithm will work fine and another will not. Then the algorithm which works for more forms is better :)

Durman (approved):

satabol:
image

Issue #4961, opened 2023-07-25
Some Sverchok nodes do not work in Blender 4.0 Alpha (Related to #4915) 


meikasahara:
Sverchok can now be installed on Blender 4.0 but the Stethoscope and the Viewer Draw nodes do not work. The following errors appear in the console: 
```

Traceback (most recent call last):
  File "/home/me/.config/blender/4.0/scripts/addons/sverchok-master/ui/bgl_callback_nodeview.py", line 148, in draw_callback_px
    draw_text_data(data)
  File "/home/me/.config/blender/4.0/scripts/addons/sverchok-master/utils/sv_stethoscope_helper.py", line 40, in draw_text_data
    blf.size(font_id, int(text_height), 72)
TypeError: blf.size() takes exactly 2 arguments (3 given)
```

and

```
Traceback (most recent call last):
  File "/home/diffeomorphist/.config/blender/4.0/scripts/addons/sverchok-master/ui/bgl_callback_3dview.py", line 83, in draw_callback_px
    bgl.glEnable(bgl.GL_DEPTH_TEST)
TypeError: 'NoneType' object is not callable

```

Durman:
Yes, this is expected, here is the task to tackle it - #4895

PR #4983, opened 2023-08-15
Bgl replacement (and other 3.5 / 4.0 curiosities)

zeffii:
alternative bgl abstraction.

continuation of https://github.com/nortikin/sverchok/pull/4936

- [x] remove bgl from current node set and modules/utils
- [x] adjust `blf.text` to handle 3 vs 2 parameters
- [ ] implement texture/image loading for texture/buffer using nodes (texture viewers and console node)
   - https://docs.blender.org/api/4.0/gpu.types.html
   - affected nodes ( TextViewer, TVLite, Stethoscope ( nice mode ), Console Node ) , These are not the most
     essential nodes but 
        - "nice mode" for `stethoscope` is pretty cool and if I get that to work then console node will also work
    automatically (they share the same code)
        - `Texture Viewer` nodes use some Gl code to place the texture in a state to receive data which i'm only tangentially 
          familiar with :)
- [x] bug fix front_facing polygon drawing thru
   - ```lines are not being obscured by polygons.```
- [x] polygon offset alternative
   -  [x] ```avoid z-fighting``` ( proof of concept: at the moment implements a normal shader in custom-shader mode )
   -  [ ] reimplement custom shader
   -  [ ] implement shader for flat/facet/smooth 

Node | works in < 4.0 | works in > 4.0 | feature parity | issue
 --- | --- | --- | --- | ---
Viewer Draw | yes | yes | no | zfighting lines
Stethoscope MK2 | yes | yes | no | advanced drawing to buffer not implemented, no colour highlighting 
Matrix View | yes | yes | maybe | plane seems to always draw
Viewer Index | yes | yes | yes | ...
Viewer 2D | yes | yes | yes | ...
Viewer Draw Curve | yes | yes | yes | ...
Viewer Draw Surface | yes | yes | yes | ...
Texture Viewer | yes | ? | ? | ?
Texture Viewer Lite | yes | ? | ? | ?
SvWaveformViewer | yes | yes | yes | ...
Console Node | yes | ? | ? | ?
Solid Viewer (did not test) | yes | ? | ? | ?
Easing 0..1 | yes | yes | yes | ...

### DIFF
--- a/utils/modules/drawing_abstractions.py
+++ b/utils/modules/drawing_abstractions.py
@@ -49,19 +49,23 @@ if bpy.app.version >= (3, 5, 0):
     class Drawing:
 
         set_wireframe_line = placeholder_function
-        set_wireframe_fill = placeholder_function
+
+        def set_wireframe_fill(self):
+            gpu.state.face_culling_set("FRONT")
+            gpu.state.front_facing_set(False)
+
         set_line_width = gpu.state.line_width_set
         reset_line_width = lambda self: gpu.state.line_width_set(1)
         set_point_size = gpu.state.point_size_set
         reset_point_size = lambda self: gpu.state.point_size_set(1)
-        enable_polygon_offset_fill = placeholder_function
+        enable_polygon_offset_fill = lambda self: gpu.state.face_culling_set("FRONT")
         disable_polygon_offset_fill = placeholder_function
         set_polygon_offset_amounts = placeholder_function
 
-        enable_blendmode = placeholder_function
-        disable_blendmode = placeholder_function
-        enable_depth_test = placeholder_function
-        disable_depth_test = placeholder_function
+        enable_blendmode = lambda self: gpu.state.blend_set("ALPHA")
+        disable_blendmode = lambda self: gpu.state.blend_set("NONE")
+        enable_depth_test = lambda self: gpu.state.depth_test_set("ALWAYS")
+        disable_depth_test = lambda self: gpu.state.depth_test_set("NONE")
 
         new_buffer_texture = placeholder_function
         get_buffer = placeholder_function
--- a/utils/modules/drawing_abstractions.py
+++ b/utils/modules/drawing_abstractions.py
@@ -91,9 +91,11 @@ class OldDrawing:
     delete_texture = lambda texture: bgl.glDeleteTextures(1, texture)
 
     def init_complex_texture(self, width, height, texname, texture, data, format):
-        if format == 'RGBA':
-            format = bgl.GL_RGBA
+
+        format = {'BW': bgl.GL_RED, 'RGB': bgl.GL_RGB, 'RGBA': bgl.GL_RGBA}.get(format)
+
         texture = self.new_buffer_texture_sized(bgl.GL_FLOAT, data.size, data.tolist())
+
         bgl.glPixelStorei(bgl.GL_UNPACK_ALIGNMENT, 1)
         bgl.glEnable(bgl.GL_TEXTURE_2D)
         bgl.glBindTexture(bgl.GL_TEXTURE_2D, texname)
--- a/utils/modules/drawing_abstractions.py
+++ b/utils/modules/drawing_abstractions.py
@@ -69,6 +69,7 @@ if bpy.app.version >= (3, 5, 0):
 
         new_buffer_texture = placeholder_function
         get_buffer = placeholder_function
+        new_buffer_texture_sized = lambda self, size, data: gpu.types.Buffer("FLOAT", size, data)
         bind_texture_2d = placeholder_function
         init_complex_texture = placeholder_function
         generate_textures = placeholder_function
--- a/utils/modules/drawing_abstractions.py
+++ b/utils/modules/drawing_abstractions.py
@@ -51,20 +51,20 @@ if bpy.app.version >= (3, 5, 0):
         set_wireframe_line = placeholder_function
 
         def set_wireframe_fill(self):
-            gpu.state.face_culling_set("FRONT")
-            # gpu.state.front_facing_set(False)
+            gpu.state.face_culling_set("BACK")
+            gpu.state.front_facing_set(False)
 
         set_line_width = gpu.state.line_width_set
         reset_line_width = lambda self: gpu.state.line_width_set(1)
         set_point_size = gpu.state.point_size_set
         reset_point_size = lambda self: gpu.state.point_size_set(1)
-        enable_polygon_offset_fill = lambda self: gpu.state.face_culling_set("FRONT")
+        enable_polygon_offset_fill = lambda self: gpu.state.face_culling_set("BACK")
         disable_polygon_offset_fill = placeholder_function
         set_polygon_offset_amounts = placeholder_function
 
         enable_blendmode = lambda self: gpu.state.blend_set("ALPHA")
         disable_blendmode = lambda self: gpu.state.blend_set("NONE")
-        enable_depth_test = lambda self: gpu.state.depth_test_set("ALWAYS")
+        enable_depth_test = lambda self: gpu.state.depth_test_set("LESS_EQUAL")
         disable_depth_test = lambda self: gpu.state.depth_test_set("NONE")
 
         new_buffer_texture = placeholder_function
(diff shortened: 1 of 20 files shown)


utils/modules/drawing_abstractions.py
```diff
@@ -0,0 +1,71 @@
+import bpy
+
+shading_3d = lambda: None
+
+if bpy.app.version <= (3, 4):
+    shading_3d.UNIFORM_COLOR = "3D_UNIFORM_COLOR"
+    shading_3d.SMOOTH_COLOR = "3D_SMOOTH_COLOR"
+else:
+    shading_3d.UNIFORM_COLOR = "UNIFORM_COLOR"
+    shading_3d.SMOOTH_COLOR = "SMOOTH_COLOR"
```
Durman, line 10: Alternative:
```py
from typing import NamedTuple

class shading_3d(NamedTuple):
    if bpy.app.version <= (3, 4):
        UNIFORM_COLOR = "3D_UNIFORM_COLOR"
        SMOOTH_COLOR = "3D_SMOOTH_COLOR"
    else:
        UNIFORM_COLOR2 = "UNIFORM_COLOR"
        SMOOTH_COLOR2 = "SMOOTH_COLOR"
```

utils/modules/drawing_abstractions.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+drawing = lambda: None
+
+
+if bpy.app.version >= (3, 5, 0):
+    drawing.set_wireframe_line = pass
+    drawing.set_wireframe_fill = pass
```
Durman, line 18: Just an alternative:
```py
class OldDrawing:
    def set_wireframe_line(self): bgl.glPolygonMode(bgl.GL_FRONT_AND_BACK, bgl.GL_LINE)
    def set_wireframe_fill(self): bgl.glPolygonMode(bgl.GL_FRONT_AND_BACK, bgl.GL_FILL)

class Drawing:
    def set_wireframe_line(self): pass
    def set_wireframe_fill(self): pass

drawing = Drawing() if bpy.app.version >= (3, 5, 0) else OldDrawing()
```

utils/modules/drawing_abstractions.py
zeffii: yes

zeffii:
image

zeffii:
;ossibly i should name `set_wireframe_line` to `wireframe_true`,   and `set_wireframe_fill` to `wireframe_false`  -- because wireframe_fill is used to reset the GL state if the user has picked the option to display wireframe on the vdmk4 ( and wireframe would be the triangles version of the polygons... not the edges as fed into the node by the user in the edges socket)

zeffii:
resolves: https://github.com/nortikin/sverchok/issues/4961

zeffii:
i'm not sure i'll be around to finish this PR, but here is one solution to the removal of the OffsetPolygon feature. We push polygons along their normal, by some small amount, from within a Geometry Shader.  I'm not certain that Blender supports Geometry Shaders, however. (fact check?)

https://www.informit.com/articles/article.aspx?p=2120983&seqNum=2

according to this each Geometry Shader call will have a default variable called `gl_in` which holds (in our case) 3 vectors. Our input primitives are all triangles. .and we can emit new vertices for each of these.

```glsl

// Calculate two vectors in the plane of the input triangle
vec3 ab = gl_in[1].gl_Position.xyz - gl_in[0].gl_Position.xyz;
vec3 ac = gl_in[2].gl_Position.xyz - gl_in[0].gl_Position.xyz;
vec3 normal = normalize(cross(ab, ac));
```
and
```glsl

for (i = 0; i < gl_in.length(); i++)
{
    gl_Position = gl_in[i].gl_Position;
    EmitVertex();
}
//EmitPrimitive();  no need for this is you only emit one primitive 
EndPrimitive();
```

zeffii:
hmmm... `shader_info = gpu.types.GPUShaderCreateInfo()`  does not have a `geometry_shader`.

```python
>>> shader_info = gpu.types.GPUShaderCreateInfo()
>>> shader_info.
                define(
                fragment_out(
                fragment_source(
                push_constant(
                sampler(
                typedef_source(
                uniform_buf(
                vertex_in(
                vertex_out(
                vertex_source(
```

zeffii:
https://docs.blender.org/api/current/gpu.types.html#gpu.types.GPUShader

here it does appear we can feed a geometry shader.
```
Parameters
vertexcode (str) – Vertex shader code.

fragcode – Fragment shader code.

geocode – Geometry shader code.

libcode – Code with functions and presets to be shared between shaders.

defines – Preprocessor directives.

name – Name of shader code, for debugging purposes.
```

zeffii:
going down the geometry shader rabbit hole https://stackoverflow.com/questions/14909796/simple-pass-through-geometry-shader-with-normal-and-color

zeffii:
this branch doesn't work in the lastest 4.0 alphas @Durman   `bpy.types.NodeSocketInterface`  doesn't exist anymore?

zeffii:
i guess this technique kind of fails when polygons are facing away from the viewer, therefore it would need to flip the normal in that case.

zeffii:
image

HARRR!!!

Durman:
> `bpy.types.NodeSocketInterface` doesn't exist anymore?

I guess it was renamed - https://docs.blender.org/api/4.0/bpy.types.NodeTreeInterfaceSocket.html#bpy.types.NodeTreeInterfaceSocket

But it seems is not the only change, there is new attribute for setting tree interface - https://docs.blender.org/api/4.0/bpy.types.NodeTree.html#bpy.types.NodeTree.interface

satabol:
Hi @zeffii !
I try this PR in Blender 4.0.1. It is not start, but I found some changes in API:

1. bpy.types.NodeSocketInterface -> bpy.types.NodeTreeInterfaceSocket

2. 
image

Now I can start this PR and it work now:

image

Your work are very awaited !!! )))

zeffii:
This branch does not yet reach feature parity, but i'm interested to see how it behaves in the wild. If anyone feels like merging this as is, I will be more likely to keep updating the 4.0+ parts (to solve the zfighting in solid/facetted/smooth modes)

zeffii:
> How can we enforse or help with tasks?

my primary goal was to get SV loading regardless of the Blender version, which it does seem to do now. Some graphical glitches (like zfighting edges/polygons) will occur in 4.0+ until a dedicated shader is implemented for 4.0+, but it needs to be executed only when edges and polygons are displayed at the same time)

it's not a big task, I purely haven't sat at a computer for much of the past month.

nortikin:
image
In 3.6 there is something backfaced

zeffii:
Weird. I probably didn't test on 3.6.. This is not good.

gerroon:
Is this merged?

nortikin:
```File "C:\Users\***\AppData\Roaming\Blender Foundation\Blender\4.1\scripts\addons\sverchok\utils\sv_3dview_tools.py", line 97, in sv_execute                                                                                                bpy.ops.view3d.view_center_cursor(ctx)                                                                                File "D:\users\***\INSTALL\blender-4.1.1-windows-x64\4.1\scripts\modules\bpy\ops.py", line 106, in __call__        C_exec, C_undo = _BPyOpsSubModOp._parse_args(args)                                                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                    File "D:\users\***\INSTALL\blender-4.1.1-windows-x64\4.1\scripts\modules\bpy\ops.py", line 60, in _parse_args      raise ValueError("1-2 args execution context is supported")                                                         ValueError: 1-2 args execution context is supported```

nortikin:
image
21:23:25.488 [ERROR] sverchok.node_error vertex_utils.py:141 - setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (34,) + inhomogeneous part.

nortikin:
4/1image
this

nortikin:
image
3.6 somehow backface culling, as in 4.1
with sverchok master blender 3.4:
image

portnov:
This branch seems to work for me with blender 3.4.1 and 4.2.1. I do not see problems, but maybe I don't know where to look?

portnov:
Eh, on 4.2 the mesh is shown transparent from inside out: 
Screenshot_20240831_150423
Still, it's better than nothing.

PR #4999, opened 2023-09-21
fix #3288 "Extrude region node - keep original"

### LINKED ISSUE #3288: Extrude region node - keep original
Keep original property does not have effect if all faces of given mesh are extruded by the node.

extrude bug

satabol:
Sverchok 1.3.0-alpha, Blender 3.6.1, Windows 11

- fix #3288
- Create library for function separate_loose_mesh

Result is good:
image

File for tests: [issue_3288.example.blend.zip](https://github.com/nortikin/sverchok/files/12693821/issue_3288.example.blend.zip)

### DIFF
--- a/nodes/modifier_change/extrude_region.py
+++ b/nodes/modifier_change/extrude_region.py
@@ -27,6 +27,8 @@ from sverchok.node_tree import SverchCustomTreeNode
 from sverchok.data_structure import updateNode, match_long_repeat, repeat_last_for_length
 from sverchok.utils.sv_bmesh_utils import bmesh_from_pydata, pydata_from_bmesh
 from sverchok.utils.nodes_mixins.sockets_config import ModifierNode
+from sverchok.utils.mesh.separate_loose_mesh import separate_loose_mesh
+import numpy as np
 
 is_290 = bpy.app.version >= (2, 90, 0)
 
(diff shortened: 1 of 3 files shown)


satabol:
Hi @portnov! Can you check it? Some code are commented. I remove it later if you approve it.

portnov:
If you think the problem is in Blender code, I'd suggest to file an issue against Blender, and add a link in code comment to that issue and write that you are doing a workaround for it.
Otherwise, no objections.

satabol:
Hi @portnov. Is it good to copy comment from Issue to here?

This is very interesting problem. I found some cases where it happens:
1. Your case with one face but there is more general case: with all extruded faces in mesh:
2. If all faces are extruded then property "Keep Original" is off has no effect:
image
Some results are good. For example ("Keep original" is off), if 4-th face if excluded from extrude then result looks good:
image

but in other cases result is not good ("Keep original" is off), if 0 and 1 faces are excluded from extrude then result is not good:
image

So why this happens?
Lets look at blender source code for function "bmesh.ops.extrude_face_region":

image

So this is problem of Blender. Shame. What solution can be here? As you can see if exists a face in shape that is not extruded then result is good:

image

So... we cannot exclude some faces from extrude... but one can append a face to every shape then exclude it from extrude then remove new faces after extrude!

image

And result is good:

image

PR #5000, opened 2023-09-22
Viewer Index+ show right text with multiple objects and indices of objects

### LINKED ISSUE #4705: Viewer Index+ show wrong text when use text input in edge and face mode with connect more than 1 object
Viz->Viewer Index+ has a 'text' input. When I connect text in vertices mode, then all look good:

image
I use mode "Draw Object Index" and all objects Id are equal. Good.

But when I use that trick in edge mode and face mode then result is wrong:

Edge mode (objects id are not equal):
image

Face mode (objects id are not equal):
image

satabol:
Sverchok 1.3.0-alpha, Blender 3.6.1, Windows 11

fix #4705

Now Modes are good:

1. Only indices for vertices, edges, faces without object indices
image
image
image

2. Indices for vertices, edges, faces with (!!!) object indices
image
image
image

3. Only text for vertices, edges, faces without object indices
image
image
image)

4. Indices for vertices, edges, faces with (!!!) object indices
image
image
image

Blend file for tests:

### DIFF
--- a/nodes/viz/viewer_idx28.py
+++ b/nodes/viz/viewer_idx28.py
@@ -217,9 +217,11 @@ class SvIDXViewer28(SverchCustomTreeNode, bpy.types.Node):
         fixed_text = []
         if geom.text:
             for obj_index, final_verts in enumerate(geom.verts):
+                fixed_text_elem = []
                 text_size = max(len(final_verts),
                                 len(geom.edges[obj_index] if geom.edges else []),
                                 len(geom.faces[obj_index] if geom.faces else []))
+
                 text_items = self.get_text_of_correct_length(obj_index, geom, text_size)
                 for text_item in text_items:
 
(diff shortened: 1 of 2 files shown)


satabol:
Hi @zeffii ! Can you test this pull request? I force to work this node in a mode with multiple objects with custom text and property "Draw Object Index" is On. But I think this node need global refactor ))) (not now) 

P.S.
Also I simplify using shaders to single shader **draw_indices_2D_wbg** without **draw_indices_2D**. **draw_indices_2D** is very simple and differ only by background rectangle. May be **draw_indices_2D** is not need any more?

satabol:
All was hidden with that configuration of params:
image
Fixed.

zeffii:
the ghost icon-button is for hiding backfacing indices, so only really works when user passes polygons too.

satabol:
Get this glitch after exclude **draw_indices_2D_wbg**.

satabol:
Hi, @zeffii! May I ask you to view this PR?

zeffii:
@satabol  sorry , i've been traveling and have not been spending much time near computers.

zeffii:
i don't see anything i disagree with, but would prefer if #4983 was pushed first.

satabol:
Ok @zeffii. No problem. I will expect it. )))

PR #5022, opened 2023-10-16
fix #5019 extend node "get object data" with options "apply matrix" and "join mesh" (merge)

### LINKED ISSUE #5019: Extend "Get Object Data" with option "Apply matrix" and "Join mesh"
Sverchok 1.3.0-Alpha, Blender 3.6.x, Windows 11.

A proposal to simplify operate with scene objects:

image

satabol:
fix #5019. 

- "Apply matrix" applied only to vectorized params vertexes/normals. Not for indexes or areas.
- "Merge" combine vertexes, edges, and faces. Do not affect to source objects matrixes, materials indexes.

- [ ] need code review or approve. Ready to merge.

**File for tests**:
[fix_5019_apply_matrix_join_mesh.source_data.v001.blend.zip](https://github.com/nortikin/sverchok/files/12919075/fix_5019_apply_matrix_join_mesh.source_data.v001.blend.zip)

image

**Apply matrix**:
image

**Merge**:

**Only merge**:
image

**Apply Matrix and Merge**:
image


zeffii:
pitty upgrading the node using a new file loses the `diff` information to clearly see the changes in the code.

nodes/scene/get_objects_data.py
```diff
@@ -296,24 +317,24 @@ def process(self):
                     # from 3dview while in edit mode when using obj.to_mesh.
                     me = obj.data
                     bm = bmesh.from_edit_mesh(me)
-                    vers, edgs, pols = pydata_from_bmesh(bm)
+                    # verts, edgs, pols = pydata_from_bmesh(bm)
 
                     if o_vs:
-                        vs.append(vers)
+                        verts = [ Vector(v.co) for v in bm.verts]  # v.co is a Vector()
```
zeffii, line 323: my only surprise here is that `verts` will be holding a list of Vectors -- obviously in anticipation for the 'if self.apply_matrix` being true.  from what I can tell .. if apply_matrix is false, then the node will be outputting lists of Vector lists into the node tree.

zeffii: > Hi @zeffii . I puzzle out now! It is convenient to rename a file node after move it to old/nodes! )))

much better !

satabol: >> if apply_matrix is false, then the node will be outputting lists of Vector lists into the node tree
Hi @zeffii . I agree with you. Fix it.

satabol: > if apply_matrix is false, then the node will be outputting lists of Vector lists into the node tree
Hi @zeffii. I agree with you. I fixed it.

PR #5050, opened 2023-11-05
fix #5049. Logic Functions cannot recreate node

### LINKED ISSUE #5049: Logic Functions cannot recreate node
Sverchok 1.3.0-Alpha, Blender 3.6.x

If try to use recreate node function of sverchok then get exception:

image

satabol:
fix #5049. Logic Functions cannot recreate node. Fixed

image

### DIFF
--- a/nodes/logic/logic_node.py
+++ b/nodes/logic/logic_node.py
@@ -118,13 +118,47 @@ class SvLogicNodeMK2(SverchCustomTreeNode, bpy.types.Node):
         layout.prop(self, "output_numpy", expand=False)
 
     def migrate_from(self, old_node):
-        self.function_name = old_node.items_
-        self.inputs['A'].default_int_property = old_node.x
-        self.inputs['B'].default_int_property = old_node.y
-        self.inputs['A'].default_float_property = old_node.i_x
-        self.inputs['B'].default_float_property = old_node.i_y
-        self.inputs['A'].default_property_type = 'float' if old_node.prop_types[0] else 'int'
-        self.inputs['B'].default_property_type = 'float' if old_node.prop_types[1] else 'int'
+        old_node_input_A = None
+        if 'A' in old_node.inputs:
+            old_node_input_A = old_node.inputs['A']
+        old_node_input_B = None
+        if 'B' in old_node.inputs:
+            old_node_input_B = old_node.inputs['B']
+        if hasattr(old_node, 'items_'):
+            self.function_name = old_node.items_
+        if hasattr(old_node, 'x'):
+            self.inputs['A'].default_int_property = old_node.x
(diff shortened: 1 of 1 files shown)


nodes/logic/logic_node.py
```diff
@@ -118,13 +118,47 @@ def rclick_menu(self, context, layout):
...
-        self.function_name = old_node.items_
-        self.inputs['A'].default_int_property = old_node.x
-        self.inputs['B'].default_int_property = old_node.y
-        self.inputs['A'].default_float_property = old_node.i_x
-        self.inputs['B'].default_float_property = old_node.i_y
-        self.inputs['A'].default_property_type = 'float' if old_node.prop_types[0] else 'int'
-        self.inputs['B'].default_property_type = 'float' if old_node.prop_types[1] else 'int'
+        old_node_input_A = None
+        if 'A' in old_node.inputs:
+            old_node_input_A = old_node.inputs['A']
```
Durman, line 123: Just shorter alternative.
`old_node_input_A = old_node.inputs.get('A')`

nodes/logic/logic_node.py
```diff
@@ -118,13 +118,47 @@ def rclick_menu(self, context, layout):
         layout.prop(self, "output_numpy", expand=False)
 
     def migrate_from(self, old_node):
```
Durman, line 120: I think I would prefer like this:
```py
def migrate_from(self, old_node):
    if old_node.bl_idname == 'SvLogicNodeMK2':
        return
```

satabol: Ok, fixed.

Durman: I meant that this is the only changes which are required for the node to work properly.

zeffii:
merge this if it works

nodes/logic/logic_node.py
```diff
@@ -118,13 +118,46 @@ def rclick_menu(self, context, layout):
...
+                self.inputs['A'].default_property_type = old_node_input_A.default_property_type
+            self.inputs['A'].use_prop = True # like sv_init
+            self.inputs['A'].show_property_type = True
+
+            if hasattr(old_node, 'prop_types') and len(old_node.prop_types)>1:
+                self.inputs['B'].default_property_type = 'float' if old_node.prop_types[1] else 'int'
+            elif old_node_input_B and hasattr(old_node_input_B, 'default_property_type'):
+                self.inputs['B'].default_property_type = old_node_input_B.default_property_type
+            self.inputs['B'].use_prop = True # like sv_init
+            self.inputs['B'].show_property_type = True
```
zeffii, line 160: wow, these 40 lines are painful :)

PR #5052, opened 2023-11-06
fix #5051. node Mesh Viewer show exception face new already exists.

### LINKED ISSUE #5051: node Mesh Viewer show exception face new already exists
Sverchok 1.3.0-alpha, Blender 3.6.x.

image

File for tests: 
[fix_5051_node_Mesh_Viewer_show_exception_face_new_is_exists.001.blend.zip](https://github.com/nortikin/sverchok/files/13271627/fix_5051_node_Mesh_Viewer_show_exception_face_new_is_exists.001.blend.zip)

satabol:
fix #5051 . node Mesh Viewer show exception face new is exists. Fixed

Source of problem. Some faces try same vertices but another sequence:
image

image

file for tests:
[fix_5051_node_Mesh_Viewer_show_exception_face_new_is_exists.001.blend.zip](https://github.com/nortikin/sverchok/files/13271656/fix_5051_node_Mesh_Viewer_show_exception_face_new_is_exists.001.blend.zip)

### DIFF
--- a/utils/sv_bmesh_utils.py
+++ b/utils/sv_bmesh_utils.py
@@ -162,7 +162,12 @@ def add_mesh_to_bmesh(bm, verts, edges=None, faces=None, sv_index_name=None, upd
         new_edges = []
     #[new_face([bm_verts[i] for i in face]) for face in faces or []] # не надо так делать
     if faces is not None:
-        for face in faces:
+        # remove double faces. Some times get faces with same verts but with another order: [62,103,102] and [102,103,62].
+        # this is not compatible with mesh and only one of that faces can be used. So convert this faces indices into dictionary.
+        # this remove faces with the same vertices but same or different orders. This may happens on some of Voronoi if
+        # it slice self intersected mesh.
+        dict_faces = { tuple(sorted(f)): f for f in list(faces) }
+        for face in dict_faces.values():
             new_face([bm_verts[i] for i in face])
     else:
         new_face = []


nortikin:
All works

utils/sv_bmesh_utils.py
```diff
@@ -162,7 +162,12 @@ def add_mesh_to_bmesh(bm, verts, edges=None, faces=None, sv_index_name=None, upd
         new_edges = []
     #[new_face([bm_verts[i] for i in face]) for face in faces or []] # не надо так делать
     if faces is not None:
-        for face in faces:
+        # remove double faces. Some times get faces with same verts but with another order: [62,103,102] and [102,103,62].
+        # this is not compatible with mesh and only one of that faces can be used. So convert this faces indices into dictionary.
+        # this remove faces with the same vertices but same or different orders. This may happens on some of Voronoi if
+        # it slice self intersected mesh.
+        dict_faces = { tuple(sorted(f)): f for f in list(faces) }
```
nortikin, line 169: nice job

nortikin (approved):

utils/sv_bmesh_utils.py
Durman: I'm interesting how much it effects performance?

satabol: Hi @Durman . I think it is not too much:
image

Is a some trick to optimize it?

Durman: I wouldn't measure the performance in this way. I mean it would be nice know how much present the node is slower, especially in cases where this extra line does nothing useful. For example: before it took 1 second now it takes 1,1 second to handle the same mesh by the node.

satabol: If exclude this line then next line will raise exceptions in case of creating face with the same vertices.

PR #5059, opened 2023-11-11
Scripting examples.

portnov:
* blender_python.sh — script to run any python script with blender's builtin python
* generate_api_documentation.py — script to generate API documentation (similar to what we have in Github Actions, but to launch locally)
* run_sverchok_script.py — script to run any python script, which probably uses Sverchok API. Similar to blender_python.sh, but does not depend on OS / shell.
* script_example.py — example of script which can be run by run_sverchok_script.py.


Durman:
Would be nice to know what is it =)

portnov:
Maybe better to put all this to scripts/ directory, for example.

Durman:
> Maybe better to put all this to scripts/ directory, for example.

Yes, this also had come into my mind.

PR #5099, opened 2024-03-30
fix 5098. Some nodes generate exceptions if output sockets are not connected

### LINKED ISSUE #5098: Some nodes generate exceptions if output sockets are not connected
Sverchok 1.3.0-alpha

image

satabol:
fix #5098
- Add check if output sockets are connected before execute nodes.
- For some nodes did exception if some input sockets are not connected. This show source of problems more precisely.

Examples if some input sockets are not connected.

image


satabol:
Ready to merge. Can somebody view changes and say your opinions?

PR #5114, opened 2024-04-29
fix 5113 Node Mesh Clustering use to many dependencies

### LINKED ISSUE #5113: Node Mesh Clustering use to many dependencies.
Sverchok 1.3-Alpha

image

Node Mesh Clustering use **[pyacvd](https://github.com/pyvista/pyacvd)** in backend and has too many dependency

image

"E:\install\Blender\blender-3.6.9-windows-x64\3.6\python\Scripts\pipdeptree.exe" --packages pyacvd
image

satabol:
fix #5113.

A try to remove pyvista dependencies.

__Update__:

No need to remove pyvista dependency. Pyvista has good subdiv method with additional params.

### DIFF
--- a/nodes/spatial/mesh_clustering.py
+++ b/nodes/spatial/mesh_clustering.py
@@ -22,6 +22,7 @@ from sverchok.dependencies import pyacvd, Cython
 from sverchok.utils.mesh import clustering
 if pyacvd is not None:
     from pyvista import PolyData
+from pyacvd import _clustering
 #import pyacvd
 
 import bpy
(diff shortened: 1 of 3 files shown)


satabol:
Founded a problem in pyacvd library. Developer did subdiv method with only linear subdiv. Additionally he used pyvista method  "clean()" to clean mesh after subdiv. But pyvista mesh has a it's own subdiv method with additional params:

image

So there is no sense to use pyacvd subdiv with only linear param.

satabol:
Cancel. Will use.

PR #5129, opened 2024-05-30
fix 5128 Voronoi On Mesh parameter Output Nesting

### LINKED ISSUE #5128: Fix Voronoi On Mesh parameter Output Nesting
Sverchok Master.

I think a parameter "Output nesting" of a node "Voronoi on Mesh" is weird:
image

Example:
image

I have a proposal to refactor this node to the next rules:

image

In result one can use a node "Voronoi on mesh" to plugin to the next node "Voronoi on Mesh". This method only work in "Flat" mode now. As I know several nodes can be plugin to each other, for example "Dual Mesh":

image

But one cannot plugin "Voronoi on Mesh" to "Voronoi on Mesh" now (only in "Flat" mode). This proposal allow connect several nodes of "Voronoi on Mesh" to the chain.

What do you think?

satabol:
fix #5128 
- fix data_structure.py. Now it select to deep not zero list but simple data (numbers, string, boolean)
- new Output nesting
- new socket names
- fix voronoi function: not do not show exceptions if sites <=1 and stay that island untouched.

Example of new view for Voronoi on Mesh:

image


satabol:
Refactor node UI:

image

image

satabol:
Compare UI of old node and new node of Voronoi on Mesh:

image

satabol:
File for tests:
https://github.com/user-attachments/files/15521386/fix_5128_VoronoiOnMesh.new.node.002.blend.zip

PR #5137, opened 2024-07-23
fix 5136 Update List Input node

### LINKED ISSUE #5136: Update List Input node (proposal)
Sverchok 1.3.0-Alpha

image

A node List Input has next disadvantages:

1. It has limit of List Length parameter (128 for int/floats and 10 for vectors params).
2. After switch mode from Numbers to Vectors then node disconnects from another nodes
3. There are another types and formats of items (Colors, Angles, Power. see https://docs.blender.org/api/current/bpy_types_enum_items/property_subtype_number_array_items.html#rna-enum-property-subtype-number-array-items)

I have several proposals:

1. Append mask column 
2. Append new index column:
3. Append preview result column
image

4. Append Bool type of data
image

5. Append subtypes to Vector column
image

6. Append Color Types and subtypes
image

**What do you think ?**

satabol:
fix #5136 

File tests:

[fix_5136_example.004.blend.zip](https://github.com/user-attachments/files/17138098/fix_5136_example.004.blend.zip)


satabol:
- text:
image

- appended unit system and units:
image

these params are building in Blender:
image


All nodes units settings are syncronized:
System units

satabol:
Mask Input Socket:
image

Added Quaternion mode:
image

Output socket type:
image

satabol:
image

satabol:
1. Mode moved to dropdown list
2. Short Invert button
3. Refactor color editor

image

satabol:
Preview:

image

Issue #5150, opened 2024-10-08
New Node Stright Skeleton. Proposal.


satabol:
Examples:

![image](https://github.com/user-attachments/assets/667af5cb-570c-450b-91af-86955bf1718f)

![image](https://github.com/user-attachments/assets/50b63ec8-727f-4ed7-bb92-99726b8909bf)

satabol:
New node Straight Skeleton 2D Extruded created and merged into the mater branch. One can update Sverchok and test node functions. See docs here: https://nortikin.github.io/sverchok/docs/nodes/CAD/stright_skeleton_2d_extrude.html

PR #5154, opened 2024-10-11
Fix 5150 stright skeleton

satabol:
#5150 

image


satabol:
see also: https://github.com/satabol/pySVCGAL

PR #5159, opened 2024-10-21
fix 5156-new node Straight Skeleton Offset (External and internal mode)

### LINKED ISSUE #5156: Straight Skeleton. Offset. proposal
Add offset to the straight skeleton.

satabol:
Sverchok 1.3

fix #5156 

image


satabol:
If offset is internal (value is positive) the no problem:
image
All internal offsets do not intersects with contours of any other polygon or other words an internal offset do not out of outer boundary. It property help to parallelize computation. 
But there is a problem if offset is external. It can overlapped with contours of external objects:
image

satabol:
Additional Error if selected shape mode "Invert holes Soft/Hard":
image

image
Here is one hole but algorithm do two holes!

satabol:
image

image

satabol:
Invert Holes work Good:

image

Issue #5189, opened 2025-01-30
CSG Boolean MK2 [ERROR] sverchok.node_error csg_geom.py:21 - maximum recursion depth exceeded


kevinsmia1939:
Hi,
I construct Gyroid mesh using marching cubes. However, I could not use CSG boolean to further processed it.
The error log is 

`[ERROR] sverchok.node_error csg_geom.py:21 - maximum recursion depth exceeded`

I search the bug report here but it does not seems to be reported before.

OS: openSUSE Tumblweed
Blender 4.2.3 LTS
Sverchok: master branch commit 6b6276470de6451da0c6920627bda984cdc51045 (latest)

Here is the file, remove the .txt before opening.
[csg_boolean_problem.blend.txt](https://github.com/user-attachments/files/18600847/csg_boolean_problem.blend.txt)

![Image](https://github.com/user-attachments/assets/bc74b971-f2a9-4967-a96c-46d0ad6e0dae)

nortikin:
thank you

PR #5190, opened 2025-02-02
Blender internal Boolean implementation node

kevinsmia1939:
To my best knowledge, the Blender implementation can perform boolean on overlapping geometries unlike the CSG boolean as well as being faster. 

I currently testing it out. I use a lot of chatgpt to help write this, so this need to be check if anyone is interested.
Cannot test CSG boolean at the moment because this issue https://github.com/nortikin/sverchok/issues/5189 but with this node it works fine.
## Solution description

Simply warp the Blender boolean implementation into a node.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.


portnov:
Hello.

We considered similar approaches long time ago and, at least back then, we decided not to go in this direction.

Unfortunately this approach has a couple of problems.
* creation of objects in Blender scene takes time; conversion from vertices to objects and back to vertices also take time. — probably not a very big problem in this particular case, as CSG Boolean node is very slow by it's nature.
* this can create a large number of objects in Blender scene if you manipulate with a large number of objects in Sverchok; you have to be very accurate to be sure that you destroy all your temporary objects in time even if there is any error or exception in python or blender code.
* manipulating with a large number of objects in scene can be problem for Blender itself.

However, I understand that lack of "normal" boolean node in Sverchok is a longstanding problem. Once we asked Blender devs to create bmesh api for booleans, but it does not look like anything is going to change in that field, unfortunately.

So maybe your approach is the only way to do this :/

Could you please describe, how did you test your node? did you, for example, check for cases when you need to intersect 10 objects with another 10 objects? I'm not sure to be honest which other corner cases we have to check, need to think.

@satabol @zeffii could you please review as well?

kevinsmia1939:
@portnov 
Thank for your feed back, I did not know that this has been tried before but I feel that it would have already been considered. Anyway, I tried to union,intersect, and diff bunch of stuff together and seems to work fine. I don't really know a how to do a more comprehensive test of this.

Screenshot_20250202_145240

nortikin:
image
image
it seems like work. Please, add some documents and we will accept commit

nortikin:
test_node_docs_existance error, so...

kevinsmia1939:
Ok, working on the doc

kevinsmia1939:
Not sure how to implement unit testing. But I think other requirements are complete.

nortikin:
> Not sure how to implement unit testing. But I think other requirements are complete.

Internal Boolean demo what supposed to be?

kevinsmia1939:
@nortikin 
Ah, just the test that we previously did? I thought it was something like running unittest or Pytest.
Anyway, do you know how to properly add screenshot? I just put the screenshot in my repo.

nortikin:
> @nortikin Ah, just the test that we previously did? I thought it was something like running unittest or Pytest. Anyway, do you know how to properly add screenshot? I just put the screenshot in my repo.

`::|Internal Boolean demo|`
has to be:
```
Examples
--------
```
and remove `.. |Internal Boolean demo|` at the end

PR #5207, opened 2025-03-20
Fix missing sockets tooltip #5206

### LINKED ISSUE #5206: sockets are missing description tooltip
1. Start Blender with sverchok.
2. Run snippet below.
3. Currently (repeats the socket name)

Image

4. Expected:
Image

```python
import bpy
from sverchok.ui.nodeview_space_menu import add_node_menu


class SvTestNode(bpy.types.Node, SverchCustomTreeNode):
    bl_idname = "SvTestNode"
    bl_label = "Test Node"

    def sv_init(self, context):
        socket = self.outputs.new("SvStringsSocket", "Entities")
        socket.description = "List of entities, list[int]"
    
    def process(self):
        pass


classes = [SvTestNode]
node_categories = [{"TestNodes": [{"Test Category": ["SvTestNode"]}]}]
add_node_menu.append_from_config(node_categories)


def register():
    add_node_menu.register()
    for cl in classes:
        bpy.utils.register_class(cl)

if __name__ == "__main__":
    register()
```

Andrej730:

Fix #5206

## Solution description

It seems the root cause for the missing description tooltip is that `description` property that's already present on [NodeSocket](https://docs.blender.org/api/current/bpy.types.NodeSocket.html) is overridden by custom property with the same name.
Removing override resolves the issue.

## Preflight checklist

- [x] Code changes complete.
- [x] Manual testing done. 
- [x] Unit-tests implemented.
- [x] Ready for merge.

### DIFF
--- a/core/sockets.py
+++ b/core/sockets.py
@@ -350,7 +350,6 @@ class SvSocketCommon(SvSocketProcessing):
     # utility field for showing number of objects in sockets data
     objects_number: IntProperty(min=0, options={'SKIP_SAVE'})
 
-    description : StringProperty()
     is_mandatory: BoolProperty(default=False)
     nesting_level: IntProperty(default=2)
     default_mode: EnumProperty(items=enum_item_4(['NONE', 'EMPTY_LIST', 'MATRIX', 'MASK']), default='EMPTY_LIST')


Andrej730:
@portnov Hi! can you please take a look?

Andrej730:
> Is this helping?

Yes, it works. 

image


> Also, did you check, isn't this override used in other places in sverchok?

I believe the only place `description` attribute is used is this `draw` method in `SvSocketCommon` and the way it's used works without an override. Perhaps adding an override wasn't completely intentional and was just an overlook missing that `description` already exist on https://docs.blender.org/api/current/bpy.types.NodeSocket.html#bpy.types.NodeSocket.description

https://github.com/Andrej730/sverchok/blob/78692c434a85b923643b968555a9df9a034f2782/core/sockets.py#L507-L514

PR #5251, opened 2025-10-06
fix 5250 Get_Objects_Data node work in Blender 4.x

### LINKED ISSUE #5250: Get Objects Data node do not work in Blender 4.x
All creases and bevels In Blender 3.6.x obtained.
If some sockets with creases and bevels In Blender 4.x are connected then no one creases and bevels do not obtained.

Image

satabol:
fix #5250.

image

for tests: 
[GetObjectsDataPreoblemCrease.001.blend.zip](https://github.com/user-attachments/files/22732036/GetObjectsDataPreoblemCrease.001.blend.zip)

Also tested in 4.5


satabol:
image

PR #5264, opened 2025-10-17
fix 5253 add on off into Matrix view

### LINKED ISSUE #5263: add on off into Matrix view
Image

satabol:
fix #5263 

image


satabol:
fix non flat list of matrixes:

image

PR #5266, opened 2025-10-19
fix 5265 Viewer Draw with Vulkan API custom shader issue

### LINKED ISSUE #5265: Viewer Draw with Vulkan API custom shader issue
(reserved number)

satabol:
fix #5265


satabol:
Edges look bad for a while:

image

satabol:
Hi @zeffii 
I try this too. It is not work for Vulkan API. I did not find solution for this case for a while.

PR #5274, opened 2025-10-25
Renew shading ViewerDraw

nortikin:
## Solution description

shift colors in defs to make lighter and affect light not to zero values, but to 0.2 of them and shift back


nortikin:
seems like i grubbed that changes to PR 
https://github.com/nortikin/sverchok/pull/5275

PR #5277, opened 2025-10-25
Fix 5271 bezier input mk2 upgrade

### LINKED ISSUE #5271: Bezier Input MK2 (Upgrade)
1. Bezier Input MK2 (Upgrade) like Get Objects Data

Image

2. Change Output into new format. Keep info about source object of bezier:

Image

satabol:
fix  #5271 

Added some UI control of list of names
image

add objects level if concatenate on or off:

image


satabol:
DO not use

PR #5278, opened 2025-10-25
Fix 5271 bezier input mk2 upgrade nodes curves for tests

satabol:
Used for tests of #5277. DO NOT MERGE WITH MASTER


satabol:
File for tests:

[Uploading BezierIn.MK2.001.blend.zip…]()

PR #5314, opened 2025-11-21
fix 5113 List Input Matrix mode Angle Exis, Euler Angles very narrow

### LINKED ISSUE #5313: List Input Matrix mode Angle Exis, Euler Angles very narrow
Image

satabol:
fix #5313

image


satabol:
fixed

PR #5334, opened 2025-12-01
fix 5333 backward and forward compatibility socket custom_draw issues

### LINKED ISSUE #5333: backward and forward compatibility socket custom_draw issues
If Sverchok in different versions will have different custom_draw for soket, it will break both backwards compatibility and future compatibility, because not only the old version does not know new function, but also the new version does not need to remember the old name (although it is desirable to remember)

If no custom_draw then property UI are not visible in the node.

Image

satabol:
- fix #5333 

(developers feature)

image


satabol:
image

PR #5348, opened 2026-01-07
fix 5347 Add Custom property into Get Objects Data for Object, Object Data and Materials

### LINKED ISSUE #5347: Add Custom property into Get Objects Data for Object, Object Data and Materials
Example with added Custom socket property of material:

Image

satabol:
fix #5347.

- add custom properties editor (copy properties, remove properties) in Get Objects Data for Object, Data, Materials
- save custom properties in node
- Output custom properties into outputs sockets

image


satabol:
Price can be calculated with 3 parameters: Area*price_for_m2*Compexity_Factor:

image

PR #5356, opened 2026-01-13
fix 5355 Offset Line 2D update

### LINKED ISSUE #5355: Offset Line 2D update
Align offset line on vertices plane matrix

satabol:
fix #5355

image


satabol:
image

PR #5398, opened 2026-04-17
fix 5397 Extend node Path Length with groups

### LINKED ISSUE #5397: Extend node Path Length with groups
Image

satabol:
fix #5397

image


satabol:
File for tests:

[Example.Blender.3.0.v.001.blend.zip](https://github.com/user-attachments/files/26838651/Example.Blender.3.0.v.001.blend.zip)

image

PR #5402, opened 2026-04-24
fix 5401 Geo Nodes Viewer Issue in Blender 5.x

### LINKED ISSUE #5401: Geo Nodes Viewer Issue in Blender 5.x
Blender 5.x

Image

satabol:
fix #5401

image


satabol:
Also work for Blender 3.x

PR #5410, opened 2026-05-03
"Optimal bezier spline" node

portnov:
I have some number of articles in computation geometry waiting in order to be read and ideas tested. It appears LLMs give an opportunity to implement ideas from articles quite quickly.

Quality of implementation is obvious concern. Let this stay in PR for some time, hoping someone will have time to review.

Otherwise, this in general works, so everyone is welcome to test.

algorithms (c) V. Borisenko 2016, code by Qwen 3.6 27B. Documentation template is also written by Qwen, waiting for me to have some time and insert proper screenshots.

Abstract from the article:

V. V. Borisenko, Construction of optimal Bézier splines, Fundamentalnaya i prikladnaya matematika, vol. 21 (2016), no. 3, pp. 57—72.

We consider a construction of a smooth curve by a set of interpolation nodes. The
curve is constructed as a spline consisting of cubic Bézier curves. We show that if we
require the continuity of the ﬁrst and second derivatives, then such a spline is uniquely
deﬁned for any ﬁxed parameterization of Bézier curves. The control points of Bézier
curves are calculated as a solution of a system of linear equations with a four-diagonal
band matrix. We consider various ways of parameterization of Bézier curves that make
up a spline and their inﬂuence on its shape. The best spline is computed as a solution
of an optimization problem: minimize the integral of the square of the second derivative
with a ﬁxed total transit time of a spline.

изображение

- [ ] Code changes complete.
- [ ] Manual testing done. 
- [ ] Unit-tests implemented.
- [ ] Ready for merge.

### DIFF
--- a/docs/nodes/curve/optimal_bezier.rst
+++ b/docs/nodes/curve/optimal_bezier.rst
@@ -1,0 +1,202 @@
+Optimal Bezier Spline
+=====================
+
+.. image:: https://github.com/nortikin/sverchok/assets/14288520/TODO-optimal-bezier-spline
+  :target: https://github.com/nortikin/sverchok/assets/14288520/TODO-optimal-bezier-spline
+
+Functionality
+-------------
+
+This node generates a C2-continuous cubic Bezier spline through the specified
+set of interpolation points. The spline is built as a sequence of cubic Bezier
+segments with continuous first and second derivatives at all joints.
+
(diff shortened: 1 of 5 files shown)


portnov:
After some thought: I moved this to sverchok-extra (already there in master).
* "Optimal Bezier spline" is not looking so much better than Hobby spline
* It is waay slower due to iterative nature of the algorithm

PR #5413, opened 2026-05-08
fix 5421 Generative Art node upgrade

### LINKED ISSUE #5412: Generative Art node upgrade
Image

Need drop down list like in the Script node to load Examples:

Image

satabol:
fix #5412

image


satabol:
Ribbon:

image

satabol:
image

portnov:
@satabol how do you think, maybe it worth implementing eisenscript (original language of structuresynth, which was an inspiration of generative art node) as well? There are much more examples of eisenscript in the internet, than of this specific xml.
I see that there are some efforts of translating eisenscript to xml, but not sure how complete they are.

satabol:
Hi @portnov 
Good idea to use eisenscript instead of xml-format. First I would like to finish fixing the nod, and then it will be possible to think about development.
I think Eisenscript is good, but managing the result isn’t flexible enough (like with XML-format too). As you can see result can has some glitches:

image

some geometry has overlapping. This can’t be used for real objects. I am not suggesting that overlapings be checked programmatically, but would like to give the user the opportunity to intervene in certain cases at a specific point. For example, if a rule rotates at a certain angle, allow the user to set that angle in a particular segment. As far as I know, neither eisenscript nor xml format allows this.

I have these plans:
1. Finish working on the errors in the current number
2. Extend its use to tube only, but also faces (add faces to the verts input sockets) (MK2)
3. Think about updating in the direction of eisenscript or xml-format (or combine both).

What do you think?

portnov:
Okay.
For now, translator from #5415 works for some number of examples for me. You can use it to translate some examples which can be found, to have something to test with.

portnov:
As for overlapping... Will be interesting to see what you can suggest to fix that. For now the only approach I see is just to use booleans :)

portnov:
@satabol if you will be doing Mk2, I'd suggest to move core logic from node file to something like sverchok.utils.modules.generative_art (or whatever inside utils). This way it will be simpler to test and reuse.

satabol:
@portnov 
> As for overlapping... Will be interesting to see what you can suggest to fix that. For now the only approach I see is just to use booleans :)

I have an idea for visually tweaking the parameters by eye, but I’m still thinking through the implementation.

satabol:
I discovered many unpredictable behaviors of the algorithm in the current implementation. A serious refactor is needed...

portnov:
@satabol could you please explain or give examples of such "unpredictables"?

Also, I have an implementation of Eisenscript interpreter in #5415. Some refactoring is pending, but it works somehow. Now I'm thinking that probably it will be better to have two different nodes, one for old Generative Art XML and another for Eisenscript.

satabol:
Hi, @portnov!

If you’re working with one instance and one call in rule, then something else is working.

Examples:
```
<rule name="curl" max_depth="80">
    <call     transforms="rx 12.5 tz 1.3 s 0.98 0.95 1.0" rule="curl"/>
    <instance transforms="tx 0.1 ry 40 tz -0.5"           shape="box"/>
    <call     transforms="tx 0.1 ty -0.45 ry 40 sa 0.25"  rule="curlsmall" />
</rule>
```
This does not work as expected: instance puts into stack and lost what call process it. If multiple instances and calls are set then they completely lose communication between themselves. And no messages are displayed if the xml file is not suitable for processing.

satabol:
I change the code for each example and it stops working in the next one. I try to do the calculation from start to finish and constantly lose the length of the tail.

portnov:
Probably it will be interesting for you to experiment with Eisenscript interpreter in parallel, to compare.

portnov:
@satabol do I correctly understand what your example should do?
(eisenscript analog, if I understand correctly):
```
rule curl maxdepth 80 {
    {rx 12.5 z 1.3 s 0.98 0.95 1.0} curl
    {x 0.1 ry 40 z -0.5} box
    {x 0.1 y -0.45 ry 40 s 0.25} curl
}

curl
```

изображение

portnov:
To be honest, it's quite hard to debug such declarative and recursive language. You have to stress your imagination quite hard :)

satabol:
image

image

satabol:
Hi @portnov 
Can you test your converter with this version of this node?

portnov:
@satabol I currently have a converter just in one direction, eisenscript -> XML. Do you need to convert in another direction?
Or I'm not sure what are you asking about :)

satabol:
@portnov 
I mean eisenscript -> XML.

portnov:
What Eisenscript source to try it with?

I have a number of tests in my PR which check compatibility of Eisenscript interpreter with old LSystem interpreter in GenerativeArt node. But those tests check very simple programs.

satabol:
@portnov 
I thought you have a converter ES->XML. I'm wrong?

portnov:
No, I have it. The question is, what program do you want to convert?

portnov:
You can play with converter, it is available in Pr I  mentioned, see pr description for usage

satabol:
I want just fix this node for a while. I’ll decide on the format later.

satabol:
@portnov 
Is this script run on windows?

satabol:
@portnov 
I got it. But something looks weird. For matrix mode it looks good.

image

image

portnov:
My eisenscript interpreter assumes that "box" primitive has size 1x1x1. I think Generative Art node with Matrices assumes the same. So if you provide boxes of another size, it will indeed look differently.

satabol:
alert message on XML error:

image

satabol:
@portnov 
I see. Fix it. Looks better:

image

satabol:
@portnov 
It seems to me that using non-standard syntax is not a good idea. But at the same time, the XML format used is not being used correctly. For example, it is not correct for "instance" objects to specify after a "call" as they did in the source node. I am more inclined to use the XML format, because XML has standard parsers. I also have a few ideas for expanding the functions of the node, and it’s not very clear how to implement them in eisenscript.

portnov:
What do you mean by "non standard format"? Well, eisenscript is not as popular, as, for example, python :) But this particular kind of XML is used only in GenerativeArt node. Eisenscript is used at least in StructureSynth, which is more or less well known.

Probably both variants have a right to exist. So probably let there be two different nodes - GenerativeArt with XML and "Eisenscript" for eisenscript interpreter. Life will show which of them gets more features and more users.

satabol:
>>What do you mean by "non standard format"?

I meant that the non-standard is when you need a parser to read a format. I understand that eisenscript has some scope, but I don’t see how it could be expanded further? If it were at least like JSON, it would in principle make no difference how to use it - xml or JSON.

But at the moment they are both not standard. For example, recursion makes it very strange that used instances are written after call. This is very bad for the algorithm.

I have now changed the order of instances and calls - first write instances, then make a call.

satabol:
image

satabol:
@portnov 
I have finish all what I planned to upgrade and will close this PR now. We can continue conversation in your PR. #5415 .
I have left the functions inside the module because it is a big, but minor change. Take away functions I plan in next update.

PR #5434, opened 2026-06-05
fix 5442 Skin Mesher Issue if doubles distance more 0

satabol:
fix #5432 Additionally fix gap on vertices with 3 and more edges.

image


satabol:
https://nortikin.github.io/sverchok/docs/nodes/viz/viewer_skin.html

image

satabol:
image

image

PR #5438, opened 2026-06-12
fix 5437 List Levels update proposal

### LINKED ISSUE #5437: List Levels update proposal
1. performance
2. update UI table view
3. difficult to understand the purpose of banning these two elements

Image

satabol:
fix #5437

image

image

### DIFF
--- a/nodes/list_struct/levels.py
+++ b/nodes/list_struct/levels.py
@@ -338,7 +338,10 @@ class SvListLevelsNodeMK3(SverchCustomTreeNode, bpy.types.Node):
         for I, socket in enumerate(self.inputs):
             if socket.is_linked==False and socket==self.inputs[-1]: # You may have a situation where the last link has just been connected and the new last unconnected link has not yet been created.
                 break
-            data = socket.sv_get(default=[], deepcopy=False)
+            if socket.name in self.outputs and self.outputs[socket.name].is_linked==True:
+                data = socket.sv_get(default=[], deepcopy=False)
+            else:
+                data=[]
             datas.append(data)
 
         if not self.levels_config or force_reload_config==True:
--- a/nodes/list_struct/levels.py
+++ b/nodes/list_struct/levels.py
@@ -73,745 +73,7 @@ class SvNestingLevelEntryMK3(bpy.types.PropertyGroup):
                 options={'SKIP_SAVE'},
                 )
 
-def recursive_unpack(data, levels_info=None, level=0):
-    res = []
-    if levels_info is None:
-        raise Exception("levels_info should be provided for recursive_unpack")
-    if len(levels_info)==0:
-        raise Exception("levels_info should contain at least one level for recursive_unpack")
-
-    if isinstance(data, ALL_TYPES):
-        res = data
-    else:
-        levels_info_level = levels_info[level+1]
-        # Вполне можно рассчитать есть ли any_flatten/wrap ниже текущего уровня, чтобы в случае, когда они не выставлены до конца глубины,
-        # то и не делать обход, а сразу заливать список в результат.
-        #any_flatten = any([levels_info[I].flatten for I in range(level+1, len(levels_info)-1)])
-        #any_wrap    = any([levels_info[I].wrap    for I in range(level+1, len(levels_info)-1)])
-        flatten = levels_info_level.flatten
-        wrap = levels_info_level.wrap
-        for elem in data:
-            if isinstance(elem, ALL_TYPES):
-                if levels_info_level.wrap:
-                    res.append([elem])
-                else:
-                    # На самом нижнем уровне, где находятся элементарные типы, flatten не выполняется, даже если стоит галочка,
-                    # так как там нет вложенных списков, которые нужно распаковать. Поэтому сохранять как есть.
-                    res.append(elem)
-                pass
-            elif isinstance(elem, (list, tuple, ndarray)):
-                val = recursive_unpack(elem, levels_info, level + 1)
-                # Одна из проблем этой функции в том, что в UI не задаётся приоритет выполнения flatten и wrap,
-                # поэтому выполняться они будут только по одной, либо когда обе выключены.
-                if flatten==True and wrap==False or flatten==True and wrap==True:
-                    # Ситуация flatten==True and wrap==True обрабатывается после цикла, так как там нужно сначала распаковать все уровни,
-                    # а потом уже обернуть результат в дополнительный список.
-                    res.extend(val)
-                elif flatten==False and wrap==True:
-                    res.append([val])
-                else:
-                    # Для случая, когда включены обе галочки, то действий не выполнять, потому что
-                    # такие настройки взаимно исключают друг друга, и когда выключены обе галочки,
-                    # то сохранять как есть. Поэтому в обоих этих случаях сохранять как есть.
-                    res.append(val)
-                pass
-            else:
-                raise Exception(f'Uncompatible value for sum: {type(elem)}')
-            pass
-
-        if flatten==True and wrap==True:
-            # Обернуть результат flatten текущего уровня:
-            res = [res]
-    
-        if level==0:
-            # На верхнем уровне flatten не выполняется, так как там нет родительского уровня, который нужно распаковать. Поэтому можно выполнить только wrap:
-            if levels_info[0].wrap:
-                res = [res]
-            pass
-    return res
-
-
-def recursive_unpack2(data, levels_info=None):
-    # res = []
-    if levels_info is None:
-        raise Exception("levels_info should be provided for recursive_unpack")
-    if len(levels_info)==0:
-        raise Exception("levels_info should contain at least one level for recursive_unpack")
-
-    level = 0
-    LIST_TYPES = (list, tuple, ndarray)
-
-    # Вызов только при погружении в уровень, так как при первом вызове функции data может быть простым числом,
-    # и тогда не будет смысла проверять галочки flatten/wrap, так как они относятся к спискам, а не к элементарным типам.
-    def _recursive_unpack(data, level):
-        res = []
-        res_append = res.append
-        res_extend = res.extend
-        levels_info_level = levels_info[level+1]
-        # Вполне можно рассчитать есть ли any_flatten/wrap ниже текущего уровня, чтобы в случае, когда они не выставлены до конца глубины,
-        # то и не делать обход, а сразу заливать список в результат.
-        #any_flatten = any([levels_info[I].flatten for I in range(level+1, len(levels_info)-1)])
-        #any_wrap    = any([levels_info[I].wrap    for I in range(level+1, len(levels_info)-1)])
-        flatten = levels_info_level.flatten
-        wrap = levels_info_level.wrap
-        for elem in data:
-            type_elem = type(elem)
-            # if isinstance(elem, ALL_TYPES):
-            if isinstance(elem, ALL_TYPES):
-                if wrap==True: # flatten==False and wrap==True
-                    res_append([elem])
-                else:
-                    # На самом нижнем уровне, где находятся элементарные типы, flatten не выполняется, даже если стоит галочка,
-                    # так как там нет вложенных списков, которые нужно распаковать. Поэтому сохранять как есть.
-                    res_append(elem)
-                pass
-            elif isinstance(elem, LIST_TYPES):
-                val = _recursive_unpack(elem, level + 1)
-                # Одна из проблем этой функции в том, что в UI не задаётся приоритет выполнения flatten и wrap,
-                # поэтому выполняться они будут только по одной, либо когда обе выключены.
-                if flatten==True:
-                    # Ситуация flatten==True and wrap==True обрабатывается после цикла, так как там нужно сначала распаковать все уровни,
-                    # а потом уже обернуть результат в дополнительный список.
-                    res_extend(val)
-                elif wrap==True:
-                    res_append([val])
-                else:
-                    # Для случая, когда включены обе галочки, то действий не выполнять, потому что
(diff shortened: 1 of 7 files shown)


satabol:
Hi, @portnov 
Do you have the ability to run a new version of Levels and comment on its work?

portnov:
Ok, will check.

Actually initially this node was using table-like UI. Later someone rewrote the UI, I believe to make the node more compact. But honestly I like the table-like UI better.

portnov:
Didn't have time to review myself yet, so here is a review by Qwen. I hope to review myself during this weekend.

# PR #5438 Review: fix 5437 List Levels update proposal

**Author:** satabol  
**Branch:** `fix_5437_List_Levels_update_proposal`  
**Status:** Open  
**Files changed:** 7  
**Additions:** ~350 lines | **Deletions:** ~70 lines  

---

## Summary

This PR introduces `SvListLevelsNodeMK3`, a significantly redesigned "List Levels" node that replaces `SvListLevelsNodeMK2`. Key changes:

1. **Multi-input/multi-output sockets** — dynamic socket management (data_1, data_2, ...)
2. **New core algorithm** — `recursive_unpack` replaces `list_levels_adjust` from `data_structure.py`
3. **New data introspection** — `data_levels_info` replaces `describe_data_shape_by_level`
4. **UI overhaul** — grid_flow layout with alert indicators
5. **Constraint** — first and last levels cannot use Flatten
6. **Migration** — MK2 moved to `old_nodes/` with `replacement_nodes` pointing to MK3

---

## ✅ Positive Aspects

- **Multi-input support** is a valuable feature — users can process multiple data streams independently
- **UI improvement** with `grid_flow` is cleaner than the old split-based layout
- **Alert system** for mixed-type levels is a nice UX touch
- **Proper migration chain**: MK → MK2 → MK3 with `replacement_nodes` pointers
- **Menu YAML files** correctly updated across all three menu configs
- **Documentation** updated with new screenshots

---

## ❌ Critical Issues

### 1. Unused imports

```python
from os import link
from platform import node
```

These imports serve **no purpose** in the file. `os.link` and `platform.node` are never used. This looks like accidental imports (possibly from IDE auto-import). Remove them.

### 2. Unused imports (continued)

```python
from sverchok.core.socket_data import sv_deep_copy
from numpy import ndarray
```

`sv_deep_copy` and `ndarray` are imported but never referenced in the code. The comment in `recursive_unpack` mentions `ndarray` was "used to have" but then removed — the import should be removed too.

### 3. `@dataclass` with `pass`

```python
@dataclass
class LevelInfo:
    TYPE: str = None
    COUNT: int = 0
    ALERT: bool = False
    SUB: bool = False
    pass
```

- Trailing `pass` is unnecessary (the class body is non-empty due to field annotations)
- Field names `TYPE`, `COUNT`, `ALERT`, `SUB` use ALL_CAPS convention which conflicts with Python's variable naming conventions (ALL_CAPS is reserved for constants). Consider `type_`, `count`, `alert`, `is_sub` or similar.
- Default `TYPE: str = None` — `None` is not a `str`. Use `TYPE: str | None = None` or `Optional[str]` for correctness.

### 4. Exception with bare `raise Exception(...)`

```python
raise Exception("levels_info should be provided for recursive_unpack")
raise Exception("levels_info should contain at least one level for recursive_unpack")
raise Exception(f"Unsupported flatten/wrap mode: {flatten_wrap}")
```

Use `ValueError` or `RuntimeError` instead of bare `Exception`. This is a Python best practice.

### 5. `data_levels_info` — premature `break`

```python
for elem in data:
    type_elem = type(elem)
    if type_elem not in levels_info_level_root:
        levels_info_level_root[type_elem] = LevelInfo(...)
    if type_elem in LIST_TYPES_EXACT:
        levels_info_level_root[type_elem].SUB = True
        _data_levels_info(elem, levels_info, level_root + 1)
    else:
        break  # ← PROBLEMATIC
```

The `break` exits the loop as soon as a non-list/tuple element is encountered. This means **mixed-type lists** (e.g., `[1, [2,3], 4]`) will only analyze up to the first non-list element. The TODO comment acknowledges this: `# TODO: разобраться, когда среди вложенных данных только list`. This is a correctness concern — the function should iterate through all elements.

### 6. Socket naming inconsistency

The new node uses `data_1` as the internal socket name but labels it `'Data'`. The old MK2 used `'Data'` as both name and label. This is fine for the first socket, but when dynamic sockets are added (`data_2`, `data_3`, ...), the pattern `f"Data_{len(self.inputs)+1}"` creates labels like `Data_3` while the first socket's label is just `Data`. Consider consistency: either all `Data`, `Data_2`, `Data_3` or `Data_1`, `Data_2`, `Data_3`.

### 7. Complex socket management in `process()`

The socket creation/removal logic in `process()` is complex and potentially fragile:

```python
while len(self.inputs) > 1 and not self.inputs[-2].links:
    socket_to_remove = self.inputs[-1]
    for link in list(socket_to_remove.links):
    self.inputs.remove(socket_to_remove)
```

- Modifying sockets during `process()` can cause issues if the node is evaluated multiple times rapidly.
- `sv_update` is the more appropriate place for socket management.
- The logic for removing links from sockets being removed (`self.id_data.links.remove(link)`) is unusual — typically Blender handles this automatically when a socket is removed.

### 8. `reload_config` called during `process()`

```python
if not self.levels_config or self.load_levels_config or force_reload_config==True:
    self.reload_config(datas)
```

Modifying node state (`levels_config`, `nesting`) during `process()` is generally discouraged in Sverchok. `process()` should be a pure computation. State changes should happen in `sv_update()` or property update callbacks. This can lead to unexpected behavior if the node is processed before the config is fully initialized.

### 9. Hardcoded constraint on first/last flatten

```python
if self.levels_config[0].flatten != False:
    self.levels_config[0].flatten = False
if self.levels_config[-1].flatten != False:
    self.levels_config[-1].flatten = False
```

This silently resets user settings without any warning. While the documentation mentions "First and last elements can't be flatten", the UI should reflect this by **disabling** the flatten checkbox for first and last levels (similar to how MK2 disabled flatten for innermost levels), rather than silently resetting the value.

### 10. Comment in Russian

```python
# Сбрасывать в любом случае
self.load_levels_config = False
```

Per AGENTS.md: "All comments must be in English". Translate to English.

---

## ⚠️ Minor Issues / Suggestions

### 11. `flatten_wrap` encoding is opaque

```python
flatten_wrap = levels_info[next_level]  # 0, 1, 2, or 3
```

The encoding `0=(False,False)`, `1=(False,True)`, `2=(True,False)`, `3=(True,True)` is a bitwise combination that works but is not immediately clear. Consider using an `enum` or named constants for readability:

```python
NONE = 0
WRAP_ONLY = 1
FLATTEN_ONLY = 2
FLATTEN_WRAP = 3
```

### 12. `force_reload_config` flag pattern

The `force_reload_config` flag is used in multiple places and passed between functions. This creates a feedback loop where `process()` → `recursive_unpack()` → flag set → `reload_config()` called → potentially changes config → affects next iteration. This could be simplified with a cleaner separation between configuration and processing phases.

### 13. `levels_config[0] in (1,3)` check

```python
if levels_info[0] in (1,3):
    res = [res]
```

This checks the encoded value directly. With the encoding, `1=WRAP_ONLY` and `3=FLATTEN_WRAP` — both have wrap=True. This is correct but fragile. Better to extract `wrap` as a boolean.

### 14. Missing `replacement_nodes` in MK3

The new `SvListLevelsNodeMK3` class does not define `replacement_nodes`. Future versions should add this for the migration chain.

### 15. `load_levels_config_` property declared but unused

```python
load_levels_config_: bpy.props.BoolProperty(default=False)
```

This property is declared but never used. The actual property is `load_levels_config` (without underscore). Remove the unused one.

### 16. `updateNodeReload` — potential infinite loop

```python
def updateNodeReload(self, context):
    if self.load_levels_config==True:
        updateNode(self, context)
    else:
        pass
    return
```

`updateNode` triggers `process()`, which sets `self.load_levels_config = False`. If `process()` then calls `reload_config()` which modifies `levels_config`, this could trigger another update chain. The `pass` in the else branch is unnecessary.

### 17. Documentation image URLs

The new documentation images point to `github.com/user-attachments/assets/...` URLs. These are ephemeral and may not persist. Consider hosting images in the repository (e.g., `docs/images/`) for long-term reliability, as the old MK2 docs did.

---

## 📋 Checklist

| Item | Status |
|------|--------|
| Unused imports (`os.link`, `platform.node`, `sv_deep_copy`, `ndarray`) | ❌ Remove |
| Russian comment | ❌ Translate to English |
| `raise Exception` → `ValueError`/`RuntimeError` | ❌ Fix |
| `@dataclass` field naming (ALL_CAPS) | ⚠️ Consider renaming |
| `data_levels_info` `break` for mixed types | ❌ Fix correctness |
| Socket management in `process()` vs `sv_update()` | ⚠️ Consider moving |
| Silent reset of first/last flatten | ⚠️ Disable in UI instead |
| `load_levels_config_` unused property | ❌ Remove |
| `pass` in `@dataclass` | ❌ Remove |
| `pass` in `else` branches (multiple) | ⚠️ Remove (Python style) |
| Documentation image persistence | ⚠️ Consider local hosting |
| `replacement_nodes` on MK3 | ⚠️ Add for future migration |
| Menu YAML files updated | ✅ OK |
| MK2 moved to old_nodes | ✅ OK |
| `register/unregister` uses factory | ✅ OK |

---

## Verdict

**Changes look directionally correct** — multi-input support and UI improvements are valuable. However, there are **several issues that should be addressed before merging**:

1. **Remove all unused imports** (lines 19-20, 32, 34)
2. **Translate Russian comment** to English
3. **Fix `data_levels_info` break** — iterate all elements, not just until first non-list
4. **Move socket management** from `process()` to `sv_update()`
5. **Replace `raise Exception`** with specific exception types
6. **Disable first/last flatten in UI** rather than silently resetting
7. **Remove unused `load_levels_config_` property**

The core algorithm (`recursive_unpack`) works but could benefit from more test coverage, especially for edge cases like mixed-type nested lists and empty sublists.

satabol:
Hi Ilya @portnov 
Thank you for your review. I refactor code. Is it ok?

nodes/list_struct/levels.py
```diff
@@ -17,115 +17,359 @@
...
                 default=False,
                 update=update_entry)
+    alert : bpy.props.BoolProperty(
+                name = "Alert",
+                description = "Alert some warning about this level",
+                default=False,
+                options={'SKIP_SAVE'},
+                )
+
+def recursive_unpack(data, levels_info=None):
```
portnov, line 68: I would suggest to move this to data_structure.py or into some separate module under utils.

nodes/list_struct/levels.py
```diff
@@ -17,115 +17,359 @@
...
+        res = data
+    
+    if max_level_reached==False:
+        # Overload the configuration if you don’t reach the maximum level (user can make flatten and then the function will not reach the bottom). Also need to be counted.
+        force_reload_config = True
+
+    return (res, force_reload_config)
+
+# Load data configuration
+def data_levels_info(data, levels_info=None, level_root=0):
```
portnov, line 150: This as well.

nodes/list_struct/levels.py
```diff
@@ -17,115 +17,359 @@
...
+            force_reload_config = True
+            # Continue to copy, without changing the data after finding that you have exceeded the number of levels you’ve set.
+            flatten_wrap = 0
+        else:
+            # If the level has reached a lower configuration level, set a flag (if this flag is not displayed, then the data configuration does not match the original configuration)
+            if len_levels_info==next_level:
+                max_level_reached = True
+            flatten_wrap = levels_info[next_level]
+
+        if flatten_wrap==0: # flatten==False and wrap==False
```
portnov, line 99: Maybe give names to these magic constants?

portnov:
Also, previous version of node indicated not only the type of nested structure ("list" / "tuple" / etc), but also it's size ("list [3]"). Is it hard to implement in the new solution? Or will it reduce performance?

Otherwise, looks good to me.

satabol:
Hi Ilia @portnov 
>I would suggest to move this to data_structure.py or into some separate module under utils.

I intend to explore this method further, so I propose to bring it to the library later.

>Maybe give names to these magic constants?

I think numbers work faster. What do you think?

>Also, previous version of node indicated not only the type of nested structure ("list" / "tuple" / etc), but also it's size ("list [3]"). Is it hard to implement in the new solution? Or will it reduce performance?

In the old version, the composition could be displayed incorrectly because the statistics were counted by the first element in the list.

This is possible, but it has a significant impact on productivity. One can add a parameter that reads statistics when enabled, but this setting also affects performance.

Additionally:
One can still increase the performance (x1.5) if replace isinstance() with type(), but I’m not sure that’s right for now.

portnov:
>> I think numbers work faster. What do you think?

I don't think so. But that's quite easy to check, no need to guess. Use a benchmark with timeit module or sverchok's profiler.

portnov:
> In the old version, the composition could be displayed incorrectly because the statistics were counted by the first element in the list.

> This is possible, but it has a significant impact on productivity. One can add a parameter that reads statistics when enabled, but this setting also affects performance.

Maybe add an option to switch to old implementation then? Yes it's slower, but sometimes information is more important.

satabol:
I decided to do away with multiple sockets because it created a lot of confusion for the output of statistics and management controls. I will cancel the dynamic creation of sockets.

image

satabol:
Hi Ilia @portnov 
Please, if possible, look at this version and test it? Given that this is a list job, I would like to see some more feedback on the outcome. Productivity is slightly lower than it was originally, but statistics are supported:

image

image

satabol:
TODO: One can increase performance by aggregating statistics and data processing

PR #5442, opened 2026-06-29
fix 5441 Simplify List Input UI for bool, int, float

### LINKED ISSUE #5441: Simplify List Input UI for bool, int, float

satabol:
fix #5441 - refactor UI

New version looks more compact

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+2026.06.29
+
+- refactor List Input:
+    - https://nortikin.github.io/sverchok/docs/nodes/number/list_input.html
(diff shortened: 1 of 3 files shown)


portnov:
Maybe make masking checkboxes optional as well?

satabol:
Hi Ilia @portnov 
I thought about it and I think that we need to upgrade the version for this, because disabling masking affects the inbound socket. The new UI interface looks much better even with the mask. What do you think?

satabol:
When using the Matrix, the space that the mask occupies is not large

image

portnov:
> The new UI interface looks much better even with the mask. What do you think?

Yes.

